PR #6, opened 2023-08-31
Enable base URL to be configurable for OpenAI

habuma:
Spring AI currently is fixed on using the official OpenAI URL when working with OpenAI. This makes it impossible to use Spring AI against a locally running OpenAI API such as [LocalAI](https://github.com/go-skynet/LocalAI).

This PR exposes a new property (`spring.ai.openai.base-url`) that allows for custom configuration of the base URL. It still defaults to the official OpenAI API, though. 

For example, to use this with a locally running LocalAI listening on port 8080, the following entry can be added to the application.properties file:

```
spring.ai.openai.base-url=http://localhost:8080/
```

### DIFF
--- a/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/openai/OpenAiAutoConfiguration.java
+++ b/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/openai/OpenAiAutoConfiguration.java
@@ -16,7 +16,14 @@
 
 package org.springframework.ai.autoconfigure.openai;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.theokanning.openai.OpenAiApi;
 import com.theokanning.openai.service.OpenAiService;
+
+import okhttp3.OkHttpClient;
+import retrofit2.Retrofit;
+import retrofit2.adapter.rxjava2.RxJava2CallAdapterFactory;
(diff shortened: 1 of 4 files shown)


spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/openai/OpenAiAutoConfiguration.java
```diff
@@ -49,7 +56,19 @@ public OpenAiService theoOpenAiService(OpenAiProperties openAiProperties) {
 			throw new IllegalArgumentException(
```
gm2552, line 56: Would it make sense to only force an API key if the URL was anything other than the default base URL (maybe that's too complex or nondeterministic)?

markpollack: I've seen two cases of model hosting projects that expose the OpenAI API locally and do not require a key.  I think we can remove the requirement for a key when the base url is not `https://api.openai.com` and see how it goes.

markpollack:
I have merged this, just adding a comment that the underlying library will likely add this support as they have an open issue for it.  I also enforced the use of the API KEY only if the baseUrl is pointing to api.openai.com

commit; 02180696133f644a2f1c3136458210e75a4c5a52

Thanks for the contribution.

PR #21, opened 2023-09-19
Add MongoDB as a vector store

Kirbstomper:
Adding MongoDB as a vector store.

Implemented using basically the same logic and math as `InMemoryVectorStore` so it's prob worth breaking some of those shared classes out.

Big thanks to @tzolov for the work on `pgvector-store`, I was able to use most of his tests as guidance

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
@@ -42,12 +42,6 @@ public class Document {
 		this.metadata = metadata;
 	}
 
-	public Document() {
-		this.id = null;
-		this.text = null;
-		this.metadata = null;
-	};
-
 	public String getId() {
 		return id;
 	}
--- a/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
@@ -42,6 +42,12 @@ public class Document {
 		this.metadata = metadata;
 	}
 
+	public Document() {
+		this.id = null;
+		this.text = null;
+		this.metadata = null;
+	};
+
 	public String getId() {
 		return id;
 	}
--- a/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/impl/InMemoryVectorStore.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/impl/InMemoryVectorStore.java
@@ -95,6 +95,14 @@ public class InMemoryVectorStore implements VectorStore {
 			this.similarity = similarity;
 		}
 
+		public double getSimilarity() {
+			return similarity;
+		}
+
+		public String getKey() {
+			return key;
+		}
+
 	}
 
 	public class EmbeddingMath {
--- a/vector-stores/spring-ai-mongodb-store/src/main/java/org/springframework/ai/vectorstore/MongoDBVectorStore.java
+++ b/vector-stores/spring-ai-mongodb-store/src/main/java/org/springframework/ai/vectorstore/MongoDBVectorStore.java
@@ -20,7 +20,9 @@ import com.mongodb.BasicDBObject;
 import org.springframework.ai.document.Document;
 import org.springframework.ai.embedding.EmbeddingClient;
 import org.springframework.ai.vectorstore.impl.InMemoryVectorStore;
+import org.springframework.data.domain.Sort;
 import org.springframework.data.mongodb.core.MongoTemplate;
+import org.springframework.data.mongodb.core.aggregation.Aggregation;
 import org.springframework.data.mongodb.core.query.Query;
 
 import java.util.Comparator;
@@ -39,7 +41,7 @@ public class MongoDBVectorStore implements VectorStore {
 
 	private EmbeddingClient embeddingClient;
 
-	private static final String VECTOR_COLLECTION_NAME = "vector_store";
+	private static final String VECTOR_COLLECTION_NAME = "SampleCollection";
 
 	public MongoDBVectorStore(MongoTemplate mongoTemplate, EmbeddingClient embeddingClient) {
 		this.mongoTemplate = mongoTemplate;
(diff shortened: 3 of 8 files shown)


markpollack:
Thanks so much, will review!

spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
```diff
@@ -42,6 +42,12 @@ public Document(String id, String text, Map<String, Object> metadata) {
 		this.metadata = metadata;
 	}
 
+	public Document() {
```
tzolov, line 45: What necessitates an empty constructor? 
Mind that currently the `Document` class doesn't have setters for the `id`, `text` and `metadata`!

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/impl/InMemoryVectorStore.java
```diff
@@ -95,6 +95,14 @@ public Similarity(String key, double similarity) {
 			this.similarity = similarity;
 		}
 
+		public double getSimilarity() {
```
tzolov, line 98: Not sure those methods in the InMemoryVectorStore are relevant to the MongoDB store?

vector-stores/spring-ai-mongodb-store/src/main/java/org/springframework/ai/vectorstore/MongoDBVectorStore.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Optional;
+
+import static org.springframework.data.mongodb.core.query.Criteria.where;
+
+/**
+ * @author Chris Smith
+ */
+public class MongoDBVectorStore implements VectorStore {
+
+	MongoTemplate mongoTemplate;
```
tzolov, line 38: should be private

vector-stores/spring-ai-mongodb-store/src/main/java/org/springframework/ai/vectorstore/MongoDBVectorStore.java
```diff
@@ -0,0 +1,128 @@
...
+
+	EmbeddingClient embeddingClient;
```
tzolov, line 40: should be a private field

vector-stores/spring-ai-mongodb-store/src/main/java/org/springframework/ai/vectorstore/MongoDBVectorStore.java
```diff
@@ -0,0 +1,128 @@
...
+
+	private final String VECTOR_COLLECTION_NAME = "vector_store";
```
tzolov, line 42: static as well

vector-stores/spring-ai-mongodb-store/src/main/java/org/springframework/ai/vectorstore/MongoDBVectorStore.java
```diff
@@ -0,0 +1,128 @@
...
+			mongoTemplate.createCollection(VECTOR_COLLECTION_NAME);
+		}
+	}
+
+	/**
+	 * Maps a basicDBObject to a Spring AI Document
+	 * @param basicDBObject
+	 * @return
+	 */
+	public Document mapBasicDbObject(BasicDBObject basicDBObject) {
```
tzolov, line 57: shouldn't this method has a `private` access?

vector-stores/spring-ai-mongodb-store/src/main/java/org/springframework/ai/vectorstore/MongoDBVectorStore.java
```diff
@@ -0,0 +1,128 @@
...
+		return document;
+	}
+
+	@Override
+	public void add(List<Document> documents) {
+		for (Document document : documents) {
+			List<Double> embedding = this.embeddingClient.embed(document);
+			document.setEmbedding(embedding);
+
+			mongoTemplate.save(document, VECTOR_COLLECTION_NAME);
```
tzolov, line 75: add `this.` prefix (e.g `this.mongoTemplate...`)

vector-stores/spring-ai-mongodb-store/src/main/java/org/springframework/ai/vectorstore/MongoDBVectorStore.java
```diff
@@ -0,0 +1,128 @@
...
+		}
+	}
+
+	@Override
+	public Optional<Boolean> delete(List<String> idList) {
+		Query query = new Query(where("_id").in(idList));
+
+		var deleteRes = mongoTemplate.remove(query, VECTOR_COLLECTION_NAME);
```
tzolov, line 83: missing field `this.` prefix

vector-stores/spring-ai-mongodb-store/src/main/java/org/springframework/ai/vectorstore/MongoDBVectorStore.java
```diff
@@ -0,0 +1,128 @@
...
+	@Override
+	public List<Document> similaritySearch(String query, int k) {
+		List<Double> queryEmbedding = this.embeddingClient.embed(query);
+
+		// TODO: explore making this query better
+		return this.mongoTemplate.findAll(BasicDBObject.class, VECTOR_COLLECTION_NAME)
+			.stream()
+			.map(this::mapBasicDbObject)
+			.map(entry -> new InMemoryVectorStore.Similarity(entry.getId(),
+					InMemoryVectorStore.EmbeddingMath.cosineSimilarity(queryEmbedding, entry.getEmbedding())))
```
tzolov, line 103: I'm confused. I've assumed, wrongly perhaps, that MongoDB provides a build-in Vector Search support? 

If it doesn't provide a native vector search capability, then I'm not convinced this Mongo DB sore would perform or scale any better than the in-memory one.  

tzolov (changes requested):
Hi @Kirbstomper, 

Thank you for your contribution. You can find my comments and requests inline. 

But my primary concern is that MongoDB does not provide native Vector Search capabilities, or at least this implementation doesn't lever them?

vector-stores/spring-ai-mongodb-store/src/main/java/org/springframework/ai/vectorstore/MongoDBVectorStore.java
Kirbstomper: Thanks for taking the time to review @tzolov!

The managed version of MongoDB (atlas), allows configuration to provide semantic search capability by automatically attaching embeddings to documents using  fields configured for each collection.
https://www.mongodb.com/products/platform/atlas-vector-search

Given that, I more of less went for making a `PersistentVectorStore` backed by MongoDB, partially out of "could I do it" and to provide another persistent  (though not optimal) out of the box.

markpollack:
This is a strange case.  All other vector stores that I'm aware of would implement the search inside the database, where as in this case is implementing a search as a linear in-memory scan.  I think we would violate the "Principle of least astonishment" with this implementation.  The implementation should be changed/updated to use AtlastVectorSearch from MongoDB.  

I'll leave the PR open for a while to discuss.

markpollack:
Maybe contributing a MongoDB implementation of `org.springframework.ai.loader.Loader` would be a good reuse of this PR for another feature.   Thoughts?

tzolov:
@Kirbstomper, 
For me would bring more value if the mongodb integration supports the `Atlast VectorSearch`.  
Don't know how challenging this task might be but will appreciate if you can explore this direction. 
The Loader is a good idea as well.

Kirbstomper:
@markpollack @tzolov 
Getting a MongoDB data loader together doesn't look like it will be too much trouble. I can explore that in a separate feature/PR.

As for better supporting the capabilities provided by Atlas I can start to explore that direction. For now I think we can close this PR until that gets to a more realized state.

jxblum:
Have a look at Spring AI Issue #48. MongoDB offers [Atlas Vector Search](https://www.mongodb.com/products/platform/atlas-vector-search), which might be a suitable way to go.

Kirbstomper:
> Have a look at Spring AI Issue #48. MongoDB offers [Atlas Vector Search](https://www.mongodb.com/products/platform/atlas-vector-search), which might be a suitable way to go.

So I was looking into this a bit last week. Using its built in similarity search seemed to be giving results I didn't expect, so need to take a deeper look on it

markpollack:
I'll close this PR for now and there can be a new one that includes support for mongodb's native capabilities.

PR #25, opened 2023-09-21
Added Record test for output parser

jexp:
Added a test for records for the beanoutputparser, as it was not obvious that it works.


markpollack:
merged as 6554bdebda009d0f10f49b23ad3805420a1e1ab6

Thanks!

PR #29, opened 2023-09-21
Add embedding model dimensions retrieval

### LINKED ISSUE #28: Retrieve embedding model dimensions
**Expected Behavior**

The `EmbeddingClient` interface should expose a method to report the dimension size of the configured embedding model. 

**Current Behavior**

Currently the users must retrieve this information manually from the documentations of the used models. 

**Context**

The dimensions are need to configure VectorStores and at the moment has to be specified explicitly. 
Instead they could obtain this information directly from the provided EmbeddingClient instance.

tzolov:
- Add dimension method to the EmbeddingClient interface. Default dimension implementation uses the embed method to produce results and counts the result dimensions.
 - Add EmbeddingUtil#dimensions utilities that look up the model dimensions from a pre-defined (static) file. If the requested model is unknown, fallback to the default behaviour.
 - Override the dimensions method in the OpenAiEmbeddingClient and AzureOpenAiEmbeddingClient to implement local caching.
 - Add unit and IT tests.

 Resolves #28


markpollack:
merged as 92773f17c041b2c140aa13039abe32fe62310c94

PR #32, opened 2023-09-25
modify message type get value

JoeCP17:
Hello, while I was looking at this project with interest, I modified the MessageType within the service method so that the item that retrieves the value through getValue can be returned directly from the AbstractMessage class, which is an abstract method.

Additionally, in the switch-case statement of convertToMessage, in addition to the value declared as the existing MessageType, I changed the part that was repeatedly declared as a string to use the value within the MessageType.


markpollack:
I'm not super convinced on the change from `getMessageType().getValue()` to `getMessageTypeValue() ` but I suppose it is a bit more readable so will merge it.  The catch on using the string instead of the type is a good one!

Thanks

markpollack:
merged as 5ff2c3bb925df7a919cb3ab890a2e4de56e57569

PR #35, opened 2023-09-28
Refactor find the last punctuation mark

hgs-study:
When truncating the token, we need to find the last punctuation mark. But in our current code, we're looping over each punctuation mark. If it can't find that punctuation mark, it will loop the length of the chunkText. I added something to define the punctuation as an enum and terminate the loop as soon as that punctuation is found. 

Thanks.


markpollack:
Closing as the current functionality has passing tests  and no feedback on questions.  Please resubmit if you want and thanks in any case.

PR #38, opened 2023-09-29
Streamline the Document API

tzolov:
Add TextFormatter and DefaultTextFormatter concept that can filter the metadata and format the Document metadata and text according to predefined templates.


spring-ai-core/src/main/java/org/springframework/ai/document/ContentFormatter.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Converts the Document text and metadata into a AI, prompt-friendly text representation.
+ *
+ * @author Christian Tzolov
+ */
+public interface ContentFormatter extends Function<Document, String> {
+
+	enum MetadataMode {
+
+		ALL, EMBED, LLM, NONE;
```
markpollack, line 30: I think we should change LLM to something like `INFERENCE` of `MODEL` as I've removed use of `LLM` generally speaking in naming things.

spring-ai-core/src/main/java/org/springframework/ai/document/DefaultContentFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+	 * placeholders.
+	 */
+	private final String textTemplate;
+
+	private final ContentFormatter.MetadataMode metadataMode;
+
+	/**
+	 * Metadata keys that are excluded from text for the LLM.
+	 */
+	private final List<String> excludedLlmMetadataKeys;
```
markpollack, line 73: similar comment wrt to naming , `excludedInferenceMetadataKeys` ?

spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
```diff
@@ -25,130 +31,101 @@ public class Document {
...
+	/**
+	 * Document content. TODO: To support binary content (image, audio ...)
+	 *
+	 * - One option is to change the content type to byte[]. Another option is to use
+	 * generics, e.g. Document<String> vs Document<byte[]>
+	 */
+	private String content;
+
+	/**
+	 * TODO: do we need the embedding field in the Document? Currently it is used only for
```
markpollack, line 43: I believe yes, usage is more widespread in llamaindex, we should review.

spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
```diff
@@ -25,130 +31,101 @@ public class Document {
...
+	/**
+	 * Mutable, ephemeral, content to text formatter. Defaults to Document text.
+	 */
+	@JsonIgnore
+	private ContentFormatter contentFormatter = DEFAULT_CONTENT_FORMATTER;
 
 	@JsonCreator(mode = JsonCreator.Mode.PROPERTIES)
-	public Document(@JsonProperty("text") String text) {
-		this(text, new HashMap<>());
+	public Document(@JsonProperty("content") String content) {
```
markpollack, line 58: we will need to change the version of the project, might as well go  for 0.7.0 snapshots, as this is a breaking api change and I don't want to disrupt the azure/spring-ai workshops and other working examples.

spring-ai-core/src/main/java/org/springframework/ai/splitter/TextSplitter.java
```diff
@@ -49,6 +64,13 @@ private List<Document> createDocuments(List<String> texts, Map<String, Object> m
 					.stream()
 					.collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()));
 				Document newDoc = new Document(chunk, metadataCopy);
+
+				if (this.copyContentFormatter) {
+					// Copies the parent content formatter to the chucks documents it was
+					// spelt into.
+					newDoc = newDoc.updateContentFormatter(formatters.get(i));
```
markpollack, line 71: multiple formatters indexed by List of strings seems unusual to me.

spring-ai-core/src/main/java/org/springframework/ai/document/ContentFormatter.java
tzolov: I like INFERENCE

spring-ai-core/src/main/java/org/springframework/ai/document/DefaultContentFormatter.java
tzolov: done

spring-ai-core/src/main/java/org/springframework/ai/splitter/TextSplitter.java
tzolov: It is because the `createDocuments` takes a list of `texts`, one per source document, and merged metadata of all source documents. We can not merge the content formatters, so we have to pass in a list of formatters (one per source document) that corresponds in order to the list of texts. 

We can simplify the createDocuments to take one document at a time, but will loose the metadata merging, or introduce an internal record class like `Cuhnk(String text, Map<String, Object> metadata, ContentFormatter formatter)`?

tzolov: Actually,  why de we merge the metadata from all parent documents? Shouldn't the chunked docs inherit only the metadata their own parent? Not a mixture of metadata of all different parents? 
Later can make sense if we can determine that all parent documents come from the same initial source for example. But, currently, we don't have a convention to reliably recognise this.

tzolov:
Replaced #45

PR #39, opened 2023-10-02
Add PdfLoader

nikitsenka:
This pull request introduces a new PdfLoader class to the project. The PdfLoader class is responsible for loading PDF documents and converting them into a list of Document objects. This is achieved by using the Apache PDFBox library to read the PDF file and extract the text.


nikitsenka:
Closing this and will create a new one as wrong account was used

PR #51, opened 2023-10-08
add "List<Double>" right bracket

lowenest:
Fix typo in concepts.adoc
before:   List<Double   
after :     List\<Double\>


sullrich84:
Would suggest to change the commit message to "fix typo"

lowenest:
> Would suggest to change the commit message to "fix typo"

The commit message was changed to 'fix typo' as suggested.

PR #64, opened 2023-10-27
Make DocumentReader stateless; pass resource to be read at read time

habuma:
Per the discussion in Slack (https://vmware.slack.com/archives/C058EAJEBU1/p1698410369053209), I've changed `DocumentReader` such that:

- No longer implements `Supplier`
- Renamed `get()` to `read()` (with 2 overloaded implementations)
- No longer requires a `Resource` (or URL string) at construction time.
- Defer provision of `Resource` or URL string to `read()` invocation. 

This allows the reader to be configured once as a bean in the Spring application context and then injected and used as many times as you'd like to read multiple documents (as opposed to creating a new instance for every document to be read).

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/document/DocumentReader.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/document/DocumentReader.java
@@ -1,8 +1,13 @@
 package org.springframework.ai.document;
 
 import java.util.List;
-import java.util.function.Supplier;
 
-public interface DocumentReader extends Supplier<List<Document>> {
+import org.springframework.core.io.Resource;
+
+public interface DocumentReader {
+
+	List<Document> read(String resourceUrl);
+
+	List<Document> read(Resource resource);
 
 }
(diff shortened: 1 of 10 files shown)


spring-ai-core/src/main/java/org/springframework/ai/document/DocumentReader.java
```diff
@@ -1,8 +1,13 @@
 package org.springframework.ai.document;
 
 import java.util.List;
-import java.util.function.Supplier;
 
-public interface DocumentReader extends Supplier<List<Document>> {
+import org.springframework.core.io.Resource;
+
+public interface DocumentReader {
```
markpollack, line 7: probably implementing Function<Resource,List<Document>> and not sure about read trying to parse the string as a resourceUrl... vs. a raw string...

markpollack:
Will include this change as part of moving away from using `Consumer` and `Supplier` to use Function instead.  

Maybe we have a `ResourceDocumentReader`,  `<Resource, List<Document>>` implementation that supports `Resource` input, but another more open ended implementation with the signature `<Void, List<Document>>` , for example in use with a `SqlDocumentReader` that gets data from a database and doesn't really need an input that lets you select from a different databases, but it tied to one upon construction.

`DocumentReader implements Function<I, List<Document>>`  with implementations like above.

markpollack:
See https://github.com/spring-projects-experimental/spring-ai/issues/106

PR #71, opened 2023-11-01
Add initial support for HuggingFace Inference Service

markpollack:


markpollack:
had an extraneous commit in there, replacing with 
 https://github.com/spring-projects-experimental/spring-ai/pull/78

PR #72, opened 2023-11-01
Add Pinecone VectorStore

tzolov:
- Based on the official pinecone java library. Later expects that indices are created externally via Ops.
 - Map Document metadata to and from Pinecone's internal Struct. Later converts the metadata into pinecone json format.
 - Add integration tests.


markpollack:
Tested and updated readme to be more of a tutorial style.  merged as 81a5eaf80167c524f2ea29a0ada00ac901fa200c

PR #80, opened 2023-11-03
Conversation Chain

habuma:
This PR builds upon `AiChain` in https://github.com/spring-projects-experimental/spring-ai/pull/73 to provide a chain with memory, useful for conversations with the model.

It can be configured as a bean, similarly to how `AiChain` is configured, although it does come with a default prompt template making it a small bit easier to configure. For example:

```
@Bean
ConversationChain conversationChain(AiClient aiClient) {
    return new ConversationChain(aiClient, "response", new StringOutputParser());
}
```

You can still provide a custom template if you want, though:

```
@Bean
ConversationChain conversationChain(AiClient aiClient, @Value("classpath:/chat-prompt.st") Resource prompt) {
    return new ConversationChain(aiClient, prompt, "question", "response", new StringOutputParser());
}
```

In the former case, the input key defaults to "input" because that's what the default template requires. In the latter case, the input key is required and must match what is in the template. This is different from `AiChain` where the input keys are derived from the template itself. That's because the template here actually has two input keys, one that will carry the user's query and another that carries the history. The history key shouldn't be treated the same as a normal input key and there's no obvious way to determine from the template which key is which, so the user must specify the name of the non-history input key.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chain/AbstractChain.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chain/AbstractChain.java
@@ -29,7 +29,7 @@ public abstract class AbstractChain implements Chain {
 	public AiOutput apply(AiInput aiInput) {
 		AiInput aiInputToUse = preProcess(aiInput);
 		AiOutput aiOutput = doApply(aiInputToUse);
-		Map<String, Object> outputMapToUse = postProcess(aiInput, aiOutput);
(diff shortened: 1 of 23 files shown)


markpollack:
This was great to focus on this high level use case.  The ideas and even some of this impl will be reused in a 'non-chain' version as discussed here - https://github.com/spring-projects-experimental/spring-ai/issues/110

Issue #82, opened 2023-11-06
Add Azure Cognitive Search as a Vector Store


gm2552:
Azure Cognitive Search provides value add capabilities to users of Azure.  In addition to supporting easy document upload and
indexing, indexes that are configured with Vector storage could be integrated and consumed via a VectorStore implementation in Spring AI.

**Expected Behavior**

Azure Cognitive Search supports a Java search API for uploading, deleting, and searching documents using vector based queries and storage.  This API could be wrapped in a Spring AI Vector Store implementation giving Spring AI and Azure users the ability to integrate their existing Cognitive Search indexes with Spring AI.  

Conceptually, a user would create a SearchClient instance using existing methods and inject the instance along with an EmbeddingClient into a VectorStore implementation.  Using bean creation methods and a PropertiesSource class, the VectorStore creation could look like the following

```java
@Bean
public SearchClient searchClient(AzureSearchClientProperties props) {

	return new SearchClientBuilder().endpoint(props.getEndpoint())
		.credential(new AzureKeyCredential(props.getApiKey()))
		.indexName(props.getIndex())
		.buildClient();
}

@Bean
public VectorStore vectorStore(SearchClient searchClient, EmbeddingClient embeddingClient) {
	return new AzureCognitiveSearchVectorStore(searchClient, embeddingClient);
}
```

**Current Behavior**

Spring AI currently does not support Azure Cognitive Search as a VectorStore implementation.

**Context**

An Azure Spring [sample project](https://github.com/Azure-Samples/spring-chatgpt-sample) currently exists that 
implements a Cognitive Search based VectorStore.  In addition, the Microsoft AI learning pages also demonstrate [sample code](https://learn.microsoft.com/en-us/azure/ai-services/openai/use-your-data-quickstart?tabs=command-line&pivots=programming-language-csharp)
that utilizes existing documents already uploaded and indexed into CognitiveSearch.  It would be desirable for Spring AI to integrate with Azure Cognitive Search as another VectorStore implementation.

PR #83, opened 2023-11-06
Initial implementation of Azure Cognitive Search vector store.

gm2552:
Implementing enhancement request #82

### DIFF
--- a/vector-stores/spring-ai-cognative-search-store/README.md
+++ b/vector-stores/spring-ai-cognative-search-store/README.md
@@ -1,0 +1,261 @@
+# Azure Cognitive Search VectorStore
+
+This README will walk you through setting up the Cognitive VectorStore to store document embeddings and perform similarity searches.
+
+[Azure Cognitive Search](https://azure.microsoft.com/en-us/products/ai-services/cognitive-search) is a versatile cloud hosted cloud information retrieval system
+that is part of Microsoft's larger AI platform.  Among other features, it allows users to query information using vector based storage
+and retrieval.
+
+## Prerequisites
+
+1. Azure Subscription: You will need an [Azure subscription](https://azure.microsoft.com/en-us/free/) to use any Azure service. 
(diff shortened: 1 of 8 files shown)


tzolov:
Hi @gm2552 , thank you for the contribution. 
For your information, I've reviewed, rebased and streamlined your PR code. 
Note that after updating to latest 11.6.0 version some braking changes popped up. 

Here is my refactoring on your PR: https://github.com/tzolov/spring-ai/tree/gh-83-azure-vector-store (it is still your pr ;).
Amongst other things i've simplified the naming convention, added `afterPropertySet()` section to setup the search index (Now we use the SearchIndexClient as a supper factory of SearchClient). 
Also i've removed some unhealthy dependencies (such as  lombok). 
Regarding the metadata i couldn't figure out a mechanism to map it to the index schema, as later is fixed while our metadata doesn't have fixed structure. 
This is unfortunate because we can't take advantage of the filtering capabilities. 
Anyway have a look and let me know if there is something missing. 

At the moment I'm waiting for my Azure Open AI account to be activated (apparently can take 10 days) to test the code and then will merge it.

gm2552:
Thanks for going through this, @tzolov.  I've reviewed your changes, and I'm good with them.

tzolov:
@gm2552 , do you know if the search-index schema could support variable length, key-value structures (e.g. Map<String, String>)? 
The [COMPLEX](https://github.com/Azure/azure-sdk-for-java/blob/0f379118d823884583bd43f1aa4f5e6ac9decab9/sdk/search/azure-search-documents/src/samples/java/com/azure/search/documents/indexes/CreateIndexExample.java#L51) type allows for nested data structures but it still requires a schema with fixed set of fields. 
I wonder if there is a way to configure are `repetitive` or `zero-or-more` key & value fields? 
If above is possible then we can map the metadata to it and leverage the OData filtering mechanism.

tzolov:
Rebased, swashed and merged at 0aa8eb651e442cdf43efebdb96354c40f8d37d93

Issue #86, opened 2023-11-09
Add Chroma Vector Store integration


tzolov:
[Chroma](https://docs.trychroma.com/) is  a popular open-source embedding database. It gives you the tools to store document embeddings, content and metadata and to search through those embeddings including metadata filtering.

tzolov:
Resolved by 98ca3e2a8fe4228b5c556e14dbdeb0b5c5ece0ae

PR #87, opened 2023-11-09
Add Chroma VectorStore support

tzolov:
- Implement ChromaApi client, based on Chroma REST API.
 - Implement ChromaVectorStore, including support for filter expression conversion.
 - Common VectorStoreUtil class to share to/from Float/Double list/array convertion as well as Json/Map convertions.
 - Add ITs including for Basic Auth and Token autheticatios.
 - Add ChromaApi security support for BasicAuth and Token.
 - Fix an issue with Text filter expression parser, related to double-quoted identifiers.
 - Add Chroma README.md.

 Resolves# #86


tzolov:
Rebased, squashed and merged at: 98ca3e2a8fe4228b5c556e14dbdeb0b5c5ece0ae

PR #89, opened 2023-11-09
Add readme for Neo4jVectorStore

meistermeier:
Heavily "inspired" by the PgVectorStore readme but I think it's useful to have the same structure and examples for maintenance.


markpollack:
Thanks so much!

PR #95, opened 2023-11-10
Simplify VectorStore interface

tzolov:
- Collapses all VectorStore similaritySearch metdods into one with SearchRequest builder.
 - Fix all affected code and tests.
 - Bump the project version to 0.7.1.
 - Adds Boot Auto-Configuraitons for milvus, pinecone and pgvector stores. 

 Resolves# https://github.com/spring-projects-experimental/spring-ai/issues/95, https://github.com/spring-projects-experimental/spring-ai/issues/96


spring-ai-core/src/main/java/org/springframework/ai/ResourceUtils.java
```diff
@@ -0,0 +1,39 @@
...
+import java.nio.charset.StandardCharsets;
+
+import org.springframework.core.io.DefaultResourceLoader;
+
+/**
+ * @author Christian Tzolov
+ */
+public class ResourceUtils {
+
+	public static String getText(String uri) {
```
markpollack, line 29: Maybe there should be a constructor to Document that takes a Resource and/or have the current String constructor try to parse the passed in text to see if it is a resource string and then take appropriate action.  I guess that would mean you can't create a document with the test "classpath:..."

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/VectorStore.java
```diff
@@ -30,107 +27,24 @@ default void accept(List<Document> documents) {
...
+	 * {@link SearchRequest}'s' search criteria.
+	 * @param query Text to use for embedding similarity comparison.
+	 * @return Returns a list of documents that have embeddings similar to the query text
+	 * embedding.
 	 */
-	default List<Document> similaritySearch(String query, int topK, double similarityThreshold,
-			String filterExpression) {
-		var filterExpressionObject = Filter.parser().parse(filterExpression);
-		return similaritySearch(query, topK, similarityThreshold, filterExpressionObject);
+	default List<Document> similaritySearch(String query) {
```
markpollack, line 46: Just a thought, but we may need another overloaded metho that returns the List<Document>s but also some 'metadata' about the query.  I see that pinecone has the class QueryResponse which brings along additional metadata.    That could maybe go inside the document class itself under 'queryMetadata' or some other class that augments/wraps Document, DocumentWithQueryData ?

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/SearchRequest.java
```diff
@@ -0,0 +1,251 @@
...
+	/**
+	 * Default value for the top 'k' similar results to return.
+	 */
+	public static final int DEFAULT_TOP_K = 4;
+
+	public String query;
+
+	private int topK = DEFAULT_TOP_K;
+
+	private double similarityThreshold = SIMILARITY_THRESHOLD_ACCEPT_ALL;
```
markpollack, line 50: is this a good default?

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/milvus/MilvusServiceClientProperties.java
```diff
@@ -0,0 +1,280 @@
...
+	/**
+	 * The keep-alive timeout value of client channel. The timeout value must be greater
+	 * than zero. Default value is 20000 ms.
+	 */
+	private long keepAliveTimeoutMs = 20000;
+
+	/**
+	 * Enables the keep-alive function for client channel.
+	 */
+	// private boolean keepAliveWithoutCalls = false;
```
markpollack, line 76: remove?

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/SearchRequest.java
```diff
@@ -0,0 +1,251 @@
...
+	 * threshold value of 1.0 means an exact match is required.
+	 */
+	public static final double SIMILARITY_THRESHOLD_ACCEPT_ALL = 0.0;
+
+	/**
+	 * Default value for the top 'k' similar results to return.
+	 */
+	public static final int DEFAULT_TOP_K = 4;
+
+	public String query;
```
markpollack, line 46: why is this public access?

markpollack:
added javadocs to the vectorstore interface and moved the code for loading a resource as a string into the test projects.  That makes for some small duplication but better than leaving it in the main code base as 'test cruft'. ;)

PR #102, opened 2023-11-15
Refine delineation of Embedding properties from OpenAI (client) properties

jxblum:
This applies Spring Boot's style to the definition of `@ConfigurationProperties` classes (for [example](https://github.com/spring-projects/spring-boot/blob/v3.1.5/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java)) to `OpenAiProperties`.

I am somewhat undecided about defining a proper, additional property namespace for the AI client, such as:

`spring.ai.openai.client.*`

For example:

`spring.ai.openai.client.api-key`.


markpollack:
I think having the additional scope after `sping.ai.openai` makes sense.  Though I'm not sure 'client' will stick, as I'm thinking to rename `AiClient` to  `GenerateClient` or `TextGenerationClient`.  We need to go through other APIs that openai offers and think up the right segmentation of configuration to plan out for the future.  Will address this during overall reviews.

PR #103, opened 2023-11-15
Make filter expression builder op public

KathrynBrusewitz:
Following the simple example for filter expressions, when I do the following:
```
val b = FilterExpressionBuilder()
val expression: Expression = b.eq("state", "Oregon").build()
```
I get these errors:
```
Type FilterExpressionBuilder.Op! is inaccessible in this context due to: public/*package*/ final class Op
Cannot access 'Op': it is package-private in 'FilterExpressionBuilder'
```

I think the fix is simply making it public.


tzolov:
Rebased, squashed  and merged at: 837c4080aa06f04b9f3c007ebfba6fc67be1b1ca

PR #117, opened 2023-11-21
Support poratlbe, filter expressions with AzureVectorStore

tzolov:
Support generic, portable [metadata filters](https://docs.spring.io/spring-ai/reference/api/vectordbs.html#_metadata_filters) with AzureVectorStore. 
For example the:
```java
vectorStore.similaritySearch(
   SearchRequest
      .query("The World")
      .withTopK(TOP_K)
      .withSimilarityThreshold(SIMILARITY_THRESHOLD)
      .withFilterExpression("country in ['UK', 'NL'] && year >= 2020"));
```
Internally will converted to the [Azure Search OData filters](https://learn.microsoft.com/en-us/azure/search/search-query-odata-filte) like: `$filter search.in(meta_country, 'UK,NL', ',') and meta_year ge 2020`.

All portable operations are supported, but you have to list explicitly the metadata fields used in filter expression.

### DIFF
--- a/vector-stores/spring-ai-azure/README.md
+++ b/vector-stores/spring-ai-azure/README.md
@@ -43,7 +43,7 @@ Add these dependencies to your project:
 1. Select an Embeddings interface implementation.
 You can choose between:
 
-* OpenAI Embedding:
+* or OpenAI Embedding:
 
    ```xml
    <dependency>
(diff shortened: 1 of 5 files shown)


markpollack:
Moved into package `org.springframework.ai.vectorstore.azure`  There are some other vector stores that need an additional level in their package name as different jars can't use the same package name.

markpollack:
merged as 69d6da2393abef7cdfe95a971f19903f3712a813

PR #118, opened 2023-11-21
Add OpenAI Stream Client implementation

### LINKED ISSUE #116: Stream support in `AiClient`
**Context**

Applications using AI often use a UI that generates text little by little in a stream instead of outputting texts all at once. The current spring-ai `AiClient` does not support stream output.

**Expected Behavior**

I hope that `generateStream` method will be added to `AiClient`. As for the return value, `Flux<AiResponse>` is promising, but if you want to avoid dependence on reactor, you can also consider the JDK standard `Flow.Publisher<AiResponse>`.

For example, openai-java used in `OpenAiClient` uses `Flowable`. This can be converted to `Flux` or `Flow.Publisher`.

https://github.com/TheoKanning/openai-java/blob/650d76b87a3246a52247b231f8d47f994ae63020/service/src/main/java/com/theokanning/openai/service/OpenAiService.java#L150

This is a little off topic, but openai-java uses outdated RxJava2 and has an unnecessary number of dependencies (e.g. kotlin sdk). Since spring-ai does not use all OpenAI APIs, I think it is better to have an in-house OpenAI client that implements only the minimum API with `WebClient` that supports the streaming.

JM-Lab:
- Add OpenAI Stream Client implementation.
- Add openAiStreamClient test in ClientIT.

Resolves #116

### DIFF
--- a/spring-ai-openai/pom.xml
+++ b/spring-ai-openai/pom.xml
@@ -56,6 +56,11 @@
 			<artifactId>spring-boot-starter-logging</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-webflux</artifactId>
+		</dependency>
+
 		<!-- test dependencies -->
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
--- a/spring-ai-openai/src/main/java/org/springframework/ai/openai/client/AiStreamClient.java
+++ b/spring-ai-openai/src/main/java/org/springframework/ai/openai/client/AiStreamClient.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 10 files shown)


making:
This PR is very specific to the openai-java implementation and does not resolve #116 .

tzolov:
@JM-Lab as @making pointed out we can not expose proprietary libraries throughout the Spring AI public API. 

Subclassing is not an option either. We need to define a new generic chunk interface or interfaces. 
Also the Spring projects is somewhat biased toward Flux instead of rx.  
Please share your design thoughts, apologies for not providing any guidelines so far.

JM-Lab:
Regarding the feedback from @tzolov, the development was revisited.
OpenAiStreamClient has been re-implemented to directly adhere to the latest Request/Response specifications of OpenAI's Chat Completion API.
Using Flux, the existing 'generate' method was re-implemented, and the 'generateStream' method was added.

The commit details are as follows:

- Add the latest Request and Response Classes of OpenAI's Chat Completion API
- Add the spring-boot-starter-webflux dependency
- Implement the AiStreamClient interface extending AiClient
- Implement OpenAiStreamClient using Reactor Flux

If you have any additional inquiries or recommendations, please don't hesitate to reach out.

tzolov:
Rebased, squashed and merged at 8139af468e9e3e8ba5f1d7f81f7e6e56c1c17fe0
@JM-Lab  thank you for the contribution and expect a follow-up changes as I'm heavily refactoring the chat generation and embedding clients.

PR #126, opened 2023-11-23
Add NOT operator to VectorStore portable filter expressions

tzolov:
- Add NOT expression type to the portable Filter.Expression model.
 - Add NOT to the Antlr grammar and implement the related parser listener method to generate Filter NOT expressions.
 - Add NOT support to the filter programming DSL.
 - Implement FilterHelper.negation for logically transform any boolean expression with NOT statements into semantically equivalent one with NOT applied to the leaf expressions.
 - Add tests for paresers, converters and vectorsores ITs.


tzolov:
replaced by https://github.com/spring-projects/spring-ai/pull/128

PR #128, opened 2023-11-24
Add NOT operator to VectorStore portable filter expressions

tzolov:
Add NOT expression type to the portable Filter.Expression model.
Add NOT to the Antlr grammar and implement the related parser listener method to generate Filter NOT expressions.
Add NOT support to the filter programming DSL.
Implement FilterHelper.negation for logically transform any boolean expression with NOT statements into semantically equivalent one with NOT applied to the leaf expressions.
Add tests for Dsl Parsers, expression converters and vector sores ITs.


markpollack:
merged as ac9ae589f41199eb0dd623efd3cd39e5bcb6b07d

PR #129, opened 2023-11-24
Fix link pointing to arxiv

harry9656:
I corrected the link to Arxiv.


tzolov:
@harry9656 , thank you for catching this.

PR #130, opened 2023-11-24
Add hint about cloning the repo

fabapp2:
I learned the hard way that GLFS must be installed to clone the repo and added a hint to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,9 @@ The Spring AI project provides a Spring-friendly API and abstractions for develo
 
 Let's make your `@Beans` intelligent!
 
+## Cloning the repo
+
+You need [Git Large File Storage](https://git-lfs.com/) installed to clone the repo.
 
 ## Project Links
 


tzolov:
Hi @fabapp2,
The intention is that the GLFS would be required only if you want to access, locally, the [default ONNX models](https://github.com/spring-projects/spring-ai/tree/main/embedding-clients/transformers-embedding/src/main/resources/onnx/all-MiniLM-L6-v2), because of their large size.
Even without GLFS the `TransformersEmbeddingClient`  looks up the default models directly from the [GH site](https://github.com/spring-projects/spring-ai/blob/c520e77472c378b49b680bcda0f24925c1b50324/embedding-clients/transformers-embedding/src/main/java/org/springframework/ai/embedding/TransformersEmbeddingClient.java#L43).

Are you experiencing something else?

fabapp2:
Hi @tzolov,
thanks for the information.
I couldn't simply `git clone`.

````
Resolving deltas: 100% (1389/1389), done.
git-lfs filter-process: git-lfs: command not found
fatal: the remote end hung up unexpectedly
warning: Clone succeeded, but checkout failed.
````

tzolov:
@fabapp2 , thank you for the update, 
Apparently, by default, the repository demands the git-lfs and you have set `export GIT_LFS_SKIP_SMUDGE=1` (for linux, macos) or `set GIT_LFS_SKIP_SMUDGE=1` (for windows) to disable it, 
or for single clone (`GIT_LFS_SKIP_SMUDGE=1 git clone git@github.com:spring-projects/spring-ai.git`)

I've incorrectly assumed that it is disabled by default.  

Given this, I'm not sure what is better 
 - (1) document the special, git clone instructions to either disable LFS or install git-lfs or
 - (2) move the ONNX files to a different GH repository.

fabapp2:
I'd move the ONNIX file to keep the expected default behavior of `git clone` without the need for additional documentation.

tzolov:
@fabapp2, I second your opinion. 
But, as it is not easy to get another spring-project GH dedicated for only for ONNX zoo, we need to figure out how to handle this. 
Until then i will merge your hint. 
Thank you!

fabapp2:
> @fabapp2, I second your opinion. But, as it is not easy to get another spring-project GH dedicated for only for ONNX zoo, we need to figure out how to handle this. Until then i will merge your hint. Thank you!

I had a feeling...
Thanks for merging the PR.

PR #143, opened 2023-12-04
fix: milvus builder configuration params setters

siddharthshankarpaul:
Fix the wrong copying of configuration params


tzolov:
Thanks for catching and fixing this @siddharthshankarpaul !

PR #169, opened 2023-12-15
Fix Prompts documentation typos

jtsnr:
Fix bullet point formatting for Function Role on the Prompts page.


markpollack:
Thanks @jtsnr

PR #170, opened 2023-12-15
Cleanup starters

eddumelendez:
starters do not provide spring-boot-starter-test


tzolov:
Thanks for cleaning this @eddumelendez

PR #172, opened 2023-12-17
Openai api sse review

JM-Lab:
@tzolov, I took a close look at the code and ran the beanStreamOutputParserRecords test in OpenAIClientIT. I'm sorry for the delayed response; I ran into an issue and needed some time to address it.

I made some changes, including defining the Delta for ChatCompletionChunk within the class to handle OpenAI's SSE for ChatCompletionChunk objects. I also adjusted the ObjectMapper for generating response JSON to align with the one used in RestClient.

Additionally, I refactored the code related to ending the Flux Stream and put it into a separate commit.

Please check out the draft pull request, and feel free to reach out if you have any questions or need further clarification.

### DIFF
--- a/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -39,6 +39,7 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
+import java.util.function.Predicate;
 
 // @formatter:off
 /**
@@ -51,7 +52,7 @@ public class OpenAiApi {
 
 	private static final String DEFAULT_BASE_URL = "https://api.openai.com";
 	private static final String DEFAULT_EMBEDDING_MODEL = "text-embedding-ada-002";
-	private static final String SSE_DONE = "[DONE]";
+	private static final Predicate<String> SSE_DONE_PREDICATE = "[DONE]"::equals;
 
 	private final RestClient restClient;
 	private final WebClient webClient;
@@ -579,10 +580,10 @@ public class OpenAiApi {
 				.body(Mono.just(chatRequest), ChatCompletionRequest.class)
 				.retrieve()
 				.bodyToFlux(String.class)
-				// cancels the flux stream after the SSE_DONE is received.
-				.takeUntil(content -> content.contains(SSE_DONE))
-				// filters out the SSE_DONE message.
-				.filter(content -> !content.contains(SSE_DONE))
+				// cancels the flux stream after the "[DONE]" is received.
+				.takeUntil(SSE_DONE_PREDICATE)
+				// filters out the "[DONE]" message.
+				.filter(SSE_DONE_PREDICATE.negate())
 				.map(content -> parseJson(content, ChatCompletionChunk.class));
 	}
 
--- a/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -16,29 +16,29 @@
 
 package org.springframework.ai.openai.api;
 
-import java.io.IOException;
-import java.util.List;
-import java.util.Map;
-import java.util.function.Consumer;
-
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import reactor.core.publisher.Flux;
-import reactor.core.publisher.Mono;
-
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.http.HttpHeaders;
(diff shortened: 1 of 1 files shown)


spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -51,7 +52,7 @@ public class OpenAiApi {
 
 	private static final String DEFAULT_BASE_URL = "https://api.openai.com";
 	private static final String DEFAULT_EMBEDDING_MODEL = "text-embedding-ada-002";
-	private static final String SSE_DONE = "[DONE]";
+	private static final Predicate<String> SSE_DONE_PREDICATE = "[DONE]"::equals;
```
tzolov, line 55: I presume that the Predicate is just a code style improvement, not due to issues with the existing conditions?

spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -533,7 +534,15 @@ public record ChatCompletionChunk(
 		public record ChunkChoice(
 				@JsonProperty("finish_reason") ChatCompletionFinishReason finishReason,
 				@JsonProperty("index") Integer index,
-				@JsonProperty("delta") ChatCompletionMessage delta) {
+				@JsonProperty("delta") Delta delta
+		) {
+			// The delta used in ChatCompletionChunk is not identical to ChatCompletion's ChatCompletionMessage, so a Delta record needs to be defined separately.
+			public record Delta(
```
tzolov, line 540: I'm afraid that this is incorrect. The message format is the same in both the [chat completion object](https://platform.openai.com/docs/api-reference/chat/object) and the [chat completion chunk object](https://platform.openai.com/docs/api-reference/chat/streaming) and it includes the tool_calls hierarchy as well.

spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -74,8 +75,8 @@ public OpenAiApi(String openAiToken) {
 	 * @param restClientBuilder RestClient builder.
 	 */
 	public OpenAiApi(String baseUrl, String openAiToken, RestClient.Builder restClientBuilder) {
-
-		this.objectMapper = new ObjectMapper();
+		// Use the same ObjectMapper for WebClient's response as the one used in RestClient.
+		this.objectMapper = Jackson2ObjectMapperBuilder.json().build();
```
tzolov, line 79: This is interesting, need to dig a bit more. 
The recent streaming response parsing failure was due to the unexpected (for me) appearance of the `LogProbs` in the chunk response.  Not sure if i missed it or OpenAI introduced it silently. Hope it is the former. 
I deliberately didn't configure the objectmapper to ignore unknown fields  so i can catch changes. But didn't expect they will come without announcement   and also didn't consider that the webclient and restclient converters are likely ignoring such fields.

tzolov (commented):
Thanks for the review @JM-Lab !
See my comments. I will pick the Predicate and have to check about the object converter.

spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
JM-Lab: I encountered an error during code review. It appears that LogProbs were added at some point. After checking OpenAI's API specification at https://platform.openai.com/docs/api-reference/chat/object, I noticed the presence of LogProbs. However, in the response for the ChatCompletion - Choice record when the stream is false, LogProbs are not included. To address this issue, I realized that a similar approach is needed for WebClient, similar to the ObjectMapper in RestClient, to ignore unknown fields. Therefore, I added the same configuration in the commit.

Instead of adding LogProbs to ChatCompletion - Choice and ChatCompletionChunk record, considering the continuous changes and potential additions or deprecations in OpenAI's spec, I believe it's more beneficial for developers using Spring AI to configure the ObjectMapper to ignore unknown fields. This way, they won't have to deal with too many specific options. Hence, I implemented the ObjectMapper configuration in the current commit.

spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
JM-Lab: I misunderstood. Thank you for letting me know.

spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
JM-Lab: The documentation at https://platform.openai.com/docs/api-reference/chat/create explicitly states that it ends with '[DONE]', so in my opinion, using 'equals' would be clearer than 'contains.

spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
tzolov: We already added the LogProbs in another commit. 
But with this PR I'll configure the object mapper to tolerate unknown properties. 
Would live out the Jackson2ObjectMapperBuilder for now.

tzolov:
Rebased, squashed and merged at b54bb8bc8a56a3a9e6fc5241b87c433fbadf25fc

PR #173, opened 2023-12-17
Refactor toString() method in Embedding class

Seol-JY:
This pull request updates the toString() method in the Embedding class.

The change replaces the usage of size() with isEmpty() when checking the embedding collection, making the code more concise and semantically accurate.


markpollack:
Thanks @Seol-JY !

PR #176, opened 2023-12-18
vertex ai aot hints, plus a fix to localize the dependency on spring web flux

joshlong:
hi i want to use vertex ai in a graalvm native image context so i crafted these aot hints to help

also: i localized the `spring-webflux` dependency, as its not used in most modules.

### DIFF
--- a/spring-ai-core/pom.xml
+++ b/spring-ai-core/pom.xml
@@ -40,10 +40,10 @@
 
 
 		<dependency>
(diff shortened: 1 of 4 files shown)


tzolov:
Thanks @joshlong 
Rebased and merged at: 95cbe54bae211da8e2aee0640ae6ff3e96bada1e

PR #178, opened 2023-12-18
Add NotNull Validation & Add Nullable Annotation

hgs-study:
While looking at the OpenAiGeneration metadata class and the OpenAiRateLimit class, we found a point of concern for NPE. 

- For the OpenAiGenerationMetadata.From() method, we verified this because the internal block uses id and usage.
- For OpenAiRateLimit classes, @Nullable Annotation has been added because null can be inserted into the long type from outside.

### DIFF
--- a/spring-ai-openai/src/main/java/org/springframework/ai/openai/metadata/OpenAiRateLimit.java
+++ b/spring-ai-openai/src/main/java/org/springframework/ai/openai/metadata/OpenAiRateLimit.java
@@ -50,8 +50,8 @@ public class OpenAiRateLimit implements RateLimit {
 
 	private final Duration tokensReset;
 
-	public OpenAiRateLimit(Long requestsLimit, Long requestsRemaining, Duration requestsReset, Long tokensLimit,
-			Long tokensRemaining, Duration tokensReset) {
+	public OpenAiRateLimit(@Nullable Long requestsLimit, @Nullable Long requestsRemaining, Duration requestsReset,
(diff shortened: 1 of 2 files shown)


markpollack:
For the OpenAiReateLimit class, these values should always come back from the openai request in the headers.

https://platform.openai.com/docs/guides/rate-limits/rate-limits-in-headers

I think it is incorrect to give a developer hints that it will be null, they should always be present.

markpollack:
Though I can see the fact that this may change beyond our control and using @nullable is a half way approach, but then perhaps we need to use Optional<> that makes for lots of verbosity.

spring-ai-openai/src/main/java/org/springframework/ai/openai/metadata/OpenAiRateLimit.java
```diff
@@ -33,20 +34,24 @@ public class OpenAiRateLimit implements RateLimit {
 
 	private static final String RATE_LIMIT_STRING = "{ @type: %1$s, requestsLimit: %2$s, requestsRemaining: %3$s, requestsReset: %4$s, tokensLimit: %5$s; tokensRemaining: %6$s; tokensReset: %7$s }";
 
+	@Nullable
```
markpollack, line 37: can this really be null?  at least in the case of openai it will always be returned I believe.

markpollack:
When we review the overall organization of response objects multiple AI models (not just openai) we will revisit this.

markpollack:
I closing this as these fields sent by openai won't be null.

PR #183, opened 2023-12-19
Refactor ContentFormatTransformer for Improved Readability

Seol-JY:
- Extract Method
- Replace explicit instanceof checks and casting with pattern matching in the apply method of ContentFormatTransformer.
- Change var to explicit type


markpollack:
merged as 698a531daed7d23b179b05e7edd8cbc993083eab

PR #203, opened 2023-12-29
Enable updating the pinecone vector store namespace

adambchouti:
When using pinecone, it would be nice to not have to create a new vectorStore instance for different namespaces. This PR is to simply:
- ~~add a setter to update the pinecone namespace used during requests~~
- Add additional functions that allow passing a custom namespace during add, search and delete


markpollack:
Thanks for this.  I've created this issue https://github.com/spring-projects/spring-ai/issues/209 to better understand the APIs of vector store providers.  Ideally I would like a mechanism to pass vector store provider options to each implementation in a generic way so that casting to a subtype of `VectorStore` is not required.  See the issue for more context.

adambchouti:
Makes perfect sense, gonna close this one. Thx a lot!

tzolov:
merged at 36278afdb31ab2a60f7374d5a69ac6c20706e1ae

PR #253, opened 2024-01-19
Use `ObjectMapper` with `FAIL_ON_UNKNOWN_PROPERTIES` false during error handling

making:
Some OpenAI compatible servers have unknown properties in the error response, so it is better to use an existing ObjectMapper in the ErrorHandler.


markpollack:
Good catch, I fixed this in another spot in the code base that was recently added as I cut-n-paste error from this code.

Thanks!

PR #270, opened 2024-01-25
Align the Embedding api with the new meta-model

tzolov:
- Craete new EmbeddingOptions -> ModelOptions, EmbeddigRequest -> ModelRequest, EmbeddingResponseMetadata -> ResponseMetadata and EmbeddignResultMetadata -> ResultMetadata.
 - Make the EmbeddigClient interface extend from ModelClient<EmbeddingRequest, EmbeddingResponse>, EmbeddingResponse implements ModelResponise and Embedding implements ModelResult.
 - Fix affected tests.
 - Streamline the EmbeddingClient interface with default method implementations based on call.


spring-ai-core/src/main/java/org/springframework/ai/embedding/EmbeddingResponseMetadata.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.ai.model.ResponseMetadata;
+
+/**
+ * @author Christian Tzolov
+ */
+public class EmbeddingResponseMetadata extends HashMap<String, Object> implements ResponseMetadata {
```
markpollack, line 27: maybe composition instead?  more for future proofing.  put this  list together of responses from the different providers.



OpenAI

metadata.put("model", model);
metadata.put("prompt-tokens", usage.promptTokens());
metadata.put("completion-tokens", usage.completionTokens());
metadata.put("total-tokens", usage.totalTokens());


Azure OpenAI


metadata.put("model", model);
metadata.put("prompt-tokens", embeddingsUsage.getPromptTokens());
metadata.put("total-tokens", embeddingsUsage.getTotalTokens());


PostgresML

Map.of(
"transformer", this.transformer,
"vector-type", this.vectorType.name(), 
"kwargs", this.kwargs));


Vertex

{
  "predictions": [
    {
      "embeddings": {
        "statistics": {
          "truncated": false,
          "token_count": 6
        },
        "values": [ ... ]
      }
    }
  ]
}

spring-ai-core/src/main/java/org/springframework/ai/embedding/EmbeddingResponseMetadata.java
```diff
@@ -0,0 +1,46 @@
...
+
+	public EmbeddingResponseMetadata(int initialCapacity) {
+		super(initialCapacity);
+	}
+
+	public EmbeddingResponseMetadata(int initialCapacity, float loadFactor) {
+		super(initialCapacity, loadFactor);
+	}
+
+	public EmbeddingResponseMetadata(Map<? extends String, ?> m) {
```
markpollack, line 42: Sonar complained about this....

However, the type parameter for the Map accepts any subtype of String which doesn't quite make sense since String is a final class and cannot have any subclasses. Hence, the ? extends keyword is unnecessary in this case and code would be cleaner and simpler as Map<String, ?>. The constructors of EmbeddingResponseMetadata could be like this:
```
public EmbeddingResponseMetadata(Map<String, ?> m) {
	super(m);
}
```

spring-ai-spring-boot-autoconfigure/src/test/java/org/springframework/ai/autoconfigure/azure/AzureOpenAiAutoConfigurationIT.java
```diff
@@ -113,11 +113,11 @@ void embedding() {
 
 			EmbeddingResponse embeddingResponse = embeddingClient
 				.embedForResponse(List.of("Hello World", "World is big and salvation is near"));
-			assertThat(embeddingResponse.getData()).hasSize(2);
-			assertThat(embeddingResponse.getData().get(0).getEmbedding()).isNotEmpty();
-			assertThat(embeddingResponse.getData().get(0).getIndex()).isEqualTo(0);
-			assertThat(embeddingResponse.getData().get(1).getEmbedding()).isNotEmpty();
-			assertThat(embeddingResponse.getData().get(1).getIndex()).isEqualTo(1);
+			assertThat(embeddingResponse.getResults()).hasSize(2);
```
markpollack, line 116: unrelated to the change, EmbeddingUtil should prob be abstract class to avoid instantiation.

tzolov: Ok, the EmbeddingUtil is merged within the AbstractEmbeddingClient now

spring-ai-core/src/main/java/org/springframework/ai/embedding/EmbeddingResponseMetadata.java
tzolov: resolved

markpollack:
merged in 5d55b68380b0c5468e4b605df5bf418f57e78275

Issue #113, opened 2023-11-18
Avoid duplicated entries in VectorStore(s) by allowing generation of Document ID based on the hashed document content.


tzolov:
Currently the `Document` if not provided with an explicit `ID`, generates a random UUID for every document. 
Even if the document content/metadata haven't changed a new ID is generated every time. 
This will lead to document content duplications in the Vector store. 

To prevent this type of unnecessary duplications we can allow generation of Document ID based on the hashed document content+metadata.

Following snippet is inspired by a langchain4j vector store implementations.

```java
import java.nio.charset.StandardCharsets;
import java.security.MessageDigest;
import java.security.NoSuchAlgorithmException;

....

public static String generateIdFrom(String contentWithMetadata) {
    try {
	    byte[] hashBytes = MessageDigest.getInstance("SHA-256").digest(contentWithMetadata.getBytes(StandardCharsets.UTF_8));
	    StringBuilder sb = new StringBuilder();
	    for (byte b : hashBytes) {
		    sb.append(String.format("%02x", b));
	    }
	    return UUID.nameUUIDFromBytes(sb.toString().getBytes(StandardCharsets.UTF_8)).toString();
    }
    catch (NoSuchAlgorithmException e) {
	    throw new IllegalArgumentException(e);
    }
}
```

markpollack:
Currenlty we have

```
	public Document(String content, Map<String, Object> metadata) {
		this(UUID.randomUUID().toString(), content, metadata);
	}
```

Perhaps we can add a strategy interface as an option to pass in with an implementation based on what is listed above.
```
	public Document(String content, Map<String, Object> metadata, IdGenerator idGenerator) {

...
```

nurlicht:
I have just created a PR for this feature (my first PR here):
https://github.com/spring-projects/spring-ai/pull/272

PR #272, opened 2024-01-25
Provide IdGenerator with multiple implementations (#113)

nurlicht:


nurlicht:
The test 'BeanOutputParserTest$FormatTest' fails locally. This behavior also exists on the 'main' branch.

tzolov:
Thank your for the contribution @nurlicht 

Is there are good reason to implement the Id generator for every hash codes algorithm? 
Wouldn't a simple MessageDigest impl. like in the snipped bellow be good enough? 
After all, the hash-based id is an heuristic approach, no 100% guaranties. 

```java
import java.security.MessageDigest;
import java.security.NoSuchAlgorithmException;

....

public static String generateIdFrom(String contentWithMetadata) {
    try {
	    byte[] hashBytes = MessageDigest.getInstance("SHA-256").digest(contentWithMetadata.getBytes(StandardCharsets.UTF_8));
	    StringBuilder sb = new StringBuilder();
	    for (byte b : hashBytes) {
		    sb.append(String.format("%02x", b));
	    }
	    return UUID.nameUUIDFromBytes(sb.toString().getBytes(StandardCharsets.UTF_8)).toString();
    }
    catch (NoSuchAlgorithmException e) {
	    throw new IllegalArgumentException(e);
    }
}
```

nurlicht:
Hi @tzolov 
It seems that the class MessageDigest is not thread-safe, whereas the Apache implementations are. As such, I tried to implement multiple implementations of an IdGenerator (the current random one, the proposed one, and seemingly thread-safe versions of the proposed one).

The code as implemented in this PR uses your proposed IdGenerator (as default), unless a client calls the new constructor of Document with a specific instance of IdGenerator.

With that said, if you still prefer to make it simpler (and not to use the Apache algorithms), feel free to change the code or let me know to do it.

markpollack:
I like the thoughtfulness with respect to thread safety but am a bit hesitant to bring in a third party library just for this functionalty *and* to offer the user multiple choices of digest options, it can be overwhelming.  Perhaps a simpler [wrapper approach](https://github.com/Glusk/caesar/blob/master/src/main/java/com.github.glusk.caesar/com/github/glusk/caesar/hashing/ImmutableMessageDigest.java) such as done here would be sufficient.  Or even a simpler approach of creating a new one every time it is needed since it is apparently not a performance penalty to do this - see this [SO post](https://stackoverflow.com/questions/17554998/need-thread-safe-messagedigest-in-java).

nurlicht:
Thanks for the feedback.
With the new commit:
- The Apache dependency has been removed.
- MessageDigest is cloned each time it is used.
- The interface for the client remains as in the old code, but the new IdGenerator is used.
- The code for the new IdGenerator is essentially the one provided in the definition of the issue #113, aside from cloning and some minor refactoring. 
- Only if the user wants to use another IdGenerator, the new cnstructor can be used (with the type of IdGenerator as an additional argument).
- The abstraction for different IdGenerator has been preserved.

Please let me know if you find issues or rooms for improvement. Thanks.

tzolov:
Thank you @nurlicht, I will have a look shortly. 

I wonder if for the sake of the cautiousness,  set the default to RANDOM (e.g. to mimic the current behaviour) and switch to hash based one later? 

Also (not related to this PR) I wish we had a better (more centralised) approach to configure stuff like the Id generator or  Document's contentFormatter.  I guess we need to discuss this after the 0.8.0 release.

tzolov:
Also @nurlicht what name should i put in the javadoc `@author`? nurlicht?

nurlicht:
@tzolov,
I changed the default IdGenerator to be the existing one (RANDOM).
@author = Aliakbar Jafarpour

tzolov:
Refactored, rebased and merged at b0d8d2b9cef37b4af988990fdc086bbae32dc2cd

@nurlicht, I took the deliberate to refactored the PR a bit according to our current needs.  
Please have a look and let me know if I have perhaps oversimplified or missed something. 

Issue #228, opened 2024-01-07
Add missing options in for OpenAI/Azure OpenAI and align autoconfiguration styles


markpollack:
The prefix for these properties should be `spring.ai.openai.chat`, currently it is `spring.ai.openai`

`OpenAIProperties` has options only for `temperature` and is missing options

* frequency_penalty
* logit_bias
* logprobs
* top_logprobs
* max_tokens
* n
* presence_penalty
* response_format 
* seed (beta)
* stop
* top_p
* tools
* tool_choice
* user

Also, compare/contrast the autoconfiguration of OpenAI and  Azure OpenAI to make sure they following a consistent approach.

markpollack:
This was fixed in 5b4784fc5673cdc82f2c121819cffa1df460d9da

PR #275, opened 2024-01-27
Add OpenAi Chat Options

tzolov:
- Add OpenAiChatOptions that implements ChatOptions and exposes all OpenAi request options, except messages and stream.
 - Add OpenAiChatOptions field (as defaultOptions) to OpenAiChatClient. Implement star-up/runtime options merging on chat request creation - Add OpenAiChatOptions options field to OpenAiChatProperties. Later is set as OpenAiChatClient#defaultOptions. Use the `spring.ai.openai.chat.options.*` prefix to set the options.
 - Add tests for properties and options merging.

 Part of #228


tzolov:
Replaced by https://github.com/spring-projects/spring-ai/pull/277

Issue #279, opened 2024-01-28
BeanOutputParserTest fails on Windows due to platform specific line endings


kirklund:
The template in BeanOutputParser contains \n line endings, the generated schema contains \r\n line endings, and the assertions in BeanOutputParserTest contain \n line endings.

This causes BeanOutputParserTest to fail when run on Windows 11 either as part of the build or by itself.

Since I'm working on both MacOS and Windows 11 I figured I'd file an Issue and submit a PR to fix it.

PR #280, opened 2024-01-28
#279: Normalize Line Endings to LF in BeanOutputParser

kirklund:
This commit modifies BeanOutputParser to utilize the newly introduced LineEndingsNormalizer utility, ensuring all platform-specific line endings are normalized to LF (Unix-style). The LineEndingsNormalizer is designed to perform a no-op if the platform's default line ending matches the specified target line ending.

This change addresses and fixes BeanOutputParserTest failures observed on the Windows platform by ensuring consistent line ending normalization across different operating systems.

---

Note: There are failures in spring-ai-weaviate-store on Windows as well but I'd prefer to submit that as its own PR since its a different sub-module. I think it's better to keep this PR focused.

* Sign the [contributor license agreement](https://cla.pivotal.io/sign/spring) [done]
* Rebase your changes on the latest `main` branch and squash your commits [done]
* Add/Update unit tests as needed [done]
* Run a build and make sure all tests pass prior to submission [done: mac and windows]

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/util/StandardLineEnding.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/util/StandardLineEnding.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.ai.util;
+
+/**
+ * Enum representing standard line endings used across different platforms. This enum
+ * provides a standardized way to reference common line separator characters such as
+ * Carriage Return (CR), Line Feed (LF), and the combination of both (CRLF).
+ */
+public enum StandardLineEnding {
+
+	/**
+	 * Carriage Return (CR) character, commonly used as a line ending in Mac OS systems up
+	 * to version 9, and in some network protocols.
+	 */
+	CR("\r"),
+
+	/**
+	 * Line Feed (LF) character, used as a line ending in Unix and Unix-like systems
+	 * including Linux and macOS X, and in many network protocols.
+	 */
+	LF("\n"),
+
+	/**
+	 * Carriage Return (CR) followed by Line Feed (LF), used as a line ending in Windows
+	 * operating systems, and in some network protocols.
+	 */
+	CRLF("\r\n");
+
+	private final String lineSeparator;
+
+	StandardLineEnding(String lineSeparator) {
+		this.lineSeparator = lineSeparator;
+	}
+
+	/**
+	 * Returns the line separator string associated with the line ending.
+	 * @return The line separator string.
+	 */
+	public String lineSeparator() {
+		return lineSeparator;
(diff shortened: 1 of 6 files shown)


spring-ai-core/src/test/java/org/springframework/ai/parser/BeanOutputParserTest.java
```diff
@@ -100,6 +122,30 @@ public void shouldReturnFormatContainingJsonSchemaIncludingPropertyAndPropertyDe
 					""");
 		}
 
+		@Test
+		public void usesLineEndingsNormalizer() {
```
jxblum, line 126: Test case methods in JUnit 5 can be package-private now (recommended).

spring-ai-core/src/test/java/org/springframework/ai/util/LineEndingsNormalizerTest.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.springframework.ai.util.StandardLineEnding.LF;
+
+/**
+ * @author Kirk Lund
+ */
+@ExtendWith(MockitoExtension.class)
+class LineEndingsNormalizerTest {
+
+	@Test
+	public void requiresNormalization_forLineEndingCR_returnsTrue_onWindows() {
```
jxblum, line 36: With JUnit 5, please make test case methods package-private.

spring-ai-core/src/test/java/org/springframework/ai/util/StandardLineEndingTest.java
```diff
@@ -0,0 +1,61 @@
...
+import static org.springframework.ai.util.StandardLineEnding.LF;
+import static org.springframework.ai.util.StandardLineEnding.findByLineSeparator;
+
+/**
+ * @author Kirk Lund
+ */
+class StandardLineEndingTest {
+
+	@Test
+	public void lineSeparator_returnsSlashR_forCR() {
```
jxblum, line 32: With JUnit 5, please make test case methods package-private.

NOTE: Not sure if this has been consistently done across the Spring AI project/codebase, if not, then I would suggest this be cleaned up in the Spring AI project in a separate PR.

spring-ai-core/src/test/java/org/springframework/ai/parser/BeanOutputParserTest.java
kirklund: Thanks John! I'll change them. I originally had it public just to match the other tests in BeanOutputParserTest.

spring-ai-core/src/main/java/org/springframework/ai/util/StandardLineEnding.java
```diff
@@ -0,0 +1,84 @@
...
+ * limitations under the License.
+ */
+package org.springframework.ai.util;
+
+/**
+ * Enum representing standard line endings used across different platforms. This enum
+ * provides a standardized way to reference common line separator characters such as
+ * Carriage Return (CR), Line Feed (LF), and the combination of both (CRLF).
+ */
+public enum StandardLineEnding {
```
jxblum, line 23: Very nice!

spring-ai-core/src/test/java/org/springframework/ai/util/StandardLineEndingTest.java
kirklund: Good idea! Looks like they're all public across most of the tests in Spring AI.

jxblum (commented):
All in all, this looks clean and intuitive.

markpollack:
I found a simpler way to do this, so modified your PR.  I think it is probably a false negative on windows anyway, since the text that would get sent over the wire would generate the correct response from the AI model (at least for chatgpt).  The line endings in in the string literal are causing this.  In any case, the behavior is now to only emit `\n` in the json schema string.

Thanks for bringing it up and the PR.

fixed in 1c222a695c7bec9e853dd213ce35c866bda16806

PR #287, opened 2024-01-30
Documentation fixes, clean and polish

mkheck:
Unsure exactly why, but I missed several changes with my PR from a couple weeks back. More docs clean and polish, no code changes with this one.


markpollack:
Hi, we changed to use `ChatClient`, so it is outdated.  Maybe some mix up.  Anyway, thanks for the proof-reading, there will certainly be other things you can catch.

mkheck:
Hi Mark,

From what I can see, AiClient is still in some docs...thus the resubmit with changes from AiClient to ChatClient. Here's what I see in my PR:

PRAiClientToChatClient

Haven't had my coffee yet this morning, so I may be looking at something weird. 😅

Anyway, moving from docs to code already. More soon!

PR #299, opened 2024-02-06
OpenAI ChatClient tools support

tzolov:
Add support for ChatClient function calling.
 - Add generic (model) `ToolFunctionCallback` and `AbstractToolFunctionCallback`  abstractions for defining and registering function callbacks with the ChatClient.
 - OpenAiChatClient Function calling implementation.
 - Unit/IT tests for OpenAiApi and OpenAiChatClient function calling.
 - Add function calling docs.


spring-ai-core/src/main/java/org/springframework/ai/chat/ChatOptions.java
```diff
@@ -35,4 +38,12 @@ public interface ChatOptions extends ModelOptions {
 
 	void setTopK(Integer topK);
 
+	default List<ToolFunctionCallback> getToolCallbacks() {
+		throw new UnsupportedOperationException("ToolCallbacks is not supported by this model");
```
markpollack, line 42: I think we should not have this default method, since the methods here should apply to every model.

spring-ai-core/src/main/java/org/springframework/ai/chat/ChatOptions.java
```diff
@@ -35,4 +38,12 @@ public interface ChatOptions extends ModelOptions {
...
+	}
+
+	default void setToolCallbacks(List<ToolFunctionCallback> toolCallbacks) {
```
markpollack, line 45: same comment as before

spring-ai-core/src/main/java/org/springframework/ai/chat/ChatResponse.java
```diff
@@ -82,4 +82,9 @@ public ChatResponseMetadata getMetadata() {
 		return this.chatResponseMetadata;
 	}
 
+	@Override
+	public String toString() {
```
markpollack, line 86: added equals and hashcode on merge.

markpollack:
removed the default methods in the chat option interface to make tests pass.  

Merged as ffd8222c209967552ea32a265837a48e72076b13   

Can review the small comments in this PR afterwards

Issue #149, opened 2023-12-06
Support for Speech-to-Text via OpenAI audio transcriptions


thomasdarimont:
**Expected Behavior**

It should be possible to send an audio sample with speech (in supported audio formats) to the OpenAI transcription endpoint and retrieve the extracted speech as text.

**Current Behavior**

Speech to text is currently not supported by the OpenAI integration and users must fallback to custom implementations.
See: https://github.com/thomasdarimont/quadropole-welcome-hero/blob/main/src/main/java/com/welcomehero/app/openai/OpenAiFacade.java#L34

**Context**

The motivating use case for this was to build a AI augmented information kiosk for hospitals to enable visitors, patients and non-native staff to ask questions with natural speech about the hospital environment, based on a controlled knowledge-base.
See: https://github.com/thomasdarimont/quadropole-welcome-hero

Greetings to the team :)

markpollack:
Hi thanks, glad our paths cross again.  There is a long list of every model types, let alone whatever happens with multi-modal support within a single model, that we should eventually cover.  Audio, Video, Voice, Code (perhaps something more specific than text), Images...   I plan to put in a sort of 'sandbox' module for people to make contributions of clients for these various models and see what abstractions and commonalities come out of it.   You sample code looks like it can be extracted into a `VoiceClient` or `SpeachClient` class.  Once I have this area setup, perhaps you can extract the essence of what did and make a PR?

ENate:
Hi. I will like to contribute to this project (spring AI) and look forward to _a good first issue_. Greetings and great work from the team.

michaellavelle:
Hi - I've created a PR to add support for audio transcriptions here :  https://github.com/spring-projects/spring-ai/pull/300

markpollack:
Closing as it as been merged. Docs will follow shortly.

PR #300, opened 2024-02-06
Adding support for transcriptions ( and OpenAI Whisper model)

michaellavelle:


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiTranscriptionClient.java
```diff
@@ -0,0 +1,140 @@
...
+			var transcription = transcriptionEntity.getBody();
+
+			if (transcription == null) {
+				logger.warn("No transcription returned for request: {}", audioResource);
+				return new TranscriptionResponse(null);
+			}
+
+			Transcript transcript = new Transcript(transcription.text());
+
+			RateLimit rateLimits = OpenAiResponseHeaderExtractor.extractAiResponseHeaders(transcriptionEntity);
```
michaellavelle, line 92: I'm re-using these RateLimit and Usage classes from the chat packages here, but given that these are now used outside of chat, should they be moved into a more generic package ?

michaellavelle:
Please see https://github.com/spring-projects/spring-ai/issues/149

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiTranscriptionClient.java
markpollack: I am not yet sure, we need to review this in the 0.9.0 timeline.  I suspect there are cross cutting concerns.  Now we are wrapping up the other direction, input options, to support a tiered structure of portable and model specific options.  The same analysis and likely same approach needs to apply for the metadata info coming out of the response.

markpollack:
Thanks for the contribution, this looks like a very strong PR on first glance!

hemeda3:
Thank you.
Your PR inspired me to do the Text To Audio (Speech API).

[Added support for OpenAI Text to Audio (Speech API )](https://github.com/spring-projects/spring-ai/pull/317/files#top)
#317

tzolov:
Thank you @michaellavelle 
I've replaced the lower-level client and removed the code from the spring-ai-core.  We will need at least few audio and transcription client implementations before we can generalise. 
Reworked, rebased, squashed and merged at 7d04167833a36de3a639d616aa996cde71e29cba

PR #313, opened 2024-02-10
Remove slash according to api

sunshiningsoo:
### Remove slash according to api

- According to `api.yaml`, `/api/v1/collections` path doesn't need slash so remove it


tzolov:
@sunshiningsoo is this breaking anything?

sunshiningsoo:
@tzolov not at all, chroma ApiRouter treats url with trailing "/" same as without. but docs only include without trailing "/"

tzolov:
@sunshiningsoo,  after the change did you run successfully all Cohere tests, including the integration tests and auto-config integration tests?

sunshiningsoo:
@tzolov yes, I've done all test include readme build side.

tzolov:
Great! thanks for the contribution @sunshiningsoo

PR #319, opened 2024-02-12
Support filter expressions in Neo4j vector store.

meistermeier:
I am not sure about the whole `StringBuilder` things there.
Also, I needed to take care of moving the `NOT` in front of the clauses. The idea of inverting the logics if using `NOT` in the abstract converter is nice but did not really match the, in my opinion, pragmatic prefix in Neo4j's Cypher with the negation.

Stole the tests from PgVector, again :)

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/filter/converter/Neo4jVectorFilterExpressionConverter.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/filter/converter/Neo4jVectorFilterExpressionConverter.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2023-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


tzolov:
This is great! Thanks @meistermeier!

> I am not sure about the whole StringBuilder things there.
Not sure what you mean? 

> Also, I needed to take care of moving the NOT in front of the clauses. The idea of inverting the logics if using NOT in the abstract converter is nice but did not really match the, in my opinion, pragmatic prefix in Neo4j's Cypher with the negation.

The inverter NOT logic provided in the abstract converter is meant to be used only for stores that does not provide  "native"  NOT support. The idea is to push down (to store's native primitives) as much of the predicates logic as possible.

tzolov:
@meistermeier , I've moved the filters from core to the project.  Something I will do for most of the other stores filters. (Only the Pinecone and Milvus have some awkward interdependencies and would likely stay in core). 

Also I've updated the Neo4j docs and while doing it realised that we might be able to optimise the auto-configuration implementation. 
Will create a separate ticket and will shamelessly assign it to you ;)

Thank you again!

Rebased and merged at: 408337e327608b328a987ef716a172c21c6f724c

PR #320, opened 2024-02-12
Configure HTTP header spring-ai user agent for reporting

mkheck:
Configured HTTP header user agent to reflect calls from Spring AI for reporting when using imperative code. Reactive code will require changes from Azure side, which I'm working to accomplish soon.

### DIFF
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java
@@ -18,8 +18,6 @@ package org.springframework.ai.azure.openai;
 
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


tzolov:
Hey @mkheck, thank for the extension. 

I've started reviewing it, perhaps I will add some additional configurations to the AzureOpenAiEmbeddingOptions. 

But this caused my  git brain to short-circuit ;)
Screenshot 2024-02-14 at 10 12 54

Could you please rebase the PR on top of the upstream/main?

mkheck:
Whaaaaa???

Absolutely. However...I've discovered a better way: updating the starter to add that identifier to upon client creation instead of within each call. This is a vastly better solution, as it leaves intact other useful info in the user agent that I've since discovered while providing visibility (and credit) for Spring-initiated Azure OpenAI calls. Win/win.

Closing this now, will submit a new, clean PR shortly. Pls ping me if you have any questions, and thank you, @tzolov!

PR #323, opened 2024-02-13
Move native hints into individual modules

markpollack:
* Use aot.factories approach for registration
* Add tests


markpollack:
merged in 596f2b06c06f6c0fa8dfd3da5a2d92ca3cfcee67

PR #330, opened 2024-02-14
Remove driver config and adjust documentation

meistermeier:
There is no need for the driver auto-config in SpringAI, it's already taken care of in Spring Boot.


tzolov:
This was quick! Thanks @meistermeier 

I wonder should we add the `spring-boot-starter-data-neo4j` dependency to the https://github.com/spring-projects/spring-ai/blob/main/spring-ai-spring-boot-starters/spring-ai-starter-neo4j-store/pom.xml ?

meistermeier:
This won't make any sense. This is for Spring Data Neo4j and the module for the `Neo4jVectorStore` does not require the object mapping. Our intention here was to keep it as simple as possible and only depend on the Neo4j Java driver for the communication.
The auto-config of the Neo4j driver is done already in Spring Boot (auto-configure https://github.com/spring-projects/spring-boot/blob/2abc25d51cbb672ef25bdb8141bb5fef25088404/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jAutoConfiguration.java#L63) and does not require any additional dependency.

tzolov:
perfect!

tzolov:
Thanks @meistermeier !
Rebased, squashed and merged at 10438c55a3da744a2d59d57d6c7c0f4e79749357

PR #333, opened 2024-02-14
Added support for OpenAI's content policy moderation API

hemeda3:
Please review

Given a input text, outputs if the model classifies it as violating OpenAI's content policy.
Related guide: [Moderations](https://platform.openai.com/docs/guides/moderation)

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiModerationClient.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiModerationClient.java
@@ -1,0 +1,182 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.openai;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.ai.model.ModelOptionsUtils;
+import org.springframework.ai.moderation.*;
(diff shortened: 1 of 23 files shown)


tzolov:
Hi @hemeda3 , thank you for your contribution. 

The safety moderation is something we've been considering for a while but still have not tackled. 

As Spring AI supports multiple LLMs providers and their APIs it important for as to come with a design that "generalise" and "unifies" (as much as possible) features such as the `safety moderation`. 
(btw, the code in the spring-ai-core/model package is meant to be generic for all models)

For example most LLM provides offer some sort of safety policies: 

Google's Vertex Gemini and PaLM2 has this:
- https://cloud.google.com/vertex-ai/docs/generative-ai/multimodal/configure-safety-attributes
- https://cloud.google.com/vertex-ai/docs/generative-ai/multimodal/configure-safety-attributes
- https://cloud.google.com/vertex-ai/docs/generative-ai/learn/responsible-ai

Microsoft Azure OpenAi: 
- https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cpython

Amazon Bedrok Guardrails
- https://aws.amazon.com/bedrock/guardrails/

OpenAI
- https://platform.openai.com/docs/guides/moderation
.....

hemeda3:
@tzolov  thanks for taking the time to check my PR.
I have tried to follow OpenAiImageApi  was good example, almost copy paste the structure file names  except metdata i left it  I think  

https://docs.spring.io/spring-ai/reference/api/index.html#generic-model-api
image

but seems you're  talking about new/ different approach I guess something more generic than  the one mentioned in this image, something  maybeshared  accross models which I believe will not diff from this structure  https://docs.spring.io/spring-ai/reference/api/index.html#generic-model-api 

 if I understood your comment correctly then I think we can close this PR  since it's too early until the design/solution that include LLMS you mentioned in your comment

tzolov:
@hemeda3, I'm truly impressed with your effort and understanding of our somewhat poorly documented concepts. 
What i'm saying is that we still have not clear understanding how we are going to extend the generic model (you've shown above) to model the safety moderations in way we can implement it of the various providers. 
We are just asking for bit more patience until we get there.

markpollack:
Hi.  I understand the concern of putting classes in the generic model package before having a second impl, but I figure we can adjust as we go considering this is a very clean PR.  Thanks @hemeda3 !  very much appreciated.

Added docs and merged into current naming conventions and use of new approach for retry/errorhandler since this PR was authored.

merged in 189468127c234ced9256f951023fc8fb29b08da2

PR #339, opened 2024-02-17
updated native hints

joshlong:
hi 

this teases out the native hints so that , for example, there's no attempt to load classes for vertex when openai is on the classpath. 

also, i expanded the algorithm to discover more things to be registered for reflection. now, the code looks for types annotated with `@JsonInclude` and types with a use of `@JsonProperty`


markpollack:
functionality provided in 596f2b06c06f6c0fa8dfd3da5a2d92ca3cfcee67

PR #364, opened 2024-02-26
Add activation/deactivation auto-config property to all Model clients

### LINKED ISSUE #363: Allow disabling all Spring AI auto-configurations using declarative properties
Currently if you have multiple Spring AI spring boot starters on the classpath, like Vertex AI PaLM2 and OpenAI it is not easy to control which part are loaded. 
For example you might want to use the OpenAI embedding client with Vertex AI PaLM2 Chat Client. 
The only why to disable the undesired auto-configurations (e.g. Vertex AI Embedding and OpenAI Chat clients) is to use the the configuration annotation `exclude` parameter or the `spring.autoconfigure.exclude`. 
But this is not covering all cases as some auto-cnfiguraitons share the same java file. For example the VertexAiAutoConfiguration defines both the VertexAiChatClient and the VertexAiEmbeddingClient clients. 

A better tailored solution would be to introduce an explicit property for each auto-configured component (e.g. clients, vector stores ...) that allow disabling the activation of this component. 
Those properties will be enabled by default (except for Bedrock - see below) and will allow selective disabling. 

NOTE: All Bedrock chat/embedding clients already have an `*.enabled` property that is set to false by default. The reason is that all Bedrock auto-configurations are sharing the same jar.

tzolov:
Property is of format `*.enabled=` and is true by default.
 Note: the Bedrock clients are disabled by default.

 - Add deactivation support for Vertex AI PaLM2 clients
 - Add deactivation support for Azure OpenAI clients
 - Add deactivation support for huggingface chat client
 - Add deactivation support for Ollama clients
 - Add deactivation support for OpenAI clients
 - Add deactivation support for Stability Image and PostgresML clients
 - Add deactivation support for ONNX Transformers Embedding client
 
 It doesn't seems necessary to provide similar functionality to the VectorStore auto-configurations. 
 Later are less likely to be mixed together and further more can use the existing `exclude` property.
 
 Resolves #363

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/clients/azure-openai-chat.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/clients/azure-openai-chat.adoc
@@ -94,6 +94,7 @@ The prefix `spring.ai.azure.openai.chat` is the property prefix that configures 
 |====
 | Property | Description | Default
 
+| spring.ai.azure.openai.chat.enabled | Enable Azure OpenAI chat client.  | true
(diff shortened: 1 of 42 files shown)


tzolov:
merged at 4e270ac75af055aa1e2f2de3f6a9fd80835b8af1

PR #373, opened 2024-02-29
Fix Mistral AI Chat Properties

ricken07:
- Fix config prefix 
- Add test methods


tzolov:
Thank you @ricken07 , but i've applied those changes in the previous PR: 31d6194d9debb079ee3c3e30ee4d15abfa7c4db1

tzolov:
Btw @ricken07 I also added support for functional calling :) 
Mistral has not published their new api yet, but I noticed that it almost identical to the OpneAiApi so I managed to reverse engineered what is needed.
I've added a few ITs (in the model and the auto-confs) that test the new functionality even managed to add one that uses the OpenAI API to talk to Mistral AI and perform functional calling: https://github.com/spring-projects/spring-ai/blob/main/spring-ai-spring-boot-autoconfigure/src/test/java/org/springframework/ai/autoconfigure/mistralai/tool/PaymentStatusBeanOpenAiIT.java :)

It seems though that their new API is still not stable (they just made a breaking change). Also unlike OpenAI/Azure/Gemini they do not support parallel function calling (yet). 
P.S. If you want you can reach me on twitter (@christzolov) or linked in (@tzolov) so we can keep in touch.

PR #382, opened 2024-03-02
Docs: Update Mistral AI Chat documentations

ricken07:
- Add Chat properties options
- Fix spring.ai.mistralai.chat.options.model descritption
- Add Mistral AI API Samples


tzolov:
Great, thank you @ricken07 !

PR #398, opened 2024-03-05
Update OpenAiApi.java

youngmoneee:
While exploring the OpenAiApi class, I identified some areas for improvement within the ChatModel and ChatCompletionFinishReason enums and proceeded with the following adjustments:

- I updated the version in ChatModel from gpt-3.5-turbo-1106 to gpt-3.5-turbo-0125.
- I corrected a typo in ChatCompletionFinishReason, changing TOOL_CAL to TOOL_CALL.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -151,12 +151,14 @@ public class OpenAiApi {
 		GPT_3_5_TURBO("gpt-3.5-turbo"),
 
 		/**
-		 * GPT-3.5 Turbo model with improved instruction following,
-		 * JSON mode, reproducible outputs, parallel function calling,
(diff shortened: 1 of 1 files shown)


markpollack:
based on this comment from https://openai.com/blog/new-embedding-models-and-api-updates

````
Customers using the unpinned gpt-3.5-turbo model alias will be automatically upgraded from gpt-3.5-turbo-0613 to gpt-3.5-turbo-0125 two weeks after this model launches.
```

it seems like we should set the default to `gpt-3.5-turbo` this way we are always on the latest version?

youngmoneee:
> I think that since the `gpt-3.5-turbo` will always point to the "unpinned" gpt-3.5-turbo model. However, users may still want to explicitly specify by name the model they want to use, even if at some point in time it matches the model that is the alias for gpt-3.5-turbo model. I'll update the PR to add the new version instead of replace the older one.

I've made modifications to keep the existing version without removing it

markpollack:
merged in 7396597678184a1a555cf1724fb65bb354e35152  Thanks!

PR #402, opened 2024-03-05
Added spring-ai-gemfire vector store

geetrawat:


vector-stores/spring-ai-gemfire/src/main/java/org/springframework/ai/vectorstore/GemFireVectorStore.java
```diff
@@ -0,0 +1,505 @@
...
+
+		@JsonProperty("max-connections")
+		private int maxConnections = 16;
+
+		@JsonProperty("vector-similarity-function")
+		private String vectorSimilarityFunction = "COSINE";
+
+		private String[] fields = new String[] { "vector" };
+
+		private int buckets = 0;
```
jhuynh1, line 197: buckets, fields and name should be configurable?

vector-stores/spring-ai-gemfire/src/main/java/org/springframework/ai/vectorstore/GemFireVectorStore.java
```diff
@@ -0,0 +1,505 @@
...
+		try {
+			String embeddingString = objectMapper.writeValueAsString(upload);
+			embeddingsJson = embeddingString.substring("{\"embeddings\":".length());
+		}
+		catch (JsonProcessingException e) {
+			throw new RuntimeException("Embedding json parsing error :" + e.getMessage());
+		}
+
+		client.post()
+			.uri("/" + INDEX_NAME + EMBEDDINGS)
```
jhuynh1, line 411: INDEX_NAME is not configurable from the rest api calls. This should probably take any index name instead of the final string (constant) one?

The question is also for the remaining api calls that use INDEX_NAME instead of a configured index name.

vector-stores/spring-ai-gemfire/src/main/java/org/springframework/ai/vectorstore/GemFireVectorStore.java
```diff
@@ -0,0 +1,505 @@
...
+
+	}
+
+	private static final int DEFAULT_PORT = 9090;
+
+	private static final int DEFAULT_TOP_K_PER_BUCKET = 10;
+
+	private static final String DEFAULT_DOCUMENT_FIELD = "document";
+
+	public static final String INDEX_NAME = "spring-ai-index";
```
jhuynh1, line 169: Is this a default value?  Or is this the only name we accept and only expect to create 1 index?

vector-stores/spring-ai-gemfire/src/main/java/org/springframework/ai/vectorstore/GemFireVectorStore.java
```diff
@@ -0,0 +1,505 @@
...
+				return new GemFireVectorStoreConfig(this);
+			}
+
+		}
+
+	}
+
+	private static final int DEFAULT_PORT = 9090;
+
+	private static final int DEFAULT_TOP_K_PER_BUCKET = 10;
```
jhuynh1, line 165: should there be a TOP_K also?

vector-stores/spring-ai-gemfire/src/main/java/org/springframework/ai/vectorstore/GemFireVectorStore.java
geetrawat: INDEX_NAME is not the default value but DEFAULT_DOCUMENT_FIELD this is a default value.

vector-stores/spring-ai-gemfire/src/main/java/org/springframework/ai/vectorstore/GemFireVectorStore.java
```diff
@@ -70,20 +74,22 @@ public static final class GemFireVectorStoreConfig {
...
 		private final String documentField;
 
 		public static Builder builder() {
 			return new Builder();
 		}
 
 		private GemFireVectorStoreConfig(Builder builder) {
-			String base = UriComponentsBuilder
-				.fromUriString("http{ssl}://{host}:{port}/gemfire-vectordb/v1/indexes/{index}")
+			String base = UriComponentsBuilder.fromUriString(DEFAULT_URI)
```
jhuynh1, line 86: The default uri isn't a 1:1 replacement.  Is that intentional?
It is now:         http{ssl}://{host}:{port}/gemfire-vectordb/v1/indexes
It used to be: http{ssl}://{host}:{port}/gemfire-vectordb/v1/indexes/{index}

Will the builder.index be used in this case?  it looks like it would have been the replacement for {index} but that no longer exists?

geetrawat: corrected it

tzolov:
Thanks for your contribution @geetrawat ! 
We are busy with the 0.8.1 milestone release at the moment. 
I hope next week we will have time to review this an other pending PR.

spring-ai-spring-boot-autoconfigure/pom.xml
```diff
@@ -308,8 +308,13 @@
 			<version>1.5.0</version>
 			<scope>test</scope>
 		</dependency>
-
-
-	</dependencies>
+        <dependency>
```
tzolov, line 311: Please add the `spring-ai-gemfire` (and the related boot starter) dependecies to the [Spring AI BOM](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-bom/pom.xml). 
Then you won't need to specify the version here. 

Also adding this dependency here is only useful if you add a geimfire vector store auto-configuration under the : https://github.com/spring-projects/spring-ai/tree/main/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore
There you can define the dedicated properties and so one. I guess you can reuse the existing Spring Boot Gemfire auto-configureations for connecting to the server.  Check how Neo4J is doing this. 

And finally once you have the auto-configuraion you can add a boot-starter under the: https://github.com/spring-projects/spring-ai/tree/main/spring-ai-spring-boot-starters

vector-stores/spring-ai-gemfire/src/main/java/org/springframework/ai/vectorstore/GemFireVectorStore.java
```diff
@@ -0,0 +1,531 @@
...
+		}
+		catch (Exception e) {
+			logger.warn("Error removing embedding: " + e);
+			return Optional.of(false);
+		}
+		return Optional.of(true);
+	}
+
+	@Override
+	public List<Document> similaritySearch(SearchRequest request) {
```
tzolov, line 464: The SearchRequest allows metadata filtering using [filter expressions](https://docs.spring.io/spring-ai/reference/api/vectordbs.html#metadata-filters) (DSL and SQL-like text). 
All Spring AI Vector Store implementations so far have opted to provide support for this feature. 
I remember that Gemfire can have a powerful predicate filtering support so It seems possible to map the Spring AI Filter expressions to the existin Gemifre one. (Just guessing). 
But if you decide not to support it, then here you have to check if `SearchRequest.hasFilterExpression()` is used and throw an exception making it explicit that Gemfire does not support metadata filter expressions.

tzolov (changes requested):
Hey @geetrawat , 
Thank you for providing Gemfire Vector Store integration! 
I've added some preliminary comments and suggestions.  Feel free to reach me for questions or help.

vector-stores/spring-ai-gemfire/src/main/java/org/springframework/ai/vectorstore/GemFireVectorStore.java
metatype: We do expect to provide prefiltering based on Lucene indexes in the future. Will be nice, just iterating towards that at the moment.

tzolov:
Rebased, squashed and merged at 7634c6b78087a708a1dff8592b1485806da8ed4d

PR #406, opened 2024-03-06
Refactor Options

youngmoneee:
In the original code, Builders and Setters were used concurrently.
This approach could compromise object immutability in multi-threaded environments,
potentially leading to undesirable side effects.

To mitigate this, Setters have been removed to enhance and ensure object immutability.

- Remove setters
- Add a builder method that references existing objects
- Enhance the immutability of collections returned by getters
- Add test code reflecting the changes

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/ChatOptions.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/ChatOptions.java
@@ -25,14 +25,8 @@ public interface ChatOptions extends ModelOptions {
 
 	Float getTemperature();
 
-	void setTemperature(Float temperature);
-
 	Float getTopP();
 
-	void setTopP(Float topP);
-
 	Integer getTopK();
-
-	void setTopK(Integer topK);
 
 }
--- a/spring-ai-core/src/main/java/org/springframework/ai/model/function/FunctionCallingOptions.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/model/function/FunctionCallingOptions.java
@@ -18,11 +18,12 @@ package org.springframework.ai.model.function;
 
 import java.util.List;
 import java.util.Set;
+import org.springframework.ai.chat.prompt.ChatOptions;
 
 /**
  * @author Christian Tzolov
  */
-public interface FunctionCallingOptions {
+public interface FunctionCallingOptions extends ChatOptions {
 
 	/**
 	 * Function Callbacks to be registered with the ChatClient. For Prompt Options the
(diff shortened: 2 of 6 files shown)


spring-ai-core/src/main/java/org/springframework/ai/model/function/FunctionCallingOptions.java
```diff
@@ -18,11 +18,12 @@
 
 import java.util.List;
 import java.util.Set;
+import org.springframework.ai.chat.prompt.ChatOptions;
 
 /**
  * @author Christian Tzolov
  */
-public interface FunctionCallingOptions {
+public interface FunctionCallingOptions extends ChatOptions {
```
markpollack, line 26: FunctionCallingOptions not extending ChatOptions was intentional in order to provide support finer grained support of what features can be opted into.  Also, the code didn't compile after this change, so be sure to do a compilation of the total project before submitting a PR please.

spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/ChatOptions.java
```diff
@@ -25,14 +25,8 @@ public interface ChatOptions extends ModelOptions {
 
 	Float getTemperature();
 
-	void setTemperature(Float temperature);
```
markpollack, line 28: removing these setters also required to remove the `@Override` annotations on all implementing classes.  As a result the project didn't compile.  Please make sure the changes you make compile before sending a PR.

spring-ai-core/src/test/java/org/springframework/ai/chat/ChatBuilderTests.java
```diff
@@ -0,0 +1,134 @@
...
+
+		ChatOptions options = ChatOptionsBuilder.builder()
+			.withTemperature(initTemperature)
+			.withTopP(initTopP)
+			.withTopK(initTopK)
+			.build();
+
+		Integer newTopK = 222;
+
+		ChatOptions newOptions = ChatOptionsBuilder.builder(options)
```
markpollack, line 74: This sort of copy constructor for the builder is interesting, but please put in a separate PR

spring-ai-core/src/test/java/org/springframework/ai/chat/ChatBuilderTests.java
```diff
@@ -0,0 +1,134 @@
...
+		assertThat(options.getFunctionCallbacks().size()).isEqualTo(1);
+		assertThat(options.getFunctionCallbacks().contains(cb));
+
+		// Functions
+		assertThat(options.getFunctions()).isNotNull();
+		assertThat(options.getFunctions().size()).isEqualTo(1);
+		assertThat(options.getFunctions().contains(func));
+
+		// Immutable
+		options.getFunctionCallbacks().add(cb);
```
markpollack, line 120: If we want the created options class to be immutable then we should create a non modifiable collection and throw an exception.  We would also need to ensure that all the other options classes behave the same way.

markpollack:
I've created https://github.com/spring-projects/spring-ai/issues/407 and https://github.com/spring-projects/spring-ai/issues/408 to discuss changes to how builders behave across the code base.

PR #413, opened 2024-03-07
Refactored to optimize Document embedding reuse:

tzolov:
- Introduced logic to utilize existing Document embeddings when not empty, avoiding unnecessary calls to embeddingClient.embed.
 - Ensured immutability of the Document#content field by making it final.
 - Implemented a default method cachedEmbed in the EmbeddingClient interface, responsible for checking existing document embeddings and adding them if absent.
 - Updated all Vector Store implementations to employ cachedEmbed upon document addition.
 - Ensured that embedding fields are returned with resolved documents in select Vector Store implementations for consistency.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
@@ -52,7 +52,7 @@ public class Document {
 	/**
 	 * Document content.
 	 */
-	private String content;
+	private final String content;
 
 	/**
 	 * Embedding of the document. Note: ephemeral field.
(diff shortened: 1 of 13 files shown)


markpollack:
I think we are going to get rid of Document having embeddings so not merging this and assigning to RC1 so it can be further discussed.

tzolov:
My opinion is that the `Document#embedding` field is useless (even harmful) and should be removed form the Document spec. 
There is not a single use-case where this field would be used.  It is is misleading to assume that it can be used a a cached vector value because there is not use case where you retrieve a Document from the VectorStore  and then save the same document again.

Issue #408, opened 2024-03-06
Add an additional builder method that takes existing options.


markpollack:
Passing into an options builder an existing set of options to start from seems like a good pattern.  If so, ensure it is followed across all builder apis.

markpollack:
No one seems to have minded not having a copy like ctor, so closing

PR #414, opened 2024-03-07
Refactor Option Builder

youngmoneee:
This change aims to conceal the implementation and facilitate handling option objects through the interface.

And enforcing immutability to ensure greater safety in multi-threaded environments.

- Add Builder and Nested Private Implemented Class.

- Add builder() to Create New Object from existing.

- Add PortableFunctionCallingOptions to extract ChatOptions, FunctionsOptions

### Open AI

- To separate the responsibilities of objects, implement a dedicated object for Properties binding.

- Hide the implementation and only allow access through the interface to increase extensibility.

Thanks

- #408


markpollack:
Thank you, this is an area of the api I would like to review.

markpollack:
There are some good things—no one is arguing against immutability—but the PR does too much, and there are some choices that I don't understand or disagree with. Let's see how we can split this up into more fine-grained issues.

I'll take your bullet points in the PR comment one by one.

1. Add Builder and Nested Private Implemented Class

The convention that I've found in Spring is not to have a top-level class with the suffix `Builder` that is accessed directly. Instead, there is a private builder class inside the class being built.

As an example, this PR has a top-level:

```java
public class OpenAiAudioTranscriptionOptionsBuilder {
```

But the current code has:

```java
public static Builder builder() {
   return new Builder();
}

public static class Builder {
   protected OpenAiAudioTranscriptionOptions options;
}
```
I do not want to break this stylistic convention, as we should adhere to the Spring ecosystem conventions.

I do see that sometimes Spring projects create a builder interface and then have a DefaultBuilder implementation. For example, UriHandlerFilter in Spring Framework does this. I think that's a good idea, so I have created a specific issue for that.

Also, we haven’t been consistent: for example, ChatOptionsBuilder is a top-level class outside the ChatOptions compilation unit. Adding tests for the options builder class is a good idea.  Created an issue to discuss.

2. Add builder() to Create New Object from Existing

I see that our current options builder already takes an existing options object, so this is already handled.

3. Add PortableFunctionCallingOptions to Extract ChatOptions, FunctionsOptions

There is already a class for this, it doesn’t follow the pattern mentioned above (i.e., the builder class is outside the class it builds).  Created an issue to discuss.

4. To Separate the Responsibilities of Objects, Implement a Dedicated Object for Properties Binding

We are currently looking to remove the Boot dependency from spring-ai-core, which mostly boils down to removing @NestedConfigurationProperty. We're not sure about the cleanest approach yet, so we are experimenting to avoid adding too much boilerplate.  This PR doesn't address that concern and also I don't see the advantage over the implementation you provided.

The separation will happen, and any Boot-specific implementation to achieve this will occur in the auto-configuration classes where the @ConfigurationProperties are located.


I appreciate the contributions, but i will close this issue and the points raised can be tackled through the new issues.

PR #415, opened 2024-03-08
Remove invalid code

zhangqian9158:
Remove invalid code

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -97,7 +97,7 @@ public class OpenAiApi {
 		this.restClient = restClientBuilder
 				.baseUrl(baseUrl)
 				.defaultHeaders(ApiUtils.getJsonContentHeaders(openAiToken))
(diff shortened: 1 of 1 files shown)


spring-ai-core/src/main/java/org/springframework/ai/reader/TextReader.java
```diff
@@ -88,12 +88,11 @@ public List<Document> get() {
...
+
 		}
 		catch (IOException e) {
 			throw new RuntimeException(e);
 		}
 	}
 
-}
\ No newline at end of file
+}
```
laglangyue, line 98: the change should be restored， it's not need.

spring-ai-core/src/main/java/org/springframework/ai/reader/TextReader.java
```diff
@@ -88,12 +88,11 @@ public List<Document> get() {
 			this.customMetadata.put(SOURCE_METADATA, this.resource.getFilename());
 
 			return List.of(new Document(document, this.customMetadata));
-			// return textSplitter.apply(Collections.singletonList(new Document(document,
-			// this.customMetadata)));
```
youngmoneee, line 92: The comment was left intentionally for smooth recovery, so it seems inappropriate to remove it.

Issue #422, opened 2024-03-10
RedisVectorStore does not properly close Jedis pipelines


Bragolgirith:
The `RedisVectorStore#add` method retrieves a Jedis `Pipeline` (which is `Closable`), but never calls the `#close` method. 
Multiple calls to `RedisVectorStore#add` thus eventually exhaust all available pool resources and lead to an exception:

```
Exception encountered during context initialization - cancelling refresh attempt: redis.clients.jedis.exceptions.JedisException: Could not get a resource from the pool
```

PR #423, opened 2024-03-10
Fix RedisVectorStore not closing Jedis pipelines

Bragolgirith:
Fixes https://github.com/spring-projects/spring-ai/issues/422 using try-with-resources.


tzolov:
@Bragolgirith , Thank you for catching and fixing this.

PR #424, opened 2024-03-10
Added Bedrock Anthropic3 support

ben-gineer:
# Created anthropic messages API for Claude 3

This package is an adaption of the existing anthropic package which supports the more recent Messages API. The API is compatible with all existing Bedrock models.

It does not currently support the Image input message type.

I would welcome guidance on how to best incorporate such support given there is already an existing anthropic package which supports only the Text Completions API. @tzolov in particular as author of the original Text Completions implementations.

### DIFF
--- a/models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/anthropic3/AnthropicChatOptions.java
+++ b/models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/anthropic3/AnthropicChatOptions.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright 2023 - 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.ai.bedrock.anthropic3;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.springframework.ai.chat.prompt.ChatOptions;
+
+import java.util.List;
+
+/**
+ * @author Christian Tzolov
+ */
+@JsonInclude(Include.NON_NULL)
+public class AnthropicChatOptions implements ChatOptions {
+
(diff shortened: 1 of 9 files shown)


tzolov:
@ben-gineer thank you very much for the great contribution. 

Recently I've added a Claude 3 support for the Anthropic's own cloud infrastructure: https://github.com/spring-projects/spring-ai/commit/ce2bb131e567d551863166329b00824fcb746d3a based on their [Message API](https://docs.anthropic.com/claude/reference/messages_post).
It seems like [Bedrock's Anthropic Message API](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html) is a strip down version of the original Anthropic API. 
So it was very interesting to compare how we've approached the API mapping to the Spring AI abstractions. 
Looks like we made a very similar choices, which i guess cloud be expected. The stream handling is perhaps slightly different as i've tried to retain all historic chunk information in the output (not sure if this is that useful though). 
I also ported my multimodality impl. so you can prompt with text and images. 
Also added auto-configuration and Anthropic 3 documentation.

tzolov:
Rebased, squashed and merged at e004751842e71c996d72e31d14992f959740ce48

tzolov:
@ben-gineer, thanks again. 
I really like your work. So please let me know if there is anything else you would be interested to contribute to?

tzolov:
fyi, https://docs.spring.io/spring-ai/reference/1.0-SNAPSHOT/api/chat/bedrock/bedrock-anthropic3.html

ben-gineer:
Thanks for picking this up @tzolov. We're looking forward to using spring-ai. @Shadowbring previously implemented our own Bedrock abstraction, having found LangChain4J bringing in too many unwanted 3rd party dependencies. We're embedded in the Spring eco-system, so spring-ai suits our needs much better.

One question, LangChain Python OutputParser supports streaming of JSON responses (see https://python.langchain.com/docs/modules/model_io/output_parsers/types/json). Do you think this is doable with streaming output support in spring-ai? Most of our use cases involved some form of structured output, and having different elements of a map streamed could be helpful.

ben-gineer:
> @ben-gineer, thanks again. I really like your work. So please let me know if there is anything else you would be interested to contribute to?

Many thanks, I'll keep this in mind. Right now, my focus is to help my team with their current AI features.

tzolov:
@ben-gineer 
> One question, LangChain Python OutputParser supports streaming of JSON responses (see https://python.langchain.com/docs/modules/model_io/output_parsers/types/json). Do you think this is doable with streaming output support in spring-ai? Most of our use cases involved some form of structured output, and having different elements of a map streamed could be helpful.

Are you referring to the structured output parsers: https://docs.spring.io/spring-ai/reference/1.0-SNAPSHOT/api/output-parser.html
I guess you've already seen the List, Map, Bean and Record output parsers in the [Anthropic3 ITs](https://github.com/spring-projects/spring-ai/blob/8e50802eb8dd9044c15d7e75ad94ae618dca5c12/models/spring-ai-bedrock/src/test/java/org/springframework/ai/bedrock/anthropic3/BedrockAnthropic3ChatClientIT.java#L85)?

PR #425, opened 2024-03-11
modified ImageMessage.java line 45: mage > Image

PARK-afk:
spring-ai > spring-ai-core > src > main > java > org > springframework > image > ImageMessage 

line 45 : 
return "mageMessage{" + "text='" + text + '\'' + ", weight=" + weight + '}';
 >>
return "ImageMessage{" + "text='" + text + '\'' + ", weight=" + weight + '}';


tzolov:
Thanks for catching and fixing this @PARK-afk

Issue #355, opened 2024-02-22
Use <> or $ as stringtemplate delimiters instead of curly brackets


pherklotz:
The class org.springframework.ai.chat.prompt.PromptTemplate configures the stringtemplate renderer ST with non default delimiters `{` and `}`. In my opinion there is no need to change the default behavior. Here are the arguments that, for me, speak against deviating from the standard:

- With curly brackets you have to escape JSON examples in the prompt (for example if you provide an expected JSON response)
- The Spring AI documentation does not state this changed behavior (except in the examples) and the stringtemplate documentation uses `<` and `>`.

In my opinion, Spring AI should either use the stringtemplate standard, make delimiters configurable or mention the changed delimiters explicitly in the documentation.

If this issue is accepted I am happy to provide a pull request with the desired changes.

pherklotz:
Hello Mark, Thank you for your answer. I can provide the PR. Just to make sure that we are on the same page. I will remove the custom delimiters in Code an samples and stick with the default delimiters of stringtemplate?!

markpollack:
Hi.  Sorry for the delay.  Yes, that sounds great, an example with JSON would be appreciated.

i’ll schedule this for the next release after 0.8.1 and we will have to highlight it as a breaking change.

ThomasVitale:
@pherklotz @markpollack I appreciated a lot the templating compatibility of Spring AI with popular frameworks from the Python world. It helps a lot when getting started with Spring AI and when doing prompt design, since it's easy to re-use the many prompts available and used in Python applications. Even outside Python, the curly brackets syntax is the most used one for LLM prompts, making it straightforward to collaborate and share across languages and frameworks.

Would it be an idea to make it configurable? The default could be "{}", but an application property could be used to replace it with something else, such as "<>" or "$".

pherklotz:
Hi @ThomasVitale, we have another similar discussion in the PR. I think it's easier if you join there :)

youngmoneee:
For Python's F-String, `injecting JSON` requires the use of `double brackets`.
Focused on this functionality, I've developed a prototype for implementing F-Strings.

Even when using a delimiter injection method, I anticipate issues arising in templates mixed with JSON, HTML, and XML. Although it may be a challenging approach, this prototype allows for injecting and utilizing JSON in F-Strings, similar to its original functionality.

The test code written includes:

1. Injecting JSON into the template.
2. Injecting JSON into the template and using template variables within JSON.
[...]

PR #444, opened 2024-03-14
Use default stringtemplate delimiters for prompt templates

pherklotz:
The current set delimiters `{` and `}` do not follow the stringtemplate default and need escaping if you want to build a prompt with JSON (as described in issue #355).  This pull request will address this issue.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java
@@ -53,7 +53,7 @@ public class PromptTemplate implements PromptTemplateActions, PromptTemplateMess
 			throw new RuntimeException("Failed to read resource", ex);
 		}
 		try {
-			this.st = new ST(this.template, '{', '}');
+			this.st = new ST(this.template);
 		}
 		catch (Exception ex) {
 			throw new IllegalArgumentException("The template string is not valid.", ex);
@@ -64,7 +64,7 @@ public class PromptTemplate implements PromptTemplateActions, PromptTemplateMess
 		this.template = template;
 		// If the template string is not valid, an exception will be thrown
 		try {
-			this.st = new ST(this.template, '{', '}');
+			this.st = new ST(this.template);
 		}
 		catch (Exception ex) {
 			throw new IllegalArgumentException("The template string is not valid.", ex);
@@ -75,7 +75,7 @@ public class PromptTemplate implements PromptTemplateActions, PromptTemplateMess
 		this.template = template;
 		// If the template string is not valid, an exception will be thrown
 		try {
-			this.st = new ST(this.template, '{', '}');
+			this.st = new ST(this.template);
 			for (Entry<String, Object> entry : model.entrySet()) {
 				add(entry.getKey(), entry.getValue());
 				dynamicModel.put(entry.getKey(), entry.getValue());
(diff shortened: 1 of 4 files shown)


markpollack:
I don't follow the comment regarding `{` and `}` offering an advantage for frontend frameworks. Do you mean sending frontend code as part of the prompt?   JSON is a more common case than embedding react code.

Between easier json embedding into the prompt or sending react code, I'd go with easier json embedding.  That said, it sounds like the delimiter should be configurable to better accommodate all cases.

My intention was to support the python F-string syntax.  How does python handle the case of inserting json into a f-string?

I'd need to reproduce the original issue, what does embedding json into the TextTemplate look like?

youngmoneee:
What I intended was similar to what you mentioned about F-String, as various languages or frameworks use `{` , `}` based variable templates. This makes it useful when importing templates from external sources or exporting them, and given its widespread use, it should be easy for developers to read.

Changing the delimiter to `<`, `>` due to the difficulty of embedding JSON in prompts might introduce new challenges when inserting HTML or XML.

To fundamentally solve these issues, developing custom parsers, renderers, or using flexible variable templates through the injection of ST with various delimiters might be a good approach.

pherklotz:
I agree, that it might be the best to introduce a configuration option for the delimiters. 
I can change the PR accordingly. The default option would still be < and >. I agree that this introduces error with HTML or XML snippets. But I assume these cases are rarer than JSON snippets.


@markpollack an example prompt could be:

`List all European countries and their capitals. Answer with the following JSON format: 
{"Country name": "capital name"}`

This prompt fails if used with the ST Template Prompt mechanism.

ThomasVitale:
I shared some thoughts on the issue for this PR about the templating strategy: https://github.com/spring-projects/spring-ai/issues/355#issuecomment-2018926008

pherklotz:
So how do we proceed with this? I am fine with changing it, but will only start working once there is a commitment to a solution ;)

youngmoneee:
> So how do we proceed with this? I am fine with changing it, but will only start working once there is a commitment to a solution ;)

I have created a code that functions similarly to F-string.
It's a simple prototype, so it's not perfect beyond the basic functionality, but based on test code, if there's a specific scenario that needs to be addressed, it could be refined into a more advanced form.
If there is a situation that needs to be handled, it would be good to talk together. :)
#355

HzjNeverStop:
I agree that PromptTemplate should provide an API for users to use different delimiters based on their own scenarios. However, can we keep the default delimiter as { } to avoid unnecessary compatibility changes for existing users?

markpollack:
I also agree that we should allow different delimiters, and I've also had requests for different template engines.  I do not want to change the default `{` and `}` as it would bring too much pain to users.  If I made bad choice there, apologies.  Bumping this to RC1 - sorry, it keeps slipping.  Will get to it eventually!  Apologies.

coderphonui:
Voted for this. The { and } characters are used a lot to put examples for the JSON output of the prompt. Using them by default will cause a lot of pains for the user.

markpollack:
Given the history and the default usage of `{` and `}` in the llm ecosystem, keeping the default to be the curly brace seems best to me.  I will look into how to make this configurable, now that we have `ChatClient` it isn't immediately obvious where to enable such a feature but will look into it.

markpollack:
Closing this PR but leaving the relevant issue https://github.com/spring-projects/spring-ai/issues/355 open

PR #449, opened 2024-03-14
Allow more customization for Neo4j store (id and constraint).

meistermeier:
Some users' use-cases going not the complete SpringAI path from document creation to search (and chat) but are working with an already existing dataset (incl. embeddings).
To align with their needs, some more customization will allow them to work with those data instead of patching their dataset to match SpringAI's `Document` definition without any benefit for them.

Unrelated to this change, the Neo4j test version increased to be current.


tzolov:
@meistermeier 
If you change the document `id` would you be still able to perform a similaritySearch? 
The [recordToDocument](https://github.com/spring-projects/spring-ai/blob/main/vector-stores/spring-ai-neo4j-store/src/main/java/org/springframework/ai/vectorstore/Neo4jVectorStore.java#L369) seems to presume that the "id" field exist?

meistermeier:
The commit also contains the change in the method you mentioned, or have I missed something?
https://github.com/spring-projects/spring-ai/pull/449/files#diff-3e9e9aca32ce4c69d1169e7a9d0c81ad17360e4d83d315e61e0968cac1044515R413

tzolov:
No, i've missed it.  Thanks for clarifying 
PR LGTM.

Issue #242, opened 2024-01-11
Create AutoConfiguration for PostgremML


markpollack:

usrivastava92:
@markpollack The PR is ready for review, Is there anything additional I should address aside from adhering to the checklist outlined in the pull request?

markpollack:
@usrivastava92 It does not appear that you made a PR from your branch.  Can you please do that?

Thanks

markpollack:
Apologies, I somehow missed the PR.  We will merge it. Thanks.

markpollack:
fixed in 5718f9683d4abd0702cfe99d754558f01996fec0

PR #454, opened 2024-03-16
Add PostgresMlAutoConfiguration to AutoConfiguration.imports

izeye:
It seems to have been missed in https://github.com/spring-projects/spring-ai/issues/242.


tzolov:
Yeh, nice catch @izeye

PR #458, opened 2024-03-16
Remove withReuse in MongoDBAtlasContainer

eddumelendez:
`Reuse Container` is a Testcontainers experimental feature. It requires `testcontainers.reuse.enable=true` in `~/.testcontainers.properties` in order to take effect but in order to avoid surprises, this commit remove it.

See https://java.testcontainers.org/features/reuse/


tzolov:
Thanks @eddumelendez

PR #460, opened 2024-03-17
Add MongoDB Atlas vector store auto-configuration

eddumelendez:
It also provides a starter.

### DIFF
--- a/spring-ai-spring-boot-autoconfigure/src/test/java/org/springframework/ai/autoconfigure/vectorstore/mongo/MongoDBAtlasVectorStoreAutoConfigurationIT.java
+++ b/spring-ai-spring-boot-autoconfigure/src/test/java/org/springframework/ai/autoconfigure/vectorstore/mongo/MongoDBAtlasVectorStoreAutoConfigurationIT.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2023 - 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


spring-ai-spring-boot-autoconfigure/src/test/java/org/springframework/ai/autoconfigure/vectorstore/mongo/MongoDBAtlasVectorStoreAutoConfigurationIT.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Eddú Meléndez
+ */
+@Testcontainers
+class MongoDBAtlasVectorStoreAutoConfigurationIT {
```
eddumelendez, line 49: Test is failing with `vector field is indexed with 1536 dimensions but queried with 384`. I wonder if dimension should be configured as well instead of using the hardcoded value. Looks like it can be up to 4096, see [docs](https://www.mongodb.com/docs/atlas/atlas-vector-search/create-index/#create-an-atlas-vector-search-index).

eddumelendez: Add OpenAI API config and it works.

tzolov:
This was a very quick response ;) 
Thanks @eddumelendez

PR #464, opened 2024-03-17
Update list of supported AI models in the main README.MD

jakkaz:
While reading the README I have noticed an incomplete list of models and updated it based on what I see in the pom.xml modules section.


markpollack:
yikes!  I think i deleted some documentation unintentionally.   Great ey.. Thanks for taking the time to make a PR.

PR #482, opened 2024-03-21
Refactor MistralAi class and Fix OpenAi ToolChoiceBuilder value

ricken07:
- Refactor MIstralAiStreamFunctionCallingHelper to MistralAiStreamFunctionCallingHelper
- Replace none to auto in OpenAi ToolChoiceBuilder 'AUTO'


tzolov:
Good catch. Thanks @ricken07

PR #483, opened 2024-03-21
Update azure-openai-embeddings.adoc

iAMSagar44:
Corrected the property prefix related to embeddings for Azure Open AI. The prefix had 'embeddings' in plural. The actual prefix is spring.ai.azure.openai.embedding

@pivotal-cla This is an Obvious Fix.


tzolov:
Thanks @iAMSagar44

PR #484, opened 2024-03-21
Add Multi-Modality Support for OpenAI

tzolov:
- Implemented a MediaContent abstraction within the OpenAiApi to handle text and image inputs.
 - Response message content remains a plain String, ensuring backward compatibility.
 - Extended the OpenAiChatClient request creation process to seamlessly map Spring AI Messages with Media content to the low-level OpenAiApi MediaContent types.
 - Added integration tests for embedded and URL images, covering both synchronous and streaming calls.
 - Updated the OpenAI class diagram to reflect the new media content types provided by the OpenAI API.
 - Incorporated a chapter on multi-modality within the openai-chat.adoc documentation.


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -407,29 +407,44 @@ public record ResponseFormat(
...
+	 * responding to.
 	 * @param toolCallId Tool call that this message is responding to. Only applicable for the {@link Role#TOOL} role
 	 * and null otherwise.
 	 * @param toolCalls The tool calls generated by the model, such as function calls. Applicable only for
 	 * {@link Role#ASSISTANT} role and null otherwise.
 	 */
 	@JsonInclude(Include.NON_NULL)
 	public record ChatCompletionMessage(
-			@JsonProperty("content") String content,
+			@JsonProperty("content") Object rawContent,
```
markpollack, line 423: I don't see the impact of this change to `Object`.  If this can be `MediaContent`, then where is the code that is surfaces that type?

markpollack:
merged in 834d2d04879c080da208bdde7cda5aea7c48f585

PR #485, opened 2024-03-21
Update getting-started documentation

ricken07:
- Adding Transcription Models and Text-To-Speech in 'Add dependencies for specific components' section
- Rename title Speech API to Text-To-Speech (TTS) API in speech.doc


tzolov:
thanks

PR #486, opened 2024-03-21
Refactor core image

youngmoneee:
### Summary

Refactored Image Object

### Description

In the case of the original image objects, we initialized them by passing both base64 and URL as constructor arguments, and returned their values through the getUrl and getBase64 methods.

However, this approach made it challenging to clearly indicate small differences between various providers leading to scalability issues when adding new types beyond base64 and URL.

To solve this, we have separated the original Image object into the Image and ImageType interfaces. Each module defines an enum that implements the ImageType interface (e.g., OpenAiImageType, etc.), allowing us to distinguish between the image's provider and format. This enum can be flexibly applied to situations like the Request's Response Format. 
``` java
//  example code
  (req, res) -> {
    return switch(req.responseFormat()) {
      case URL -> new MyUrlImage(res.data());
      case B64 -> new MyBase64Image(res.data());
      default -> throw new MyException("Unsupported type")
  }
}
```
Moreover, through the Image interface, all image objects can be uniformly treated, making the design easily adaptable even when new response formats or providers are added.

Please kindly review the proposed structure change to resolve this issue.

I am requesting your review as you are the most recent committer to this file.
@markpollack 

Thank you.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiImageClient.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiImageClient.java
@@ -17,10 +17,10 @@ package org.springframework.ai.openai;
 
 import java.util.List;
 
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.springframework.ai.image.Image;
 import org.springframework.ai.image.ImageClient;
 import org.springframework.ai.image.ImageGeneration;
 import org.springframework.ai.image.ImageOptions;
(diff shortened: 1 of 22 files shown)


markpollack:
Thanks, trying to catch up on many things, i'll get to it.  Appreciate the PR.

youngmoneee:
I know that just by looking at the number of merges and commits, this project is rapidly changing and you are working hard.
Although it might be insufficient, I hope it helps.

markpollack:
This is old and there is no spring-ai-core anymore! :)  Thanks though, too bad we didn't get to it MUCH earlier.

PR #487, opened 2024-03-21
Update azure-openai-chat.adoc

iAMSagar44:
In v0.8.1, the property spring.ai.azure.openai.chat.options.model does not exist anymore. I have tried with spring.ai.azure.openai.chat.options.deployment-name and am able to connect to the Azure Open AI service.

Considering spring.ai.azure.openai.chat.options.deployment-name is the correct property name, I have updated the document to reflect this.

I have also removed this statement "In a subsequent release, Spring AI will rename the property spring.ai.azure.openai.chat.options.model to spring.ai.azure.openai.chat.options.deployment-name to avoid confusion." as this has already been renamed. I have added an appropriate note related to this.

@pivotal-cla This is an Obvious Fix.


tzolov:
Yeh we did this change in https://github.com/spring-projects/spring-ai/commit/1e98f82e03529506c9afaa650717b3819633d573 and I'm surprised we've missed to update the doc. 
Thanks for catching it.

PR #491, opened 2024-03-21
Fix TitanChatBedrockApi.java

### LINKED ISSUE #490: Inconsistent constructor visibility of the TitanChatBedrockApi class
The constructor 
```java
TitanChatBedrockApi(String modelId, String region) {
  super(modelId, region);
}
```
of the TitanChatBedrockApi class is currently package-private and not accessible as a user of the library.

The corresponding other [...]Api class constructors of the bedrock clients are public.

I assume this should be public as well.

youngmoneee:
fix Access modifier

close #490


tzolov:
Thanks

PR #496, opened 2024-03-22
Azure openai tool streaming

### LINKED ISSUE #478: Azure OpenAI Client Streaming with Function (tools)
The AzureOpenAI client doesn't support function calling (now called tools in Azure SDK) in streaming response.

See:
https://github.com/spring-projects/spring-ai/blob/1e98f82e03529506c9afaa650717b3819633d573/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java#L168

To reproduce, create a simple Client with a tool, and invoke, the choice contains the function call, but is ignored.

Grogdunn:
Try to fix #478 issue, now works for Azure.

Due Microsoft API we need to "clone" the DTO structure. All MS DTO are final class with private constructors.

### DIFF
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java
@@ -18,7 +18,6 @@ package org.springframework.ai.azure.openai;
 import com.azure.ai.openai.OpenAIClient;
 import com.azure.ai.openai.models.*;
 import com.azure.core.util.BinaryData;
-import com.azure.core.util.IterableStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.ai.azure.openai.dto.AccessibleChatChoice;
@@ -44,7 +43,6 @@ import reactor.core.publisher.Flux;
 
 import java.util.*;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * {@link ChatClient} implementation for {@literal Microsoft Azure AI} backed by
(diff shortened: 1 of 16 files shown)


radhakrishna67:
@tzolov , this will solve latency issues. Could you please prioritize it?

tzolov:
Hey @radhakrishna67, 
Thank you for taking your time for addressing this. 
I've started reviewing it 2 weeks ago, but hit some IT test failures and something like multiple flux pipeline executions for the same call. 
Otherwise the approach seems right. 
I will be able to look at it again, hopefully next week, after we coock out the chat-history feature.

imGruuu:
@tzolov Hello Dear Sir, I'm blocked by the same question above. Will the next release version fix the ability to stream Azure OpenAi to make function calls? This is important for us as we work based on the resources provided by Azure, thank you very much! Best regards.

Grogdunn:
keep updated with main

tzolov:
Hi @Grogdunn , Finally i had some time to review and refactor this PR.  You can find the changes here: https://github.com/tzolov/spring-ai/commits/gh-496-pr
Basically I fixed the few small issues causing failures for OpenAI, Gemini and Mistral models (Anthropic doesn't support streaming function calling). 
Also i've replaced the wrapping with java reflection for calling the private constructors and fields. I believe this is less intrusive and easier to maintain. 

Let me know if you agree with those changes so i can merge it. 
Also, please let me know what name to put in the @author javadoc for you?

tzolov:
@Grogdunn thank you again for the great contribution! 
I went ahead and merged the PR with the above mentioned  changes and using your GH handle as javadoc @author. 
Feel free to submit follow up PRs if  you need to make some adjustments.

tzolov:
Rebased, squashed and merged at a7eb28ac17debbaaaf227802eb0385f3daea4fd8

PR #525, opened 2024-03-31
Add the default model to the initial Vertex AI Gemini autoconfiguration chat properties

zucchivan:
I'm not sure if there is a reason for it, but Vertex AI Gemini basic autoconfiguration didn't carry a default model, unlike some other implementations (like OpenAI's). 

Setting the default model makes it a little bit more _plugandplay'able_.


tzolov:
Makes sense. Thanks @zucchivan

tzolov:
can you confirm that you've run `./mvnw spring-javaformat:apply` ?

zucchivan:
Yups, it's done!

PR #527, opened 2024-04-01
Fixing metadata for text splitting

ricken07:
A correction has been made to the "page_number" value, which corresponded to the last page number.

Each chunk of text on a page will now have the page number as the 'page_number' metadata, instead of the last page number for the entire block.


tzolov:
Thanks for the fix @ricken07 !

tzolov:
@ricken07 please run the tests and format the code before submitting a PR

ricken07:
OK I will add test classes and I will not forget to use the plugin provided to format the code. Thanks

PR #536, opened 2024-04-02
[WIP] Add Chat History support

tzolov:
- Add ChatHistory interface and in-memory token bounded implementation: TokenCountSlidingWindowChatHistory.java
 - Fix a bug with AssistantMessage hash/equality implementation.
 - Add HistoryChatClient unit tests.
 - Implement a ChatClient decorator with history support.
 - expose PromptHistoryAgumenter hook with default implementation.
 - Special handling of System messages.
 
Tested with OpenAI only.


spring-ai-core/src/main/java/org/springframework/ai/chat/history/ChatExchange.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * Represents single call request and response set of messages.
+ *
+ * @author Christian Tzolov
+ */
+public class ChatExchange {
+
+	private final String sessionId;
+
+	private final List<Message> messages;
```
michaelsembwever, line 34: is it expected to have more than a pair (user request and assistant response) here? 

will this eventually overlap with LLM caching, i.e. the need to also store the context of the exchange, e.g. the request's prompt

spring-ai-core/src/main/java/org/springframework/ai/chat/history/ChatHistory.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Core abstraction for storing and retrieving chat history.
+ *
+ * @author Christian Tzolov
+ */
+public interface ChatHistory {
+
+	void add(ChatExchange historyGroup);
+
+	List<ChatExchange> get(String sessionId);
```
michaelsembwever, line 30:
```suggestion
	List<ChatExchange> get(String sessionId, int lastN);
```

the limit will want to be a pushed-down predicate, e.g. in non-memory impls.

spring-ai-core/src/main/java/org/springframework/ai/chat/history/ChatExchange.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.springframework.ai.chat.messages.Message;
+
+/**
+ *
+ * Represents single call request and response set of messages.
+ *
+ * @author Christian Tzolov
+ */
+public class ChatExchange {
```
michaelsembwever, line 30: why not a java record ? (there's a number of javabeans/pojos in this PR that could be simplified as records)

tzolov:
Replaced by #634

PR #539, opened 2024-04-02
Feat: Add additional parameter to WatsonxAiChatOptions

PabloSanchi:
Some custom models may include extra parameters, so the idea is to define all the common parameters but add and additional attribute that will store all these custom configuration parameters. This will also help with its maintaniablitiy as if new parameters are added we can still use it while the options are not updated.

- Add "additional" attribute to WatsonxAiChatOptions

### DIFF
--- a/models/spring-ai-watsonx-ai/src/main/java/org/springframework/ai/watsonx/WatsonxAiChatOptions.java
+++ b/models/spring-ai-watsonx-ai/src/main/java/org/springframework/ai/watsonx/WatsonxAiChatOptions.java
@@ -89,7 +89,7 @@ public class WatsonxAiChatOptions implements ChatOptions {
     /**
      * Sets how many tokens must the LLM generate. (Default: 0)
      */
-    @JsonProperty("min_new_tokens") private Integer minNewTokens = 0;
+    @JsonProperty("min_new_tokens") private Integer minNewTokens;
 
     /**
      * Sets when the LLM should stop.
@@ -103,7 +103,7 @@ public class WatsonxAiChatOptions implements ChatOptions {
      * (e.g., 1.8) will penalize repetitions more strongly, while a lower value (e.g.,
      * 1.1) will be more lenient. (Default: 1.0)
      */
-    @JsonProperty("repetition_penalty") private Float repetitionPenalty;
+    @JsonProperty("repetition_penalty") private Float repetitionPenalty = 1.0f;
 
     /**
      * Produce repeatable results, set the same random seed value every time. (Default: randomly generated)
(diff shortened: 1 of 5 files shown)


models/spring-ai-watsonx-ai/src/main/java/org/springframework/ai/watsonx/WatsonxAiChatOptions.java
```diff
@@ -99,7 +103,7 @@ public class WatsonxAiChatOptions implements ChatOptions {
      * (e.g., 1.8) will penalize repetitions more strongly, while a lower value (e.g.,
      * 1.1) will be more lenient. (Default: 1.0)
      */
-    @JsonProperty("repetition_penalty") private Float repetitionPenalty;
+    @JsonProperty("repetition_penalty") private Float repetitionPenalty = 1.0f;
```
tzolov, line 106: Do not set default values here but your default instance in https://github.com/spring-projects/spring-ai/blob/9236913696298bd42838bab0b72416ab413bf8d3/models/spring-ai-watsonx-ai/src/main/java/org/springframework/ai/watsonx/WatsonxAiChatClient.java#L57
or https://github.com/spring-projects/spring-ai/blob/9236913696298bd42838bab0b72416ab413bf8d3/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/watsonxai/WatsonxAiChatProperties.java#L42

tzolov (changes requested):

tzolov (approved):

models/spring-ai-watsonx-ai/src/main/java/org/springframework/ai/watsonx/api/WatsonxAiRequest.java
```diff
@@ -62,7 +63,7 @@ public WatsonxAiRequest withProjectId(String projectId) {
     public static Builder builder(String input) { return new Builder(input); }
 
     public static class Builder {
-        private final String input;
+        public static final String MODEL_PARAMETER_IS_REQUIRED = "Model parameter is required";private final String input;
```
tzolov, line 66: This looks awkward. 
Please run ./mvnw spring-javaformat:apply

PabloSanchi: Right, apologies, I did the build and it passed so I thought the code linting was right

tzolov:
Rebased, squashed and merged at ae6a019ad17fe157346f6dafbdbf9d2713c764eb

Issue #553, opened 2024-04-06
[Ollama] Missing fields in API


ThomasVitale:
The `keep_alive` field in the Ollama requests objects are missing. (option that might become more configurable in https://github.com/spring-projects/spring-ai/issues/549)
The `images` field in the Generation request object is missing.

See: https://github.com/ollama/ollama/blob/main/api/types.go

PR #554, opened 2024-04-06
Ollama: Added missing fields in API

ThomasVitale:
Fixes gh-553

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
@@ -150,7 +150,9 @@ public class OllamaApi {
 			@JsonProperty("template") String template,
 			@JsonProperty("context") List<Integer> context,
 			@JsonProperty("stream") Boolean stream,
-			@JsonProperty("raw") Boolean raw) {
+			@JsonProperty("raw") Boolean raw,
+			@JsonProperty("images") List<String> images,
+			@JsonProperty("keep_alive") String keepAlive) {
 
 		/**
 		 * Short cut constructor to create a CompletionRequest without options.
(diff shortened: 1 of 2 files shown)


models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
```diff
@@ -445,21 +461,26 @@ public Builder withFormat(String format) {
 				return this;
 			}
 
+			public Builder withKeepAlive(Duration keepAlive) {
+				this.keepAlive = keepAlive;
+				return this;
+			}
+
 			public Builder withOptions(Map<String, Object> options) {
-				Objects.requireNonNullElse(options, "The options can not be null.");
```
ThomasVitale, line 449: I fixed the method name here to prevent `NullPointerExceptions`

tzolov:
Hey @ThomasVitale , thanks for improving this. 

It looks related to the aca8ffd400fd3e3851aec8dc6705c6613d456073 PR I've just merged, but concerns the legacy `generate` endpoint only?

Though the ChatClient impl.  uses only the new `chat completion` endpoint I guess we should to keep the entire OllamaApi up to date. 
Could you please rebase and resolve the possible conflicts?

tzolov:
yeah, I did a mistake with my previous merge. Please rebase after 15628a34b258ef637878a32e0867420ed252f039

ThomasVitale:
@tzolov thanks, I'll rebase the PR shortly. I can see that a `template` field has been added as an option to the `ChatRequest` model, but it doesn't look like it's part of the Ollama API for the `chat completion` endpoint. It's only used in the `generate` endpoint. (https://github.com/ollama/ollama/blob/main/api/types.go#L40)

The Ollama documentation includes it also for the `chat completion` endpoint, but that might be a copy/paste mistake in the docs. Looking at the code of the Ollama server, the handler for chat completion requests is always using the template configured in the model and doesn't allow overwriting (https://github.com/ollama/ollama/blob/main/server/routes.go#L1292). From what I could test, the `template` field included in the `ChatRequest` Java class will simply be ignored by Ollama. I raised the issue on the Ollama project: https://github.com/ollama/ollama/issues/3514

ThomasVitale:
I've just rebased the PR on top of the current `main`.

tzolov:
Sorry for the mess I caused earlier today. Obviously lacking testing for the options/parameter merging/mapping logic. Will add some today. 
Thanks  again to for the alignment.

ThomasVitale:
@tzolov no problem! Thanks for the review!

Issue #555, opened 2024-04-06
[OpenAI] Missing fields in API


ThomasVitale:
The Chat API is missing these optional fields (https://platform.openai.com/docs/api-reference/chat):

* logprobs
* top_logprobs

The Embedding API is missing this optional field (https://platform.openai.com/docs/api-reference/embeddings):

* dimensions

PR #556, opened 2024-04-06
OpenAI: Added missing fields to API

ThomasVitale:
Fixes gh-555


tzolov:
It is nice to keep the APIs in sync. Thanks @ThomasVitale

PR #558, opened 2024-04-07
Ollama: Remove unused option in chat completion

ThomasVitale:
The Ollama documentation used to include a `template` parameter for the `chat completion` endpoint. However, that parameter only exists for the `generation` endpoint. It was probably a copy/paste mistake in the documentation. I have submitted a fix to the Ollama project which has now been merged (https://github.com/ollama/ollama/pull/3515).

This pull request is for removing the `template` parameter from the `ChatRequest` class and from the `OllamaOptions` class since it's not part of the Ollama API for Chat Completion. I have also updated the related documentation. The Ollama Server will not fail if the parameter is included, but it won't have any effect.


ThomasVitale:
@tzolov I created this PR as a followup to my comment from yesterday in https://github.com/spring-projects/spring-ai/pull/554#issuecomment-2041062684

PR #575, opened 2024-04-10
Cassandra Vector Store initial impl follow up

michaelsembwever:
.
- CassandraVectorStoreAutoConfiguration uses CassandraAutoConfiguration
- driver profiles for production stability+performance,
- CassandraVectorStore.addStmts map must be concurrent-safe
- code cleanups and naming fixes (field->column),
- main doc tidy-up
- astradb compatibility (protocol V4)


michaelsembwever:
A question: 
when calling vectorStore.add(docs) are all the docs expected to need embeddings created for them, or may this be optional ?

Currently all the stores are creating the embeddings in their `add(docs)` method.

But the spring-ai api does make it possible for the user to do the embeddings first and then call `add(docs)`, shouldn't stores be honouring this…?

michaelsembwever:
> when calling vectorStore.add(docs) are all the docs expected to need embeddings created for them, or may this be optional ?

got an answer from @tzolov , basically all in https://github.com/spring-projects/spring-ai/pull/413 

adding a squash commit that provides the same functionality inside CassandraVectorStore (it's a one line change).

tzolov:
Hey @michaelsembwever , thanks for the great contribution. All looks good!

tzolov:
rebased, squashed and merged at 0eaf7d05c9bc509f8ae2d9e16594ec2db4fd2b4a

PR #592, opened 2024-04-16
Enhancing Elasticsearch vector store implementation

l-trotta:
This PR provides a more performant search function for the Elasticsearch vector store and removes the bean autoconfiguration.

### Autoconfiguration

The current implementation of `afterPropertiesSet()` automatically creates a new index with a set of properties that only works with OpenAI, or any other model that works with vectors with a dimension of 1536; users adopting other models would currently have to manually delete the index. By default, Elasticsearch automatically creates the correct index settings when it receives the first PUT request for vectors, so in our opinion there's no need for such autoconfiguration. 

### Search

The function used now is [script_score](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-script-score-query.html), which can be slow for large data samples since it does a brute force comparison with all vectors using the similarity function. This PR replaces it with the approximate [knn search](https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html), more performant because it only scans the closest neighbours. The similarity functions available for `knn` can be easily configured in the index mapping by setting the correct name. 

##

We would also like to contribute to the documentation, should that be done in a different PR?  

Thank you @JM-Lab for the original implementation, would you like to review these changes?

 (Disclosure: I work for Elastic)

### DIFF
--- a/spring-ai-spring-boot-autoconfigure/pom.xml
+++ b/spring-ai-spring-boot-autoconfigure/pom.xml
@@ -260,6 +260,7 @@
 			<optional>true</optional>
 		</dependency>
 
+		<!-- Elasticsearch Vector Store-->
 		<dependency>
 			<groupId>org.springframework.ai</groupId>
 			<artifactId>spring-ai-elasticsearch-store</artifactId>
@@ -279,6 +280,12 @@
 			<artifactId>spring-ai-zhipuai</artifactId>
 			<version>${project.parent.version}</version>
 			<optional>true</optional>
+		</dependency>
+
+		<dependency>
+			<groupId>co.elastic.clients</groupId>
+			<artifactId>elasticsearch-java</artifactId>
+			<version>${elasticsearch-java.version}</version>
 		</dependency>
 
 		<!-- test dependencies -->
--- a/vector-stores/spring-ai-elasticsearch-store/src/main/java/org/springframework/ai/vectorstore/ElasticsearchVectorStore.java
+++ b/vector-stores/spring-ai-elasticsearch-store/src/main/java/org/springframework/ai/vectorstore/ElasticsearchVectorStore.java
@@ -21,6 +21,7 @@ import co.elastic.clients.elasticsearch.core.BulkRequest;
 import co.elastic.clients.elasticsearch.core.BulkResponse;
 import co.elastic.clients.elasticsearch.core.SearchResponse;
 import co.elastic.clients.elasticsearch.core.search.Hit;
+import co.elastic.clients.elasticsearch.indices.CreateIndexRequest;
 import co.elastic.clients.elasticsearch.indices.CreateIndexResponse;
 import co.elastic.clients.json.jackson.JacksonJsonpMapper;
 import co.elastic.clients.transport.rest_client.RestClientTransport;
@@ -42,6 +43,8 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import static java.lang.Math.sqrt;
+
 /**
  * @author Jemin Huh
  * @author Wei Jiang
@@ -56,6 +59,8 @@ public class ElasticsearchVectorStore implements VectorStore, InitializingBean {
 	private final ElasticsearchClient elasticsearchClient;
 
 	private final ElasticsearchVectorStoreOptions options;
+
+	private String similarityFunction = SIMILARITY_DEFAULT;
 
 	private final FilterExpressionConverter filterExpressionConverter;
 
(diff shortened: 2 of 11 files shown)


JM-Lab:
Hi @l-trotta,

Thanks for the update. I liked how script_score was quick to implement due to available examples, but I'm pleased we're switching to KNN for better performance.

Could you check if the code at https://github.com/spring-projects/spring-ai/blob/main/vector-stores/spring-ai-elasticsearch-store/src/main/java/org/springframework/ai/vectorstore/ElasticsearchAiSearchFilterExpressionConverter.java needs any improvements?

I'm looking forward to the official documentation for the Elasticsearch vector store of the Spring AI project, which I think @tzolov will address.

tzolov:
@l-trotta , thank you for the improvements. 
Could you please rebase and update your PR after the https://github.com/spring-projects/spring-ai/pull/633 merge.

l-trotta:
Updated! I'd like to explain why I removed the `dense-vector-indexing` property: quoting the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html#dense-vector-params), 

> If true, you can search this field using the [kNN search API](https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search-api.html). Defaults to true.

And since this implementation uses kNN search, setting this to false would make the data unsearchable.

ezimuel:
@tzolov what's the status of this PR? Thanks.

tzolov:
@l-trotta thanks for your contribution and thank you for your patience.

While reviewing the PR i've noticed that you have removed the capability to override the similarityFunction at runtime? 
Is there any particular reason to reduce this flexibility?  Either way it renders this [ElasticsearchVectorStoreAutoConfigurationIT](https://github.com/spring-projects/spring-ai/blob/879f99239410cd44515708e4ac6ab75456a0c812/spring-ai-spring-boot-autoconfigure/src/test/java/org/springframework/ai/autoconfigure/vectorstore/elasticsearch/ElasticsearchVectorStoreAutoConfigurationIT.java#L74) almost useless? 
If there are not strong reasons to have this immutable perhaps we should re-enable the `withSimilarityFunction`  setter?

spring-ai-spring-boot-autoconfigure/pom.xml
```diff
@@ -281,6 +282,12 @@
 			<optional>true</optional>
 		</dependency>
 
+		<dependency>
```
tzolov, line 285: What is the reason to add the elasticsearch-java explicitly here? As it is already defined in the vector-store dependenies?

vector-stores/spring-ai-elasticsearch-store/src/main/java/org/springframework/ai/vectorstore/ElasticsearchVectorStore.java
```diff
@@ -84,30 +79,22 @@ public ElasticsearchVectorStore(ElasticsearchVectorStoreOptions options, RestCli
 		this.embeddingModel = embeddingModel;
 		this.options = options;
 		this.filterExpressionConverter = new ElasticsearchAiSearchFilterExpressionConverter();
-		// the potential functions for vector fields at
-		// https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-script-score-query.html#vector-functions
-		this.similarityFunction = COSINE_SIMILARITY_FUNCTION;
-	}
-
-	public ElasticsearchVectorStore withSimilarityFunction(String similarityFunction) {
```
tzolov, line 92: Is it required to remove the ability to change the similarity function at runtime? 
This by the way breaks the `ElasticsearchVectorStoreAutoConfigurationIT` tests.

tzolov (changes requested):
Thanks @l-trotta , 
It looks good but there few things that needs clarification/fixing. Please check my comments. 
Looking forward for merging this improvement.

l-trotta:
@tzolov Thank you for the review, I'll explain my changes:

- Removal of the `withSimilarityFunction` setter: in elasticsearch the similarity function is defined as a mapping property of the index and cannot be changed, trying to update the property in the mapping will result in the following error: 
  ```
  {
    "error": {
      "root_cause": [
        {
          "type": "illegal_argument_exception",
          "reason": """Mapper for [embedding] conflicts with existing mapper:
	  Cannot update parameter [similarity] from [dot_product] to [cosine]"""
        }
      ],
      "type": "illegal_argument_exception",
      "reason": """Mapper for [embedding] conflicts with existing mapper:
	  Cannot update parameter [similarity] from [dot_product] to [cosine]"""
    },
    "status": 400
  }
  ```
  so the only way to change mapping is to create another index, and in this case a new vector store instance.
- Having `elasticsearch-java` explicitly defined in autoconfiguration: while testing the autoconfiguration I noticed it was importing an older version of the client  (8.10.4), which breaks the current implementation of the vector store since the knn query was recently fixed with version 8.13.3, so I wanted to make sure that this would be the version used, but maybe it's an issue with my local environment? If everything works without that it can be removed.

tzolov:
Thanks for the clarification @l-trotta !

It seems latest Spring Boot 3.3.x (the one we use for our auto-configs) is already at [Elasticsearch Client 8.13.4](https://github.com/spring-projects/spring-boot/blob/v3.3.0/spring-boot-project/spring-boot-dependencies/build.gradle). 
So I guess we are fine without the forced dependencies?  

Perhaps you can mention (in a separate PR) this requirement in the docs. E.g. if someone is using pre 3.3. Boot they have to add the elastic search `8.13.4` dependency to their poms?

tzolov:
rebased, squashed and merged at 78f3797f8d2c06cb429e6705f7b5eade6259dcf3

tzolov:
@l-trotta thank you for the great work! 
Just merged the PR. 
Please check if the dependencies work as expected and consider adding a NOTE in the documentation as suggested above.

l-trotta:
@tzolov thank you! I'll test with an older boot version, if there could be version problems I'll open another PR for the documentation.

ezimuel:
@tzolov do you know when this PR will be public released? We would like to prepare some materials to promote this integration for Elasticsearch. Thanks!

PR #599, opened 2024-04-18
Add Gemfire VectorStore Auto-Configuration

### LINKED ISSUE #488: Add Auto-configuration and Boot starter for the GemFire Vector Store
GemFire Vector Store: https://github.com/spring-projects/spring-ai/tree/main/vector-stores/spring-ai-gemfire

- Make sure to register the AutoConfig class to the [org.springframework.boot.autoconfigure.AutoConfiguration.imports](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-spring-boot-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports)
- Add the boot-starter dependency to the [BOM](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-bom/pom.xml).

PhilKes:
This PR closes #488 

- Adds Auto-Configuration for GemFire VectorStore
- Adds `GemFireVectorStoreProperties` to set all config parameters

## TODO:
As suggested in https://github.com/spring-projects/spring-ai/issues/488#issuecomment-2042738738 I tried to implement the integration tests with the https://hub.docker.com/r/gemfire/gemfire Image with a custom `docker-compose.yml` to add the [GemFire VectorDB extension](https://network.tanzu.vmware.com/products/gemfire-vectordb).

The first problem I encountered is that **you have to be logged in or provide an VMWare API Key in order to download the extension** (.gfm file), therefore I had to manually download an add the `.gfm` file to the repo.

The second problem is that the extension is mounted into the `/gemfire/extensions` which is successfully picked up in the class-path:
```
gemfire-gemfire-1  | Class Path:
gemfire-gemfire-1  |   /gemfire/lib/geode-core-9.15.11.jar
gemfire-gemfire-1  |   /gemfire/lib/geode-server-all-9.15.11.jar
gemfire-gemfire-1  |   /gemfire/extensions/vmware-gemfire-vectordb-1.1.0.gfm
```
But the [VectorDB REST API](https://docs.vmware.com/en/VMware-GemFire-VectorDB/1.0/gemfire-vectordb/rest_endpoints.html) is **only returning 404s**.
This setup works when I start GemFire locally and with the `.gfm` file in the `extensions` folder, I can then request e.g. `GET  http://localhost:7070/gemfire-vectordb/v1/indexes` and get a 200 with an empty list, but it does not work with the Docker image, does anybody have any suggestions why?
The general GemFire API does work with Docker (e.g. `GET http://localhost:7070/gemfire-api/v1/ping`), but not the VectorDB API.


PhilKes:
@markpollack could you take a lot at it? 🙂

PR #603, opened 2024-04-18
[WIP] Add ChatAgent and Evaluator

markpollack:
This is the bare bones of a higher level abstraction that can support various higher level functionality with an AI Model.

The main abstractions are a `ChatAgent` that takes uses multiple `PromptContextTransformer`s as collaborating objects.  A `PromptContextTransformer` is a function that takes a `PromptContext` as input and output.

The implementation DefaultChatAgent can handle many use cases.  It allows one to build an agent that can invoke functions (at the moment contained in the prompt) and execute various RAG algorithms.  Augmentation can come from Node implementations, which are Documents and Messages.  Augmenting Messages would be done for the case of including chat history into the prompt.

A fluent builder API for `DefaultChatAgent` puts `PromptContextTransformer`s into different collections so that they can be executed at the right time in the default workflow inside the agent's `call` method.  The DefaultChatAgent also contains a list of listeners that are invoked at the end of the ChatAgent's work so that state in external systems (such as a MessageHistory or MessageStore) can be updated.

There is a new `Evaluator` interface and one implementation `RelevancyEvaluator`

The test class `OpenAiDefaultChatAgentIT` minimally exercises this new functionality.

Open issues to consider:

* Adding a list of functions availalbe for the Agent to call can be achieved by transforming the prompt and adding function, but mayb
* PromptContext is currently mutable, this made implementation easy but we may want to consider returning a copy of a PromptContext from the transformer.
* MessageHistory has not been implemented.


tzolov:
Replaced by #634

PR #604, opened 2024-04-19
Fix bean name conflicts when using BedrockAnthropic and BedrockAnthropic3 at the same time.

wmz7year:
When enabling bedrock anthropic and anthropic3 at the same time, it will throw the bean name conflict exception.
```
spring:
  ai:
    bedrock:
      anthropic:
        chat:
          enabled: true
      anthropic3:
        chat:
          enabled: true

```

>
>The bean 'anthropicApi', defined in class path resource [org/springframework/ai/autoconfigure/bedrock/anthropic3/BedrockAnthropic3ChatAutoConfiguration.class], could not be registered. A bean with that name has already been defined in class path resource [org/springframework/ai/autoconfigure/bedrock/anthropic/BedrockAnthropicChatAutoConfiguration.class] and overriding is disabled.
>

So change the anthropic3 to a proper bean name to avoid this issue.


tzolov:
another good catch. Thanks @wmz7year

Issue #600, opened 2024-04-18
BedrockAnthropic3ChatClient returns ChatResponse without AnthropicUsage


lameroot:
I'am using BedrockAnthropic3ChatClient and i can not get info about input and output tokens. But these parameters i can get from AnthropicChatResponse.

Could you transfer these parameters to ChatResponseMetadata ?

tzolov:
This problem will resolved by the https://github.com/spring-projects/spring-ai/pull/605 indeed.

markpollack:
closing as the PR has been merged

PR #605, opened 2024-04-19
Bedrock AI models add usage information.

maxjiang153:
This PR closes: https://github.com/spring-projects/spring-ai/issues/600

When Bedrock invokes model API, Bedrock will return input, output token count, and latency information from the `SDKHttpResponse` header.

extract this information as `ChatResponseMetadata` to `ChatResponse`

### DIFF
--- a/models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/BedrockUsage.java
+++ b/models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/BedrockUsage.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.ai.bedrock;
 
+import org.springframework.ai.bedrock.api.AbstractBedrockApi.AmazonBedrockInvocationMetadata;
 import org.springframework.ai.bedrock.api.AbstractBedrockApi.AmazonBedrockInvocationMetrics;
 import org.springframework.ai.chat.metadata.Usage;
 import org.springframework.util.Assert;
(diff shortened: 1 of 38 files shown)


tzolov:
@wmz7year thank you for looking at this. It would be a valuable contribution to add more, structured usage metadata. 

From the PR code, I have the impression that we can implement (most of) this inside `AnthropicChatBedrockApi.java` without having to expose it via the  AmazonBedrockInvocationContex and reduce some necessary code repetition. 

For example It looks like the `BedrockAnthropicChatResponseMetadata` , `BedrockAnthropic3ChatResponseMetadata`, `CohereChatResponseMetadata`, `BedrockAi21Jurassic2ChatResponseMetadata`, `BedrockLlama2ChatResponseMetadata` and `BedrockTitanChatResponseMetadata` are the same? Then why not create one instance and use it inside the AnthropicChatBedrockApi to generate the metadata inside the `AnthropicChatBedrockApi`? 

What do you think about this approach?

wmz7year:
The idea behind `AmazonBedrockInvocationContext` is that the Amazon Bedrock invoke model API will return input/output tokens and latency information from HTTP headers, not only from the response body. So I wrap it up as an `AmazonBedrockInvocationContext` to include more information besides the API response.

Check this out:
https://github.com/spring-projects/spring-ai/blob/65e6880fa0f8db22ebcf780c3c840dc5e063b37f/models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/api/AbstractBedrockApi.java#L220

And also, the reason why I create multiple metadata classes like `BedrockAnthropicChatResponseMetadata` is because each model has a different response structure. Currently, it seems to look similar, but for example, the response message ID is different. So I decided to keep this up for future changes in each model.

How do you think about this? btw I can reduce metadata classes to a single one maybe we can add a dynamic field like `Map<String, Object> attribute` fields to cover the different metadata fields?

tzolov:
Thanks, @wmz7year I need to look into this further, but wont be able to do it in coming 2 weeks.

wmz7year:
> Thanks, @wmz7year I need to look into this further, but wont be able to do it in coming 2 weeks.

Sure, if you have any thoughts just let me know

maxjiang153:
re-implement with: https://github.com/spring-projects/spring-ai/pull/813

PR #606, opened 2024-04-19
Initial chatcall implementation

markpollack:
`ChatCall` encapsulates the steps of using `ChatClient` into an object named `ChatCall`.  The `ChatCall` is created using a fluent-api, often registered as a Spring `@Bean` and then calling an `execute` method with some runtime parameters, such as the variables to substitute into the prompt.

Similar to Spring's `RdbmsOperation` interface, which models RDBMS operations as objects with the `SqlCall` implementation handling SQL-based calls like stored procedures or functions, and akin to `JdbcClient` offering a fluent API for JDBC operations, we introduce a helper class named `ChatCall`.

This class aims to streamline the process of gathering all necessary parameters for making calls to an AI model. It provides an easily shareable object encapsulating various options for calling an AI model and also helping to more easily chaining multiple calls to the model.

```java
ChatCall chatCall = new ChatCall.Builder().
                                    .withChatClient(chatClient)
                                    .withSystemMessage("you are a helpful assistant") // not likely to change
                                    .withSystemMap(Map.of)  // likely to change at runtime
                                    .withUserMessage("tell me a joke")  // not likely to change
                                    .withUserMap(Map.of)  // likely to change at runtime
)                                   .withChatResponseMapper(chatResponseMapper) // not likely to change
                                    .withChatOptions(chatOptions)  // only likely to change during development
                                    .build()

String result = chatCall.execute(Map.of)  //User map
String result = chatCall.execute(Map.of, Map.of)  //User and System map
String result = chatCall.execute("user message", Map.of);
String result = chatCall.execute("user message", Map.of, "system message", Map.of);
<T> T result  = chatCall.execute("user message", Map.of, Class<T> returnType)
<T> T result  = chatCall.execute("user message", Map.of, Class<T> returnType, "system message", Map.of)

ChatResponse chatResponse = chatOperation.call()
```

open issues

* Change execute signature to have `T...` as last parameter so don't have to specify `MyDomainObject.class`
* Ensure support for returning collections of objects, need to use `ParameterizedTypeReference`

### DIFF
--- a/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/chat/OpenAiChatCallIT.java
+++ b/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/chat/OpenAiChatCallIT.java
@@ -1,0 +1,147 @@
+package org.springframework.ai.openai.chat;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+import org.springframework.ai.chat.ChatClient;
+import org.springframework.ai.chat.call.ChatCall;
+import org.springframework.ai.chat.prompt.ChatOptions;
+import org.springframework.ai.chat.prompt.ChatOptionsBuilder;
+import org.springframework.ai.openai.OpenAiChatClient;
+import org.springframework.ai.openai.api.OpenAiApi;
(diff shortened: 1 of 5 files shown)


markpollack:
replaced by fluent api added to ChatClient

PR #618, opened 2024-04-22
Fix CassandraVectorStoreAutoConfiguration's @ConditionalOnClass to not include EmbeddingClient

michaelsembwever:
@tzolov discovered this slip-up (from me) while working on a use case

### DIFF
--- a/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/cassandra/CassandraVectorStoreAutoConfiguration.java
+++ b/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/cassandra/CassandraVectorStoreAutoConfiguration.java
@@ -36,7 +36,7 @@ import org.springframework.context.annotation.Bean;
  * @since 1.0.0
  */
 @AutoConfiguration(after = CassandraAutoConfiguration.class)
-@ConditionalOnClass({ CassandraVectorStore.class, EmbeddingClient.class, CqlSession.class })
+@ConditionalOnClass({ CassandraVectorStore.class, CqlSession.class })
 @EnableConfigurationProperties(CassandraVectorStoreProperties.class)
 public class CassandraVectorStoreAutoConfiguration {
 


spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/cassandra/CassandraVectorStoreAutoConfiguration.java
```diff
@@ -36,7 +36,7 @@
  * @since 1.0.0
  */
 @AutoConfiguration(after = CassandraAutoConfiguration.class)
-@ConditionalOnClass({ CassandraVectorStore.class, EmbeddingClient.class, CqlSession.class })
+@ConditionalOnClass({ CassandraVectorStore.class, CqlSession.class })
```
eddumelendez, line 39: If it is not required then `EmbeddingClient embeddingClient` should be wrapped with `ObjectProvider`

```java
public CassandraVectorStore vectorStore(ObjectProvider<EmbeddingClient> embeddingClient, CassandraVectorStoreProperties properties,
			CqlSession cqlSession) {
```

otherwise, it will be a required bean at it will fail if not present

michaelsembwever: that was what I first thought.

but when you specify an EmbeddingClient (in a constructor) so that that bean is instantiated, then the CassandraVectorStore bean is automatically instantiated too.

that isn't desired.  a bean that's not needed shouldn't be instantiated.

this code works with this patch.  note that i don't declare or configure the EmbeddingClient bean anywhere…
https://github.com/datastax/ai-agent-java/blob/main/src/main/java/com/datastax/ai/agent/AiApplication.java 

I also noted that other StoreAutoConfigurations don't have it:
e.g. https://github.com/spring-projects/spring-ai/blob/main/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/pgvector/PgVectorStoreAutoConfiguration.java 
(also HanaDb)

eddumelendez: Please ignore my first comment, you don't need to do that unless the bean is not available. However, in the project the bean is created because [spring-ai-openai-spring-boot-starter is present](https://github.com/datastax/ai-agent-java/blob/main/pom.xml#L50-L54). I wonder if `@ConditionalOnBean(EmbeddingClient.class)` is needed.

PR #627, opened 2024-04-24
Add AWS Bedrock Meta LLama3 AI model support.

wmz7year:
AWS Bedrock just released the Meta Llama3 model:  https://aws.amazon.com/about-aws/whats-new/2024/04/meta-llama-3-amazon-bedrock/

Meta Llama3 and Llama2 models have the same request parameters and responses: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-meta.html

So I made a break change from llama2 to llama to support both versions of these models.


tzolov:
Thanks @wmz7year 
Rebased, re-enable some llama output parser ITs, squash and merge at 9e865f0b0f9aa03360d59f2974d0e1620d9fba77

PR #633, opened 2024-04-25
ElasticsearchVectorStore add support for configuration.

wmz7year:
Add ElasticsearchVectorStore support for creating index configuration. such as `dims`, `denseVectorIndexing` and `similarity`.

Also, include support for adding documents to ES with a failed response; this change will result in an exception.


tzolov:
Thanks @wmz7year , 
We're still considering adding the concept of runtime options to the VectorStores (similar to the model clients). Anyway this PR is a step in this direction if we get there eventually. 

@

tzolov:
Rebased, squashed and merged at f90c51a72937680f77f9072080b19b84766b5340

Additionally rename the dims property into dimension.
Add javadoc and antora ES properties documentation.

PR #634, opened 2024-04-25
[WIP] Chat agent 2

tzolov:


spring-ai-core/src/main/java/org/springframework/ai/chat/history/ChatMemory.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public interface ChatMemory {
+
+	default void add(String conversationId, Message messages) {
+		this.add(conversationId, List.of(messages));
+	}
+
+	void add(String conversationId, List<Message> messages);
+
+	List<Message> get(String conversationId);
```
michaelsembwever, line 34:
```suggestion
	List<Message> get(String conversationId, int lastN);
```
the limit will want to be a pushed-down predicate, e.g. in non-memory impls.

spring-ai-core/src/main/java/org/springframework/ai/evaluation/RelevancyEvaluator.java
```diff
@@ -0,0 +1,82 @@
...
+		return evaluationRequest.getChatResponse().getResult().getOutput().getContent();
+	}
+
+	protected String doGetSupportingData(EvaluationRequest evaluationRequest) {
+		List<Content> data = evaluationRequest.getDataList();
+		String supportingData = data.stream()
+			.filter(node -> node != null && node.getContent() instanceof String)
+			.map(node -> (Content) node)
+			.map(Content::getContent)
+			.collect(Collectors.joining("\n"));
```
thesurlydev, line 69:
```suggestion
			.collect(Collectors.joining(System.lineSeparator()));
```

spring-ai-core/src/main/java/org/springframework/ai/tokenizer/JTokkitTokenCountEstimator.java
```diff
@@ -0,0 +1,85 @@
...
+
+			for (Media media : content.getMedia()) {
+
+				tokenCount += this.estimate(media.getMimeType().toString());
+
+				if (media.getData() instanceof String textData) {
+					tokenCount += this.estimate(textData);
+				}
+				else if (media.getData() instanceof byte[] binaryData) {
+					tokenCount += binaryData.length; // This is likely incorrect.
```
thesurlydev, line 68: Jtokkit is specific to OpenAI. This might be more accurate by:

1. Normalizing whitespace and split based on spaces and punctuation.
2. Replacing Unicode non-breaking spaces and trimming excess whitespace.

spring-ai-core/src/test/java/org/springframework/ai/chat/history/ChatHistoryTests.java
```diff
@@ -0,0 +1,152 @@
...
+
+		AgentResponse response1 = chatAgent.call(promptContext);
+
+		assertThat(response1.getChatResponse().getResult().getOutput().getContent()).isEqualTo("assistant:1");
+
+		// List<Message> messages2 = promptCaptor.getValue().getInstructions();
+		// assertThat(messages2)
+		// 		.isEqualTo(List.of(new UserMessage("user:1"), new UserMessage("user:2"), new UserMessage("user:3"),
+		// 				new UserMessage("user:4"), new UserMessage("user:5")));
+
```
thesurlydev, line 113:
```suggestion
```

spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/Prompt.java
```diff
@@ -92,4 +96,31 @@ public int hashCode() {
...
+			}
+			else if (message instanceof AssistantMessage) {
+				messagesCopy.add(new AssistantMessage(message.getContent(), message.getMetadata()));
+			}
+			else if (message instanceof FunctionMessage) {
+				messagesCopy.add(new FunctionMessage(message.getContent(), message.getMetadata()));
+			}
+			else {
+				throw new IllegalArgumentException("Unsupported message type: " + message.getClass().getName());
+			}
```
thesurlydev, line 120: There are several areas in the code where the list of messages are being filtered. Would it make sense to introduce a map representation of messages by type? Perhaps something like: 
```
Map<Class<? extends Message>, List<Message>> messagesByType
```

spring-ai-core/src/main/java/org/springframework/ai/chat/history/ChatMemory.java
tzolov: Agree. Current ChatMemory is likely very naive and the moment we try to implemnt a Repository around it I guess more changes will come.

spring-ai-core/src/main/java/org/springframework/ai/tokenizer/JTokkitTokenCountEstimator.java
tzolov: @thesurlydev, can you please elaborate a bit more, perhaps with code examples? 
Btw, we need to add the metdata to the estimations as well. 

Do you know any alternatives of Jtokiit?

spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/Prompt.java
tzolov: Not sure I completely undestand the suggestion?  Where is this map going to be? Also the order of the messages is important and must be preserved (e.g. User followed by Assistant then another pair of user/assistant ... ).

spring-ai-core/src/main/java/org/springframework/ai/tokenizer/JTokkitTokenCountEstimator.java
thesurlydev: Jtokkit may be good enough for now. Here's some (untested) code examples:
```
public static int estimateTokenCount(byte[] byteArray) {
    String text = new String(byteArray, StandardCharsets.UTF_8);
    // Normalize whitespace and split based on spaces and punctuation
    String[] tokens = normalizeText(text).split("\\s+|(?=\\p{Punct})|(?<=\\p{Punct})");
    return tokens.length;
}

public static String normalizeText(String text) {
    // Replace Unicode non-breaking spaces and trim excess whitespace
    return text.replace("\u00A0", " ").trim();
}
```

spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/Prompt.java
thesurlydev: Using a `LinkedHashMap` as the Map implementation would preserve ordering. I was suggesting this in place of or in addition to the `List<Message>` as part of `Prompt`.

spring-ai-core/src/main/java/org/springframework/ai/chat/history/InMemoryChatMemory.java
```diff
@@ -0,0 +1,50 @@
...
+
+	@Override
+	public void add(String conversationId, List<Message> messages) {
+		this.conversationHistory.putIfAbsent(conversationId, new ArrayList<>());
+		this.conversationHistory.get(conversationId).addAll(messages);
+	}
+
+	@Override
+	public List<Message> get(String conversationId, int lastN) {
+		List<Message> all = this.conversationHistory.get(conversationId);
```
michaelsembwever, line 41: of note: persistent implementations will need some way of serialising and deserialising messages.

the best i have for now is something along the lines of 
```
            MessageType messageType = MessageType.valueOf(row.getString("messageType"));
            String messageContent = row.getString("messageContent");

            Message message = new AbstractMessage(messageType, messageContent) {};
```

spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/transformer/VectorStoreRetriever.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Transforms the PromptContext by retrieving documents from a VectorStore
+ */
+public class VectorStoreRetriever implements PromptTransformer {
+
+	private final VectorStore vectorStore;
+
+	private final SearchRequest searchRequest;
+
+	public VectorStoreRetriever(VectorStore vectorStore, SearchRequest searchRequest) {
```
habuma, line 38: Was just kicking the tires on this PR and this constructor tripped me up a bit because it requires a `SearchRequest` as a parameter. 

My thinking was that I'd declare `ChatAgent` as a bean and then inject and use it in place of `ChatClient`. But you won't know how to create a `SearchRequest` at the time the application context is being created, so...

Is the intention that  instead of as a bean, the `ChatAgent` (specifically, the `DefaultChatAgent`) be created at the time that the query is being handled, because at that time you'd know how to build the `SearchRequest`?

spring-ai-core/src/main/java/org/springframework/ai/chat/history/InMemoryChatMemory.java
michaelsembwever: the alternative approach i see persistent stores taking is storing all the types of messages explicitly in the same row,
e.g.
```
            String assistant = r.getString("assistant_column");
            String user = r.getString("user_column");
            messages.add(new UserMessage(user));
            messages.add(new AssistantMessage(assistant));
```
this approach requires the properties in the messages to accessible and mutable (and you have done in this PR), so to be able to keep a "conversationId" that correlates the above messages.

spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/transformer/VectorStoreRetriever.java
habuma: Thanks to @markpollack , I was made aware of `SearchRequest.defaults()` which lets me create a `VectorStoreRetriever` as a bean rather than in the controller. And, as Mark said, that same bean could use a `SearchRequest` with other filter info rather than the defaults, but I can still create it as a bean. So I withdraw my question.

PR #636, opened 2024-04-25
Fix sample in "Manual Configuration" section of OpenAI Embeddings API

izeye:
This PR fixes the sample in the "Manual Configuration" section of OpenAI Embeddings API.


tzolov:
Thanks @izeye

PR #640, opened 2024-04-26
Fix column creation when adding additional normal and embedding columns, and make index name unique (for when there are multiple vector indexes in the same keyspace)

michaelsembwever:
Fix column creation when adding additional normal and embedding columns, and make index name unique (for when there are multiple vector indexes in the same keyspace).
…

And change stream to for-loop when converting List<Double> to Float[] for performance


tzolov:
Thanks @michaelsembwever 
rebased and merged at daf131be2cc668fe8f1b536b3025dff14e1d231f

PR #647, opened 2024-04-28
Docs: Fix typos in embedding and function calling documentation

ThomasVitale:
Fix typos in the function calling and embedding model documentation for specific models.

There are many duplicated sections across the different model docs. It might be worth looking into extracting those sections into AsciiDoc fragments to re-use them (and simplify the maintenance).


tzolov:
Thank you @ThomasVitale

Issue #619, opened 2024-04-22
Gemini Function call with multiple functions doesn't work


Grogdunn:
**Bug description**
If you define multiple function in a "call" with Gemini, lead to INVALID_ARGUMENT server error.

**Environment**
tested on spring-ai 0.8.1 bug spotted on code of 1.0.0-SNAPSHOT

**Steps to reproduce**
Pass multiple function to a "VertexAICliente.call"

**Expected behavior**
No error on server, hopefully all function called.

From a trial Vertex seems support one tool with multiple function, not multiple tools with a single funcion each. So this function "getFunctionTools" must produce a single tool with all function inside
https://github.com/spring-projects/spring-ai/blob/563f9b48af60a176b32ee4fa03877123f076b9a7/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatClient.java#L350

tzolov:
@Grogdunn, thank you for the feedback. 
Can you please write a reproducible example?

Grogdunn:
Sure, when I've spere time I'll try to fix it, but meanwhile put that test in `VertexAiGeminiChatClientFunctionCallingIT`
```java
 @Test
    public void apiFail() {

        UserMessage userMessage = new UserMessage(
                "What's the weather like in San Francisco, in Paris and in Tokyo, Japan? Use Multi-turn function calling. Provide answer for all requested locations.");

        List<Message> messages = new ArrayList<>(List.of(userMessage));

        final var weatherFunction = FunctionCallbackWrapper.builder(new MockWeatherService())
                .withSchemaType(SchemaType.OPEN_API_SCHEMA)
                .withName("getCurrentWeather")
                .withDescription("Get the current weather in a given location")
                .build();
        final var theAnswer = FunctionCallbackWrapper.<String, Integer>builder(o -> 42)
                .withSchemaType(SchemaType.OPEN_API_SCHEMA)
                .withName("theAnswerToTheUniverse")
                .withDescription("the answer to the ultimate question of life, the universe, and everything")
                .build();
        var promptOptions = VertexAiGeminiChatOptions.builder()
                .withModel(VertexAiGeminiChatClient.ChatModel.GEMINI_PRO.getValue())
                .withFunctionCallbacks(List.of(weatherFunction, theAnswer))
                .build();

        Flux<ChatResponse> response = vertexGeminiClient.stream(new Prompt(messages, promptOptions));

[...]

PR #651, opened 2024-04-29
Currently Vertex API uses one Tool per function definition, but tools can have multiple function inside

Grogdunn:
as described in issue #619 There are an API error with Gemini-Vertex API with multiple functions.

The "model" can admit multiple tools with multiple function inside. But if you specify two or more tools the Gemini API throw an 400 error.

Avoid this generating a single tool with all function inside.


tzolov:
Yeh, i appears that Gemini Pro model has broken  functional calling support. 
You can non anymore specify function names like "getWeatherInfo" as the model rewrites those into get_weather_info".
Furthermore the  parallel function call is gone as well. 
Very disappointing. Not sure if there is any Google Gemini channel to report those issues.

tzolov:
I've updated the PR to make at least some of the IT test to work.  
Also added a document warning about the functionality degradation: https://docs.spring.io/spring-ai/reference/1.0-SNAPSHOT/api/chat/vertexai-gemini-chat.html#_function_calling

tzolov:
Merged at 9f32a3ca9cd119116b112e050b942e5e94855914

Issue #652, opened 2024-04-29
Avoid second roundtrip when call functions


Grogdunn:
**Expected Behavior**
For some customers we do not have necessity to pass the function-result to LLMs again. 

Eg: extract some information from a blob of text and save it in structured mode. That can be accomplished with a function call with in input the "data structure" but the output is not necessary, so we can avoid totally the "second call" to LLM.

**Current Behavior**
Every function must return a value that is passed to LLM to produce a nice output.

**Context**
As workaround we implement some function with "spying" capability (like tests) and use it to grab the "first call" data. And ignore completly the second call-result.

We try to add this feature in spring-ai I think that can be useful to others. Or at least anyone can decide.

tzolov:
@Grogdunn , are you referring to providing support for `Function<SomeRequest, Void>`? E.g. function with Void response? 
If so shouldn't the implementation detect the Void response definition and handle it accordingly? rather than masking it via fake string response and additional property flags?

Grogdunn:
Ok! I'll try good point

Grogdunn:
@tzolov Ok Done, if you use a Function<Input, Void> or a Consumer<Input> the second round-trip is skipped

markpollack:
The PR #656 has some discussion, but perhaps this can be solved via clever prompt engineering.  A prompt such as `Respond only with required function calls and a single word "yes" or "no". Do not provide explanations or additional text.`  may work since you want to suppress the final response. 

The contract with AI models and function calling is a contract driven by the AI model itself, so one does need to reply to finish the conversation.   

The lower level API that lets you control the conversation is demonstrated [here](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/chat/OpenAiChatModelProxyToolCallsIT.java).    Perhaps after the first call you can simply not reply. Don't know what state this will leave things in but you can experiment with these two approaches.

Grogdunn:
For me this issue can be closed (as pull request). 

There are 2 workarounds to handle this:
1. using low level API
2. using `org.springframework.ai.converter.BeanOutputConverter` with a correct prompt

Option 1 is less prone to hallucination, but "disrupts" the standard AI workflow/contract.
Option 2 can have some hallucination but it's clear, and respects the contract with the AI.

So feel free to close this issue.

PR #656, opened 2024-04-29
[FEATURE] avoid second round trip for function call

Grogdunn:
Reefer issue #652 

For some use case after local function is called no need to pass the results to LLMs. (eg: data extraction from documents)

### DIFF
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java
@@ -28,6 +28,7 @@ import com.azure.ai.openai.models.ChatRequestMessage;
 import com.azure.ai.openai.models.ChatRequestSystemMessage;
 import com.azure.ai.openai.models.ChatRequestToolMessage;
 import com.azure.ai.openai.models.ChatRequestUserMessage;
+import com.azure.ai.openai.models.ChatResponseMessage;
 import com.azure.ai.openai.models.CompletionsFinishReason;
 import com.azure.ai.openai.models.ContentFilterResultsForPrompt;
 import com.azure.ai.openai.models.FunctionCall;
@@ -50,6 +51,7 @@ import org.springframework.ai.chat.prompt.ChatOptions;
 import org.springframework.ai.chat.prompt.Prompt;
 import org.springframework.ai.model.ModelOptionsUtils;
 import org.springframework.ai.model.function.AbstractFunctionCallSupport;
+import org.springframework.ai.model.function.FunctionCallback;
 import org.springframework.ai.model.function.FunctionCallbackContext;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
@@ -514,6 +516,15 @@ public class AzureOpenAiChatClient
 	}
 
 	@Override
+	protected boolean hasReturningFunction(ChatRequestMessage responseMessage) {
+		return ((ChatRequestAssistantMessage) responseMessage).getToolCalls()
+			.stream()
+			.map(toolCall -> ((ChatCompletionsFunctionToolCall) toolCall).getFunction().getName())
+			.map(functionName -> Optional.ofNullable(this.functionCallbackRegister.get(functionName)))
+			.anyMatch(functionCallback -> functionCallback.map(FunctionCallback::returningFunction).orElse(false));
+	}
+
+	@Override
 	protected ChatCompletionsOptions doCreateToolResponseRequest(ChatCompletionsOptions previousRequest,
 			ChatRequestMessage responseMessage, List<ChatRequestMessage> conversationHistory) {
 
@@ -530,8 +541,10 @@ public class AzureOpenAiChatClient
 
 			String functionResponse = this.functionCallbackRegister.get(functionName).call(functionArguments);
 
-			// Add the function response to the conversation.
-			conversationHistory.add(new ChatRequestToolMessage(functionResponse, toolCall.getId()));
+			if (functionResponse != null) {
+				// Add the function response to the conversation.
+				conversationHistory.add(new ChatRequestToolMessage(functionResponse, toolCall.getId()));
+			}
 		}
 
 		// Recursively call chatCompletionWithTools until the model doesn't call a
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatClient.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatClient.java
@@ -20,6 +20,7 @@ import java.util.Base64;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
@@ -48,6 +49,7 @@ import org.springframework.ai.chat.prompt.ChatOptions;
 import org.springframework.ai.chat.prompt.Prompt;
 import org.springframework.ai.model.ModelOptionsUtils;
 import org.springframework.ai.model.function.AbstractFunctionCallSupport;
+import org.springframework.ai.model.function.FunctionCallback;
 import org.springframework.ai.model.function.FunctionCallbackContext;
 import org.springframework.ai.retry.RetryUtils;
 import org.springframework.http.ResponseEntity;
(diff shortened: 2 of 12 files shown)


models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java
```diff
@@ -15,10 +15,7 @@
  */
 package org.springframework.ai.azure.openai;
 
-import java.util.Collections;
```
tzolov, line 18: Wildcard imports such as import java.util.* or import static org.assertj.core.api.Assertions.* are forbidden, even in test code: https://github.com/spring-projects/spring-framework/wiki/Code-Style#import-statements

models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java
```diff
@@ -443,8 +442,11 @@ protected ChatCompletionsOptions doCreateToolResponseRequest(ChatCompletionsOpti
 
 			String functionResponse = this.functionCallbackRegister.get(functionName).call(functionArguments);
 
-			// Add the function response to the conversation.
-			conversationHistory.add(new ChatRequestToolMessage(functionResponse, toolCall.getId()));
+			if (functionResponse != null) {
```
tzolov, line 445: I believe that the CompleteRoundTripBox and the runtime detection of needCompleteRoundTrip are unnecessary.
Since in the FunctionCallbackContext you have already detected Void response (and Consumer)  functions it should be possible add new `boolean noReturnFunction()` method to the `FunctionCallback` interface. Then in the  AbstractFunctionCallSupport#handleFunctionCallOrReturn you should be able to use this to decide how to handle the return? 
This way you can implement everything inside the abstract support classes without changing the Model implementations?

tzolov (changes requested):
Thank you @Grogdunn, 
I believe we are getting closer and this will be very useful function calling feature. 

Still I believe that the `CompleteRoundTripBox` and the runtime detection of `needCompleteRoundTrip` are unnecessary.
Since in the `FunctionCallbackContext` you have already detected `Void` response (and `Consumer`)  functions it should be possible add new `boolean noReturnFunction()` method to the `FunctionCallback` interface. 
Then in the  `AbstractFunctionCallSupport#handleFunctionCallOrReturn` you should be able to use this to decide how to handle the return? 
This way you can implement everything inside the abstract support classes without changing the Model implementations?
Maybe I'm missing some detail? 
Let me know if you would be interested to explore the suggested approach? or should I give it a try.
Thank you very much for the contribution!

models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java
Grogdunn: removed all star imports from all sources

models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java
Grogdunn: yes but in model implemetation we "extract" the function to call and call it. If the model support more than one function or we change the implementation to extract all involved functions and in AbstractFunctionCallback we decide if a second roundtrip is needed... 

Try to clean model implementation :thinking:

tzolov:
@Grogdunn i'm afraid you've committed many changes unrelated to this PR. 
Please fix the PR or create new one that includes only the related changes. 

To prevent such situations, please don't reformat the project code. Especially code not related to the PR. 
Also use pull rebase requests (`e.g. git pull -r upstream main`) to rebase to the top of main. Never merge with upstream main.

Grogdunn:
@tzolov removed commit with all import * fixed and re-played only for touched subproject.
rebased on top on upstream master (as requested)

conflicts are gone! :tada:

tzolov:
Thanks @Grogdunn.
Still the majority of the changes (> 100 files)  in the PR are code reformatting unrelated to the PR goal. For example import reordering in AnthropicChatOptions.java, AnthropicApi.java ....
I'm presuming that actual changes would affect much fewer files.
Can you please undo those changes (i guess your IDE is doing them automatically) or copy only the related changes in new PR.

Grogdunn:
Sure! Today I'll try to reduce PR changeset.

Grogdunn:
Ok, squashed, rebased onto main, cleaned up.

tzolov:
Hi @Grogdunn , 
I've been reviewing the PR and doing some [improvements](https://github.com/tzolov/spring-ai/tree/gh-656-pr-2) when I realised that this approach (and likely any approach) will be incorrect. 

First, I noticed that the PR doesn't support streaming function calling and while reasoning about ways to provide such support I realised that we are on a wrong path. 

Basically we can not safely abruptly stop the function calling message exchange  and leave the conversation in a safe state. 

Simply we don't know what other actions the LLM might need to perform before safely answering the user prompt.  
For example if you your prompt is " Cancels my flight and let me know what is the weather in Amsterdam" and the `Function<FlightToCancel, Void>` doesn't return value. The PR will abruptly stop the conversation and will not allow the  LLM to call the second WeatherInfo function and complete the conversation. 
It gets even more complicated for streaming case. 

So IMO the only way this can ever work is if the LLM Function Calling API provide a protocol to send `EmptyResponse` for a Tool call. Any other heuristics will be incorrect. 

Please let me know what do you think.

Grogdunn:
:thinking: 
Well the PR was made before the other PR with function streaming come to main, so I can port the changes to function-streaming.

I think that is possible because before call local function we wait for streaming-stop signal (I don't remeber the correct terms), then the stream stops and we call the local functions.
After that we have, with your example, two results: 1 - `null` and 2 - `{"temperature": 30, "sky": "sunny", "blabla": "lorem ipsum"}`
So we evalutate all the local responses and complete the second roundtrip because we have the answer number 2.

I try to spike that today.

tzolov:
@Grogdunn , the problem is not related to the streaming but to the sync calls as well. 
Simply, current LLM API demands that you should always return a function call result message back to the  LLM. Even for a void functions. 
Perhaps a way to emulate a void function call behaviour is to  return an empty string  or something like "DONE" in case of void functions.

Grogdunn:
Well in terms of conversation you are right. But you can use LLM to make something else instead of "chatbot". Some of my customers need to grab structured data from unstructured data, like name, surname, other information from emails, attachments, extract information from datasheet ad so on... 
In this case we do not have a "conversation" but a single-shot request with a lot of context that ends with the function call.

A simple prompt can be, for instance:
```
U: Extract name, surname, emotional state, address, city, and summarize the following text, and call the function SAVEITNOW with extracted data:
--
Lorem ipsum dolor sit amet...[a lot of text]
```
the function is described as usual. No other interaction with this data will be done in the future.

Grogdunn:
I add, in my case, the second roundtrip is useless, and make double the costs for any interaction.

tzolov:
I understand your use case and the goal to reduce the cost. 
But the solution is sort of  a hack that breaks the LLM APIs/protocols. Therefore we will be reluctant to add such hacks to the general purpose Spring AI. 
Perhaps the [Structured Output Converter](https://docs.spring.io/spring-ai/reference/1.0-SNAPSHOT/api/structured-output-converter.html) can help to returns one-shot structured response for other use case? 
Other options is to think of some extension/customisation points that might allow one hack the spring ai function calling API in their applications. 

At the same time i've reworked our PR to returns "DONE" in case of void function definition, which is still a valid user case and modified the test  like this:
https://github.com/tzolov/spring-ai/blob/8f832e337c6d4fc7a244de68cd61c5bd8e0c5876/models/spring-ai-azure-openai/src/test/java/org/springframework/ai/azure/openai/function/AzureOpenAiChatClientFunctionCallIT.java#L123

This seems to work (most of the time) for sync and stream calls.

markpollack:
While doing triage for the M1 release, we will move this issue to the M2 release.  There is a Pr that @tzolov will submit that takes this discussion further.  Thanks for your patience.

Grogdunn:
I've done some experiment with structured output, and with GPT-4o simply works. But remain hallucination prone.

Maybe that feature is not necessary anymore?

markpollack:
I'm afraid I have to bump this again, apologies.

tzolov:
@Grogdunn  I think that https://github.com/spring-projects/spring-ai/commit/501774925c809fb47e02c73688092e46cdb78099 might help addressing this issue? 
Please, check the [OpenAiChatModelProxyToolCallsIT.java](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/chat/OpenAiChatModelProxyToolCallsIT.java) for examples how to run the function calling entirely on the client side. 
Let me know if you have further questions.

Grogdunn:
@tzolov Thanks for the hint! I see the intent of this commit and if I understand well is what I need! I try to use as soon as possible

markpollack:
Going to close this issue.  Please open a new one should the current feature set not work out.  Thanks @Grogdunn

PR #676, opened 2024-05-04
Docs: Fix filter expression reference to Elasticsearch Query string query

JM-Lab:
I fixed a mistake in our documentation that incorrectly mentioned SQL's WHERE. We actually use Elasticsearch's Query string query. This update makes sure our guide is right.


tzolov:
Thanks @JM-Lab

PR #680, opened 2024-05-05
fix : use ollamacontainer endpoint

rajadilipkolli:
Instead of manually setting the url use containers OOTB URL

### DIFF
--- a/models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatClientIT.java
+++ b/models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatClientIT.java
@@ -73,7 +73,7 @@ class OllamaChatClientIT {
 		ollamaContainer.execInContainer("ollama", "pull", MODEL);
 		logger.info(MODEL + " pulling competed!");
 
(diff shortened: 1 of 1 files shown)


models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatClientIT.java
```diff
@@ -73,7 +73,7 @@ public static void beforeAll() throws IOException, InterruptedException {
 		ollamaContainer.execInContainer("ollama", "pull", MODEL);
 		logger.info(MODEL + " pulling competed!");
 
-		baseUrl = "http://" + ollamaContainer.getHost() + ":" + ollamaContainer.getMappedPort(11434);
+		baseUrl = ollamaContainer.getEndpoint();
```
f3rnandomoreno, line 76: you are smashing the host configuration. I think would be better to allow both configurations. If the endpoint is set, the get the endpoint, but if the endpoint is not set, then leave the code as it is to allow setting by host

rajadilipkolli: Hi @f3rnandomoreno , since this is using `TestContainers` under the hood, getEndPoint() will set the host. No need to explicitly set it. 

Also, getEndpoint() cannot be null if container is started

PR #682, opened 2024-05-05
Expose `QdrantClient` and `WeaviateClient` as beans

eddumelendez:
Currently, `QdrantClient` and `WeaviateClient` are not exposed as beans. Having access to those would benefit to perform operations with an already configured client.

This would be great for having a actuator module with HealthContributor for each integration.

This is a breaking change but IMHO it is fine for 1.0.0.


tzolov:
Thank you @eddumelendez , 
The QdrantVectorStoreCloudAutoConfigurationIT is failing for me

tzolov:
Thank you again @eddumelendez 
I've added some documentation updates for the qdrant and weaviate stores.

tzolov:
Rebased, squashed and merged at 5beef21a4eca4fff54a4a8395419fbf36acf60cc
Additionally deprecate the qdrant config class and update the qdrant and weavient documentation.

PR #686, opened 2024-05-06
Vector Store polish

joshlong:
general low hanging optimization and polish
- refactored the `VectorStore` implementations so that if they implemented `InitializingBean`, they now implement `ApplicationListener<ApplicationReadyEvent>`, instead. The result is that the vector stores now initialize their own state _after_ the boot app has started. this is important because a) nobody wants to wait 2x as long for their app to run SQL statements or create collections or whatever, especially when developing, and nobody wants their Kubernetes health check to fail for stuff that can be done after the application's fully online. 
- who is the author of `ChromaVectorStore`? Should we add their author tag? do we need ChromaVectorStore#SIMILARITY_THRESHOLD_ALL and ChromaVectorStore#SIMILARITY_THRESHOLD_ALL
- renamed the module/artifact for `spring-ai-azure-vector-store` to `spring-ai-azure-store` to match other vector stores
- renamed a few directories to match their `artifactId`s 
- made sure `pom.xml` `name` elements looked consistent  (try running `./mvnw clean` and you'll see the output is nicer)


tzolov (changes requested):
I like the idea of replacing the `InitializingBean ` with the `ApplicationListener<ApplicationReadyEvent>`. Unfortunately this change breaks the IT tests. I'm not sure if this is  test configuration issue or a problem with the lifcycle initializaiton.

markpollack:
Half of the changes in the PR were applied, however the approach to initializing the vector store though the use of a Spring Boot application event is concerning as ideally we do not want to take a dependency in the vector stores on Spring Boot.  There is perhaps another approach that can be used.  Moving this to M2 to look into it more deeply another time.

markpollack:
I think all topics/feature changes in this PR have been addressed except for the default in the common properties

```
public class CommonVectorStoreProperties {

	private boolean initializeSchema = true;

     ...
}
```

Closing this issue and opening up another now to address changing the default initialization from true to false.

markpollack:
https://github.com/spring-projects/spring-ai/issues/907 is the new issue.

PR #689, opened 2024-05-06
Fix AWS Bedrock Cohere embedding truncate type error.

wmz7year:
According to the AWS Bedrock cohere embedding document: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html

The request parameter `truncate` should have the following values:
```
truncate – Specifies how the API handles inputs longer than the maximum token length. Use one of the following:

NONE – (Default) Returns an error when the input exceeds the maximum input token length.

START – Discards the start of the input.

END – Discards the end of the input.

If you specify START or END, the model discards the input until the remaining input is exactly the maximum input token length for the model.
```

But currently we have `LEFT` and `RIGHT` values. It will be throw the exception below:

```
Caused by: software.amazon.awssdk.services.bedrockruntime.model.ValidationException: Malformed input request: #/truncate: LEFT is not a valid enum value, please reformat your input and try again. (Service: BedrockRuntime, Status Code: 400
```

### DIFF
--- a/models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/cohere/api/CohereEmbeddingBedrockApi.java
+++ b/models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/cohere/api/CohereEmbeddingBedrockApi.java
@@ -162,11 +162,11 @@ public class CohereEmbeddingBedrockApi extends
 			/**
 			 * Discard the start of the input.
 			 */
-			LEFT,
+			START,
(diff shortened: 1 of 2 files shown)


tzolov:
Thank you for catching and fixing this @wmz7year

tzolov:
rebased, squashed and merged at 3497a1ec7b1aa57141652fb18b94e79063310e45
Additionally fix test compilation errors and javadoc issues.

PR #713, opened 2024-05-11
Fix ChromaDB basic authentication

eddumelendez:
Currently, `username` is added as a password in `BasicAuthenticationInterceptor`. This commit fixes the issue and also make sure the integration test setup is correct.


tzolov:
Nice catch @eddumelendez

PR #719, opened 2024-05-13
Optimize Request for Embedding in Vector Store

ricken07:
Currently, vector store automatically calls the embedding client to generate the document embedding without checking whether the document already had an embedding.

In this PR, I first check if the document doesn't already have an embedding before calling the client to generate an embedding. This prevents too many calls to generate an embedding.

- Tests are green for impacted vector stores

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/SimpleVectorStore.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/SimpleVectorStore.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.ai.document.Document;
 import org.springframework.ai.embedding.EmbeddingClient;
 import org.springframework.core.io.Resource;
+import org.springframework.util.CollectionUtils;
 
 import java.io.File;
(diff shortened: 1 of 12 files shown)


tzolov:
If not mistaken this is the same or related to https://github.com/spring-projects/spring-ai/pull/413 ? 

But this change comes with some risks. For example, it is not clear when one would have to invalidate the pre-computed embedding (e.g. the index). Likely when 
Also I'm not sure how useful this feature would be. What is the use case where you will use repeatedly the same Documents (with pre-computed embeddings) for searching? Maybe I'm missing some interesting use cases?

tzolov:
If the pre-computed embedding are not applicable/useful for any real use case, IMO, we need to remove the `embedding` field form the Document.

markpollack:
See https://github.com/spring-projects/spring-ai/issues/1781

I think this was a design mistake to begin with, we shouldn't be caching/storing the embedding in the document in the first place.

markpollack:
we have removed embedding from Document, it isn't needed.  I was a copy of the design from langchain way back when I started the project.  This simplifes the flow of getting data into the vector database, no confusion

PR #722, opened 2024-05-14
OpenAI: Add gpt-4o to chat model enum

ThomasVitale:
Add gpt-4o to OpenAI chat models enum and update JavaDoc


tzolov:
Thanks @ThomasVitale

PR #735, opened 2024-05-18
Fix: watsonx.ai | refresh token issue and implement retry mechanism

PabloSanchi:
Refreshing the token constantly per request threw an error in load tests (and increased the overall response time), so refresh when needed and apply a retry policy to avoid 429 (too many requests) errors.


tzolov:
Thank you @PabloSanchi

PR #737, opened 2024-05-19
Add getContent() method to ChatResponse

habuma:
Simple convenience method on `ChatRespoonse` to return the content of the first generation.

Although `ChatResponse` has many useful bits of information carried in it, the most common thing needed by all applications is the content of the first generation. E.g.

```
String content = chatResponse.getResult().getOuptut().getContent();
```

Because it's the most commonly needed information from a `ChatResponse`, it would be helpful if there were a convenience method that returns the content directly instead of having to navigate through a few other properties. That's what this PR provides. The above code can then be replaced with:

```
String content = chatResponse.getContent();
```


markpollack:
the branch `modelcaller` has a new fluent api that covers this use case and more.

PR #750, opened 2024-05-22
Add JSON type detection to handle both arrays and objects

### LINKED ISSUE #749: Support JSON Type Detection for Parsing Arrays and Objects
**Expected Behavior**

The JSON data should be correctly read and parsed regardless of whether it is an array or an object. The implementation should check the type of the root node and handle it accordingly. For instance, if the JSON data is an array, it should be read into a List<Map<String, Object>>. If it is an object, it should be processed as a single Map<String, Object>

// Todo
1. ~~Modifying Json objects
The objects you've added are all objects with a depth of 1.
I will modify the Json objects a bit more, so that the test case works with different types of objects.~~ Done

2. Look at how other open source Json readers are using it(ex [springboot JsonReader](https://github.com/spring-projects/spring-boot/blob/ba9708d7545eae70b851f0bfccf7b4cb685d6e75/spring-boot-project/spring-boot-tools/spring-boot-configuration-metadata/src/main/java/org/springframework/boot/configurationmetadata/JsonReader.java)), and reference on this.

**Current Behavior**

Currently, the code assumes that the JSON data is always in array format and tries to read it into a List<Map<String, Object>>. This causes failures when the JSON data is an object, as the current implementation does not handle this case.

**Context**

- JSON Type Check: The current implementation does not check whether the JSON data is an array or an object, leading to potential failures. The modified code uses JsonNode to inspect the type of the root node.
- ObjectMapper Reuse: The code reuses the ObjectMapper instance for reading and converting JSON data.
- Stream API: When the JSON data is an array, the modified code uses the Stream API to traverse the JsonNode, parse each node, convert it to a Document object, and collect these objects into a list.
- Type Conversion: The modified code uses the ObjectMapper's convertValue method to convert JsonNode to Map<String, Object>.

Please see as below
* #750

rivkode:
- Handle both JSON arrays and Objects
- Type checking for JSON using JsonNode
- Utilize Stream API to process JSON arrays

I am considering the following for future improvements as below.

- Looking at how other open source Json readers are using it (e.g. [springboot JsonReader](https://github.com/spring-projects/spring-boot/blob/ba9708d7545eae70b851f0bfccf7b4cb685d6e75/spring-boot-project/spring-boot-tools/spring-boot-configuration-metadata/src/main/java/org/springframework/boot/configurationmetadata/JsonReader.java)), and referencing it.
- Modify the constructor to make the ObjectMapper a bit more flexible. Allow users to create any object they want.

<br>

issue link
* #749 

close #749


spring-ai-core/src/main/java/org/springframework/ai/reader/JsonReader.java
```diff
@@ -58,41 +62,37 @@ public JsonReader(Resource resource, JsonMetadataGenerator jsonMetadataGenerator
 
 	@Override
 	public List<Document> get() {
-		ObjectMapper objectMapper = new ObjectMapper();
-		List<Document> documents = new ArrayList<>();
 		try {
 			// TODO, not all json will be an array
```
injae-kim, line 66: can we also remove this TODO?

rivkode: Thanks, I will remove it!

spring-ai-core/src/test/java/org/springframework/ai/reader/JsonReaderTests.java
```diff
@@ -30,12 +29,26 @@
...
+		}
+	}
 
 	@Test
-	void loadJson() {
-		assertThat(resource).isNotNull();
-		JsonReader jsonReader = new JsonReader(resource, "description");
+	void loadJsonObject() {
+		assertThat(ObjectResource).isNotNull();
+		JsonReader jsonReader = new JsonReader(ObjectResource, "description");
```
rivkode, line 51: Before

After
image

markpollack:
Thanks for the contribution, will review after we release M1, moving this to M2.

markpollack:
thanks to the both of you, I've merged it in 1d6cce2419a1621f393906765a838c304bfb9b80

PR #760, opened 2024-05-23
Add ChatClient.Builder auto-configuraiton support

tzolov:
- add ChatClinetCustomizer support.
  - add enable/disable property - enabled by default.
  - add ITs.


tzolov:
squashed and merged at 0571aac44c2a99be439dac371c022c706bfe52dc

PR #763, opened 2024-05-24
fix:when invoke ZhiPuAiEmbeddingModel.call method request body format error

wb04307201:
when invoke ZhiPuAiEmbeddingModel.call method, I get some error,like this  
image  

then check ZhiPuAiEmbeddingModel class and ZhiPuAi website  
image  
image  
in ZhiPuAiEmbeddingModel class use a Collection, but in website input is String  

so I transform List to String, then everything is right!

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,27 @@ For further information go to our [Spring AI reference documentation](https://do
 
 On our march to release 1.0.0 M1 we have made several breaking changes.  Apologies, it is for the best!
 
+**(22.25.2024)**
+
+Vector stores that have a schema are now *not* initialized by default.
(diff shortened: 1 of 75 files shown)


mxsl-gr:
Hi, thanks for reporting and PR.
﻿
I have reproduced the bug, the unit tests do not cover multiple text embeddings. 
﻿
but simply concatenating texts does not meet the API's intent, as passing in multiple text embeddings should return multiple embedding vectors, not a single merged one.
﻿
I think we can create a new issue. what do you think?

wb04307201:
OK, I will create a new issue.
I think we should convert the EmbeddingRequest class to a generic type to adapt to this situation.

PR #768, opened 2024-05-25
Add ChatClient plugable advisors support

tzolov:
- Add RequestResponseAdvisor interface with adviseReqeust and adviseResponse methods. The adviseRequest method takes and returns AdvisedRequest. The adviseResponse method takes and returns ChatRequest.
  - Add ChatClient#ChatClientRequets advisor(...) methods to register advisros. ChatClient call the registered advisors in order before sealing the ChatClientRequest into a Prompt and call the model and after the model response.
  - Implement PromptChatMemoryAdvisor that uses the ChatMemory and the systemem prompt.
  - Implement MessageChatMemoryAdvisor that usees the ChatMemory and the prompt messages.
  - Implement VectorStoreChatMemoryAdvisor that uses VecrStore for long term message history.
  - Add tests.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/ChatClient.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/ChatClient.java
@@ -230,6 +230,34 @@ public interface ChatClient {
 
 		private final Map<String, Object> systemParams = new HashMap<>();
 
+		private List<RequestResponseAdvisor> advisors = new ArrayList<>();
+
+		/* copy constructor */
+		ChatClientRequest(ChatClientRequest ccr) {
+			this(ccr.chatModel, ccr.userText, ccr.userParams, ccr.systemText, ccr.systemParams, ccr.functionCallbacks,
+					ccr.messages, ccr.functionNames, ccr.media, ccr.chatOptions, ccr.advisors);
+		}
+
+		public ChatClientRequest(ChatModel chatModel, String userText, Map<String, Object> userParams,
+				String systemText, Map<String, Object> systemParams, List<FunctionCallback> functionCallbacks,
+				List<Message> messages, List<String> functionNames, List<Media> media, ChatOptions chatOptions,
+				List<RequestResponseAdvisor> advisors) {
+
+			this.chatModel = chatModel;
+			this.chatOptions = chatOptions != null ? chatOptions : chatModel.getDefaultOptions();
+
+			this.userText = userText;
+			this.userParams.putAll(userParams);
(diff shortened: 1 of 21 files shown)


spring-ai-core/src/main/java/org/springframework/ai/chat/client/ChatClient.java
```diff
@@ -455,48 +577,65 @@ public <T> T entity(Class<T> type) {
...
+						functionCallingOptions.setFunctions(new HashSet<>(advisedRequest.functionNames));
 					}
-					if (!this.request.functionCallbacks.isEmpty()) {
-						functionCallingOptions.setFunctionCallbacks(this.request.functionCallbacks);
+					if (!advisedRequest.functionCallbacks.isEmpty()) {
+						functionCallingOptions.setFunctionCallbacks(advisedRequest.functionCallbacks);
+					}
+				}
+				var prompt = new Prompt(messages, advisedRequest.chatOptions);
+				var chatResponse = this.chatModel.call(prompt);
```
markpollack, line 627: This is where we can add a chain of 'around advice' so that one can help with the reliability of calls to the model, for example to enforce json, add 'the functions available to call are' and other types of validation with actions to take on failure to try and remediate.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/RequestResponseAdvisor.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Advisor called before and after the {@link ChatModel#call(Prompt)} and
+ * {@link ChatModel#stream(Prompt)} methods calls. The {@link ChatClient} maintains a
+ * chain of advisors with chared execution context.
+ *
+ * @author Christian Tzolov
+ * @since 1.0.0 M1
+ */
+public interface RequestResponseAdvisor {
```
markpollack, line 36: maybe add `aroundRequest`

spring-ai-core/src/main/java/org/springframework/ai/chat/model/MessageAggregator.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package org.springframework.ai.chat.service;
+package org.springframework.ai.chat.model;
```
markpollack, line 17: I think I had moved this out of the model package to avoid a dependency cycle, will need to revisit.

markpollack:
merged in c4784a070f13e0f8714e5ac71f358980bcdb3e82

Issue #764, opened 2024-05-24
Streaming chat client does not seem to work out of the box for Azure OpenAI, but, does for Bedrock Anthropic3


bruno-oliveira:
**Bug description**

_(Note: this is all a local test for now, but, the actual models being called are "production-ready", so to say so its a representative example)_

When calling the below code from an API client, like Postman:

```
 @GetMapping(value = "/streamed-completions-get", produces = TEXT_EVENT_STREAM_VALUE)
    public ResponseEntity<Flux<ChatResponse>> liveCompletionsStreamed(HttpServletResponse httpServletResponse) {
        httpServletResponse.setHeader("Content-Security-Policy",
            "default-src 'self'; connect-src 'self' http://localhost:9841");
        httpServletResponse.setStatus(HttpServletResponse.SC_OK);
        httpServletResponse.setContentType(TEXT_EVENT_STREAM_VALUE);

        return ResponseEntity.ok(((StreamingModel)chatModel).stream(new Prompt("Explain me how to configure the server port in springboot with examples")));
    }
```

the expected behavior would be that the response is always streamed, meaning that chunks are received over time instead of all at once.

This works as expected when connecting to Anthropic models:
```
spring:
  ai:
    bedrock:
      anthropic3:
        chat:
          enabled: true
          model: anthropic.claude-3-haiku-20240307-v1:0
```

See below the screenshot for evidence that the response is streamed back:

<img width="1392" alt="Screenshot 2024-05-24 at 17 15 08" src="https://github.com/spring-projects/spring-ai/assets/4722412/d11528c1-9023-4c0a-a1c3-75d1d6d4f657">

However, then, I simply stop the app, "swap" the credentials to use the Azure OpenAI Chat Client and the answer is instead, all "streamed at once":

<img width="1393" alt="Screenshot 2024-05-24 at 17 17 33" src="https://github.com/spring-projects/spring-ai/assets/4722412/6fc34b5b-8bc1-45e0-8497-5878d071756f">

What I expected is that the answer would also be streamed just like it is when using Anthropic models.

**Environment**

Using:
Spring AI: 1.0.0-SNAPSHOT (bleeding edge release, received yesterday's breaking changes)
SpringBoot: 3.2.4
Java: 21 

**Steps to reproduce**
A streaming call with a ChatClient configured for Bedrock Anthropic3 models will stream.
The exact same code for an Azure OpenAI client will not stream in chunks, but all at once.

**Expected behavior**
[...]

PR #770, opened 2024-05-26
Fix improper streaming behavior in data handling

bruno-oliveira:
Fix improper streaming behavior in data handling

Resolved an issue where data was not being streamed correctly due to improper use of the reduce function in the stream(Prompt prompt) method. This was causing data to be processed in a blocking manner, contrary to the expected reactive streaming behavior.

Changes made:
- Removed the reduce operation which aggregated all data into one chunk, thus blocking the streaming.
- Replaced the reduce with a direct handling of each individual item within the window to maintain continuous stream flow.
- Added comprehensive tests to ensure streaming is performed as intended, and to verify the non-blocking behavior with multiple emitted items.

Resolves: #764

### DIFF
--- a/models/spring-ai-azure-openai/pom.xml
+++ b/models/spring-ai-azure-openai/pom.xml
@@ -58,6 +58,11 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>io.projectreactor</groupId>
+			<artifactId>reactor-test</artifactId>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 
 </project>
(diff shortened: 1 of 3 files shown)


bruno-oliveira:
@joshlong @markpollack I have taken the liberty of opening a PR that attempts to address a problem I've been seeing with production usage where the streaming mode for the Azure OpenAI completions would, instead, return the entire response at once, which was not the desired behavior.

I will be happy to fix any potential issues and/or addressing any comments you might have!

didalgolab:
I could be wrong, but wasn't the removed `concatMapIterable` code with `window.reduce` part needed for function calling?

Also, I'm under impression that the only thing needed to fix streaming issue in Azure OpenAI model was in this code from `AzureOpenAIChatModel`:
```
			.windowUntil(chatCompletions -> {
				if (isFunctionCall.get() && chatCompletions.getChoices()
					.get(0)
					.getFinishReason() == CompletionsFinishReason.TOOL_CALLS) {
					isFunctionCall.set(false);
					return true;
				}
				return false;
			}, false)

```

to replace `return false;` with `return !isFunctionCall.get();`. At least this is what I've used in my code.

bruno-oliveira:
Seems fixed by: https://github.com/spring-projects/spring-ai/commit/178a607cf6fc65e302b7420fe50cb8dff7e2df2d

PR #773, opened 2024-05-27
fix ZhiPu model batch embeddings

### LINKED ISSUE #771: [BUG]The request body format for the ZhiPuAiEmbeddingModel.call method is incorrect
when invoke ZhiPuAiEmbeddingModel.call method, I get some error,like this  
image  

then I check ZhiPuAiEmbeddingModel class and ZhiPuAi website  
image
image

in ZhiPuAiEmbeddingModel class use a Collection, but in website input is String  

Should convert the org.springframework.ai.embedding.EmbeddingRequest class to a generic type to adapt to this situation?

mxsl-gr:
fixes #771

the PR content:

1. ZhiPu embedding client makes individual calls to embed document collection, fixes #771.
2. add batch embeddings unit tests


markpollack:
merged in 23e448cf2a550a392e6135310cc5ad01ece6f702

Issue #537, opened 2024-04-02
[Malicious, Package Squatting Attack] Do Not Use Packages from the "io DOT springboot" domain!!!


yuluo-yx:
hi, community. If I want to develop based on spring core module to adapt new ai big model, how pom.xml is selected.

Docs：

```xml
  <repositories>
    <repository>
      <id>spring-milestones</id>
      <name>Spring Milestones</name>
      <url>https://repo.spring.io/milestone</url>
      <snapshots>
        <enabled>false</enabled>
      </snapshots>
    </repository>
    <repository>
      <id>spring-snapshots</id>
      <name>Spring Snapshots</name>
      <url>https://repo.spring.io/snapshot</url>
      <releases>
        <enabled>false</enabled>
      </releases>
    </repository>
  </repositories>
```

```xml
<dependencyManagement>
    <dependencies>
        <dependency>
            <groupId>org.springframework.ai</groupId>
            <artifactId>spring-ai-bom</artifactId>
            <version>0.8.1-SNAPSHOT</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
    </dependencies>
</dependencyManagement>
```

Maven repo： https://mvnrepository.com/artifact/io DOT springboot DOT ai/spring-ai-core/1.0.0

```xml
<dependency>
    <groupId>io DOT springboot DOT ai</groupId>
    <artifactId>spring-ai-core</artifactId>
    <version>1.0.0</version>
</dependency>
```

Which one should I choose and is there any difference between them?

yuluo-yx:
Community Issue Questions go unanswered?  There's already a backlog of questions.  @tzolov

anbusampath:
> ```
> <dependencyManagement>
>     <dependencies>
>         <dependency>
>             <groupId>org.springframework.ai</groupId>
>             <artifactId>spring-ai-bom</artifactId>
>             <version>0.8.1-SNAPSHOT</version>
>             <type>pom</type>
>             <scope>import</scope>
>         </dependency>
>     </dependencies>
> </dependencyManagement>
> ```

you should continue to use above one.



> ```
> 
> <dependency>
>     <groupId>io.springboot.ai</groupId>
>     <artifactId>spring-ai-core</artifactId>
>     <version>1.0.0</version>
> </dependency>
> ```

This looks like some accidental publish.

yuluo-yx:
> > ```
> > <dependencyManagement>
> >     <dependencies>
> >         <dependency>
[...]

PR #778, opened 2024-05-27
Modified dependency

cemdrman:
Changed dependency. Because this dependency is outdated.


eddumelendez:
This is not valid coordinates. See https://github.com/spring-projects/spring-ai/issues/537#issuecomment-2041455227

markpollack:
this is not a valid change, that is not the official dependency.

PR #785, opened 2024-05-29
Extend format to improve generate JSON

pgerhard:
- Add an additional instruction to the format for the model not to include markdown syntax for JSON quotes.
- Extend the conversion to strip markdown quotes in case the model included them in the response


tzolov:
Thanks for the contribution @pgerhard !
Could you please confirm that `./mvnw spring-javaformat:apply` has been run?

pgerhard:
Just ran it locally but it didn't make any changes so I haven't updated the branch

PR #787, opened 2024-05-29
[WIP] Add possibility to register functions in a type-safe manner

pgerhard:
**Goal:** The goal of this PR is to extend the existing function registration methods to also accept types of beans. The advantage would be that this is type-safe. A type-safe interface improves intelli-sense and refactoring


pgerhard:
Hi Christian (@tzolov),

As discussed i've updated the PR with a way to define functions via a factory class. Compared to my initial implementation using the `BeanNameResolver` this feels like a more consistent approach.
However, as I mentioned in my comment (see `Tools`) the different ways that functions are currently handled by the ChatClient makes this more difficult to do. I think a discussion around how we can have a more consistent approach to adding functions to the ChatClient might be helpful
If you have some time to discuss this just let me know (also open to jumping on a call)

ThomasVitale:
Thanks for your PR! Declarative tool definition via annotation has been introduced in https://github.com/spring-projects/spring-ai/pull/2064, making it possible to pass to ChatClient an object containing @Tool-annotated methods. Does that handle your use case?

Example:

```java
class MyTools {

    @Tool(description = "Get the list of books written by the given author available in the library")
    List<Book> booksByAuthor(String author) {
        return bookService.getBooksByAuthor(new Author(author));
    }

}
```

```java
@GetMapping("/chat")
String chat(String authorName) {
    return chatClient.prompt()
            .user("What books written by %s are available in the library?".formatted(authorName))
            .tools(myTools)
            .call()
            .content();
}
```

PR #789, opened 2024-05-29
Overload QuestionAnswerAdvisor to use SearchRequest.defaults()

habuma:
Although it's helpful to be able to customize the `SearchRequest` given to `QuestionAnswerAdvisor`, it's not uncommon to just rely on `SearchRequest.defaults()`. Therefore, this PR overloads the constructor with a version that only requires a `VectorStore` and uses `SearchRequest.defaults()`...in the interest of simplifying the application code a little.


tzolov:
Makes perfect sense. Thanks @habuma

PR #790, opened 2024-05-30
Use List.of instead of Arrays.asList In the SEARCH_OUTPUT_FIELDS field.

Lee-ChungMu:
Use List.of instead of Arrays.asList In the SEARCH_OUTPUT_FIELDS field.

The Arrays.asList method can change elements. SEARCH_OUTPUT_FIELDS is an immutable field, so it has been improved to util List.of.


tzolov:
Thanks @Lee-ChungMu

PR #799, opened 2024-05-30
fixe the wrong word 'builderh' to 'builder' in chatclient.adoc file

shihaoxiong:
There is a wrong word 'builderh' in chatclient.adoc file, it should be 'builder'.


tzolov:
Thank you @shihaoxiong . It seems though that this type has been fixed already by some other PR.
Nothing to fix

PR #807, opened 2024-06-01
Different improvements, regarding final modifiers, try with resources, and other

alxkm:
There several of different improvements. 
Regarding final modifiers 
Try with resources 
Lambda reference 
And other


tzolov:
Thank you @alxkm 
Could you please rebase (not merge) your PR to of upstream/main and resolve the conflicts. 
E.g. `git pull -r upstream main `

spring-ai-core/src/main/java/org/springframework/ai/ResourceUtils.java
```diff
@@ -25,6 +25,10 @@
  */
 public class ResourceUtils {
 
+	private ResourceUtils() {
```
markpollack, line 28: Spring conventions are to make the class abstract, e.g. StringUtils

spring-ai-core/src/main/java/org/springframework/ai/aot/AiRuntimeHints.java
```diff
@@ -40,6 +40,10 @@
  */
 public class AiRuntimeHints {
 
+	private AiRuntimeHints() {
```
markpollack, line 43: Similar to before, will create as abstract.  Likely should change the name to RuntimeHintUtils, but that will be a big commit, so will do another time.

spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
```diff
@@ -203,7 +203,7 @@ else if (!content.equals(other.content))
 
 	@Override
 	public String toString() {
-		return "Document{" + "id='" + id + '\'' + ", metadata=" + metadata + ", content='" + new String(content) + '\''
+		return "Document{" + "id='" + id + '\'' + ", metadata=" + metadata + ", content='" + content + '\''
```
markpollack, line 206: regenerated toString to also include media

markpollack:
refined and merged in cb6662a61e421af43801c4090e3330d2b1811edf

thanks!

PR #815, opened 2024-06-03
Add ChatClient support for returning ResponseEntity<ChatResponse, T>

tzolov:
ChatClient already provides the .chatResponse() method to return the entire ChatResponse instance.
 It also provides a set of overloaded .entity(Type) methods to provide Type-converted responses.
 The new .responseEntity(Type) method returns a ResponseEntity<ChatResponse, T> instance, encapsulating
 both the ChatResponse and the requested Type-converted response entity.

 This change allows for more flexibility when handling different response types and facilitates
 easier integration with other components that expect ResponseEntity instances.


tzolov:
Rebased and merged at d6a0dffd3ec4eea7dd1489b1ced2275ee868e98d

PR #820, opened 2024-06-04
Introduce .editorconfig

piotrooo:
To help the rest of us who are not using tabs for indentation.

Taken from https://github.com/spring-projects/spring-framework/blob/main/.editorconfig


tzolov:
Thanks @piotrooo !

PR #821, opened 2024-06-04
Update mistralai-chat.adoc

rubin0:
In the manual configuration section of MistralAI SpringAI documentation is used a `withMaxToken` method that doesn't exist in the current version of the API.

Replaced with correct `withMaxTokens` method name.


tzolov:
Thanks @rubin0

PR #826, opened 2024-06-05
[Mistral] Update models

ThomasVitale:
* Align model naming with Mistral AI documentation
* Add missing model for mixtral 22B
* Deprecate mistral-medium-latest because Mistral will remove it soon


tzolov:
Cool!

PR #858, opened 2024-06-12
Implement CassandraChatMemory

michaelsembwever:
And, fix initialize-schema docs and so that it actually works.

In action here: https://github.com/datastax/ai-agent-java/blob/mck/step-1-new-chat-model/src/main/java/com/datastax/ai/agent/history/AiAgentSession.java

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,5 @@ package.json
 shell.log
 
 .profiler
+/spring-ai-spring-boot-autoconfigure/nbproject/
+/vector-stores/spring-ai-cassandra-store/nbproject/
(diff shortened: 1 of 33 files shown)


tzolov:
This is really cool! Thanks @michaelsembwever 

Let me know if stumble on some design flows with simplistic ChatMemory interface.

michaelsembwever:
@tzolov ready for review.

Second commit is refactoring of `CommonVectorStoreProperties` to what i believe is a better package location.

tzolov:
Thanks @michaelsembwever 
LGTM. you've forgotten to add the CassandraAutoConfiguration to the AutoConfiguration.imports. But I will fix it during the merge. 

FYI, after M2, we will explore how to implement a "generic" Spring Data Chat Memory support that might affect your impl. as well. 
Anyway, this is the first persistent chat memory implementation! :)

tzolov:
Rebased, squashed and merged at 4aacab020b2fa6932ac3a69433a60c282eec617b

Issue #861, opened 2024-06-13
Spring AOT fails for dev service-enabled auto-configuration 


ThomasVitale:
**Bug description**
The Spring AOT compilation step fails because the property-based implementation of the `ConnectionDetails` interface is private instead of being package-scoped.

* Chroma
* Milvus
* Ollama
* Redis.

**Environment**
Java 21
Spring AI 1.0.0-SNAPSHOT

**Steps to reproduce**
Add the "GraalVM Native Support" to a Spring Boot 3.3.0 application using any of the modules mentioned above.
During the compilation process, the AOT steps fails with the following error:

```
app/build/generated/aotSources/org/springframework/ai/autoconfigure/ollama/OllamaAutoConfiguration__BeanDefinitions.java:31: error: PropertiesOllamaConnectionDetails has private access in OllamaAutoConfiguration
    return BeanInstanceSupplier.<OllamaAutoConfiguration.PropertiesOllamaConnectionDetails>forFactoryMethod(OllamaAutoConfiguration.class, "ollamaConnectionDetails", OllamaConnectionProperties.class);
```

**Expected behavior**
The AOT compilation step succeeds.

**Minimal Complete Reproducible example**
See  https://github.com/ThomasVitale/llm-apps-java-spring-ai/tree/main/00-use-cases/text-classification

PR #782, opened 2024-05-28
Mark Testcontainers ContainerConnectionDetailsFactory as non-public

eddumelendez:

PR #862, opened 2024-06-13
Fix Spring AOT for ConnectionDetails beans

ThomasVitale:
Fixes gh-861


eddumelendez:
I think #782 will fix this 😁 I noticed after the PR was merged

ThomasVitale:
@eddumelendez nice, I hadn't noticed that! Without any public classes in the Testcontainers module, I believe the maven publishing task might fail the Maven Central criteria. I got similar issues before with other libraries. It's a bit counterintuitive.

PR #865, opened 2024-06-15
Update FunctionCallbackWrapper docs for Anthropic, Mistral, ZhiPu, MiniMax, and OpenAI

habuma:
I updated the documents for Anthropic, Mistral, ZhuPu, MiniMax, and OpenAI to match up with what is in the docs for Azure OpenAI to use the FunctionCallbackWrapper builder. In doing so the call to set to response converter was lost, although it could have been specified...it just wasn't in the Azure example.

I also noticed that the Gemini document also sets the schema type, as it's required for Gemini. I confirmed that it's not required for OpenAI, Anthropic, or Mistral. But while confirming, I discovered that the documentation for Mistral functions refers to "mistral_large_latest", but it should be "mistral-large-latest". The same fix was also needed for "mistral_small_latest" -> "mistral-small-latest".


tzolov:
thans @habuma

Issue #866, opened 2024-06-15
Add "role": "function" to Gemini API Function Calling Results


e1arikawa:
**Expected Behavior**
I want to set role to "function" in the content of the Gemini API function call result. That is because in the "Multi-turn function calling curl example request" example in the Gemini API documentation, role is set to "function".

[https://ai.google.dev/gemini-api/docs/function-calling#expandable-7](https://ai.google.dev/gemini-api/docs/function-calling#expandable-7)

"role":"function" attribute is consistent with best practices observed in other major APIs such as OpenAI's GPT. This clear role assignment improves readability and debugging and makes it easier for developers to understand and manage the flow of function calls and responses.

Currently, the lack of a "role" attribute in the result of a function call makes it difficult to distinguish between different parts of the interaction, potentially leading to confusion and response processing errors. Explicitly setting "role" to "function" will make the structure of the response clearer and more consistent with the examples in the Gemini API documentation.
### 
```
{
  "contents": [
    {
      "role": "user",
      "parts": [
        {
          "text": ...
        }
      ]
    },
    {
      "role": "model",
      "parts": [
        {
          "functionCall": {
            ...
          }
        }
      ]
    },
    {
      "role": "function",
      "parts": [
        {
          "functionResponse": {
            ...
          }
        }
      ]
    }
  ]
}
```

**Current Behavior**
The content of the Function Calling result of the Gemini API does not include a "role".

```
{
  "contents": [
    {
      "role": "user",
      "parts": [
        {
          "text": ...
        }
      ]
    },
    {
      "role": "model",
      "parts": [
        {
          "functionCall": {
            ...
          }
        }
      ]
    },
    {
      "parts": [
        {
          "functionResponse": {
            ...
          }
        }
      ]
    }
  ]
}
```

e1arikawa:
PR created.
thanks.
https://github.com/spring-projects/spring-ai/pull/870

PR #870, opened 2024-06-15
Update FUNCTION role case property to GeminiRequest

BlackBean99:
Description:
This pull request adds the FUNCTION role property to the GeminiAiRequestModel class. This enhancement allows the model to handle messages that require a function role, enhancing the flexibility and extensibility of the chat application.

The reason for adding the FUNCTION case to the toGeminiContent method is to ensure extensibility. For example, while currently handling only user, model, and function roles, there is a possibility of adding other roles in the future, such as error, notification, command, and meta.

If the FUNCTION filter condition is not needed, please comment below. Thank you.

reference
- #866


tzolov:
Addressed by the https://github.com/spring-projects/spring-ai/pull/1038

Issue #747, opened 2024-05-21
PgVectorStore should allow table and index name to be specified


johnsonr:
It should be possible to specify a custom name for the table used by `PgVectorStore`, which is currently hardcoded to  `vector_store`. The associated index name could default to being generated based on the name while also allowing custom specification.

`PgVectorStore` is effectively a database singleton. This is not Spring-like.

Concrete reasons this would be desirable:

- It should be possible to have multiple vector stores in the same database.
- It should be possible to have a vector store within a schema. It's good practice for an app's tables to be within a schema.

muthuishere:
I have been closely following this issue regarding enabling custom table names for the PGvector Store, and I’ve developed a solution that I believe addresses the requirements discussed here.

The solution includes:
- Adding properties to configure custom table and index names.
- Modifying the PgVectorStore class to use these configurable properties.
- Ensuring backward compatibility with setups using default settings.

I have thoroughly tested these changes to ensure they work seamlessly with both new configurations and existing deployments. I will be submitting the PR for the same

tzolov:
Hi @johnsonr ,

Regarding the 
> It should be possible to have multiple vector stores in the same database. 
> It should be possible to have a vector store within a schema. It's good practice for an app's tables to be within a schema.

The reason for not providing parametrisable table name/schema/index is that , for convenience, we pre-create the table when not available. 
And we cannot afford dynamic DDLs as no support for prepared DDL is provided.

We could either drop the pre-creation all together or run it only when no custom table/schema/index were provided. 
For the later we would assume that the user would take care.

Would this make sense?

muthuishere:
@tzolov Without this modification, a user will be able to create only one table in any given database. If we want to accommodate multiple embeddings of different dimensions, the user would need to have separate databases. This is because the table name remains constant in any given database.

I could see two  options:

1. Add a table name validation using matches("^[a-zA-Z0-9_]+$"). to ensure no SQL injection is there

2. Upon receiving the property spring.ai.vectorstore.pgvector.vectorTableName, ensure the table exists and the fields are available, or we can throw an exception. if its not specified use the existing flow

I would like to go for option 2

muthuishere:
@tzolov / @johnsonr 

When revising the approach, I encountered a recurring issue related to using table names from configurations. Whether it's selecting from or deleting tables, similar to using in DDL statements, the same limitation arises: we cannot embed table names into prepared statements directly. This constraint leads us to two viable solutions:

[...]

PR #872, opened 2024-06-16
feat: PgVectorStore should allow table and index name to be specified…

muthuishere:
PGvector Store's restriction to a single hardcoded table name per database limits the development of multiple applications; this PR introduces configurable table names to improve database flexibility and management.

This change allows users to specify custom table name within a single database instance.  

The update ensures backward compatibility and adheres to discussions in issue #747.

**Changes:**
- Added `vectorTableName` and `vectorIndexName` properties.
- Updated `PgVectorStore` class to utilize these new properties.

**Testing:**
- Integration tests to ensure custom table names are handled correctly.


tzolov:
@muthuishere thanks you for the contribution. 

Reviewing you PR reminded me why we haven't implemented the parametrisation so far. Check this https://github.com/spring-projects/spring-ai/issues/747#issuecomment-2172456938

Parametrising the DDL with string concatenations (the approach you've taken) is prone for sql injections. So we need to find different approach. 
Let me know what do you think for my suggestions in the  https://github.com/spring-projects/spring-ai/issues/747#issuecomment-2172456938

muthuishere:
> @muthuishere thanks you for the contribution.
> 
> Reviewing you PR reminded me why we haven't implemented the parametrisation so far. Check this [#747 (comment)](https://github.com/spring-projects/spring-ai/issues/747#issuecomment-2172456938)
> 
> Parametrising the DDL with string concatenations (the approach you've taken) is prone for sql injections. So we need to find different approach. Let me know what do you think for my suggestions in the [#747 (comment)](https://github.com/spring-projects/spring-ai/issues/747#issuecomment-2172456938)



PR updated with commits to  introduce the ability to specify custom table name for PgVectorStore, and stop pre create tables when custom table names are provided. also added tablename validations, field validations to ensure things are okay

Key changes include:
- Adding properties to configure custom table names.
- Modifying PgVectorStore to conditionally pre-create tables only when custom table names are not provided.
- Implementing table & field validation for custom configurations.
- Extensive tests to ensure compatibility with both new configurations and existing deployments.

With these changes, automatic table pre-creation is removed, and users are expected to manage table setup in custom scenarios, consistent with production environment practices.

tzolov:
Hi @muthuishere , 
I can see that you went above and beyond to try to validate the DDL parameters !

I had internal discussion with our spring data team and the conclusion is that there is not reliable solution. So we should assume that this is user's configuration responsibility. E.g. no need to try to do the validations ourself. 

Perhaps we can keep the validation you've added but  make it disabled by default. Also can enable the auto-generation for custom table names as well.  

Additionally in https://github.com/spring-projects/spring-ai/issues/747 Rod mentions custom schema definition as well. I guess we need to add yet another parameter (e.g. schema name)? 

What do you think?

muthuishere:
I am okay with your suggestions. I will be working on the following tasks:

- A property to specify the database schema (schemaName).
- A property to enable database validations; the default setting will be false (validations.enabled).
- The custom table name can support auto-generation. If the table does not exist, we will automatically generate it as we did previously. We will create the index ourselves by appending _index to it, instead of asking users to specify this, as it is no longer relevant. (I would appreciate your thoughts on this.)

Let me know if you need any more.

tzolov:
Sounds like a compete plan. Thanks @muthuishere

muthuishere:
> Sounds like a compete plan. Thanks @muthuishere

@tzolov 

- A property to specify the table name (spring.ai.vectorstore.pgvector.vectorTableName) , defaults to existing table name(vector_store)
- A property to specify the database schema (spring.ai.vectorstore.pgvector.vectorTableName) defaults to public
 - A property to enable database validations (spring.ai.vectorstore.pgvector.vectorTableValidationsEnabled) defaults to false 


Added tests to ensure things are okay.

tzolov:
Thanks you @muthuishere . Great contribution! 

rebased, squashed and merged at f0ca61252b0de2a3e3ab4a7cb493844eafd82139

Additional changes and fixes applied:
     - Rename properties to schemaName, tableName, and schemaValidation
     - Update pgvector documentation with new properties
     - Add schema/table name tests to PgVectorStoreAutoConfigurationIT and PgVectorStorePropertiesTests
     - Create standalone PgVectorSchemaValidator class for schema/table validation
     - Add missing 'CREATE SCHEMA IF NOT EXISTS' when initializeSchema=true
     - Remove redundant code and classes

PR #879, opened 2024-06-17
Enhancing Elasticsearch vector store: docs + dependency fix

l-trotta:
As previously discussed with @tzolov, removing the unnecessary elasticsearch dependency from the autoconfigure pom and adding a note in the documentation to explain possible version problems with older spring-boot releases.


tzolov:
Thanks @l-trotta 
rebased, formatted and merged at aca3720f5b27761c4f1ec51dc5ecbb062f7861ee

Issue #880, opened 2024-06-18
Transfer the query when to call the vector store


showpune:
Please do a quick search on GitHub issues first, the feature you are about to request might have already been requested.

**Expected Behavior**

To support multiple language, normally we need to provide RAG document for multiple languages for search. It is too cost and a possible solution is that we can just provide one RAG document, and before the search the document, we can convert the customer input into the target language

**Current Behavior**

Use customer input to query the vectore store directly, if the customer input different language with the document, we can't get the content

**Context**

ThomasVitale:
This use case is now supported by the `TranslationQueryTransformer` component of the new (experimental) Modular RAG features. You can find a full example here: https://github.com/ThomasVitale/llm-apps-java-spring-ai/blob/main/rag/rag-sequential/rag-advanced/src/main/java/com/thomasvitale/ai/spring/RagControllerTranslation.java

PR #881, opened 2024-06-18
Feat: Transfer the query when to call the vector store

showpune:
This PR is to provide the feature https://github.com/spring-projects/spring-ai/issues/880

Change:
1) New class QueryTransformerQuestionAnswerAdvisor extends QuestionAnswerAdvisor
2) Call the AI model to transfer the query before call the vector search

 Done
Done
Done
Done

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/QueryTransformerQuestionAnswerAdvisor.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/QueryTransformerQuestionAnswerAdvisor.java
@@ -1,0 +1,79 @@
+package org.springframework.ai.chat.client.advisor;
+
+import org.springframework.ai.chat.client.AdvisedRequest;
+import org.springframework.ai.chat.model.ChatModel;
+import org.springframework.ai.chat.prompt.PromptTemplate;
+import org.springframework.ai.vectorstore.SearchRequest;
+import org.springframework.ai.vectorstore.VectorStore;
+import org.springframework.util.StringUtils;
+
+import java.util.Map;
+
+/**
+ * Context for the question is retrieved from a Vector Store with Query Transformer and
+ * added to the prompt's user text.
+ *
+ * @author Zhiyong Li
+ * @since 1.0.0
+ */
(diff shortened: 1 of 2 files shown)


markpollack:
Hi.  Thanks for you contribution.  I feel that features such as checking for preconditions on the user text should be done in their own advisor and not to extend the `QuestionAnswerAdvisor`.  The reasoning is that checking for some precondition on the user text doesn't involve using a vector store - at least in the current implementation you propose - and that we have a chain of advisors that get executed one after another.  This allows for flexibility in combining different advisor functionality.

showpune:
Hi @markpollack thanks for your response.

I extend the QuestionAnwserAdvisor because I just want to optimize the vector store query by refine the query using model according to customer input. I even can ask model to "generate 3 different query from different perspective but still keep the same meaning" to optimize my search in vector store.

I don't want to put a preadvisor because I don't want to change the orignal customer input. In my example, if the customer ask in Chinese, I just want to using the query in English to query the vector store, and I still want to use the orignal customer input in Chinese to talke with model and model will return in Chinese. If we use preadvisor to translate the customer input  into English, the return will be in English

markpollack:
Thanks for the explanation, I understand the issue better now.  I want to do a bit of research into this area of more advanced RAG techniques, certainly something we need to support.  The technique you have here looks to be in the category of 'query rewriting' and then, as you mention, asking multiple questions and taking the 'best' response from them.  I want to see how well the advisor chain itself can be setup to perform this in sequential steps - including keeping the original query - or if it needs to be encapsulated into a single advisor.  I think some modularity should be possible with advisors that are known to work well together and data sharing between advisor's occuring with the shared context each advisor is given.

ThomasVitale:
This use case is now supported by the `TranslationQueryTransformer` component of the new (experimental) Modular RAG features. More info in the documentation: https://docs.spring.io/spring-ai/reference/api/retrieval-augmented-generation.html#_translationquerytransformer

>  I even can ask model to "generate 3 different query from different perspective but still keep the same meaning" to optimize my search in vector store.

That use case is supported by the `MultiQueryExpander` component. More info in the documentation: https://docs.spring.io/spring-ai/reference/api/retrieval-augmented-generation.html#_multiqueryexpander

PR #884, opened 2024-06-18
Fix:base-url Error in openai-embeddings.adoc

JavaProgrammerLB:
- fix base-url error in openai-embeddings.adoc
- When I set `spring.ai.openai.base-url=api.openai.com` as the document, I meet the Error `URI is not absolute`
- It is necessary to set the base-url with `https`(`spring.ai.openai.base-url=https://api.openai.com` )

image

image


tzolov:
Thanks @JavaProgrammerLB

PR #892, opened 2024-06-19
Support Qdrant's Api Key in Docker Compose and Testcontainers

eddumelendez:
Read `QDRANT__SERVICE__API_KEY` env var from compose file and Testcontainers definition to configure the connection details for Qdrant.


tzolov:
excelent :)

PR #896, opened 2024-06-19
Polish OracleVectorStoreAutoConfiguration

eddumelendez:


tzolov:
Hey @eddumelendez , 
Is there a problem with the explicit `@ImportAutoConfiguration(classes = JdbcTemplateAutoConfiguration.class)` ?

eddumelendez:
@tzolov sorry for not providing more information but the autoconfiguration is already declared as `@AutoConfiguration(after = JdbcTemplateAutoConfiguration.class)`. So, the `@ImportAutoConfiguration(classes = JdbcTemplateAutoConfiguration.class)` is not needed but only for test purposes.

Issue #907, opened 2024-06-20
Change default of schema initialization of vector stores from true to false


markpollack:
public class CommonVectorStoreProperties {

	private boolean initializeSchema = true;

     ...
}

This should be `false`.   

The integration tests need to change so that they are set to true.

Documentation needs to be updated in each vector store to indicate the change in behavior.

PR #918, opened 2024-06-21
GH-907: Change vector store schema init default

sobychacko:
* Change default schema initialization of vector stores from `true` to `false.` Users need to explicitly opt-in for schema initialization by setting the `initialize-schema` property on the corresponding vector store.
* Update integration tests
* Update copyright years

Resolves https://github.com/spring-projects/spring-ai/issues/907

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
@@ -5,6 +5,11 @@
 
 * The configuration prefix for the Chroma Vector Store has been changes from `spring.ai.vectorstore.chroma.store` to `spring.ai.vectorstore.chroma` in order to align with the naming conventions of other vector stores.
 
+* The default value of the `initialize-schema` property on vector stores capable of initializing a schema is now set to `false`.
+This implies that the applications now need to explicitly opt-in for schema initialization on supported vector stores, if the schema is expected to be created at application startup.
+Not all vector stores support this property.
+See the corresponding vector store documentation for more details.
+
 == Upgrading to 1.0.0.M1
 
 On our march to release 1.0.0 M1 we have made several breaking changes.  Apologies, it is for the best!
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
@@ -9,6 +9,13 @@
 This implies that the applications now need to explicitly opt-in for schema initialization on supported vector stores, if the schema is expected to be created at application startup.
 Not all vector stores support this property.
 See the corresponding vector store documentation for more details.
+The following are the vector stores that currently don't support the `initialize-schema` property.
+
+1. Cassandra
+2. Gemfire
+3. Hana
+4. Pinecone
+5. Weaviate
 
 == Upgrading to 1.0.0.M1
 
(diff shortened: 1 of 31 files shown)


spring-ai-spring-boot-autoconfigure/src/test/java/org/springframework/ai/autoconfigure/vectorstore/azure/AzureVectorStoreAutoConfigurationIT.java
```diff
@@ -81,8 +82,8 @@ public static void beforeAll() {
 	public void addAndSearchTest() {
 
 		contextRunner
-			.withPropertyValues("spring.ai.vectorstore.azure.indexName=my_test_index",
-					"spring.ai.vectorstore.azure.defaultTopK=6",
+			.withPropertyValues("spring.ai.vectorstore.azure.initializeSchema=true",
```
markpollack, line 85: this test passes when `initializeSchema=false`.  I would expect it to fail.

spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
```diff
@@ -5,6 +5,11 @@
 
 * The configuration prefix for the Chroma Vector Store has been changes from `spring.ai.vectorstore.chroma.store` to `spring.ai.vectorstore.chroma` in order to align with the naming conventions of other vector stores.
 
+* The default value of the `initialize-schema` property on vector stores capable of initializing a schema is now set to `false`.
+This implies that the applications now need to explicitly opt-in for schema initialization on supported vector stores, if the schema is expected to be created at application startup.
+Not all vector stores support this property.
```
markpollack, line 10: we should list the one's that don't so they don't have to search.  Also, I don't think we are very clear on this in the ref docs itself, will need to double check.

markpollack: My quick check of those that don't are 

* `Cassandra` - I have a feeling though it should.  There is a property `disallowSchemaChanges` that seems like it is performing the same task as `initializeSchema`
* `Gemfire` - The implementation of `afterPropertiesSet` looks problematic, which is a separate issue, but seems like we would need to update that method to work off `initializeSchema`  @geetrawat can you comment. 
* `Hana` 
* `OpenSearch` implements `InitializingBean` with an index being created if it doesn't exist.  Should be retrofitted to use `initializeSchema`
* `Pinecone`
* `Typesense` implements  `InitializingBean` with an collection being created if it doesn't exist.  Should be retrofitted to use `initializeSchema`

spring-ai-spring-boot-autoconfigure/src/test/java/org/springframework/ai/autoconfigure/vectorstore/azure/AzureVectorStoreAutoConfigurationIT.java
sobychacko: As we discussed, this happens because the index is already on the Azure vector store and is never deleted.

spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
```diff
@@ -5,6 +5,18 @@
...
+See the corresponding vector store documentation for more details.
+The following are the vector stores that currently don't support the `initialize-schema` property.
+
+1. Cassandra
```
markpollack, line 14: Cassandra does support this property.

`CassandraVectorStoreProperties extends CommonVectorStoreProperties` and the bean definition for the `CassandraVectorStore` in `CassandraVectorStoreAutoConfiguration` has the following code
```
		if (!properties.isInitializeSchema()) {
			builder = builder.disallowSchemaChanges();
		}
```

So the

markpollack:
fixed in 50d34b8a486b83d5d380ce6dfd70dcbe219444af

PR #919, opened 2024-06-21
Consistent vector store property prefix for Chroma store

sobychacko:
* The Chroma vector store config property prefix uses .store at the end, which is not consistent with the other vector stores—fixing this.


markpollack:
I updated the `upgrade-notes.adoc` and merged in b206b37dfd7f9b938961265bb4fb8e217c6873af

PR #829, opened 2024-06-06
Add support for getting usage tokens for openaiAPI 

YanjiuHe:
fix #814

PR #920, opened 2024-06-22
OpenAI: Add support for getting token usage when using streaming

### LINKED ISSUE #814: Feature Request: Support for get tokens
I want get Usage stats for streaming

https://community.openai.com/t/usage-stats-now-available-when-using-streaming-with-the-chat-completions-api-or-completions-api/738156

`data: {"id":"chatcmpl-9VvDzy405iABmaFVxI4lrE5Ho6rPl","object":"chat.completion.chunk","created":1717395167,"model":"gpt-4o-2024-05-13","system_fingerprint":"fp_319be4768e","choices":[],"**usage**":{"prompt_tokens":19,"completion_tokens":10,"total_tokens":29}}`

didalgolab:
Fixes #814

I'm aware of PR #829, but I would like to propose a more complete solution here.

This PR:
- Adds support for the `stream_options` chat completion request property.
- Prevents sending `stream_options` if the `call` instead of the `stream` method is used (otherwise, OpenAI API returns an error).
- Adds an integration test comparing token usage obtained with the `call` and `stream` methods.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -222,7 +222,12 @@ public class OpenAiChatModel extends
 							return generation;
 						}).toList();
 
-						return new ChatResponse(generations);
+						if (chatCompletion.usage() != null) {
+							return new ChatResponse(generations, OpenAiChatResponseMetadata.from(chatCompletion));
+						}
+						else {
+							return new ChatResponse(generations);
+						}
 					}
 					catch (Exception e) {
 						logger.error("Error processing chat completion", e);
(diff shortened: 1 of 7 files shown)


tzolov:
Thanks @didalgolab , 
I wonder if the include usage shouldn't be enabled by default. In which case the question of opting out became relevant. 
We can introduce a shortcut `withStreamUsage(boolean)` that will set the `StreamOptions.INCLUDE_USAGE`.
Will try to do this while merging your PR.

didalgolab:
I wondered about that too. I'd love to make it enabled by default! However, there is an issue with using OpenAiApi with different providers (e.g., Groq, OpenRouter...) in such case. As far as I know, none of them support `stream_options`, although some stream token usage out of the box without any opt-in. Enabling the `include_usage` option by default would make the default chat options incompatible with different providers' endpoints, requiring opt-out in the code each time. If you are okay with that, I am too.

Or perhaps `OpenAiChatModel` could detect if the default baseUrl is used and enable or disable `stream_options` accordingly, although I must admit I generally dislike conditional logic like that.

Feel free to share any thoughts or suggestions about this.

Adding a shortcut `withStreamUsage(boolean)` is an excellent idea.

tzolov:
@didalgolab, I've added the Builder shortcut along with a field-less boolean getter/setters to provide a`stream-usage` property. 
Updated the docs, tests and did some small amendments.

extended, rebased, squashed and merged at 20e4b5609a35c204b0e93ac6fcd38ae821a86d1d

PR #929, opened 2024-06-22
Support Chroma's CHROMA_SERVER_AUTH_CREDENTIALS env var from version 0.4.x

eddumelendez:
Currently, Docker Compose and Testcontainers support `CHROMA_SERVER_AUTHN_CREDENTIALS`, which is provided by Chroma 0.5.x. Previous versions use `CHROMA_SERVER_AUTH_CREDENTIALS` instead.


tzolov:
Thanks @eddumelendez !

Issue #936, opened 2024-06-23
Include "id" and "model" in ChatResponseMetadata for better traceability


ThomasVitale:
**Expected Behavior**

Many LLM providers (including OpenAI, Mistral AI, and Anthropic), include an ID in their responses to uniquely identify an operation, and also include the exact model used to process the request (which differs from the one in the request, e.g. requested: `gpt-3.5-turbo` => actual: `gpt-3.5-turbo-0125`). I would expect them to be available in the `ChatResponseMetadata` interface. They will be critical when introducing observability in Spring AI (I'll soon share a draft solution about that, but this feature request is a pre-requisite).

**Current Behavior**

These two pieces of information are available in the lower LLM provider API, but they are not surfaced through the Spring AI abstractions. The "id" is actually already included in some customisations of `ChatResponseMetadata`, but it should be moved up to the generic interface.

**Context**

When using cloud-based LLM providers, it's important to be able to trace a specific operation for troubleshooting, evaluation, or further analysis. Knowing the unique operation ID and the actual model used to process the request helps towards that goal.

PR #550, opened 2024-04-04
[WIP] Add missing response fields to Spring AI ChatResponse, Generation and AssistantMessage

tzolov:
Ensure that Generation and AssistantMessage has Id, index, isCompleted attributes.
  Also the ChatResponse should have an unique ID as well.
  Refactor all LLM APIs to introduce the response ID and index and hasFinished (isCompleted) status.
  Some of the APIs provide build in unique response ID and finished reason. But many does not.
  For later we generate a synthetic IDs.

  Add ID, index and isCompleted to the following LLM API responses:

  - Anthropic, Azure, Mistral, OpenAI
  - Vertex AI Gemini. Resolve API deprivation issue. Streamline Geminie output code. Serialise citation and safety output in message  properties.
  - Ollama support
  - Bedrock cohere result metadata propagation.
  - Bedrock Llama2 and Titan
  - Berock Jurassic2
  - Bedrock Anthropic2
  - WatsanxAI
  - Vertex PaLM2, HuggingFace
  - Remove Generation.withGenerationMetadata
  - Remove AssistantPromptTemplate
  - Make all Generation field final.
  - Make all AssistantMessage fields final.

 Experiment with Message and Token window chat history

PR #937, opened 2024-06-23
Introduce id and model in ChatResponseMetadata

ThomasVitale:
* Extend ChatResponseMetadata for Anthropic (blocking, streaming)
* Add ChatResponseMetadata for Mistral AI (blocking)
* Extend ChatResponseMetadata for OpenAI (blocking)
* Deprecate gpt-4-vision-preview and replace its usage in tests because OpenAI rejects the calls, making the tests fail (see: https://platform.openai.com/docs/deprecations)

Fixes gh-936

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/metadata/AnthropicChatResponseMetadata.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/metadata/AnthropicChatResponseMetadata.java
@@ -30,6 +30,7 @@ import java.util.HashMap;
  * {@link ChatResponseMetadata} implementation for {@literal AnthropicApi}.
  *
  * @author Christian Tzolov
+ * @author Thomas Vitale
  * @see ChatResponseMetadata
  * @see RateLimit
  * @see Usage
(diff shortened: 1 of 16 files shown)


ThomasVitale:
The `OpenAiChatModel` doesn't currently support getting the `Usage` metadata when using the streaming mode. After https://github.com/spring-projects/spring-ai/pull/920 is fixed, the `OpenAiChatResponseMetadata` can be included in the streaming response as well.

Similarly, `MistralAiChatModel` doesn't currently support getting the `Usage` metadata when using the streaming mode.

tzolov:
Hey @ThomasVitale , 
Glad you've taken the responsibility of fixing this gap in our response format. The IDs are important. 
Here: https://github.com/spring-projects/spring-ai/pull/550 is an early attempt to address the missing IDs and other response handling issue (like correct handling for multiple choices response), but didn't like the result and haven't had time to give it another try. So thank your for looking at it.
Btw, in my experiments I've noticed that not all Ai Models provide ID in their response, so In those cases i've opted for generating synthetic Ids. But maybe there are better solutions.

ThomasVitale:
@tzolov thanks for the link, I'll have a look at that PR. Some providers don't return a response ID, but in that case I would keep it empty in `ChatResponseMetadata `since it's a provider-specific piece of information to trace a specific request on the provider side. If it's not provided, then it should be fine to ignore it.

For other client-driven scenarios where an ID is required on all requests (supplemented by synthetic IDs), we might need a separate concept to model that, distinct from the server-driven response ID. What do you think?

tzolov:
@ThomasVitale you have a point. 
When did my experiments I had the wrong assumption that need the server side ID in order to aggregate streaming messages. 
But this assumptions was incorrect we can do [this](https://github.com/spring-projects/spring-ai/blob/25a03728e6b4382faf7c65440804c24705ba2feb/spring-ai-core/src/main/java/org/springframework/ai/chat/model/MessageAggregator.java#L40).

tzolov:
Rebased, squashed and merged at 3227311314ca49f0ed9c36f4a67df86822c6055c

ThomasVitale:
Thank you!

PR #946, opened 2024-06-24
Use Weaviate's readiness endpoint

eddumelendez:
Currently, Testcontainers Weaviate implementation relies on port availability.
This commit switch to Weaviate's readiness endpoint.


tzolov:
Thanks @eddumelendez . Lets see how would this play.

PR #948, opened 2024-06-24
Accept String answer in BasicEvaluationTest instead of ChatResponse

habuma:
Per a discussion with @markpollack, perhaps the `evaluateQuestionAndAnswer()` method in `BasicEvaluationTest` should accept a `String` instead of a `ChatResponse`. 

By doing so, this enables tests to be written against services whose implementations are based on Spring AI (or potentially any AI/non-AI implementation) and use the evaluation test to evaluate the response, rather than requiring that the test work with a `ChatResponse`, which would require the service to expose how it's implemented.

This PR makes that change.


markpollack:
merged in 9956367fc842334dffdf2c3cfa078c9255895b97  also added not null check to question method argument.

Issue #595, opened 2024-04-17
ContentFormatterTests failed on Windows


alexcheng1982:
**Bug description**
ContentFormatterTests failed on Windows. This is a Windows specific issue.

```
org.opentest4j.AssertionFailedError: 
expected: 
  "llmKey2: value4
embedKey1: value1
embedKey2: value2
embedKey3: value3

The World is Big and Salvation Lurks Around the Corner"
 but was: 
  "llmKey2: value4
  embedKey1: value1
  embedKey2: value2
  embedKey3: value3

The World is Big and Salvation Lurks Around the Corner"
```

**Environment**
Spring AI 0.8.1
Java 17
Windows

**Steps to reproduce**
Just `mvn test` to run the tests

**Expected behavior**
ContentFormatterTests  should pass

**Minimal Complete Reproducible example**
Just the main repo

I did some investigation and found out the issue was with the `DEFAULT_METADATA_SEPARATOR`. On Windows, the default metadata separator is `\r\n`, not `\n`.

```java
private static final String DEFAULT_METADATA_SEPARATOR = System.lineSeparator();
```

The test uses a text block as the expected result, which always uses `\n`. So on Windows the assertion failed, as the formatted content has an extra `\r` in it.

The `customTextFormatter` test can be fixed by overriding the default metadata separator using `withMetadataSeparator("\n")`.  However, the `defaultConfigTextFormatter` test cannot be fixed in this way, as there is no way to configure the default `DefaultContentFormatter`.

Another option is to set the `DEFAULT_METADATA_SEPARATOR` to `\n`. Not sure about the impact of changing the default metadata separator.

markpollack:
planning to remove this, so this issue maybe no longer valid.

markpollack:
Finally closing!!!

PR #944, opened 2024-06-24
Bug fix: fix failed prompt template validation if a variable name contains hyphen(-)

nichozhan:
Bug fix for #943 .

After some troubleshooting, I find that the bug was introduced by old-versioned StringTemplate4.
There is an issue about this: https://github.com/antlr/stringtemplate4/issues/94
I upgrade StringTemplate4 to the latest version and add a UT to cover this.

Thank you for taking time to contribute this pull request!
You might have already read the [contributor guide][1], but as a reminder, please make sure to:

* Sign the [contributor license agreement](https://cla.pivotal.io/sign/spring)
* Rebase your changes on the latest `main` branch and squash your commits
* Add/Update unit tests as needed
* Run a build and make sure all tests pass prior to submission

PR #949, opened 2024-06-25
Fix Windows line separation.

szarpul:
Bug fix for #595 
* `BeanOutputConverter` - replace all Windows line separation with \n.
* `BeanOutputConverterTest` - explicitly check line separation using \n.
* Introduce `TextBlockAssertion` that asserts text blocks with line separation set to \n.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
@@ -200,7 +200,7 @@ public class BeanOutputConverter<T> implements StructuredOutputConverter<T> {
 				Here is the JSON Schema instance your output must adhere to:
 				```%s```
 				""";
-		return String.format(template, this.jsonSchema).replaceAll("\r\n", "\n");
+		return String.format(template, this.jsonSchema);
 	}
 
 }
--- a/spring-ai-core/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
@@ -200,7 +200,7 @@ public class BeanOutputConverter<T> implements StructuredOutputConverter<T> {
 				Here is the JSON Schema instance your output must adhere to:
 				```%s```
 				""";
-		return String.format(template, this.jsonSchema);
+		return String.format(template, this.jsonSchema).replaceAll("\r\n", "\n");
 	}
 
 }
--- a/spring-ai-core/src/test/java/org/springframework/ai/assertj/TextBlockAssertion.java
+++ b/spring-ai-core/src/test/java/org/springframework/ai/assertj/TextBlockAssertion.java
@@ -1,9 +1,12 @@
 package org.springframework.ai.assertj;
 
 import org.assertj.core.api.AbstractAssert;
+import org.assertj.core.api.AbstractCharSequenceAssert;
 import org.assertj.core.api.Assertions;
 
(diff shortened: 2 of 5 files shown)


tzolov:
@szarpul I'm afraid your fix is not multiplatform? have you tested it on Mac, Linux and Wind?

szarpul:
Hey @tzolov , I tested it only on Windows. Nevertheless, the fix should not affect other platforms as they don't have \r\n line separation.

nichozhan:
Hi @szarpul & @tzolov , How about using `System.lineSeparator()` as the target of replacement? 
I have a similar issue in https://github.com/spring-projects/spring-ai/pull/944.

szarpul:
@nichozhan I would avoid using `System.lineSeparator()` at all. The test intention is clear, we want to have just `\n` as a line separation. On the other hand,  `System.lineSeparator()`  produces different results on different OS thus making the test flaky (failing on Windows for instance).

szarpul:
@tzolov I've tested it on Linux using docker image [maven:eclipse-temurin-17](https://github.com/carlossg/docker-maven/blob/43cc67eb6ec38bfc7394a51f43e39ca3f24af65c/eclipse-temurin-17/Dockerfile) with a command `clean package` and all tests have passed.

szarpul:
Hey @markpollack  @tzolov 
* After the fix done in https://github.com/spring-projects/spring-ai/pull/944, the tests were still not passing on Windows.
* Following the fix for https://github.com/spring-projects/spring-ai/pull/944, we should compare text blocks against normalized EOL.
* If our goal is to have normalized EOL, then there is no sens to hardcode a `\n` as the target value in `BeanOutputConverterTest.normalizesLineEndingsClassType` test.

WDYT?

markpollack:
Thanks for hanging in there on this.  I'll dive back in and take a look.  The way this reads now has me a bit confused.  It looks like there isn't portability in text blocks across platforms?  

The related issue https://github.com/spring-projects/spring-ai/issues/595 is about `DEFAULT_METADATA_SEPARATOR` use in Documents.  

I've created a 'cross-platform' label to track / group better

szarpul:
Hey @markpollack 

To clarify the situation, it all comes to one thing, somwhere in the production code the `System.lineSeparator()` is used, which produces different result for each OS: 
* `\n` UNIX
* `\r\n` WINDOWS
* 
### BeanOutputConverterTest 
In the method ` BeanOutputConverter.generateSchema` the `System.lineSeparator()` is used:
> ObjectWriter objectWriter = new ObjectMapper().writer(new DefaultPrettyPrinter()
			.withObjectIndenter(new DefaultIndenter().withLinefeed(System.lineSeparator())));

### ContentFormatterTests
This test runs the `DefaultContentFormatter` with a `DEFAULT_METADATA_SEPARATOR` set to `System.lineSeparator()`. 
> private static final String DEFAULT_METADATA_SEPARATOR = System.lineSeparator();

### PromptTemplateTest
`PromptTemplate` class is using [StringTemplate library](https://www.stringtemplate.org/) that probably is also using the  `System.lineSeparator()` unde the hood, as the `PromptTemplate.render` metod is returning the results with Windows specific line separator.

----
### Wrap up
We could either explicitly specify to have the `\n` in the production code (don't know exactly how to enforce that for the ST library, but lets assume there is a way) or make our test adjust the line separator to the system separator, like in the preposition in this PR (maybe the `TextBlockAssertions` could be changed to something else). 

WDYT?

markpollack:
I made a pass to fix this in the tests.  https://github.com/spring-projects/spring-ai/commit/06ea88312a1788734aed849da6f8a5050dee8dce

PR #950, opened 2024-06-25
Azure AI: Support OpenAI key and MS Entra auth

### LINKED ISSUE #260: Microsoft Entra authentication
Please add Microsoft Entra authentication for Azure:
https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#authentication

tzolov:
- Add azure-identity dependency
- Enable OpenAI API key and TokenCredential auth
  - add the spring.ai.azure.openai.openai-api-key, non-azure property to allow connection to OpenAI service
  - DefaultAzureCredential bean when the AZURE_CLIENT_ID, AZURE_TENANT_ID, AZURE_CLIENT_SECRET variables are present
- Update Azure Chat and Embedding docs
- Add integration tests for OpenAI connection

Resolves #260

### DIFF
--- a/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/azure/openai/AzureOpenAiAutoConfiguration.java
+++ b/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/azure/openai/AzureOpenAiAutoConfiguration.java
@@ -17,17 +17,14 @@ package org.springframework.ai.autoconfigure.azure.openai;
 
 import java.util.List;
 
-import com.azure.ai.openai.OpenAIClient;
-import com.azure.ai.openai.OpenAIClientBuilder;
-import com.azure.core.credential.AzureKeyCredential;
-import com.azure.core.util.ClientOptions;
-
 import org.springframework.ai.azure.openai.AzureOpenAiChatModel;
 import org.springframework.ai.azure.openai.AzureOpenAiEmbeddingModel;
 import org.springframework.ai.azure.openai.AzureOpenAiImageModel;
 import org.springframework.ai.model.function.FunctionCallback;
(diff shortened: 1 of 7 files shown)


spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/azure/openai/AzureOpenAiAutoConfiguration.java
```diff
@@ -36,26 +33,82 @@
...
+	 * @param clientSecret Azure client secret
+	 * @return new TokenCredential instance
+	 */
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnProperty(name = { "azure.client.id", "azure.tenant.id", "azure.client.secret" },
+			matchIfMissing = false)
+	public TokenCredential createTokenCredential(@Value("${azure.client.id}") String clientId,
+			@Value("${azure.tenant.id}") String tenantId, @Value("${azure.client.secret}") String clientSecret) {
+		return new DefaultAzureCredentialBuilder().build();
```
markpollack, line 109: The application properties are not being used to create the TokenCredential.  The `DefaultAzureCredentialBuilder` does have a tenantId property, but does not have a property such as `secret` and there are two types of client id's so it doesn't match up.

Maybe we need to leverage spring cloud azure and it's support for microsoft entra id [here](https://learn.microsoft.com/en-us/azure/developer/java/spring-framework/authentication#other-credential-types).

tzolov: Please remove the  `createTokenCredential` and add the spring cloud azure documentation references to the Azure Chat and Embedding model documentations

PR #952, opened 2024-06-25
Provide Anthropic Claude streaming Function Calling support

### LINKED ISSUE #930: Provide Anthropic Claude streaming Function Calling support
Currently provide Anthropic support only non-streaming function calling. 
Extend the support for streaming similar to OpenAI, MistralAI, Gemini and Azure OpenAI.

didalgolab:
Fix #930


models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/AnthropicApi.java
```diff
@@ -388,6 +389,7 @@ public record MediaContent( // @formatter:off
 		@JsonProperty("id") String id,
 		@JsonProperty("name") String name,
 		@JsonProperty("input") Map<String, Object> input,
+		@JsonProperty("input_json") String inputJson,
```
tzolov, line 392: I understand the reason for this, but It is unfortunate that we need yet another field just to aggregate the partial json.

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/AnthropicStreamFunctionCallingHelper.java
```diff
@@ -0,0 +1,118 @@
...
+
+	public boolean isStreamingToolFunctionCall(StreamResponse response) {
+		if (response == null || response.contentBlock() == null) {
+			return false;
+		}
+
+		return response.contentBlock().type() == MediaContent.Type.TOOL_USE;
+	}
+
+	public boolean isStreamingToolFunctionCallFinish(StreamResponse response) {
```
tzolov, line 44: Not sure this is correct. 
Isn't it possible to have multiple "tool_use" content blocks in the same message? Perhaps for parallel function calling? 
According to the [streaming spec](https://docs.anthropic.com/en/api/messages-streaming#streaming-request-with-tool-use) a message can include multiple content blocks including tools such. 

Maybe we should use the content block stop event instead?

tzolov (commented):
Thanks @didalgolab 
Test pass and the implementation seems to be doing what is expected. 
I've left few comments. The important one is about the the function call finish trigger.

tzolov:
rebased, squashed and merged at f5b72242c03fd52688dcc3d741a188019e07170b

Issue #953, opened 2024-06-25
[Observability] Initial Observability for Models


ThomasVitale:
In the GenAI ecosystem, many solutions have been introduced that are centered around the concepts of observability and evaluation of Large Language Models. Unlike more traditional applications, **observability becomes even more a concern to be addressed already as part of the local development workflow**, because it enables refining prompts and tuning the model integration to fit the case at hand. As such, I think it would be really great to start introducing some basic observability features in Spring AI before cutting the 1.0.0 GA release. 

### Background

From a Spring AI perspective, I believe it's useful to **categorise the available GenAI observability products** as follows:

* **Products relying on OpenTelemetry** to standardize the instrumentation, collection, and visualisation of telemetry data (traces, metrics, logs, and more). Examples: [OpenLit](https://openlit.io), [OpenLLMetry](https://www.traceloop.com/openllmetry)/[TraceLoop](https://www.traceloop.com), [Arize Phoenix](https://docs.arize.com/phoenix), [LangTrace](https://langtrace.ai).
* **Products relying on proprietary/custom strategies** for instrumentation, collection, and visualisation of telemetry data. Examples: [LangSmith](https://www.langchain.com/langsmith), [LangFuse](https://langfuse.com), [Lunary](https://lunary.ai).

Considering the consolidation around OpenTelemetry (the industry standard for observability) and the many integration possibilities with the rest of the ecosystem, I would recommend drawing inspiration from the first group of products to understand better the types of features needed for observability, with a possible goal to integrate them with Spring AI at some point. There is also work in progress to agree on [**semantic conventions for GenAI applications**](https://opentelemetry.io/docs/specs/semconv/gen-ai/llm-spans/) in the OpenTelemetry project. Those conventions are still under active development, but they are being already experimentally adopted in products like OpenLit and OpenLLMetry/TraceLoop. A similar attempt of agreeing on standard naming conventions is also brought forward within the [OpenInference](https://github.com/Arize-ai/openinference/tree/main/python/openinference-semantic-conventions) group.

### Context

What does it mean observing an LLM-powered applications? There are many use cases, going from more general observability to specific evaluation scenarios. For starters, I want to focus on the fundamentals, as also highlighted in this [blog post](https://opentelemetry.io/blog/2024/llm-observability/) from OpenTelemetry. In particular, what kind of telemetry do we need?

#### LLM Request

* Provider (mistral)
* Operation (chat.completion)
* Model (open-mistral-7b)
[...]

PR #954, opened 2024-06-25
Initial Observability for Models in Spring AI

ThomasVitale:
* Observation APIs for chat, embedding and image models
* Conventions based on OpenTelemetry Semantic Conventions for GenAI
* Instrumentation for OpenAI chat, embedding, and image models
* Autoconfiguration for observability for OpenAI

See proposal and design in gh-953.

### DIFF
--- a/models/spring-ai-openai/pom.xml
+++ b/models/spring-ai-openai/pom.xml
@@ -89,6 +89,12 @@
 		</dependency>
 
 		<dependency>
+			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-observation-test</artifactId>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
 			<groupId>org.testcontainers</groupId>
 			<artifactId>qdrant</artifactId>
 			<scope>test</scope>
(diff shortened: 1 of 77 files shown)


codefromthecrypt (commented):
interesting and thanks for adding IT tests. FYI, I went around to other public projects like openlit, langtrace, openinference, openllmetry, litellm and agiflow to see how far folks are from the current semantics and right now langtrace and openlit are very close.

Note this eval was based on the current main code which should be cut as the next semver release in the next day or two https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md

A quick glance here I noticed some drifts maybe because you were looking at last published vs latest (to be published). If you want second eyes, lemme know when ready and I'll try to help validate.

ThomasVitale:
@codefromthecrypt thanks a lot for reviewing the pull request, much appreciated. This will be a step 1 towards a larger observability arrangement. There's a few things right now that need to be synced with the SemConv, mostly those that are currently under discussions (like how to shape the prompt and completion content) and some metrics, but we'll get there. 

I'm currently working on some documentation, including an overview of the status of the GenAI SemConv adoption in Spring AI, for better tracking the process. I'm planning to publish that in the upcoming days.

OpenLit and OpenLLMetry are great projects, I'm also following their adoption journey of the semantic conventions.

PR #969, opened 2024-06-28
Correctness evaluator

habuma:
This PR adds a new evaluator to judge correctness of a response. It is based loosely on LlamaIndex's `correctness.py` evaluator.

Note that I couldn't decide the best way to provide the reference answer. I saw three choices:

 - Add it directly to `EvaluationRequest` - This seemed like the worst choice, as it would be a property that is probably only used for this evaluator.
 - Override the `evaluate()` method to take an `EvaluationRequest` as well as the reference answer. I started this way, but it felt clunky.
 - Extend `EvaluationRequest` with `CorrectnessEvaluationRequest` and implement `evaluate()` to check for a `CorrectnessEvaluationRequest` and use its reference answer. This is the option I chose.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/evaluation/CorrectnessEvaluationRequest.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/evaluation/CorrectnessEvaluationRequest.java
@@ -1,0 +1,27 @@
+package org.springframework.ai.evaluation;
+
+import java.util.List;
+
+import org.springframework.ai.model.Content;
+
+/**
+ * Represents an evaluation request for correctness evaluation.
+ *
+ * @author Craig Walls
+ * @since 1.0.0 M1
+ */
+public class CorrectnessEvaluationRequest extends EvaluationRequest {
+
+	private final String referenceAnswer;
+
+	public CorrectnessEvaluationRequest(String userText, List<Content> dataList, String responseContent,
+			String referenceAnswer) {
+		super(userText, dataList, responseContent);
+		this.referenceAnswer = referenceAnswer;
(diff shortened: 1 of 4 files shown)


ilopezluna (commented):
Some time ago, I worked on a project very similar to this one. I found it very useful to include the reason in the response. You can see an example here:
[ValidatorAgent.java](https://github.com/AtomicJar/genai-test/blob/main/src/test/java/com/example/ragchat/ValidatorAgent.java)

I believe using a score to evaluate the answers might not be the best option, as the expected output for a test is typically a boolean (test passes or not). Using a score could make it harder to properly test an answer because the developer would need to decide on a threshold, which could be very arbitrary.

habuma:
This returns both a pass/fail boolean and an explanation. The pass/fail is determined by the score threshold, such that if the score is below a certain threshold, then the test fails. And the explanation is provided in the `feedback` property of the `EvaluationResponse`.

ilopezluna:
> This returns both a pass/fail boolean and an explanation. The pass/fail is determined by the score threshold, such that if the score is below a certain threshold, then the test fails. And the explanation is provided in the `feedback` property of the `EvaluationResponse`.

I believe it makes more sense to rely on the judgment of the LLM to determine if the test passes or not. However, I've seen many examples where evaluations are based on scores, so I might be wrong.

If you always want the explanation included in the response, I suggest being more explicit in the prompt. Currently, you are asking for:

_Output a single score that represents a holistic evaluation._

I believe this could lead the LLM to omit the explanation. When I was working on this, I realized that I had to be extremely explicit about what I wanted.

PR #975, opened 2024-06-28
Fix OpenAiChatModel NPE

### LINKED ISSUE #970: OpenAiChatModel NullPointerException

**Bug description**
image
image
org.springframework.ai.openai.OpenAiChatModel#stream

when roleMap is Empty throw NullPointerException

devholic22:
* Fix OpenAiChatModel NPE using getOrDefault

closed #970 

I fixed this issue using `Map.getOrDefault`


tzolov:
@devholic22 thanks for addressing the issue

PR #976, opened 2024-06-28
Bump Tika to avoid PDFBox conflicts

Grogdunn:
Apache tika 2.x.x wants old version of PDFBox 2.x, bump to tika 3.0.0-BETA that want PDFBox 3.x. With this we can use Tika as PDF reader.


tzolov:
Thanks @Grogdunn 
I will merge the version update, but for my understanding did you see an actual failures while using the previous version?

PR #1000, opened 2024-07-03
Add comment in MessageType for readability

### LINKED ISSUE #972: Add comment in MessageType for readability
When I first saw MessageType, I didn't understand easily what each value meant. So I wrote an comment by referring to the description of each implementation, which I thought would be easy for beginners to understand, so I wrote it. I'd appreciate it if you could review it.

please see #1000

devholic22:
- Add comment in MessageType for readability

resolved #972


tzolov:
thanks @devholic22

PR #1001, opened 2024-07-03
Missed call to setEmbedding on specific VectorStore implementations

fjtorres:
During the use of spring-ai we detected that the `PgVectorStore` class didn't call the method `setEmbedding` for each document added to the store, we were using it to do some clustering work based on each document added to the store. After check the source code we detected the same problem on `MilvusStore`.

The unique change on this PR is adding the call to `setEmbedding` in each store implementation.


markpollack:
cool, thanks.

merged in 867bd173e5e400d9a3f4a9317bbf42afcd41a0b0

Issue #405, opened 2024-03-06
Why are embeddings stored as List<Double>?


agoerler:
Hi,

I am wondering why the class [Embedding](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/embedding/Embedding.java) internally stored the vector as `List<Double>`. I think `double[]` or `float[]` would be more memory efficient.

youngmoneee:
### List

Firstly, List is a basic collection in Java, capable of storing only **objects** instead of primitive types like double or float.
Using collections allows for **more flexible** usage than basic arrays of primitive types, and depending on the use case, you can choose implementations such as ArrayList or LinkedList.

If you need to add a new value exceeding the predefined size of a double[n] array, you must allocate a new double[n+1] and copy the existing values before adding the new one. However, when using a List, you can **easily** handle adding values to the array through the abstracted add() method.

### Double

A primitive type double variable defaults to 0.0. Can you determine if an operation was successful or failed when the result is 0.0?
In such cases, **'null'** can serve as a means to express the absence of a value or an impossible calculation, and by using the Double wrapper class, which inherits from Object, you can **represent null**.

aseovic:
@youngmoneee I believe that's exactly the point @agoerler is trying to make.

While the list may be more _flexible_, it is significantly less _efficient_ than a simple `float[]` which is why pretty much LLM and most frameworks that work with them, including Langchain and Langchain4j, use primitive `float[]` to represent embeddings.

For one, the additional flexibility offered by a list is completely unnecessary, as the operations you need to do with the embeddings, such as calculations necessary for the similarity search, are just as easy, if not easier to do with arrays than with lists, and you can actually leverage SIMD support in the modern CPUs to do them more efficiently. Not to mention that you can actually use classes like `FloatBuffer` to access `float[]` allocated in native memory by ONNX, for example, directly, but can't do that with a list and have to box every single value into a `Double` on the way in (and likely unbox into a `float` on the way out).

But more importantly, storing 1,536 dimension vector returned by OpenAI, for example, requires 6,144 bytes when stored as a `float[]`, and 36,864 bytes when stored as a `List<Double>`, for the exact same payload (ignoring the space used by the array or list instance itself, which is approximately the same and irrelevant in this case). That's 6x the memory cost, for no good reason, as it is neither faster nor easier to work with. Quite the opposite, actually.

[...]

PR #1002, opened 2024-07-03
Replace the Embedding format from List<Double> to float[]

tzolov:
- Adjust all affected classes including the Document.
 - Update docs.

 Related to #405

### DIFF
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiEmbeddingModel.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiEmbeddingModel.java
@@ -29,7 +29,9 @@ import org.springframework.ai.embedding.Embedding;
 import org.springframework.ai.embedding.EmbeddingRequest;
 import org.springframework.ai.embedding.EmbeddingResponse;
 import org.springframework.ai.embedding.EmbeddingResponseMetadata;
+import org.springframework.ai.model.EmbeddingUtils;
 import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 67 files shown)


aseovic:
Well, it's (slightly) better than `List<Double>`... It's using 16 (instead of 24) bytes per vector dimension, which is still a far cry from 4 bytes per dimension a `float[]` would use. And it still requires boxing and unboxing to do anything useful with, which adds up.

I understand the reluctance to use `float[]` directly, as it ties you to it and makes it difficult to use any other vector type (and models such as Cohere are already capable of creating int8 and bit embeddings, and any float32 vector can be quantized into similar embeddings to save space, regardless of the model used), but what we've done to address that is introduce `Vector<T>` interface, where `T` is one of `float[]`, `byte[]`, or `BitSet` via corresponding `Float32Vector`, `Int8Vector` and `BitVector` implementations.

It's really not that difficult to do, and it allows you to store the embeddings using the optimal memory representation, and provides a place to add common vector properties and operations on, such as `dimension`, `magnitude`, `dotProduct`, etc.

tzolov:
@aseovic, I guess your assumptions are based on the presumption that all Vector Store APIs take float[]. My experience integrating with those shows that many use List<Float> and some even List<Double> to represent vectors. So for those the float[] won't make difference or likely introduce additional boxing/unboxing steps. 

Having said this i've updated the PR to float[]. Lets see how this plays out.

markpollack:
@aseovic 

> but what we've done to address that is introduce Vector<T> interface

Where?

markpollack:
Fixed a few tests and the typesense impl.  Merged as d538e00643b3dcab0a3e0595aa406775733aaf3e

aseovic:
> @aseovic, I guess your assumptions are based on the presumption that all Vector Store APIs take float[]. My experience integrating with those shows that many use List and some even List to represent vectors. So for those the float[] won't make difference or likely introduce additional boxing/unboxing steps.

I guess my take on it as that you should optimize for the vector stores that use _optimal_ data structure to represent vectors in Java, not for the ones that don't ;-) 

In other words, make the inefficient stores pay the price of that conversion if they need to, instead of punishing everyone. So yes, `float[]` is a much better option. Thank you!

aseovic:
> @aseovic
> 
> > but what we've done to address that is introduce Vector interface
> 
> Where?

In [Coherence](https://github.com/oracle/coherence/blob/main/prj/coherence-core/src/main/java/com/oracle/coherence/ai/Vector.java).

PR #1006, opened 2024-07-04
Use gson instead of fastjson

m1ngyuan:
Use gson instead of fastjson


codespearhead:
I wonder why not just use Jackson since it's already defined in the parent POM?

markpollack:
What is the motivation?  @tzolov will need to confirm, but we have had several issues with incompatible versions of libraries that different 3rd party libraries requires - in particular at autoconfiguration support is contained in a single project.

markpollack:
reopen if you can make a strong case for this.

PR #1015, opened 2024-07-08
Extend EmbeddingResponseMetadata with model and usage

ThomasVitale:
Model name and token usage are common response metadata that are currently passed as free-text key/value pairs. This change make them part of the interface so that they can be programmatically and safely used
for evaluation and observability purposes.

@tzolov This change enables a better solution for https://github.com/spring-projects/spring-ai/issues/953, in particular this part: https://github.com/ThomasVitale/spring-ai/blob/834679a42c089502e2db07419242e96cf8f6bb44/spring-ai-core/src/main/java/org/springframework/ai/embedding/observation/DefaultEmbeddingModelObservationConvention.java#L124


markpollack:
reviewed and merged in e601f1436451c02345363107f3dd33d400eb7477

PR #1019, opened 2024-07-09
Add self-query implementation

florind:
* Add a `SelfQueryAdvisor` that follows LangChain's [SelfQueryRetriever](https://python.langchain.com/v0.1/docs/modules/data_connection/retrievers/self_query/):
`Specifically, given any natural language query, the retriever uses a query-constructing LLM chain to write a structured query and then applies that structured query to its underlying VectorStore. `
* The prompt is taken from [here](https://github.com/langchain-ai/langchain/blob/859e43493262ff6f54356b5ecd8c0ef7ebfe2307/libs/langchain/langchain/chains/query_constructor/prompt.py#L118)
* To be able to extract query attributes, a list of `AttributeInfo` objects must be specified. This instructs the LLM to identify attributes in the given user query then query the vector store using the metadata attributes.
* Example: 
```
        var selfQueryAdvisor = SelfQueryAdvisor(
          listOf(
                AttributeInfo(
                name = "firstName",
                type = "string",
                description = "A person's first name"),
          ),
          vectorStore,
          SearchRequest.defaults(),
          chatModel,
        );
       
        var response = chatClient.prompt()
                         .advisors(selfQueryAdvisor)
                         .user("Return a list of persons whose first name is Kylian").chatResponse();
```

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/SelfQueryAdvisor.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/SelfQueryAdvisor.java
@@ -1,6 +1,8 @@
 package org.springframework.ai.chat.client.advisor;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.ai.chat.client.AdvisedRequest;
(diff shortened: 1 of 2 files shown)


tzolov:
Hi @florind and thank you for the contribution. 
Interesting stuff! Let me dig a bit through the LangChain docs

PR #1021, opened 2024-07-10
Allow custom metadata in all document reader implementations

rubin0:
Not all implementations of DocumentReader allow adding metadata to a document during the document loader step. Only TextReader and JsonReader allow it, using different methods.

I have added support for custom metadata in the following implementations, using the same structure as in the TextReader implementation:

- PagePdfDocumentReader
- ParagraphPdfDocumentReader
- TikaDocumentReader

I have updated the tests as well. However, the ParagraphPdfDocumentReader currently only checks for a negative case because I don't have any PDF with the correct license and structure to create a positive case.


stohirov:
@rubin0 conflicts between main and current branch should be resolved. If help needed, I can update the PR

Issue #698, opened 2024-05-09
MongoDB is unable to persist `Document` types


tzolov:
MongoDB is unable to persist `Document` types. The problem is the new `Media` class and particularly the `MimeType` filed is contains: 

```
java.lang.reflect.InaccessibleObjectException: Unable to make protected java.nio.charset.Charset(java.lang.String,java.lang.String[]) accessible: module java.base does not "opens java.nio.charset" to unnamed module @4ee22985
 at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible([AccessibleObject.java:354](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html))
 at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible([AccessibleObject.java:297](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html))
 at java.base/java.lang.reflect.Constructor.checkCanSetAccessible([Constructor.java:188](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html))
 at java.base/java.lang.reflect.Constructor.setAccessible([Constructor.java:181](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html))
 at org.springframework.util.ReflectionUtils.makeAccessible([ReflectionUtils.java:201](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html))
 at org.springframework.data.mapping.PreferredConstructor.([PreferredConstructor.java:56](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html))
 at org.springframework.data.mapping.model.PreferredConstructorDiscoverer$Discoverers.buildPreferredConstructor([PreferredConstructorDiscoverer.java:265](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html))
[...]

PR #1022, opened 2024-07-10
Added MimeTypeConverters class

rozza:
Allows the converters to be shared between the autoconfigure codebase and the vector-stores code.

Reenabled the vector-stores code and fixed the TestApplication (so that it uses the MimeTypeConverters).

Annotated tests to check for the `OPEN_API_KEY` environment variable.
Updated to `mongodb/atlas:v1.24.0` in vector-stores so inline with the autoconfigure
test codebase.

Related to #698 and the auto configure fix in a1afb0d


markpollack:
I had made a [commit](https://github.com/spring-projects/spring-ai/commit/a1afb0db4644f216b3fead6142357fa468b7ca19) for https://github.com/spring-projects/spring-ai/issues/698 and I somehow missed this.  

Aside from where the converters should live (in a separate class or not) the bigger issue I found was that 

```
@AutoConfiguration(after = MongoDataAutoConfiguration.class)
@ConditionalOnClass({ MongoDBAtlasVectorStore.class, EmbeddingModel.class, MongoTemplate.class })
@EnableConfigurationProperties(MongoDBAtlasVectorStoreProperties.class)
public class MongoDBAtlasVectorStoreAutoConfiguration {
```

The use of `after` in the autoconfiguration is what was leading to the converter not being replaced.  Removing that fixed the overall issue and no need to use `BeanPostProcessor`  I am inquiring if that config should use `before`.   @rozza would you be able to verify that?

rozza:
> The use of after in the autoconfiguration is what was leading to the converter not being replaced. Removing that fixed the overall issue and no need to use BeanPostProcessor I am inquiring if that config should use before. @rozza would you be able to verify that?

Hi @markpollack - verified using `@AutoConfiguration(before = MongoDataAutoConfiguration.class)` works. I've updated the PR.

I'm not sure where the converters should live, I opted to move them into the `org.springframework.ai.vectorstore.convert` namespace so that they could be reused by the `TestApplication` which doesn't have dependency on `rg.springframework.ai.autoconfigure.vectorstore.mongo`. Not sure what is the correct idiomatic path forward here.

sobychacko:
@rozza, I missed the existence of this PR while fixing the MongoDB vector store test failures. I think we can still take some ideas in this PR and refactor some code duplications. For now, I will close this PR, and we can consider opening a new one on top of the changes from https://github.com/spring-projects/spring-ai/pull/1255. Thanks!

PR #1023, opened 2024-07-10
Fix leaked dependencies in autoconfigure module

ThomasVitale:
The `spring-ai-spring-boot-autoconfigure` module had a few dependencies that were included in the final JAR instead of being marked as optional (it's the starter dependency that is supposed to include compile-scope dependencies when needed).

Also, the PostgreSQL dependency was duplicated.

### DIFF
--- a/spring-ai-spring-boot-autoconfigure/pom.xml
+++ b/spring-ai-spring-boot-autoconfigure/pom.xml
@@ -25,6 +25,7 @@
 			<groupId>com.google.protobuf</groupId>
 			<artifactId>protobuf-java</artifactId>
 			<version>${protobuf-java.version}</version>
+			<optional>true</optional>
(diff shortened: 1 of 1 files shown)


spring-ai-spring-boot-autoconfigure/pom.xml
```diff
@@ -370,17 +373,12 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-jdbc</artifactId>
 			<scope>test</scope>
 		</dependency>
-		<dependency>
```
ThomasVitale, line 378: This was a duplicate

Issue #1017, opened 2024-07-09
Qianfan's chatmode returns an error issue due to the presence of system messages in the prompt.


llggookkokok:
Use the following code to execute correctly。

org.springframework.ai.qianfan.QianFanChatModel

public ChatCompletionRequest createRequest(Prompt prompt, boolean stream) {
        var chatCompletionMessages = prompt.getInstructions()
                .stream()
                .map(m -> new ChatCompletionMessage(m.getContent(),
                        ChatCompletionMessage.Role.valueOf(m.getMessageType().name())))
                .toList();
        var systemMessageList = chatCompletionMessages.stream().filter(msg -> msg.role() == Role.SYSTEM).toList();
        **var chatMessages = chatCompletionMessages.stream().filter(msg -> msg.role() != Role.SYSTEM).toList();**

......

 var request = new ChatCompletionRequest(chatMessages, systemMessage, stream);

....

mxsl-gr:
hi, i have reproduced the issue and will fix it later. thank you very much for reporting it

mxsl-gr:
fixed in PR https://github.com/spring-projects/spring-ai/pull/1029

PR #1029, opened 2024-07-11
fix: QianFan role based chat model & unit tests

mxsl-gr:
fix `QianFan` chat model error on role based message call and add role based unit tests

related issue https://github.com/spring-projects/spring-ai/issues/1017


markpollack:
thanks!

PR #1040, opened 2024-07-12
Improve new function calling strategy

ThomasVitale:
Small implementation changes to the function calling mechanism. Right now, in case there is a “tool call” request from a model, we directly do the recursive call. What I suggest in the PR is actually building the final `ChatResponse` object first. Then, we can return it directly if there’s no tool call. If there is, we trigger the recursive call.

With this change, we can correctly handle the observation scopes, making sure each observation contains the correct request/response pair in the context.

We can also generalize a bit more the function calling logic. Working directly with a `ChatResponse` object, I managed to remove the need for making the abstract class generic, moving even more logic up and reducing the amount of code that is provider-specific.

One more thing I changed is wrapping only the call to OpenAI within `RetryTemplate`. Right now, it wraps the entire implementation, catching many different exceptions not related to the HTTP call, and retrying even though it shouldn’t (impacting both runtime code but also tests). I would suggest progressively making this change for all implementations. Besides reaching a more predictable behaviour in terms of retries, it helps making the code more readable without unnecessary nesting, especially after delivering the observability changes (which would add one more nested level).

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -35,7 +35,6 @@ import org.springframework.ai.model.function.FunctionCallbackContext;
 import org.springframework.ai.openai.api.OpenAiApi;
 import org.springframework.ai.openai.api.OpenAiApi.ChatCompletion;
 import org.springframework.ai.openai.api.OpenAiApi.ChatCompletion.Choice;
-import org.springframework.ai.openai.api.OpenAiApi.ChatCompletionFinishReason;
 import org.springframework.ai.openai.api.OpenAiApi.ChatCompletionMessage;
 import org.springframework.ai.openai.api.OpenAiApi.ChatCompletionMessage.ChatCompletionFunction;
 import org.springframework.ai.openai.api.OpenAiApi.ChatCompletionMessage.MediaContent;
@@ -79,7 +78,7 @@ import java.util.concurrent.ConcurrentHashMap;
  * @see StreamingChatModel
  * @see OpenAiApi
  */
-public class OpenAiChatModel extends AbstractToolCallSupport<ChatCompletion> implements ChatModel {
+public class OpenAiChatModel extends AbstractToolCallSupport implements ChatModel {
 
 	private static final Logger logger = LoggerFactory.getLogger(OpenAiChatModel.class);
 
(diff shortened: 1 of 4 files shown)


spring-ai-core/src/main/java/org/springframework/ai/model/function/AbstractToolCallSupport.java
```diff
@@ -149,6 +151,21 @@ protected List<ToolResponseMessage> executeFuncitons(AssistantMessage assistantM
...
+		}
+
+		var generations = chatResponse.getResults();
+		if (CollectionUtils.isEmpty(generations)) {
+			return false;
+		}
+
+		var generation = generations.get(0);
+		return !CollectionUtils.isEmpty(generation.getOutput().getToolCalls())
+				&& toolCallFinishReason.equalsIgnoreCase(generation.getMetadata().getFinishReason());
```
ThomasVitale, line 168: This check could be improved further if we introduced some abstractions of the FinishReason, too

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -143,135 +142,134 @@ public ChatResponse call(Prompt prompt) {
...
-
-		List<AssistantMessage.ToolCall> assistantToolCalls = nativeAssistantMessage.toolCalls()
-			.stream()
-			.map(toolCall -> new AssistantMessage.ToolCall(toolCall.id(), "function", toolCall.function().name(),
-					toolCall.function().arguments()))
-			.toList();
-
-		AssistantMessage assistantMessage = new AssistantMessage(nativeAssistantMessage.content(), Map.of(),
-				assistantToolCalls);
+	private static Generation buildGeneration(Choice choice, Map<String, Object> metadata) {
```
ThomasVitale, line 246: This part was duplicated in both "call()" and "stream()". I moved it out to avoid adding more duplications with the tool call handling.

tzolov:
Thanks @ThomasVitale
Streamline the Generation constructor implementation, rebase, squash and merge at: 40d8671f3e4ec12b26f37a017e706da3a16b7ab5

tzolov:
Now we have to repeat this exercise for the remaining Function Calling models :)

ThomasVitale:
@tzolov great, thank you! Tomorrow, I'll update the observability PR and work on extending this functional calling strategy to the other models.

PR #1041, opened 2024-07-12
Registering JavaTimeModule to ObjectMapper

v891:
Small fix to register the JavaTimeModule to ObjectMapper in BeanOutputConverter.


tzolov:
Thanks @v891

Issue #613, opened 2024-04-22
[Azure OpenAI] Function calling - Can't execute multiple function calls


tha2015:
**Bug description**
When there should be multiple function calls for a single question, only one call is made because AzureOpenAiChatClient doesn't copy tools value from first request to second request, as a result, the next function call will not happen (loosing context in the function call response). 

**Environment**
spring-ai-azure-openai-1.0.0-20240420.063719-112.jar

**Steps to reproduce**
- AzureOpenAiChatClient  makes a call to gpt-4 with question "What's the weather like in San Francisco, Tokyo, and Paris?" and function "weatherFunction"
- AzureOpenAiChatClient  receives a function call in response for "San Francisco"
- AzureOpenAiChatClient  calls the function and builds a second request including the result, sends new request 
- AzureOpenAiChatClient  receives the result for "San Francisco" and stops

**Expected behavior**
Multiple function calls should happen (for SanFrancisco, Tokyo, Paris).

**Analysis**
This issue doesn't happen if using OpenAIChatClient. And
comparing [OpenAiChatClient.doCreateToolResponseRequest](https://github.com/spring-projects/spring-ai/blob/65e6880fa0f8db22ebcf780c3c840dc5e063b37f/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatClient.java#L327) and [AzureOpenAiChatClient.doCreateToolResponseRequest](https://github.com/spring-projects/spring-ai/blob/65e6880fa0f8db22ebcf780c3c840dc5e063b37f/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java#L430) shows that OpenAiChatClient uses `ModelOptionsUtils.merge(newRequest, previousRequest, ChatCompletionRequest.class)` to copy attributes (including `tools`) from first request to second request, while AzureOpenAiChatClient uses [AzureOpenAiChatClient.merge](https://github.com/spring-projects/spring-ai/blob/65e6880fa0f8db22ebcf780c3c840dc5e063b37f/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatClient.java#L386) to copy attributes (missing `tools`).
So it looks like missing `tools` property is causing the issue.

**Minimal Complete Reproducible example**

```kotlin
   // Kotlin code snippet
    val response: ChatResponse = chatClient.call(
        Prompt(
[...]

PR #1042, opened 2024-07-12
Multiple Roundtrips for functions calls are failing in Azure OpenAI Chat Chatcompletion

timostark:
If the user question (or the available functions) force multiple roundtrips of the user flow, the existing Azure OpenAI implementation only calls functions in the first roundtrip.

**Example Question**: What's the weather like in San Francisco? If the weather is above 25 degrees, please check the weather in Tokyo and Paris

==> Will return: 
grafik

The reason for this issue is, that the second call to Azure OpenAI (including the tool_id response) does not include the tools of the original request anymore. This of course makes the model unable to call another tool if required.

Unfortunatly there are additional refactorings required to enable streaming support - see first comment.

Should/Might also fix:
- https://github.com/spring-projects/spring-ai/discussions/588
- https://github.com/spring-projects/spring-ai/issues/613 (at least partly)

### DIFF
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatModel.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatModel.java
@@ -44,7 +44,9 @@ import com.azure.core.util.IterableStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.ai.azure.openai.metadata.AzureOpenAiChatResponseMetadata;
+import org.springframework.ai.chat.messages.AssistantMessage;
 import org.springframework.ai.chat.messages.Message;
+import org.springframework.ai.chat.messages.ToolResponseMessage;
 import org.springframework.ai.chat.metadata.ChatGenerationMetadata;
 import org.springframework.ai.chat.metadata.PromptMetadata;
 import org.springframework.ai.chat.metadata.PromptMetadata.PromptFilterMetadata;
@@ -59,13 +61,9 @@ import org.springframework.ai.model.function.FunctionCallbackContext;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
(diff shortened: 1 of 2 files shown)


timostark:
remark.. that is not working for streaming unfortunatly :-(

timostark:
One last remark.. The upgrade to azure-ai-openai 1.0.0.-beta.10 (from 8) causes a ton of JsonEOFException during streaming && tool-calls (reproducible also before my change).

      <artifactId>azure-ai-openai</artifactId>
      <version>1.0.0-beta.10</version>

These issues do not occur with beta 8.

grafik

tzolov:
Hi @timostark , I'm so grateful for your efforts in identifying and solving this issue.  Great job!

Currently, I'm refactoring, the Azure OpenAI function calling support to raise the API to the Spring AI API level. Similar to what I did recently fro OpenAI, Anthropic and VertexAI Gemini. 
Have the feeling that this might solve the issue for both call and streaming. Will reuse your ITs to verify it.

Btw, what name should I use for the javadoc `@author`? Your timostark handle or something else?

timostark:
@tzolov  i've quickly raised https://github.com/spring-projects/spring-ai/pull/1054 to avoid double efforts.. I was refactoring Azure AI to be compliant to OpenAI API and also switch to the Async API. By that the streaming is actually working...

I've also reverted to beta 8 and reported an azure bug ( https://github.com/Azure/azure-sdk-for-java/issues/41164 )

tzolov:
@timostark , I'm not sure what causes the json reading errors, but it seems to be some Azure API bug. 

Even if I iterate the direct openAIClient response without and processing:

```java
IterableStream<ChatCompletions> chatCompletionsStream = this.openAIClient.getChatCompletionsStream(options.getModel(), options);
chatCompletionsStream.stream().forEach(cc -> System.out.println(cc));
``` 

I still see errors like this, showing that the underlining json have been cut in the middle

```
ERROR 82446 --- [oundedElastic-1] c.a.c.i.s.DefaultJsonSerializer          : com.azure.json.implementation.jackson.core.io.JsonEOFException: Unexpected end-of-input within/between Object entries at [Source: (byte[])"{"choices":[{"content_filter_results":{"hate":{"filtered":false,"severity":"safe"},"self_harm":{"filtered":false,"severity":"safe"},"sexual":{"filtered":false,"severity":"safe"},"violence":{"filtered":false,"severity":"safe"}},"delta":{"content":" Story"},"finish_reason":null,"index""; line: 1, column: 285]
```

tzolov:
@timostark, 
please feel free to review this PR before merged: https://github.com/spring-projects/spring-ai/pull/1055

the https://github.com/spring-projects/spring-ai/pull/1054 come bit late for me.  But if there is something  in 1054 missing in 1055 feel free to rise another PR. 

Again , thank you very much for the effort.

tzolov:
Most of it is addressed by the feb036d2f6534505ad0bdac12925f93a52737e73

PR #1043, opened 2024-07-13
Fix not-null assertion in PdfDocumentReaderConfig

alexcheng1982:
This appears to be a copy-and-paste issue.


tzolov:
Thanks @alexcheng1982

PR #1054, opened 2024-07-16
Support actual streaming for AzureAI 

timostark:
Follow up of https://github.com/spring-projects/spring-ai/pull/1042

As already raised in a few PRs the streaming is not really working at the moment for azure.. This PR contains the following fixes:

1. Refactor Azure to actually work identically to OpenAI. The code is copied from the OpenAIModel. Retry template is not yet implemented. This looks like a lot of changes in the PR, but in general it is just a copy/paste from OpenAI (with some adjustments so that is compliant to the Azure Library)
2. Switch to Async Azure Client, so that streaming is *actually* working (by switching to the Azure Async Client). The documentation is super hard to understand (by Azure), but after testing it is clear that only the async library is actually async and streams at least line by line (not token by token) -  see https://github.com/spring-projects/spring-ai/pull/796 ++ https://github.com/spring-projects/spring-ai/pull/796 --> Note: this change might be breaking for spring ai consumers, as this will cause function al
3. rollback to azure beta 8, due to issues in the JSON decoder errors ( https://github.com/Azure/azure-sdk-for-java/issues/40629 )

Can you have a look if that is the correct direction for you?

### DIFF
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatModel.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatModel.java
@@ -16,31 +16,8 @@
 package org.springframework.ai.azure.openai;
 
 import com.azure.ai.openai.OpenAIClient;
-import com.azure.ai.openai.models.ChatChoice;
-import com.azure.ai.openai.models.ChatCompletions;
-import com.azure.ai.openai.models.ChatCompletionsFunctionToolCall;
-import com.azure.ai.openai.models.ChatCompletionsFunctionToolDefinition;
-import com.azure.ai.openai.models.ChatCompletionsJsonResponseFormat;
-import com.azure.ai.openai.models.ChatCompletionsOptions;
-import com.azure.ai.openai.models.ChatCompletionsResponseFormat;
-import com.azure.ai.openai.models.ChatCompletionsTextResponseFormat;
-import com.azure.ai.openai.models.ChatCompletionsToolCall;
-import com.azure.ai.openai.models.ChatCompletionsToolDefinition;
-import com.azure.ai.openai.models.ChatMessageContentItem;
-import com.azure.ai.openai.models.ChatMessageImageContentItem;
-import com.azure.ai.openai.models.ChatMessageImageUrl;
-import com.azure.ai.openai.models.ChatMessageTextContentItem;
-import com.azure.ai.openai.models.ChatRequestAssistantMessage;
-import com.azure.ai.openai.models.ChatRequestMessage;
-import com.azure.ai.openai.models.ChatRequestSystemMessage;
-import com.azure.ai.openai.models.ChatRequestToolMessage;
-import com.azure.ai.openai.models.ChatRequestUserMessage;
-import com.azure.ai.openai.models.CompletionsFinishReason;
-import com.azure.ai.openai.models.ContentFilterResultsForPrompt;
-import com.azure.ai.openai.models.FunctionCall;
-import com.azure.ai.openai.models.FunctionDefinition;
+import com.azure.ai.openai.models.*;
 import com.azure.core.util.BinaryData;
-import com.azure.core.util.IterableStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.ai.azure.openai.metadata.AzureOpenAiChatResponseMetadata;
@@ -56,7 +33,7 @@ import org.springframework.ai.chat.model.Generation;
 import org.springframework.ai.chat.prompt.ChatOptions;
 import org.springframework.ai.chat.prompt.Prompt;
 import org.springframework.ai.model.ModelOptionsUtils;
-import org.springframework.ai.model.function.AbstractFunctionCallSupport;
+import org.springframework.ai.model.function.AbstractToolCallSupport;
 import org.springframework.ai.model.function.FunctionCallbackContext;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
@@ -64,7 +41,7 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import java.util.*;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * {@link ChatModel} implementation for {@literal Microsoft Azure AI} backed by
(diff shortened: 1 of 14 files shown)


timostark:
Also remark: This PR is nicely working with a pretty big professional project with >50 functions and streaming. I do not have access to the Images service though. So would be nice if you can also check if that is working for you.

tzolov:
Reopening this as a reference util @timostark submits the new PR. Looking forward for it :)

tzolov:
> I do not have access to the Images service though. So would be nice if you can also check if that is working for you.

Not sure I follow?

timostark:
> Reopening this as a reference util @timostark submits the new PR. Looking forward for it :)

@tzolov: So I had to debug that for quite a while now.. The point is that streaming is actually working (unfortunatly not really smooth, but that is due to veeeery strange code inside the byte-buffer / event handling of azure.....). This is especially noteable when just running the stream as a unit test and logging the timestamps.

The main "issue" is that for reactive web-services to actually work, the returned stream apperantly needs an own subscribeOn/publishOn with own Schedulers.
I am unfortunatly not an expert at all in the usage of Schedulers, but my "naive" assumption is, that the Event-Stream implementation / sending is somewhat blocked by the Azure OpenAI call on the same thread and can not send out the events until the event stream is done. 

The usage of Schedulers is not required when using the Azure OpenAI Async library - because probably there is no blocking thread in the thread of the rest call.
I would normally suggest, to go to the async API approach, which internally creates a own thread for the Azure OpenAI call (which avoids that blocking behavior) - but I am not sure if the async library is a little slower (also for the non-streaming scenario). I will give that a try in the nighttime to compare the performance of both approaches (at the moment,the performance is generally not that good..)

Alternatives would be
a) to just update the documentation and use Schedulers in the consumer code. What is your opinion about that?
b) switch to Azure Async API (possibly using a spring property to enable/disable)
c) just start a new thread on our own inside the stream method and continue to use the standard Azure API (will prevent any downgrade of the non-streaming scenarios). Maybe using the Schedulers or virtual threads fuctionality inside the Azure stream function.

What is your opinion here? I would tend for option (c) but first would like to hear your opinion..

Code what is working using normal client.

`    @GetMapping(value = "/chat/{chatId}/start", produces = MediaType.TEXT_EVENT_STREAM_VALUE)
    public Flux<ChatMessageDto> chat(@PathVariable Long chatId) {
        log.info("Starting with chatId: {}", chatId);

        return chatBotService.getMessageProcessingStreamForChat(chatId)
                .subscribeOn(Schedulers.boundedElastic())
                .publishOn(Schedulers.parallel());
    }
`

"Wrong" documentation: https://docs.spring.io/spring-ai/reference/api/chat/azure-openai-chat.html

grafik

timostark:
@tzolov OK, would need a little help to create a PR.. using subscribeOn / publishOn i was not able to find a functional solution. When using this approach (either in SpringAI or in my own code) the WebFlux implementation will be stuck after ~~100-300 tokens. Might be threading or backpressure.. I am not deep enough into the details of Flux/Threading of Schedulers to understand the root cause.

For me the WebFlux stream is working nicely if i isolate the handling to an own thread (probably the same what the OpenAiAsync lib by azure is doing) - See example: ( sorry, that will probably work much nicer with some integrated stuff of Flux - as said I am really not an expert here).

```
        Sinks.Many<ChatResponse> sink = Sinks.many().multicast().onBackpressureBuffer();
        ExecutorService executor = Executors.newSingleThreadExecutor();
        Runnable task = () -> {
            var response = azureOpenAiChatModel.stream(prompt);
            response.doOnEach(x -> sink.tryEmitNext(x.get())).doOnComplete(() -> sink.tryEmitComplete()).doOnError((error) -> {
                sink.tryEmitError(error);
            }).collectList().block();
        };
        return sink;
```

This is of course very ugly to have outside of the lib code.. So as said above: Do you have any preferred solution?

markpollack:
Just a quick comment, we should avoid an implementation that uses `block` to give the appearance of a sync call stack.  Using both client classes, one for sync, one for async/reactive would be best.   I am not sure if this addresses the issues you are observing as described above.  Is the issue that sync calls aren't working when you use the async client but async calls are?  If that is the case, using the sync client for sync calls would work well.

tzolov:
@timostark if the Async Client is working as expected for streaming response can't we use both clients, the AzureClient for call and Async Azure Client for stream?

timostark:
@tzolov yup certainly.. i will adjust the code accordingly in the next days and raise a PR.

hy4470:
The code improvement was completed a month ago, but when can it be reflected in the package?

hy4470:
> @tzolov yup certainly.. i will adjust the code accordingly in the next days and raise a PR.

It looks like the bug fix was done a month ago, when can we expect it to make it to the PR?

hy4470:
@tzolov  @timostark 
Is no one following up on this issue?

timostark:
@hy4470  @tzolov sorry i forgot about that issue. i am currently blocked with my work projects, but just fyi there is a functional workaround to get streaming working with the current status of SpringAI (Azure) by simply running the azureOpenAiChatmodel.stream inside an isolated runnable with a seperated EventSink.. That is reaaally no nice code, but imo good enough as a functional workaround.
Changing everything to the Async Client of Azure felt a little bit risky for me, because the documentation does not really tell much about the difference..

Extremly simplified functional code for the consumer service attached.

```
    private Flux<ChatResponse> sampleCode() {
        Sinks.Many<ChatResponse> eventSink = Sinks.many().multicast().onBackpressureBuffer();
        ExecutorService executor = Executors.newSingleThreadExecutor();

        try {
            Runnable task = () -> {
                var response = azureOpenAiChatModel.stream(new Prompt());
                var allResponses = response.onBackpressureBuffer().doOnError((error) -> {
                    eventSink.tryEmitError(error);
                }).doOnEach(x -> {
                    eventSink.tryEmitNext(x.get());
                }).collectList().block();

                eventSink.tryEmitComplete();
            };
            executor.execute(task);
            executor.shutdown();
        } catch (Exception ex) {
        }
        return eventSink.asFlux();
    }

```

You can return this value in your TEXT_EVENT_STREAM_VALUE GetMapping..

`    @GetMapping(value = "/chat/{chatId}/start", produces = MediaType.TEXT_EVENT_STREAM_VALUE)`

bruno-oliveira:
Hello all!
Do you foresee this will make it to Milestone 3, and, if yes, do you also think that the expected deadline of releasing on October 4th will be met?

Great work on the library all around, by the way! We are happily using it in production and it works like a charm! Asking about the deadline, because we have some use cases that would benefit from streaming being integrated in the Azure bean, just like it works for the Bedrock/Anthropic ones!

Once again, really great work so far!

markpollack:
This is working now, see this PR for the details.  https://github.com/spring-projects/spring-ai/pull/1447 

Thanks for everyone's patience!

hy4470:
> This is working now, see this PR for the details. #1447
> 
> Thanks for everyone's patience!

I conducted a test with the modified version and confirmed that many tokens are being lost. Did anyone else not experience this issue?

bruno-oliveira:
@hy4470 I’ll give this a try when milestone 3 is officially released.
For now, at my team, we’re just going with the low level client for Azure, the “direct one” from the Azure SDK that works well enough…

bruno-oliveira:
@tzolov @markpollack Is milestone 3 officially available? I’ll try this out if it is 🥳

sobychacko:
Yes, M3 is available now. https://spring.io/blog/2024/10/08/spring-ai-1-0-0-m3-released

bruno-oliveira:
Yes, very awesome!! I’ve just seen this now, was knee deep in implementing stuff and this fell under my radar, will try it out today! Thank you! 🙏

PR #1055, opened 2024-07-16
Add high-level function calling support for Azure OpenAI

### LINKED ISSUE #613: [Azure OpenAI] Function calling - Can't execute multiple function calls

tzolov:
Resolves #613, #1042

### DIFF
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatModel.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatModel.java
@@ -15,6 +15,33 @@
  */
 package org.springframework.ai.azure.openai;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.springframework.ai.azure.openai.metadata.AzureOpenAiChatResponseMetadata;
+import org.springframework.ai.chat.messages.AssistantMessage;
+import org.springframework.ai.chat.messages.Message;
+import org.springframework.ai.chat.messages.ToolResponseMessage;
+import org.springframework.ai.chat.metadata.ChatGenerationMetadata;
+import org.springframework.ai.chat.metadata.PromptMetadata;
+import org.springframework.ai.chat.metadata.PromptMetadata.PromptFilterMetadata;
+import org.springframework.ai.chat.model.ChatModel;
+import org.springframework.ai.chat.model.ChatResponse;
+import org.springframework.ai.chat.model.Generation;
+import org.springframework.ai.chat.prompt.ChatOptions;
+import org.springframework.ai.chat.prompt.Prompt;
+import org.springframework.ai.model.ModelOptionsUtils;
+import org.springframework.ai.model.function.AbstractToolCallSupport;
+import org.springframework.ai.model.function.FunctionCallbackContext;
+import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
+
 import com.azure.ai.openai.OpenAIClient;
 import com.azure.ai.openai.models.ChatChoice;
 import com.azure.ai.openai.models.ChatCompletions;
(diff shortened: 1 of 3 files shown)


timostark:
Thanks, that PR is actually very similar to mine.. Generally two remarks:

- The streaming does not really work with that PR too, as the internals of Azure AI rest calls seem to make blocking calls.. Yes tokens are streamed, but they are streamed when the chat model finishs answering. I think that we have to go to the Async OpenAI builder instead of the "normal" one. With the Async the streaming actually has sense and the tokens are retrieved directly after generation. A good way to test that, is that you make a simple Prompt: "Please give me an abstract about the US geography - minimum 1000 words". You will notice a long delay and than all tokens will be sent. If you use the async lib it will be token by token (as expected). 

- I don't really like that (from Azure official) "hack":
			// Note: the first chat completions can be ignored when using Azure OpenAI
			// servie which is a known service bug.
			.skip(1)
 As far as I've seen the only real effect of the first message is, that the getCHoices return value is null. As soon as Azure fixes their API the skip(1) will be dangerous and we will miss stuff. If you simply filter out stuff where getCHoices return null, it wil be safer..

- As long as the JSON parsing is corrputed in azure ( https://github.com/Azure/azure-sdk-for-java/issues/41164 ) i would really tend to stay with 1.0.8.. The logs are really polluted in streaming scenario..Or is there a big reason for 1.0.10?

I will raise a PR as soon as this one is merged to main (i only have time on friday probably, but will do it than), adressing (1) and (3) of my suggestions.

Thanks for your work!

tzolov:
Thank you for the details @timostark 

I've just merged the PR also replaying the skip with filter as you've suggested. 
Maybe will we will have to revert the version to beta8. But lets do this in a separate PR. 

Feel free to submit your PR. Very interested to see how the Async client will pay.

tzolov:
rebased, squashed and merged at feb036d2f6534505ad0bdac12925f93a52737e73

PR #1065, opened 2024-07-17
Ensure that ChatClient copies the input chatoptions

### LINKED ISSUE #1064: ChatClient mutates the default chat options instead of using a copy of it

**Bug description**
It looks like there is something setting function bean names” to defaultChatClientRequest.chatOptions after setting them in an individual chat client request.

Therefore, function calling is also activated with this function for each chatClient request, even if it's not configured.

**Environment**
Java 21, Spring AI 1.0.0-M1, Azure OpenAI Model

**Steps to reproduce**
chatClientBuilder.build().prompt()
        .messages(promptMessage)
        .call()
        .entity(Recipe.class);
-> No function call

chatClientBuilder.build().prompt()
        .messages(promptMessage)
        .functions("fetchIngredientsAvailableAtHome")
        .call()
        .entity(Recipe.class);

-> Function call

chatClientBuilder.build().prompt()
        .messages(promptMessage)
        .call()
        .entity(Recipe.class);
-> Function call

**Expected behavior**
The configuration for a function (or something else) for a request shouldn't have any effect on the configuration of a different request.

A sample with [a workaround](https://github.com/timosalm/spring-ai/blob/8bc4b0c199b2160be36bb56bc823d1df69fe5f91/src/main/java/com/example/RecipeService.java#L127) is available here: https://github.com/timosalm/spring-ai
[Here](http://github.com/timosalm/spring-ai) are the instructions in the README to enable Azure OpenAI.

tzolov:
- introduce copy() method to the ChatOptions.
 - make sure that the DefaultChatClientRequestSpec takes a copy of the input chat options to prevent multation.
 - add a OpenAiChatClientMultipleFunctionCallsIT to reproduce the problem and verify the solution.

 Resolves #1064

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatOptions.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatOptions.java
@@ -229,6 +229,11 @@ public class AnthropicChatOptions implements ChatOptions, FunctionCallingOptions
 		this.functions = functions;
 	}
 
+	@Override
+	public AnthropicChatOptions copy() {
+		return fromOptions(this);
+	}
+
 	public static AnthropicChatOptions fromOptions(AnthropicChatOptions fromOptions) {
 		return builder().withModel(fromOptions.getModel())
 			.withMaxTokens(fromOptions.getMaxTokens())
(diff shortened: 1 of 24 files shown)


tzolov:
Resolved by 018bb2c60056c323f0119d12bae2f90d942d59d1 and edf943ec97b30cfaf718066f16adb84a26755e84

PR #1079, opened 2024-07-19
Update enum to include OpenAI's GPT-4o-mini model

dongfeng3692:


tzolov:
Thank you @dongfeng3692

PR #1084, opened 2024-07-19
Make chat and embedding paths configurable

markpollack:


tzolov:
@markpollack please update the OpenAI properties documentation: https://docs.spring.io/spring-ai/reference/api/chat/openai-chat.html#_chat_properties

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -23,6 +23,7 @@
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import io.micrometer.common.util.StringUtils;
```
ThomasVitale, line 26: This should probably be imported from Spring and not Micrometer

`import org.springframework.util.StringUtils;`

ThomasVitale: Actually, it might be an unused import

markpollack: removed

markpollack:
added docs and merged in ee696f33ddc1ef0dd1f4b0953a2178403f589333

PR #1102, opened 2024-07-23
Consolidate retry config for OpenAI

ThomasVitale:
Extend the retry configuration strategy introduced in https://github.com/spring-projects/spring-ai/pull/1040 to the rest of the OpenAI model integration.

Main aspects:

- A `RetryTemplate` can be autoconfigured for each model type integration, or else a default one is used.
- The `RetryTemplate` is used to wrap the call to the model service for targeted HTTP retry logic instead of wrapping the entire implementation logic.
- The model request object is assembled in a dedicated method for improved readability and for sharing it between standard and streaming implementations.


tzolov:
Thanks @ThomasVitale

PR #1104, opened 2024-07-24
Remove flush operations to prevent excessive segment creation in Milvus

dongfeng3692:
tupian

[Typically, this method is called once all the data is ingested. You are advised not to call this method frequently since it could generate a lot of tiny segments and lead to unstable problems.
](https://milvus.io/api-reference/java/v2.4.x/v1/Collection/flush.md)

I found that the flush operation is time-consuming, and frequent flush operations result in the creation of many segments, which can impact performance. This also increases the pressure during the compaction process.


markpollack:
Thanks!

PR #1099, opened 2024-07-23
Support function calling for reflection Method with multiple arguments

tzolov:

PR #1105, opened 2024-07-24
Support for declarative build function callbacks in the Spring container

KAMO030:
* add FunctionCallbackMethodProcessor to support parsing. 
* add the FunctionCalling annotation to declare
* add FunctionCallbackMethodProcessorIT as test

Dependent on https://github.com/spring-projects/spring-ai/pull/1099

define:
```java
public  class FunctionCallConfig {
		@FunctionCalling(name = "dateTime", description = "get the current date and time")
		public String dateTime(String location) {
			return location + " dateTime:" + LocalDateTime.now();
		}
}
``` 

use:
```java
client.prompt().user("What is the time in New York?").functions("dateTime").call().content();
```


ThomasVitale:
Declarative tool definition via annotation has been introduced in https://github.com/spring-projects/spring-ai/pull/2064

ilayaperumalg:
@KAMO030 Please re-open this if you still see the issue and need help. Thanks!

PR #1113, opened 2024-07-25
fix: AbstractToolCallSupport wrong method name

mxsl-gr:
fix `AbstractToolCallSupport` wrong method name `executeFuncitons` -> `executeFunctions`


tzolov:
Thanks @mxsl-gr

PR #1140, opened 2024-08-01
Optimize documents embedding in the add method from the MilvusVectortore

solenyk:
Today I was using Spring AI and ran into a problem with org.springframework.ai.vectorstore.MilvusVectorStore. In method add(List<Document> documents), when going through each document separately, a request embeddingModel.embed(document); is sent for each of them. I have 4,000 documents in my case, so the method takes a very long time (50 minutes).

To speed up the process, I extended from MilvusVectorStore and already after executing the loop over the documents I used the List<List<Double>> embed(List<String> texts) method from org.springframework.ai.embedding.EmbeddingModel with contentArray, which significantly accelerated execution (now performance lasts 1 minute).

### DIFF
--- a/vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/MilvusVectorStore.java
+++ b/vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/MilvusVectorStore.java
@@ -45,7 +45,6 @@ import io.milvus.param.RpcStatus;
 import io.milvus.param.collection.CreateCollectionParam;
 import io.milvus.param.collection.DropCollectionParam;
 import io.milvus.param.collection.FieldType;
-import io.milvus.param.collection.FlushParam;
 import io.milvus.param.collection.HasCollectionParam;
 import io.milvus.param.collection.LoadCollectionParam;
 import io.milvus.param.collection.ReleaseCollectionParam;
@@ -59,7 +58,7 @@ import io.milvus.response.QueryResultsWrapper.RowRecord;
 import io.milvus.response.SearchResultsWrapper;
 
 /**
- * @author Christian Tzolov
+ * @author Christian Tzolov, Iryna Kopchak
  */
 public class MilvusVectorStore implements VectorStore, InitializingBean {
 
@@ -68,6 +67,8 @@ public class MilvusVectorStore implements VectorStore, InitializingBean {
 	public static final int OPENAI_EMBEDDING_DIMENSION_SIZE = 1536;
 
 	public static final int INVALID_EMBEDDING_DIMENSION = -1;
+
+	private static final int MAX_OPENAI_EMBEDDING_ARRAY_DIMENSIONS = 2048;
 
 	public static final String DEFAULT_DATABASE_NAME = "default";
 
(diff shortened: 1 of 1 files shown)


vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/MilvusVectorStore.java
```diff
@@ -69,6 +69,8 @@ public class MilvusVectorStore implements VectorStore, InitializingBean {
 
 	public static final int INVALID_EMBEDDING_DIMENSION = -1;
 
+	private static final int MAX_EMBEDDING_ARRAY_DIMENSIONS = 2048;
```
sobychacko, line 72: Is there a reason for this limit? Milvus documentation suggests much higher limits for dimensionality - https://milvus.io/docs/limitations.md#Dimensions-of-a-vector.

vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/MilvusVectorStore.java
```diff
@@ -69,6 +69,8 @@ public class MilvusVectorStore implements VectorStore, InitializingBean {
```
sobychacko, line 69: Please add your name as an author to the class.

sobychacko (changes requested):

vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/MilvusVectorStore.java
i-kopchak: The reason I added this dimensions limit is that when using OpenAiEmbeddingModel implementation of the EmbeddingModel is an OpenAi API limit:
[](https://platform.openai.com/docs/api-reference/embeddings/create)

I attached the link for the first version of the API, but there is the same situation for the second one!

solenyk: The reason I added this dimensions limit is that when using OpenAiEmbeddingModel implementation of the EmbeddingModel is an OpenAi API limit:
https://platform.openai.com/docs/api-reference/embeddings/create

I attached the link for the first version of the API, but there is the same situation for the second one!

vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/MilvusVectorStore.java
solenyk: Thanks!

vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/MilvusVectorStore.java
solenyk: I renamed MAX_EMBEDDING_ARRAY_DIMENSIONS to the MAX_OPENAI_EMBEDDING_ARRAY_DIMENSIONS for greater clarity

sobychacko:
@solenyk Based on the ideas in this PR, we decided to make some API changes that are overarching to how we embed documents and then use those embeddings in the vector stores. The core of those changes is a new `BatchingStrategy` API. Take a look at this PR: https://github.com/spring-projects/spring-ai/pull/1215

These changes were included in the recently released `M2`. Currently, the batching call is only applied to the Milvus vector store. We are going to evaluate other vector stores for similar usage. 

Thank you for the PR;  it certainly geared us toward providing it as a general pattern that other vector stores can use. I am closing the PR now. More PR contributions are welcomed!

PR #1146, opened 2024-08-01
Fix Azure chat model multimodality issue when Media data is byte array

ochnios:
**Checklist**
- [x] Contributor license agreement signed
- [x] Unit tests added
- [x] Build run and tests for the affected module passed

**Problem**
I need to send images to the Azure OpenAI chat model that are not publicly available. The existing implementation of the `fromSpringAiMessage` method in the `AzureOpenAiChatModel` wasn't entirely correct - it assumed that media data is always a URL. When a Media instance was created using a Resource object instead of a URL, `media.getData().toString()` caused a byte array reference string to be passed in the request:

```json
{
  "messages": [
    {
      "content": [
        {"text": "Explain what you see in this picture?", "type": "text"},
        {"image_url": {"url": "[B@d65982"}, "type": "image_url"}
      ],
      "role": "user"
    }
  ],
  "temperature": 0.699999988079071,
  "stream": false,
  "model": "gpt-4o"
}
```

This resulted in api error:

```json
{
  "error": {
    "code": "BadRequest",
    "message": "Invalid image URL. The URL must be a valid HTTP or HTTPS URL, or a data URL with base64 encoding.",
    "param": null,
    "type": null
  }
}
```

**Solution**
As a solution, I've added a method that takes a `Media` object and returns:
a) Just the image URL when data is a string
b) A base64 encoded string when data is a byte array

The mentioned method is called when `ChatMessageImageUrl` is created. I've also added a missing test case for the scenario when media data is a resource.

**Notes**
Possibly example with ClassPathResource can be added to docs.
This is my first contribution to the Spring projects, so please let me know if something is wrong or if any improvements are needed.


markpollack:
Thanks and congrats on your first contribution!  Keep them coming...  I've updated the docs with a sample for this case.  Merged in 197fe8105ce3184de88d68c95009ce60149dcaa2

PR #1149, opened 2024-08-02
Make MeterObservationHandler conditional on MeterRegistry bean

ThomasVitale:
Right now, it's conditional on the class. It must be conditional on the presence of a `MeterRegistry` bean instead.


tzolov:
Thanks @ThomasVitale

Issue #1148, opened 2024-08-02
Streamlined ModelOptions APIs for chat, embedding, and image


ThomasVitale:
While working on the initial observability support in Spring AI (delivered in https://github.com/spring-projects/spring-ai/pull/954), more common behaviour emerged across the different model providers with regards to the options used to customize model requests. It's worth surfacing those common options through the `ChatOptions` (e.g. model name, max tokens, stop sequences), `EmbeddingOptions` (e.g. dimensions, encoding format), and `ImageOptions` (e.g. image style) abstractions.

As agreed with @tzolov, I'll submit a PR for each of the three option variants and simply the observation contextual data accordingly.

- [x]  `ChatOptions`
- [x]  `EmbeddingOptions` => https://github.com/spring-projects/spring-ai/pull/1183
- [x]  `ImageOptions` => https://github.com/spring-projects/spring-ai/pull/1151

tzolov:
@ThomasVitale, I guess the https://github.com/spring-projects/spring-ai/pull/1151 is related to `ImageOptions` not to the ChatOptions as shown above?

ThomasVitale:
Right, I added the link to the wrong bullet point. I'll fix that and proceed with the 2 other PRs completing this task

PR #1151, opened 2024-08-02
Streamline ImageOptions

ThomasVitale:
* Add style to option abstraction
* Use abstraction in Observations directly instead of dedicated implementation
* Clean-up the merge of runtime and default image options in OpenAI and Stability AI

Fix gh-1148


tzolov:
LGTM, like the fact that you've fixed the builder convention (using the with prefix) and the ` mergeOption` helper cleans merge code. 
I believe this PR is related to #1148 but does not fix as there two more tasks to complete there.

Issue #1152, opened 2024-08-02
OpenAiUsage throws `NullPointerException` in EmbeddingModel


ThomasVitale:
**Bug description**
The methods implemented by `OpenAiUsage` throw `NullPointerException` whenever one of the usage type is not reported by OpenAI. For example, that's the case with the "generation tokens" for Embedding models.

**Environment**
Spring AI: 1.0.0-SNAPSHOT

**Steps to reproduce**
- Use the `OpenAiEmbeddingModel` to make a call to OpenAI
- On the `EmbeddingResponse`, get the `Usage` and call the `getGenerationTokens()` method.
- You'll get a `NullPointerException`.

**Expected behavior**
No `NullPointerException` when the usage tokens are not reported by the model provider.

ThomasVitale:
Submitted a PR. @tzolov @ashni-mongodb the OpenAI embedding model is currently broken without this fix

PR #1153, opened 2024-08-02
Fix NPE in OpenAiUsage

ThomasVitale:
Fix gh-1152


tzolov:
Good catch. Thanks @ThomasVitale

PR #1170, opened 2024-08-06
Docs: fix pinecone example

JayPark7821:
fix  `vectorStore.add(List.of(document));` to `vectorStore.add(documents);`


spring-ai-docs/src/main/antora/modules/ROOT/pages/api/vectordbs/pinecone.adoc
```diff
@@ -93,7 +93,7 @@ List <Document> documents = List.of(
     new Document("You walk forward facing the past and you turn back toward the future.", Map.of("meta2", "meta2")));
 
 // Add the documents
-vectorStore.add(List.of(document));
+vectorStore.add(documents);
```
sobychacko, line 96: `List.of(document)` is probably clearer. It is up to the user if they want to extract that to a variable. What do you think?

sobychacko (changes requested):

PR #1171, opened 2024-08-06
docs : fix chroma.adoc example

jo-kim:
Since `documents` is the list containing  `Document`objects 
`document` variable should be `documents` 

`vectorStore.add(List.of(document));`  -> `vectorStore.add(documents);`


spring-ai-docs/src/main/antora/modules/ROOT/pages/api/vectordbs/chroma.adoc
```diff
@@ -98,7 +98,7 @@ List <Document> documents = List.of(
     new Document("You walk forward facing the past and you turn back toward the future.", Map.of("meta2", "meta2")));
 
 // Add the documents
-vectorStore.add(List.of(document));
+vectorStore.add(documents);
```
sobychacko, line 101: `List.of(document)` is probably clearer. It is up to the user if they want to extract that to a variable. What do you think?

sobychacko (changes requested):

PR #1175, opened 2024-08-06
Add observability support for ChatClient

tzolov:
- Introduce ObservationRegistry to ChatClient creation and builder methods
- Implement observable chat responses in DefaultChatClient
- Add ChatClientObservationContext and related classes for metrics
- Update ChatClientAutoConfiguration to support ObservationRegistry
- Enhance RequestResponseAdvisor with getName() method

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
@@ -1,0 +1,64 @@
+/*
+* Copyright 2024 - 2024 the original author or authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* https://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package org.springframework.ai.chat.client.observation;
+
+import org.springframework.ai.chat.client.DefaultChatClient.DefaultChatClientRequestSpec;
+import org.springframework.ai.observation.AiOperationMetadata;
+import org.springframework.ai.observation.conventions.AiOperationType;
+import org.springframework.ai.observation.conventions.AiProvider;
+
+import io.micrometer.observation.Observation;
+
+/**
+ * @author Christian Tzolov
+ * @since 1.0.0
+ */
+
+public class ChatClientObservationContext extends Observation.Context {
+
+	private final boolean stream;
+
+	private final String format;
+
+	private final DefaultChatClientRequestSpec request;
+
+	private final AiOperationMetadata operationMetadata = new AiOperationMetadata(AiOperationType.FRAMEWORK.value(),
+			AiProvider.SPRING_AI.value());
+
+	public ChatClientObservationContext(DefaultChatClientRequestSpec requestSpec, String format, Boolean isStream) {
+
+		this.request = requestSpec;
+		this.format = format;
+		this.stream = isStream;
+	}
+
+	public DefaultChatClientRequestSpec getRequest() {
+		return this.request;
+	}
+
+	public AiOperationMetadata getOperationMetadata() {
+		return this.operationMetadata;
+	}
+
+	public boolean isStream() {
+		return this.stream;
+	}
+
+	public String getFormat() {
+		return this.format;
+	}
+
+}
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
@@ -1,0 +1,152 @@
+/*
+* Copyright 2024 - 2024 the original author or authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* https://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package org.springframework.ai.chat.client.observation;
+
+import io.micrometer.common.docs.KeyName;
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationConvention;
+import io.micrometer.observation.docs.ObservationDocumentation;
+
+/**
+ * @author Christian Tzolov
+ * @since 1.0.0
+ */
+public enum ChatClientObservationDocumentation implements ObservationDocumentation {
+
+	/**
+	 * AI Chat Client observations
+	 */
+	AI_CHAT_CLIENT {
+		@Override
+		public Class<? extends ObservationConvention<? extends Observation.Context>> getDefaultConvention() {
+			return DefaultChatClientObservationConvention.class;
+		}
+
+		@Override
+		public KeyName[] getLowCardinalityKeyNames() {
+			return LowCardinalityKeyNames.values();
+		}
+
+		@Override
+		public KeyName[] getHighCardinalityKeyNames() {
+			return HighCardinalityKeyNames.values();
+		}
+
+	};
+
+	public enum LowCardinalityKeyNames implements KeyName {
+
+		/**
+		 * Spring AI kind.
+		 */
+		SPRING_AI_KIND {
+			@Override
+			public String asString() {
+				return "spring.ai.kind";
+			}
+		},
+
+		/**
+		 * Is the chat model response a stream.
+		 */
+		STREAM {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.stream";
+			}
+		}
+
+	}
+
+	public enum HighCardinalityKeyNames implements KeyName {
+
+		/**
+		 * Enabled tool function names.
+		 */
+		CHAT_CLIENT_TOOL_FUNCTION_NAMES {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.tool.function.names";
+			}
+		},
+		/**
+		 * List of configured chat client function callbacks.
+		 */
+		CHAT_CLIENT_TOOL_FUNCTION_CALLBACKS {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.tool.functioncallbacks";
+			}
+		},
+
+		/**
+		 * List of configured chat client advisors.
+		 */
+		CHAT_CLIENT_ADVISORS {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.advisors";
+			}
+		},
+		/**
+		 * Map of advisor parameters.
+		 */
+		CHAT_CLIENT_ADVISOR_PARAMS {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.advisor.params";
+			}
+		},
+		/**
+		 * Chat client user text.
+		 */
+		CHAT_CLIENT_USER_TEXT {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.user.text";
+			}
+		},
+		/**
+		 * Chat client user parameters.
+		 */
+		CHAT_CLIENT_USER_PARAMS {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.user.params";
+			}
+		},
+		/**
+		 * Chat client system text.
+		 */
+		CHAT_CLIENT_SYSTEM_TEXT {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.system.text";
+			}
+		},
+		/**
+		 * Chat client system parameters.
+		 */
+		CHAT_CLIENT_SYSTEM_PARAM {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.system.params";
+			}
+		};
+
+	}
+
+}
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/DefaultChatClientObservationConvention.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/DefaultChatClientObservationConvention.java
@@ -1,0 +1,155 @@
+/*
+* Copyright 2024 - 2024 the original author or authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* https://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package org.springframework.ai.chat.client.observation;
+
+import java.util.stream.Collectors;
+
+import org.springframework.ai.chat.client.observation.ChatClientObservationDocumentation.LowCardinalityKeyNames;
+import org.springframework.ai.chat.observation.ChatModelObservationDocumentation;
+import org.springframework.lang.Nullable;
+import org.springframework.util.CollectionUtils;
+
+import io.micrometer.common.KeyValue;
+import io.micrometer.common.KeyValues;
+
+/**
+ * @author Christian Tzolov
+ * @since 1.0.0
+ */
+
+public class DefaultChatClientObservationConvention implements ChatClientObservationConvention {
+
+	public static final String DEFAULT_NAME = "spring.ai.chat.client.operation";
+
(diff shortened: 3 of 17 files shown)


spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.ai.chat.model.ChatResponse;
+import org.springframework.ai.model.observation.ModelObservationContext;
+import org.springframework.ai.observation.AiOperationMetadata;
+
+/**
+ * @author Christian Tzolov
+ * @since 1.0.0
+ */
+
+public class ChatClientObservationContext extends ModelObservationContext<DefaultChatClientRequestSpec, ChatResponse> {
```
ThomasVitale, line 28: I'm not sure about defining this context as a subtype of `ModelObservationContext`, which is more focused on the "request/response" single interactions with a model service. It could be its own type without inheritance. What do you think?

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		@Override
+		public KeyName[] getLowCardinalityKeyNames() {
+			return LowCardinalityKeyNames.values();
+		}
+
+		@Override
+		public KeyName[] getHighCardinalityKeyNames() {
+			return new KeyName[] {};
```
ThomasVitale, line 45: Should this reference the high cardinality values?

```suggestion
			return HighCardinalityKeyNames.values();
```

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
```diff
@@ -0,0 +1,158 @@
...
+		CHAT_CLIENT_USER_TEXT {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.user.text";
+			}
+		},
+		/**
+		 * Chat client user parameters.
+		 */
+		CHAT_CLIENT_USER_PARAM {
```
ThomasVitale, line 126: For naming consistency, this could be `CHAT_CLIENT_USER_PARAMS`

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
```diff
@@ -0,0 +1,158 @@
...
+		CHAT_CLIENT_SYSTEM_TEXT {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.system.text";
+			}
+		},
+		/**
+		 * Chat client system parameters.
+		 */
+		CHAT_CLIENT_SYSTEM_PARAM {
```
ThomasVitale, line 144: For naming consistency, this could be `CHAT_CLIENT_SYSTEM_PARAMS`.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
```diff
@@ -0,0 +1,158 @@
...
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.system.params";
+			}
+		};
+
+		@Override
```
ThomasVitale, line 151: Is this needed?

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
```diff
@@ -0,0 +1,158 @@
...
+		CHAT_CLIENT_ADVISOR {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.advisors";
+			}
+		},
+		/**
+		 * Map of advisor parameters.
+		 */
+		CHAT_CLIENT_ADVISOR_PARAM {
```
ThomasVitale, line 108:
```suggestion
		CHAT_CLIENT_ADVISOR_PARAMS {
```

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
```diff
@@ -0,0 +1,158 @@
...
+		TOOL_FUNCTION_NAMES {
+			@Override
+			public String asString() {
+				return "spring.ai.chat.client.tool.function.names";
+			}
+		},
+		/**
+		 * List of configured chat client advisors.
+		 */
+		CHAT_CLIENT_ADVISOR {
```
ThomasVitale, line 99:
```suggestion
		CHAT_CLIENT_ADVISORS {
```

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
```diff
@@ -0,0 +1,158 @@
...
+			public String asString() {
+				return "chat.client.stream";
+			}
+		},
+
+		/**
+		 * ChatModel response raw status code, or {@code "IO_ERROR"} in case of
+		 * {@code IOException}, or {@code "CLIENT_ERROR"} if no response was received.
+		 */
+		STATUS {
```
ThomasVitale, line 76: Is this used?

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/client/ChatClientBuilderProperties.java
```diff
@@ -36,12 +36,48 @@ public class ChatClientBuilderProperties {
...
 	public void setEnabled(boolean enabled) {
 		this.enabled = enabled;
 	}
 
+	public static class Observation {
+
+		/**
+		 * Whether to include the input content in the observations.
+		 */
+		private boolean includeInput = false;
```
ThomasVitale, line 58: Can we reuse the same property we have already for the low-level model observations, and control in one place whether we want the print out the prompt content?

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/client/ChatClientBuilderProperties.java
```diff
@@ -36,12 +36,48 @@ public class ChatClientBuilderProperties {
...
+
+		/**
+		 * Whether to include the output content in the observations.
+		 */
+		private boolean includeOutput = false;
```
ThomasVitale, line 63: I wonder if we need this. The output is already observed from the low-level model, including the completion content.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
```diff
@@ -0,0 +1,60 @@
...
+
+	private ChatResponse chatResponse;
```
ThomasVitale, line 30: Do we need the ChatResponse at this level? I would assume all data from the response are already observed from the last model call in the "chain" underlying the ChatClient, including the final completion content.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
ThomasVitale: We might just need the `DefaultChatClientRequestSpec` object here, and perhaps additional metadata like `format` and `isStream`

tzolov: Agree. I was tempted because it thought we can reuse some code related to the ChatResponse. But this is not relevant.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
tzolov: It should of course.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
tzolov: which one?

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
tzolov: nope, will remove it

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
tzolov: I think we would need some global response indicator for the pipeline and related status. But not sure how it would look like. So i will remove the ChatResponse until we figure it out.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
tzolov: yep

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
tzolov: +

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/client/ChatClientBuilderProperties.java
tzolov: Thought about it but my feeling is that we should not. the low level controls all chat models while the chat client is one. Furthermore it is nice to keep more fine grain control. 
We can change this in the future if we find it useful to control everything with a single property.

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/client/ChatClientBuilderProperties.java
tzolov: I will remove this along the ChatResponse context field.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
ThomasVitale: The `STATUS` enum value (`chat.client.status`)

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
```diff
@@ -0,0 +1,63 @@
...
+
+	private final boolean stream;
+
+	private final String format;
+
+	private final DefaultChatClientRequestSpec request;
+
+	private final AiOperationMetadata operationMetadata;
+
+	public ChatClientObservationContext(DefaultChatClientRequestSpec requestSpec, AiOperationMetadata operationMetadata,
```
ThomasVitale, line 38: Since this context is used specifically for ChatClient observations, we could simplify the constructor by removing AiOperationMetadata, building an instance inside the constructor with fixed values "spring_ai" and "framework". I simplified in a similar way the constructors for the model contexts (e.g.: https://github.com/spring-projects/spring-ai/blob/bf84d5945e9499d01a9deff9435b2e772804afab/spring-ai-core/src/main/java/org/springframework/ai/chat/observation/ChatModelObservationContext.java#L38)

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
```diff
@@ -0,0 +1,157 @@
...
+			}
+		},
+
+		/**
+		 * Is the chat model response a stream.
+		 */
+		STREAM {
+			@Override
+			public String asString() {
+				return "chat.client.stream";
```
ThomasVitale, line 68: We might want to use the prefix also here:

```suggestion
				return "spring.ai.chat.client.stream";
```

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/DefaultChatClientObservationConvention.java
```diff
@@ -0,0 +1,144 @@
...
+import io.micrometer.common.KeyValues;
+
+/**
+ * @author Christian Tzolov
+ * @since 1.0.0
+ */
+
+public class DefaultChatClientObservationConvention implements ChatClientObservationConvention {
+
+	public static final String DEFAULT_NAME = "gen_ai.chat.client.operation";
```
ThomasVitale, line 35: This might be something more specific to Spring AI

```suggestion
	public static final String DEFAULT_NAME = "spring.ai.chat.client.operation";
```

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/DefaultChatClientObservationConvention.java
```diff
@@ -0,0 +1,144 @@
...
+
+	@Override
+	public String getName() {
+		return this.name;
+	}
+
+	@Override
+	@Nullable
+	public String getContextualName(ChatClientObservationContext context) {
+		return "%s %s".formatted(context.getOperationMetadata().operationType(),
```
ThomasVitale, line 69: Since the framework operations are at a higher abstraction level than the OTel semantic conventions, I would suggest to go with our own naming strategy here. Otherwise, all Spring AI own operations would all have the same span name (`framework spring_ai`).

I would suggest a naming strategy such as `spring_ai chat_client` (the first value being the "provider", the second value being the "Spring AI Kind") perhaps even normalised without the underscore: `spring ai chat client`.

Other frameworks like OpenLit do something similar.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/DefaultChatClientObservationConvention.java
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	@Override
+	public KeyValues getHighCardinalityKeyValues(ChatClientObservationContext context) {
+		return KeyValues.of(toolFunctionNames(context), toolFunctionCallbacks(context), chatClientAvisor(context),
+				chatClientAvisorParam(context));
+	}
+
+	protected KeyValue springAiKind(ChatClientObservationContext context) {
+		return KeyValue.of(ChatClientObservationDocumentation.LowCardinalityKeyNames.SPRING_AI_KIND, "chat.client");
```
ThomasVitale, line 85: I would rename `chat.client` to `chat_client` because that's the typical naming convention for attribute values of this type (whereas the key itself uses the dots convention)

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/DefaultChatClientObservationConvention.java
```diff
@@ -0,0 +1,144 @@
...
+		if (CollectionUtils.isEmpty(context.getRequest().getFunctionCallbacks())) {
+			return CHAT_CLIENT_TOOL_FUNCTION_CALLBACKS_NONE;
+		}
+		return KeyValue.of(
+				ChatClientObservationDocumentation.HighCardinalityKeyNames.CHAT_CLIENT_TOOL_FUNCTION_CALLBACKS,
+				context.getRequest()
+					.getFunctionCallbacks()
+					.stream()
+					.map(fc -> fc.getName())
+					.collect(Collectors.joining(",")));
```
ThomasVitale, line 120: For all values containing lists, I would recommend using the same format we used in the lower-level observations and put them between quotes, inside square brackets. That will make it easier to process list values consistently across different observations (for example, when drawing dashboards in Grafana or when reading the observability data via external integrations)

https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/chat/observation/DefaultChatModelObservationConvention.java#L150

spring-ai-core/src/main/java/org/springframework/ai/observation/conventions/AiOperationType.java
```diff
@@ -32,7 +32,8 @@ public enum AiOperationType {
 	CHAT("chat"),
 	EMBEDDING("embedding"),
 	IMAGE("image"),
-	TEXT_COMPLETION("text_completion");
+	TEXT_COMPLETION("text_completion"),
+	FRAMEWORK("framework");
```
ThomasVitale, line 36: Consider keeping the alphabetical order

spring-ai-core/src/main/java/org/springframework/ai/observation/conventions/AiProvider.java
```diff
@@ -33,7 +33,8 @@ public enum AiProvider {
 	MISTRAL_AI("mistral_ai"),
 	OLLAMA("ollama"),
 	OPENAI("openai"),
-	VERTEX_AI("vertex_ai");
+	VERTEX_AI("vertex_ai"),
+	SPRING_AI("spring_ai");
```
ThomasVitale, line 37: Consider keeping the alphabetical order

ThomasVitale:
This is super nice!

tzolov:
Thanks for the super valuable review and improvements @ThomasVitale !
Pushed the changes that hopefully address most of the comments.

PR #1176, opened 2024-08-06
Update convention for chat finish reason in LLM observations

ThomasVitale:
Adhere to semantic convention for the "finish reason" attribute in LLM observations.


tzolov:
LGTM, thanks @ThomasVitale

PR #1185, opened 2024-08-07
feat: Add responseMimeType option in vertexAiGeminiChatOptions

jyami-kim:
The Gemini model provides the `responseMimeType` parameter, as documented in the [Gemini Model Reference](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini?_ga=2.232454687.1873628075.1702319245-1182285243.1701938872&hl=ko#parameters). 

However, when I attempting to call the Gemini model using Spring AI, there is no direct option to set this parameter. 
To work around this, I used reflection in my project:
```kotlin
        val chatModel = VertexAiGeminiChatModel(
            VertexAI(System.getenv("GEMINI_PROJECT_ID"), System.getenv("GEMINI_LOCATION"))
        )

        val generationConfig = GenerationConfig.newBuilder()
            .setResponseMimeType("application/json") // responseMimeType 을 설정할 수 없어서. 직접 접근하여 설정하였다.
            .setTemperature(0.8f)
            .build()

        val generationConfigField = VertexAiGeminiChatModel::class.java.getDeclaredField("generationConfig")
        generationConfigField.isAccessible = true
        generationConfigField.set(chatModel, generationConfig)
```

Therefore, it would be beneficial if the `responseMimeType` parameter was officially supported. The model generated by default in `VertexAiGeminiChatModel` is `GEMINI_1_5_PRO`, so this option is available from this version onwards, so it should be added as needed.

### DIFF
--- a/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatModel.java
+++ b/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatModel.java
@@ -359,6 +359,9 @@ public class VertexAiGeminiChatModel extends AbstractToolCallSupport implements 
 		if (options.getStopSequences() != null) {
 			generationConfigBuilder.addAllStopSequences(options.getStopSequences());
 		}
+		if (options.getResponseMimeType() != null) {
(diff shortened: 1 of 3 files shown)


markpollack:
Thanks!  Important feature indeed. I updated the docs.  Merged in bc55bc7196f9d02115c32fa1d831b85d7c35e86f

PR #1192, opened 2024-08-08
docs(onnx): add a note when the onnxruntime_gpu dependency is missing.

dongfeng3692:


tzolov:
Thanks you @dongfeng3692
Valuable insight. I was wondering if  we should add the onnxruntime_gpu dependency by default, but it appears that it is not supported on all platforms yet.

dongfeng3692:
@tzolov 
I do not recommend adding onnxruntime_gpu as a default dependency. Different versions of the dependency support different CUDA versions, and the onnxruntime_gpu package is large, for example, version 1.18.0 is 357 MB. Such a large dependency could cause inconvenience to users who do not use GPUs.
image

tzolov:
Thanks for the elaboration @dongfeng3692 
Agree that it will be too heavy to add this deps

PR #1203, opened 2024-08-10
Streamline ChatOptions

ThomasVitale:
* Surface more configuration APIs to ChatOptions
* Use abstraction in Observations directly instead of dedicated implementation
* Simplify metadata config in observations for defined models
* Improve merging of runtime and default options in OpenAI
* Fix missing option in Mistral AI

Relates to gh-1148


tzolov:
Great job @ThomasVitale , thanks. 
I scanned through the code and have not remarks or suggestion. 
I guess in the future we can improve some repetitive code in the call and stream observation instrumentation. But this is minor and not related to this PR.
Great stuff!

PR #1205, opened 2024-08-11
Add observability support to VectorStore

### LINKED ISSUE #1204: [Observability] Add observability support for VectorStores
To continue the https://github.com/spring-projects/spring-ai/issues/953 story we would like to provide observability support for the Spring AI VectorStore abstractions.
We would like to observe the `add`, `delete` and `similaritySearch` endpoints.

tzolov:
- Introduce AbstractObservationVectorStore for instrumentation. Add observe instrumentation for the add, delete and the similaritySearch methods.
 - Add VectorStoreObservationContext for capturing operation details.
 - Implement DefaultVectorStoreObservationConvention for naming and tagging
 - Create filters for add and delete request content observation Implement VectorStoreQueryResponseContentObservationFilter.
 - Update VectorStore interface with getName() method, returning the class name by default.
 - Add VectorStoreObservationDocumentation for defining observation keys.
 - Create VectorStoreObservationAutoConfiguration for auto-configuring observations.
 - Add VectorStoreObservationProperties for configuring observation behavior. They are used to control the optional, additional, observationcontent filters.
 - Update PgVectorStoreAutoConfiguration to support observations.
 - Modify PgVectorStore to extend AbstractObservationVectorStore
 - Add observation support to PgVectorStore's Builder.

 Resolves #1204


spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java
```diff
@@ -0,0 +1,204 @@
...
+				return "spring.ai.kind";
+			}
+		},
+		/**
+		 * The name of the operation or command being executed.
+		 */
+		DB_OPERATION_NAME {
+			@Override
+			public String asString() {
+				return "db.operation.name";
```
ThomasVitale, line 66: Consider adding this attributes coming from the OTel Conventions (db.*) to the `AiObservationAttributes` enum.

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java
```diff
@@ -0,0 +1,204 @@
...
+				return "db.vector.query.response";
+			}
+		},
+		/**
+		 * The database query being executed.
+		 */
+		QUERY {
+			@Override
+			public String asString() {
+				return "db.query.text";
```
ThomasVitale, line 117: I can see that this is used to add the text that is eventually translated into embeddings, and not the actual database query. Perhaps we should rename this to something else? How about `db.vector.query.content`? I expect the instrumentation for the lower-level database to use the `db.query.text` attribute for the db-specific query like `SELECT *, embedding <=> ? AS distance FROM public.vector_store WHERE embedding <=> ? < ?  ORDER BY distance LIMIT ?` (getting this from PGVector when using the "net.ttddyy.observation:datasource-micrometer-spring-boot" library)

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java
```diff
@@ -0,0 +1,204 @@
...
+
+	public enum HighCardinalityKeyNames implements KeyName {
+
+		/**
+		 * Add documents request content.
+		 */
+		DELETE_REQUEST {
+			@Override
+			public String asString() {
+				return "db.vector.delete.request";
```
ThomasVitale, line 90: Do we need to include the content of the documents we add/delete? I'm not sure about a possible use case for this data, it's typically not something included in database instrumentation. I can see it's also not included in LLM products like OpenLLMetry/TraceLoop and OpenLit.

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java
```diff
@@ -0,0 +1,204 @@
...
+				return "db.vector.add.request";
+			}
+		},
+		/**
+		 * Similarity search response content.
+		 */
+		QUERY_RESPONSE {
+			@Override
+			public String asString() {
+				return "db.vector.query.response";
```
ThomasVitale, line 108: If this is used for the documents returned in the response, I would suggest changing it to `db.vector.query.response.documents`, and establish a namespace (`db.vector.query.response.*`) we extend in the future to observe response metadata, response dimension, response vectors, and so on.

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java
```diff
@@ -0,0 +1,204 @@
...
+				return "db.query.text";
+			}
+		},
+		/**
+		 * The metadata filters used in the query.
+		 */
+		QUERY_METADATA_FILTER {
+			@Override
+			public String asString() {
+				return "db.query.metadata.filter";
```
ThomasVitale, line 126: I would keep using the vector namespace here and name the attribute `db.vector.query.metadata.filter` or even just `db.vector.query.filter`

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationContext.java
```diff
@@ -0,0 +1,266 @@
...
+		public final String value;
+
+		Operation(String value) {
+			this.value = value;
+		}
+
+	}
+
+	// DELETE
+	private List<String> deleteRequest;
```
ThomasVitale, line 57: This list of fields might grow a lot once Spring AI add support for more operations (like for updating embeddings for documents that changed). Consider if they can be grouped into some Records perhaps, or even maybe have a base `VectorStoreObservationContext` (that might be enough for add/delete operations if we remove the need to track the Documents) and specialised implementation whenever relevant (e.g. `QueryVectorStoreObservationContext` for semantic search)

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreAddRequestContentObservationFilter.java
```diff
@@ -0,0 +1,54 @@
...
+import io.micrometer.observation.ObservationFilter;
+
+/**
+ * An {@link ObservationFilter} to include the Vector Store add request content in the
+ * observation.
+ *
+ * @author Christian Tzolov
+ * @since 1.0.0
+ */
+public class VectorStoreAddRequestContentObservationFilter implements ObservationFilter {
```
ThomasVitale, line 32: I would consider not having this and remove the related attribute, unless it's needed for some specific use case

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreDeleteRequestContentObservationFilter.java
```diff
@@ -0,0 +1,55 @@
...
+import io.micrometer.observation.ObservationFilter;
+
+/**
+ * An {@link ObservationFilter} to include the Vector Store delete request content in the
+ * observation.
+ *
+ * @author Christian Tzolov
+ * @since 1.0.0
+ */
+public class VectorStoreDeleteRequestContentObservationFilter implements ObservationFilter {
```
ThomasVitale, line 32: Like for the "add" scenario, I would consider not having this and remove the related attribute, unless it's needed for some specific use case

vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/vectorstore/PgVectorStore.java
```diff
@@ -535,12 +558,35 @@ public Builder withInitializeSchema(boolean initializeSchema) {
...
 	}
 
+	@Override
+	public VectorStoreObservationContext.Builder createObservationContextBuilder(String operationName) {
+
+		return VectorStoreObservationContext.builder("pg_vector", operationName)
+			.withDimensions(this.embeddingDimensions())
+			.withCollectionName(this.vectorTableName)
+			.withNamespace(this.schemaName)
+			.withSimilarityMetric(this.getDistanceType().name())
```
ThomasVitale, line 587: The OTel Conventions have defined a list of well known values for the similarity metric. I would suggest defining a `VectorStoreSimilarityMetric` enum in the same package as `AiOperationType` and use those values if possible. That will allow LLM observability backend to recognise the known values. Here, we would need to map between the DB specific name and the convention name.

vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/vectorstore/PgVectorStore.java
```diff
@@ -535,12 +558,35 @@ public Builder withInitializeSchema(boolean initializeSchema) {
...
+			.withIndexName(this.createIndexMethod.name())
+			.withModel(this.embeddingModel.getClass().getSimpleName());
```
ThomasVitale, line 589: This attribute should be the model name used for computing embedding (e.g. `text-embedding-3-small`). It's optional, so if we cannot get the value for now, I would leave this field empty. Also, since we have already instrumented the `EmbeddingModel`, we already have the model name in the trace as part of the embedding span, so we might even argue whether we want/need this here. It might be needed for cases where it's the vector database itself providing the embedding model, but as far as I know it's not something supported by the Spring AI APIs yet.

vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/vectorstore/PgVectorStore.java
```diff
@@ -535,12 +558,35 @@ public Builder withInitializeSchema(boolean initializeSchema) {
...
-					initializeSchema);
+					initializeSchema, observationRegistry, searchObservationConvention);
 		}
 
 	}
 
+	@Override
+	public VectorStoreObservationContext.Builder createObservationContextBuilder(String operationName) {
+
+		return VectorStoreObservationContext.builder("pg_vector", operationName)
```
ThomasVitale, line 583: Consider introducing a `VectorStoreProvider` enum in the same package as `AiProvider`, and have the "pg_vector" value as part of that enum (but still leaving the field type as String, similar to how we handle the model provider in the LLM observations).

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/observation/VectorStoreObservationProperties.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Configuration properties for vector store observations.
+ *
+ * @author Christian Tzolov
+ * @since 1.0.0
+ */
+@ConfigurationProperties(VectorStoreObservationProperties.CONFIG_PREFIX)
+public class VectorStoreObservationProperties {
+
+	public static final String CONFIG_PREFIX = "spring.ai.vector.store.observations";
```
ThomasVitale, line 29: The config properties for vector stores use the prefix `spring.ai.vectorstore.*`. Should this be using the same namespace?
```suggestion
	public static final String CONFIG_PREFIX = "spring.ai.vectorstore.observations";
```

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java
markpollack: that sounds like a good idea.

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java
markpollack: `db.vector.query.content` sounds quite reasonable to me.

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/observation/VectorStoreObservationProperties.java
markpollack: Agree, matching the extra 'dot' is not needed, it implies a sort of grouping under 'vector' as if there could be something else other than 'store'

vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/vectorstore/PgVectorStore.java
markpollack: I will create an issue to consider this change once this PR is merged.

tzolov: Added entry for the Simple Vector Store as well.

vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/vectorstore/PgVectorStore.java
tzolov: I left this open, until the Options PR is merged. 
But even now it the `EmbeddingModel` does not expose `getDefaultOptions` method (similar to the ChatOptions) to show, at least, the default method name. 
So we should either add the get options to the EmbeddingModel or drop this attribute all together.

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java
tzolov: The add/delete content is optional (e.g. requires filters) and is disabled by default. 
My though was that if you want to debug/trace or provide providence for you app you might need to enable logging/trace this data as well.
But if you and Mark consider that this is too rare need I'm Ok to drop it?

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreAddRequestContentObservationFilter.java
tzolov: same as https://github.com/spring-projects/spring-ai/pull/1205#discussion_r1718056142

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreDeleteRequestContentObservationFilter.java
tzolov: same as https://github.com/spring-projects/spring-ai/pull/1205#discussion_r1718056142

vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/vectorstore/PgVectorStore.java
tzolov: I've decide to remove the Model attribute for now. We can consider adding it later if needed.

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationContext.java
tzolov: Actually intended this initially but then seemed like unnecessary complexity because of the shared context attributes. 
I will rethink it again.

ThomasVitale:
Looks great!

Issue #1214, opened 2024-08-13
Introduce batching support for embedding documents


sobychacko:
Introduce a way for `EmbeddingModel` implementations to embed `Document` objects using a batching strategy.

sobychacko:
Using this https://github.com/spring-projects/spring-ai/pull/1140 as a starting point, introduce a new API for batching for all vector stores and embedding models to use.

impactCn:
@sobychacko hi, I also encountered the same problem in PG. The segmentation process I use , https://github.com/spring-projects/spring-ai/pull/1200 .
Can you guide me on how to use this batch API?

sobychacko:
@impactCn Sorry for the delay in responding. We just merged similar batching changes for the PG vector store. Can you take a look and see if that satisfies your use case? If there is a gap, we can improve on that. Thanks!

PR #1215, opened 2024-08-13
GH-1214: Batching strategy for embedding documents

sobychacko:
- When embedding documents, allow batching the documents using some criteria.
 - `BatchingStrategy` interface with a `TokenCountBatchingStrategy` implementation that uses the openai max input token size of 8191 as the default.
 - Add a default method in EmbeddingModel to embed document using this new batching strategy.
 - Change `MilvusVectorStore` to make use of this new batching API.
 - Adding unit tests for `TokenCountBatchingStrategy`.
 - Adding openai integration test to call the embed API that uses batching.

Resolves https://github.com/spring-projects/spring-ai/issues/1214


markpollack:
merged in 949f1ed4e80345fe45acb0ba75de32f09989f111

PR #1237, opened 2024-08-17
Observability - Prompt content and completion as span events

ThomasVitale:
Currently, prompt content and completion can optionally be included as span attributes when model observability is enabled. That was a temporary solution because span attributes are not ideal for potentially long text.

This PR stores such content as span events instead whenever the underlying tracing technology supports them (i.e. OpenTelemetry), and it falls back on span attributes when the underlying tracing technology doesn't support span events (i.e. Zipkin/Brave).

The OpenTelemetry Semantic Conventions for GenAI also defines span events for the prompt content and completion data in the current release (1.27.0): https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#events 

The solution in this PR relies on Reflection. It can be simplified if and when this issue is solved: https://github.com/micrometer-metrics/tracing/issues/808.

In the future, when Micrometer supports span events, the fallback solution based on span attributes can be removed entirely (see: https://github.com/micrometer-metrics/micrometer/issues/5238).

### DIFF
--- a/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/observation/ChatObservationAutoConfiguration.java
+++ b/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/observation/ChatObservationAutoConfiguration.java
@@ -16,20 +16,17 @@
 package org.springframework.ai.autoconfigure.chat.observation;
 
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.tracing.otel.bridge.OtelTracer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.ai.chat.model.ChatModel;
-import org.springframework.ai.chat.observation.ChatModelCompletionObservationFilter;
-import org.springframework.ai.chat.observation.ChatModelMeterObservationHandler;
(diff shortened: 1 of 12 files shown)


spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/observation/ChatObservationAutoConfiguration.java
```diff
@@ -16,20 +16,17 @@
...
-import org.springframework.ai.chat.observation.ChatModelMeterObservationHandler;
-import org.springframework.ai.chat.observation.ChatModelPromptContentObservationFilter;
+import org.springframework.ai.chat.observation.*;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.condition.*;
```
markpollack, line 26: need to set your ide to not use .*

markpollack:
merged in 3fa102e78ff7cc1a19d8d10da61cf4936c871ee1

Issue #1243, opened 2024-08-19
Add observability support for the existing Vector Store implementations. 


tzolov:
The https://github.com/spring-projects/spring-ai/pull/1205 provided the infrastructure for VectorStore Observability support. 
Building on it we should add observability support for the existing Vector Store implementations. 

- [x] Azure
- [x] Cassandra
- [x] Chroma
- [x] ElasticSearch
- [x] GemFire 
- [x] HanaDB (NOTE: have not been able to create Hana account)
- [x] Milvus
- [x] MongoDB (NOTE: unable to run the MongoDB test containers on my Mac/M3)
- [x] Neo4j
- [x] OpenSearch
- [x] Oracle (NOTE: unable to run the Oracle test containers on my Mac/M3)
- [x] PGVector
- [x] Pinecone
- [x] Qdrant
- [x] Redis
- [x] Typesense
- [x] Weaviate

tzolov:
Resolved by 93fa2bf45ad37d0455b3f0d3a11b211f839f454f

PR #1244, opened 2024-08-19
Add observability support for the existing Vector Stores

tzolov:
Related to https://github.com/spring-projects/spring-ai/issues/1243


vector-stores/spring-ai-azure-store/src/main/java/org/springframework/ai/vectorstore/azure/AzureVectorStore.java
```diff
@@ -379,4 +406,13 @@ public void afterPropertiesSet() throws Exception {
 		this.searchClient = this.searchIndexClient.getSearchClient(this.indexName);
 	}
 
+	@Override
+	public Builder createObservationContextBuilder(String operationName) {
+
+		return VectorStoreObservationContext.builder(VectorStoreProvider.AZURE_VECTOR_STORE.value(), operationName)
+			.withDimensions(this.embeddingModel.dimensions())
+			.withSimilarityMetric(this.initializeSchema ? VectorStoreSimilarityMetric.COSINE.value() : "Unknown")
```
ThomasVitale, line 414: Should it be `null` instead of `"Unknown"`, so that the observation value will end up being `KeyValue.NONE`? For other attributes, when we can't infer the value, we fall back to "none".

vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/vectorstore/ChromaVectorStore.java
```diff
@@ -162,4 +187,12 @@ public void afterPropertiesSet() throws Exception {
 		this.collectionId = collection.id();
 	}
 
+	@Override
+	public Builder createObservationContextBuilder(String operationName) {
+		return VectorStoreObservationContext.builder(VectorStoreProvider.CHROMA_VECTOR_STORE.value(), operationName)
+			.withDimensions(this.embeddingModel.dimensions())
+			.withCollectionName(this.collectionName + ":" + this.collectionId)
+			.withFieldName(this.initializeSchema ? DISTANCE_FIELD_NAME : KeyValue.NONE_VALUE);
```
ThomasVitale, line 195: I would suggest keeping `KeyValue` concepts outside the vector store implementation and instead set it to `null`, so that we let the conventions handle "null" values (users might provide their implementation of the Conventions and decide to handle "null" values differently than `KeyValue.NONE_VALUE `).

spring-ai-core/src/main/java/org/springframework/ai/observation/conventions/VectorStoreProvider.java
```diff
@@ -23,6 +23,22 @@ public enum VectorStoreProvider {
 
 	// @formatter:off
         PG_VECTOR("pg_vector"),
+        AZURE_VECTOR_STORE("azure_vector_store"),
```
ThomasVitale, line 26: I would keep these names framework-agnostic for better interoperability with LLM observability backends. For example, I would use `chroma` and `elasticsearch` instead of `chroma_vector_store` and `elasticsearch_vector_store`

tzolov: Agree. It also improves the readability.

tzolov:
Thanks for the suggestions @ThomasVitale 
Hopefully last commit address them all.

ThomasVitale:
@tzolov super! It's so great to get observability instrumentation for all vector stores!

markpollack:
merged in 93fa2bf45ad37d0455b3f0d3a11b211f839f454f

PR #1255, opened 2024-08-20
Fixing MongoDB vector store support test failures

sobychacko:
Remove the requirement of MongoDB vector store auto-configuration only after `MongoDataAutoConfiguration`. This enables the `MongoDBAtlasVectorStoreAutoConfiguration` to properly provide custom Mongo conversions.

In `MongoDBAtlasVectorStoreIT` and `MongoDbVectorStoreObservationIT` tests, properly provision the `MongoTemplate` with custom conversions as these tests are not relying on auto-configuration.


tzolov:
This is great! Thanks @sobychacko !

rozza:
@tzolov & @markpollack I think you can probably close out #1022.

It took an alternative approach of adding a single `MimeTypeConverters` class which was used by both `MongoDBAtlasVectorStoreAutoConfiguration` and in `MongoDBAtlasVectorStoreIT`. Although I think the `TestApplication` in #1022 is much simplier / more readable.

PR #1262, opened 2024-08-21
Enhance vector store observability support

ThomasVitale:
- Consolidate usage of “db.collection.name” attribute to track table name, collection name, index name, document name, or whatever concept a vector database uses to store data. Removed “db.index” that was use sometimes instead of “db.collection.name”. This usage is in line with the OpenTelemetry Semantic Conventions.
- Configure query response content to be included as a “span event” instead of a “span attribute” if the backend system supports that, similar to how we do for the model observations.
- Structure vector store observation attributes in dedicated enums, including one for the Spring AI Kinds to avoid hard-coding the same value in a lot of places. This follows the OpenTelemetry Semantic Conventions as much as possible. Also, adopt Spring usual non-null-by-default strategy as much as possible.
- Align vector store conventions to the chat model ones, and follow alphabetical order for values. This is particularly useful for the convention classes, for which the Micrometer performance of exporting telemetry data improves when key values are added already sorted to the context.
- Fix flaky test in Mistral AI.
- Improve Qdrant integration tests.

### DIFF
--- a/models/spring-ai-mistral-ai/src/test/java/org/springframework/ai/mistralai/MistralAiChatModelObservationIT.java
+++ b/models/spring-ai-mistral-ai/src/test/java/org/springframework/ai/mistralai/MistralAiChatModelObservationIT.java
@@ -139,7 +139,8 @@ public class MistralAiChatModelObservationIT {
 			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.REQUEST_TEMPERATURE.asString(), "0.7")
 			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.REQUEST_TOP_K.asString(), KeyValue.NONE_VALUE)
 			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.REQUEST_TOP_P.asString(), "1.0")
-			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.RESPONSE_ID.asString(), responseMetadata.getId())
+			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.RESPONSE_ID.asString(),
+					StringUtils.hasText(responseMetadata.getId()) ? responseMetadata.getId() : KeyValue.NONE_VALUE)
 			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.RESPONSE_FINISH_REASONS.asString(), "[\"STOP\"]")
 			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.USAGE_INPUT_TOKENS.asString(),
 					String.valueOf(responseMetadata.getUsage().getPromptTokens()))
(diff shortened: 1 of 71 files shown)


models/spring-ai-mistral-ai/src/test/java/org/springframework/ai/mistralai/MistralAiChatModelObservationIT.java
```diff
@@ -139,7 +139,8 @@ private void validate(ChatResponseMetadata responseMetadata) {
 			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.REQUEST_TEMPERATURE.asString(), "0.7")
 			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.REQUEST_TOP_K.asString(), KeyValue.NONE_VALUE)
 			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.REQUEST_TOP_P.asString(), "1.0")
-			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.RESPONSE_ID.asString(), responseMetadata.getId())
+			.hasHighCardinalityKeyValue(HighCardinalityKeyNames.RESPONSE_ID.asString(),
```
ThomasVitale, line 142: This was flaky, Mistral AI doesn't always return a Response ID in streaming mode.

spring-ai-core/src/main/java/org/springframework/ai/chat/observation/ChatModelCompletionObservationFilter.java
```diff
@@ -36,7 +37,7 @@ public Observation.Context map(Observation.Context context) {
 
 		chatModelObservationContext
 			.addHighCardinalityKeyValue(ChatModelObservationDocumentation.HighCardinalityKeyNames.COMPLETION
-				.withValue(ChatModelObservationContentProcessor.concatenateStrings(completions)));
+				.withValue(TracingHelper.concatenateStrings(completions)));
```
ThomasVitale, line 40: Extracted the logic into the `TracingHelper` to re-use it for the vector store observations.

spring-ai-core/src/main/java/org/springframework/ai/observation/conventions/SpringAiKind.java
```diff
@@ -0,0 +1,44 @@
...
+ * Types of Spring AI constructs which can be observed.
+ *
+ * @author Thomas Vitale
+ * @since 1.0.0
+ */
+public enum SpringAiKind {
+
+	// @formatter:off
+
+	CHAT_CLIENT("chat_client"),
```
ThomasVitale, line 28: The first two values here are not used yet, I didn't want to make the PR even bigger. We can adopt them in a separate PR.

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/DefaultVectorStoreObservationConvention.java
```diff
@@ -24,41 +25,41 @@
 import io.micrometer.common.KeyValues;
 
 /**
+ * Default conventions to populate observations for vector store operations.
```
ThomasVitale, line 28: The main change here is a resorting to feed Micrometer already sorted list of key values for a performance improvement, and updating the observation name to be closer to the semantic conventions

tzolov:
I've added  .withCollectionName(this.pineconeIndexName) to the PineconeVectorStore

tzolov:
Rebased and merged at 036093a42b719d2506b93bec9dfea3cc977d730a

PR #1264, opened 2024-08-22
Customize `SearchRequest` sent to `VectorStore` in `QuestionAnswerAdvisor`

alexcheng1982:
In the current implementation of  `QuestionAnswerAdvisor`, the user text is always used as the query sent to `VectorStore`'s `similaritySearch` method. This may not be the case for some advanced RAG scenarios,  where the original user query may be expanded, transformed, or rewritten to get the actual query sent to the vector store.

The new `customizeSearchRequest` method provides an extension point to customize the `SearchRequest` used by `VectorStore`'s `similaritySearch` method. This allows subclasses of  `QuestionAnswerAdvisor` to apply custom logic to update the `SearchRequest`.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/QuestionAnswerAdvisor.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/QuestionAnswerAdvisor.java
@@ -103,6 +103,8 @@ public class QuestionAnswerAdvisor implements RequestResponseAdvisor {
 			.withQuery(request.userText())
 			.withFilterExpression(doGetFilterExpression(context));
 
+		searchRequestToUse = customizeSearchRequest(searchRequestToUse, request, context);
+
 		// 2. Search for similar documents in the vector store.
 		List<Document> documents = this.vectorStore.similaritySearch(searchRequestToUse);
 
@@ -117,12 +119,7 @@ public class QuestionAnswerAdvisor implements RequestResponseAdvisor {
 		Map<String, Object> advisedUserParams = new HashMap<>(request.userParams());
 		advisedUserParams.put("question_answer_context", documentContext);
 
-		AdvisedRequest advisedRequest = AdvisedRequest.from(request)
-			.withUserText(advisedUserText)
-			.withUserParams(advisedUserParams)
-			.build();
-
-		return advisedRequest;
+		return AdvisedRequest.from(request).withUserText(advisedUserText).withUserParams(advisedUserParams).build();
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/QuestionAnswerAdvisor.java
```diff
@@ -151,4 +148,18 @@ protected Filter.Expression doGetFilterExpression(Map<String, Object> context) {
...
+	 * Customize {@link SearchRequest} for each request. The returned
+	 * {@link SearchRequest} will be used in
+	 * {@link VectorStore#similaritySearch(SearchRequest)} to find similar documents.
+	 * @param searchRequest the original {@link SearchRequest}
+	 * @param request the {@link AdvisedRequest} representing the current request
+	 * @param context the shared data between advisors in the chain
+	 * @return the customized {@link SearchRequest}
+	 */
+	protected SearchRequest customizeSearchRequest(SearchRequest searchRequest, AdvisedRequest request,
+			Map<String, Object> context) {
```
youngmoneee, line 161: Why is it necessary to inherit and reimplement the implementation to return a SearchRequest method, even though QuestionAndAnswerAdvisor is already receiving a SearchRequest object through the constructor?

markpollack:
@alexcheng1982 This is indeed a very important use case to address.  Previous iterations of what became the Advisor approach better accommodated this, it in fact kept a history of the changes to the user message.   Instead of a subclass approach, what if we make use of the context in the advisor and have a well known key which is 'transformedUserText` or some name like that.  This way there can be a chain of advisors such that before the `QuestionAnswerAdvisor` advisor executes, a `QueryRewriteAdvisor` executes and puts the new updated query under the key `transformedUserText` in the context.  Then the `QuestionAnswerAdvisor` would look in that area of the context first and use the text there if present and if not, fall back to the current impl.

Thoughts?

youngmoneee:
I’m implementing a proxy object to intercept requests from the QuestionAndAnswerAdvisor to customize a Similarity Search.

```java
.prompt()
.advisors(spec -> spec
    .param(CHAT_MEMORY_CONVERSATION_ID_KEY, chatId)
    .param(SEARCH_REQUEST, searchRequest)
    .param(QUERY, queryString)
    // pass query or SearchRequest
)
.user(userMessage)
```
Wouldn't it be more natural to pass the `SearchRequest` object through `AdvisorSpec` rather than using inheritance?

```java
request.advisorParams().getOrDefault(
        SEARCH_REQUEST,
        this.searchRequest
    )
```

ThomasVitale:
Using the new (experimental) RAG features, advanced strategies for query transformation/expansion are supported. More info in the documentation: https://docs.spring.io/spring-ai/reference/api/retrieval-augmented-generation.html#_retrievalaugmentationadvisor_incubating and https://docs.spring.io/spring-ai/reference/api/retrieval-augmented-generation.html#_pre_retrieval

ilayaperumalg:
Thanks @ThomasVitale for the input. 
@alexcheng1982 Please re-open this if you still see the issue. Thanks!

PR #1276, opened 2024-08-26
fix streaming function call for MiniMax chat model

mxsl-gr:
this PR fixed streaming function call for MiniMax chat model

sometimes, the `MiniMax` stream mode function calls might get split, resulting in an empty tool call ID, which indicates that the previous call is not finished, this api design is quite strange.

For example, tool calls return like `[{"id":"1","function":{"name":"a"}},{"id":"","function":{"arguments":"[1]"}}]`, need to be merged into `[{"id":"1","name":"a","arguments":"[1]"}]`


markpollack:
thanks, what a tricky business!   merged in 8bfb25da86d241f7602575de97c107493214c078

PR #1278, opened 2024-08-27
Add structured response for Azure AudioTranscription

piotrooo:
## Motivation

For the Azure `AudioTranscription`, we need to obtain a structured response in the `verbose_json` response format. Currently, it is not being returned.

## Description

I have introduced a core, general-purpose `StructuredResponse` to handle a detailed response (`verbose_json`).

I've assumed that the data are associated with `ResponseMetadata`. I'm also considering using `ResultMetadata`, but I'm not sure which is a better fit? @tzolov and @markpollack do you have any thoughts on this?


piotrooo:
Gently pinging @tzolov and @markpollack, as we need this for one of our project :pray:

markpollack:
Thanks, we feel behind and catching up now.

PR #1296, opened 2024-08-30
Autowire WebClient.Builder into OllamaApi

habuma:
`OllamaApi` creates its own `WebClient.Builder` rather than take one via its constructor. (This is in contrast to its `RestClient.Builder` which it accepts via a constructor. )

As such, this prevents customization of the `WebClient.Builder` outside of `OllamaApi`. 

This PR changes `OllamaApi` to accept a `WebClient.Builder` via the constructor the same way that it already accepts a `RestClient.Builder` via the constructor.


markpollack:
tx, merged in 3cab5bd00928437c635d2ae31b90bab0486477d1

PR #1305, opened 2024-09-03
Fix problem with VectorStoreChatMemoryAdvisor using pgvector

fabapp2:
When `VectorStoreChatMemoryAdvisor`  is used the `SELECT` from the pgvector store failed with a SQL error.
This PR fixes the SQL by removing the `'` as reported.

I tried to provide a useful test that verifies the correct behaviour when `VectorStoreChatMemoryAdvisor` is used.
It verifies that chat messages are enriched with similar messages from the vector store. This query was failing.


tzolov:
Thanks you @fabapp2 . Especially for the additional test !
I have some concerns that some of the Vector stores out there might not be able to handle camel case name without quotes. But we will see. Maybe it would be safer to rename the `conversationId` to `conversationid` but only if we see issues later.

Issue #1316, opened 2024-09-05
Elasticsearch vector store - Wrong error reported when missing index


l-trotta:
Hello, I'm the author of parts of the implementation of the Elasticsearch Vector store (see #592).

I was testing the new release and I noticed that [this PR](https://github.com/spring-projects/spring-ai/pull/918) introduced a [check](https://github.com/sobychacko/spring-ai/blob/99b010a094f24338fbac45f9772f4bb4880aae03/vector-stores/spring-ai-elasticsearch-store/src/main/java/org/springframework/ai/vectorstore/ElasticsearchVectorStore.java#L101) that prevents documents to be added to the bulk only if the index already exists.

This is problematic because if the user set `initializeSchema` as false in the initial configuration and did not previously create the index (which is allowed by Elasticsearch, since it will automatically configure and create the index), the user will receive the following error: 

```
Caused by: co.elastic.clients.util.MissingRequiredPropertyException: Missing required property 'BulkRequest.operations'
	at co.elastic.clients.util.ApiTypeHelper.requireNonNull(ApiTypeHelper.java:76) ~[elasticsearch-java-8.13.4.jar:na]
	at co.elastic.clients.util.ApiTypeHelper.unmodifiableRequired(ApiTypeHelper.java:141) ~[elasticsearch-java-8.13.4.jar:na]
	at co.elastic.clients.elasticsearch.core.BulkRequest.<init>(BulkRequest.java:122) ~[elasticsearch-java-8.13.4.jar:na]
	at co.elastic.clients.elasticsearch.core.BulkRequest.<init>(BulkRequest.java:77) ~[elasticsearch-java-8.13.4.jar:na]
	at co.elastic.clients.elasticsearch.core.BulkRequest$Builder.build(BulkRequest.java:518) ~[elasticsearch-java-8.13.4.jar:na]
	at org.springframework.ai.vectorstore.ElasticsearchVectorStore.doAdd(ElasticsearchVectorStore.java:130) ~[spring-ai-elasticsearch-store-1.0.0-20240904.212955-466.jar:1.0.0-SNAPSHOT]
	at com.example.demo.Service.init(Service.java:88) ~[classes/:na]
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103) ~[na:na]
	at java.base/java.lang.reflect.Method.invoke(Method.java:580) ~[na:na]
[...]

PR #1319, opened 2024-09-06
Elasticsearch vector store - Wrong error reported fix

l-trotta:
Fixes https://github.com/spring-projects/spring-ai/issues/1316 as discussed in the issue, also updates the documentation to better explain the various index initialization possibilities. 

Added `user-agent` header to the elasticsearch client configuration. This is what we at Elastic will use to keep track of the usage of Spring AI, so that we can see statistics in the [Elastic Cloud](https://cloud.elastic.co/) metrics, better understand the needs of Spring users and contribute more to the project.

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/vectordbs/elasticsearch.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/vectordbs/elasticsearch.adoc
@@ -50,16 +50,14 @@ TIP: Refer to the xref:getting-started.adoc#repositories[Repositories] section t
 
 
 The vector store implementation can initialize the requisite schema for you, but you must opt-in by specifying the `initializeSchema` boolean in the appropriate constructor or by setting `...initialize-schema=true` in the `application.properties` file.
-Alternatively you can opt-out the initialization and either:
-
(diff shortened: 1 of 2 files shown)


sobychacko:
Looks good. Squashed the commits and merged upstream. Thanks!

l-trotta:
I forgot to add a unit test I wrote :') can I create another PR?

sobychacko:
@l-trotta, of course!

l-trotta:
[here](https://github.com/spring-projects/spring-ai/pull/1339) it is, thank you!

sobychacko:
@l-trotta When you added the `user-agent` header, is there a reason you didn't do it on the `RestClient` dIrectly? Also, we are considering adding this header across all the vector stores, and I was wondering how you tested this for Elasticsearch. Thanks!

l-trotta:
@sobychacko the RestClient would usually be the place where to put headers, but it this case its configuration is either:
- up to the user when creating the bean
- up to spring-boot when using autoconfigure

So the only way to add a header is to use the override that we have in the high level client.
I tested it by checking our internal monitoring tool where requests are saved and catalogued, so not with a simple Elasticsearch instance (and noticed that I actually set it wrong, [here](https://github.com/spring-projects/spring-ai/pull/1346)'s the fix ^^").

About other vector stores, I'd probably check whether they use user-agent as a relevant header and if possible prepend `spring-ai` to the original value, or rebuilt it like I did in the new PR.

PR #1332, opened 2024-09-09
[WIP] Refactor AdvisedRequest and introduce AdvisedResponse

tzolov:
- Introduce AdvisedResponse to encapsulate ChatResponse and advice context
- Refactor RequestAdvisor and ResponseAdvisor interface to use AdvisedRequest/AdvisedResponse records
- Update DefaultAroundAdvisorChain to work with new advisor interfaces
- Modify DefaultChatClient to use new advisor framework
- Adjust MessageChatMemoryAdvisor for compatibility with new advisor structure
- Update integration tests to reflect changes in advisor framework

- Update AdvisedRequest to include adviseContext and add toPrompt() method
- Modify advisor interfaces to remove explicit context parameter
- Introduce AdvisedResponse to encapsulate ChatResponse with adviseContext
- Update DefaultChatClient to use new advisor interfaces and AdvisedResponse
- Adjust integration tests in OpenAiPaymentTransactionIT and VertexAiGeminiPaymentTransactionIT
- Refactor the RequestResponseAdvisor interface to make it a backward compatible adabror for ReqeustAdvisor and ResponseAdvisors. Aslo mark it as deprecated.
- Update various advisor implementations to work with new AdvisedRequest and AdvisedResponse
- Modify AroundAdvisorChain to use AdvisedResponse in both call and stream methods
- Implement new adviseResponse method for streaming responses in RequestResponseAdvisor
- Update MessageChatMemoryAdvisor, PromptChatMemoryAdvisor, QuestionAnswerAdvisor, SimpleLoggerAdvisor, and VectorStoreChatMemoryAdvisor to use new advisor interfaces
- Add new AdvisedResponse class to encapsulate ChatResponse with advice context
- Update CallAroundAdvisor, StreamAroundAdvisor, RequestAdvisor, and ResponseAdvisor interfaces to use AdvisedResponse and remove adviceContext parameter
- Modify CacheAroundAdvisor and SafeGuardAroundAdvisor to work with new advisor interfaces and AdvisedResponse
- Update AdvisorObservableHelper to use AdvisedResponse and remove explicit context parameter
- Refactor MessageAggregator to support aggregation of AdvisedResponse objects
- Rename extractRequestAdvisors and extractResponseAdvisors methods to requestAdvisors and responseAdvisors in AdvisorObservableHelper
- Add null checks and return empty lists for advisors in AdvisorObservableHelper


spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/observation/AdvisorObservableHelper.java
```diff
@@ -37,57 +36,62 @@ public abstract class AdvisorObservableHelper {
...
 
-	public static List<RequestAdvisor> extractRequestAdvisors(List<Advisor> advisors) {
+	public static List<BeforeAdvisor> requestAdvisors(List<Advisor> advisors) {
+		if (CollectionUtils.isEmpty(advisors)) {
+			return Collections.emptyList();
+		}
 		return advisors.stream()
-			.filter(advisor -> advisor instanceof RequestAdvisor)
-			.map(a -> (RequestAdvisor) a)
+			.filter(advisor -> advisor instanceof BeforeAdvisor)
```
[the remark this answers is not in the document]
chemicL: Perhaps in `ChatClient` the signature `ChatClientRequestSpec advisors(Consumer<AdvisorSpec> consumer)` could be the one offered and the `AdvisorSpec` would offer type-safe, yet specific, advisor registrations for particular types. But without changing the API the `DefaultChatClient` could internally store particular implementations into well-defined slots upon specification by the user.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/api/AfterAdvisor.java
```diff
@@ -77,21 +72,18 @@ public enum StreamResponseMode {
...
 	 * Represents the row {@link ChatResponse} stream data after the
 	 * {@link ChatModel#stream(Prompt)} method is called.
-	 * @param adviseContext the shared data between the advisors in the chain. It is
-	 * shared between all request and response advising points of all advisors in the
-	 * chain.
 	 * @return the advised {@link ChatResponse} flux.
 	 */
-	default Flux<ChatResponse> adviseResponse(Flux<ChatResponse> fluxResponse, Map<String, Object> adviseContext) {
-		return fluxResponse;
+	default Flux<AdvisedResponse> afterStream(Flux<AdvisedResponse> advisedResponseStream) {
```
chemicL, line 85: This API feels that it should belong elsewhere, to a specific custom advisor type. Currently, from my understanding, this is only called for the custom case. For most regular advisors this would be confusing that this can be overridden.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/api/AfterAdvisor.java
```diff
@@ -33,23 +31,20 @@
...
 	 * @param response the {@link ChatResponse} data to be advised. Represents the row
 	 * {@link ChatResponse} data after the {@link ChatModel#call(Prompt)} method is
 	 * called.
-	 * @param adviseContext the shared data between the advisors in the chain. It is
-	 * shared between all request and response advising points of all advisors in the
-	 * chain.
 	 * @return the advised {@link ChatResponse}.
 	 */
-	ChatResponse adviseResponse(ChatResponse response, Map<String, Object> adviseContext);
+	AdvisedResponse afterCall(AdvisedResponse advisedResponse);
```
chemicL, line 42: This name might be better aligned with `BeforeAdvisor`'s `before()` signature if the `Call` part is skipped, or the other one also has the `Call` suffix.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/api/StreamAroundAdvisor.java
```diff
@@ -31,11 +26,9 @@ public interface StreamAroundAdvisor extends Advisor {
 	/**
 	 * Around advice that wraps the invocation of the advised request.
 	 * @param advisedRequest
-	 * @param adviceContext
 	 * @param chain the chain of advisors to execute
 	 * @return the result of the advised request
 	 */
-	Flux<ChatResponse> aroundStream(AdvisedRequest advisedRequest, Map<String, Object> adviceContext,
-			AroundAdvisorChain chain);
+	Flux<AdvisedResponse> aroundStream(AdvisedRequest advisedRequest, AroundAdvisorChain chain);
```
chemicL, line 32: What's stopping the "Custom" implementations to be considered as the "around" advisors, making the "custom" abstraction unnecessary?

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/api/AfterAdvisor.java
```diff
@@ -77,21 +72,18 @@ public enum StreamResponseMode {
 
 	}
 
-	default StreamResponseMode getStreamResponseMode() {
-		return StreamResponseMode.ON_FINISH_ELEMENT;
+	default AfterStreamMode getAfterStreamMode() {
+		return AfterStreamMode.ON_FINISH_ELEMENT;
```
chemicL, line 76: Would it not be more natural if a default behaviour was to apply the advisor per each chunk in the stream using the imperative implementation? Or perhaps applying it over the aggregation? If there's no easy answer, perhaps picking a default should not happen at all and stream advisors should not belong to the imperative type at all?

chemicL (commented):
Left a few comments. I really like the direction of having the context embedded in the `Advised(Request|Response)` and the reactive manipulations look well optimized. There are a few more things that can be improved I think.

tzolov:
Replaced by #1422

PR #1346, opened 2024-09-11
Fixing Elasticsearch's user agent header

l-trotta:
Adding version and actually adding the header to the client


tzolov:
Thanks @l-trotta

PR #1361, opened 2024-09-14
OpenAI: Add o1/o1-mini to chat model enum

PARK-afk:

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -184,6 +184,26 @@ public class OpenAiApi {
 	public enum ChatModel implements ChatModelDescription {
 
 		/**
+		 * Points to the most recent snapshot of the o1 model:o1-preview-2024-09-12
+		 */
+		O_1_PREVIEW("o1-preview"),
+
+		/**
+		 * Latest o1 model snapshot
+		 */
+		O_1_PREVIEW_2024_09_12("o1-preview-2024-09-12"),
+
+		/**
+		 * Points to the most recent o1-mini snapshot:o1-mini-2024-09-12
+		 */
+		O_1_MINI("o1-mini"),
+
+		/**
+		 * Latest o1-mini model snapshot
+		 */
+		O_1_MINI_2024_09_12("o1-mini-2024-09-12"),
+
+		/**
 		 * Multimodal flagship model that’s cheaper and faster than GPT-4 Turbo. Currently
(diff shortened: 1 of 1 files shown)


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -183,6 +183,26 @@ public OpenAiApi(String baseUrl, String apiKey, MultiValueMap<String, String> he
 	 */
 	public enum ChatModel implements ChatModelDescription {
 
+		/**
```
ThomasVitale, line 186: These models are a bit different compared to the rest. For example, they don't support function calling nor system messages. I would suggest considering how to convey this information to warn users that by switching model value they might get a broken application. It might be worth considering whether adding them to this enum is necessary. Even if they're not added to the enum, it's always possible to use these new models, either via configuration properties or via the ChatModel/ChatClient APIs.

markpollack: good point.  we need to think about it.  in anycase, i merged the enums in the interest of having folks explore.

markpollack:
merged in 40eaa9974e2ca84763e7eb9032392c241a6145a2

will create an issue to discuss how 'common options' are not getting i'll defined.

dafriz:
Note that the commit that got merged has the wrong values - https://github.com/spring-projects/spring-ai/commit/40eaa9974e2ca84763e7eb9032392c241a6145a2  . It was his follow up commit that had the names that match OpenAI docs - https://github.com/spring-projects/spring-ai/pull/1361/commits/0bd27e4dc9a2259802fd40c4c5b68a812f999dd7 

See https://platform.openai.com/docs/models/o1 - they start with o1, not gpt.

markpollack:
I've updated the enums to be 

```

		/**
		 * Points to the most recent snapshot of the o1 model:o1-preview-2024-09-12
		 */
		O1_PREVIEW("o1-preview"),
		/**
		 * Latest o1 model snapshot
		 */
		O1_PREVIEW_2024_09_12("o1-preview-2024-09-12"),
		/**
		 * Points to the most recent o1-mini snapshot:o1-mini-2024-09-12
		 */
		O1_MINI("o1-mini"),
		/**
		 * Latest o1-mini model snapshot
		 */
		O1_MINI_2024_09_12("o1-mini-2024-09-12")

markpollack:
closing this PR and created #1378  for deeper analysis of rationalizing chatoptions, in particular due to how different the `o1` model is regarding options and also functionality (e.g. no system messages)

PR #1372, opened 2024-09-19
Overload EvaluationRequest constructor

habuma:
I find that more often than not, I create an `EvaluationRequest` by passing in `List.of()` or `Collections.emptyList()` as the 2nd constructor argument (data list). 

As a convenience, I thought it would be nice if there were a constructor that defaulted to an empty list.


markpollack:
tx, merged in 62feb38b6cac2a113ba772a390b5e71353df62e7

PR #1380, opened 2024-09-19
Add/Verify Azure user-agent header

sobychacko:
* Adding test to verify the user-agent header in the Azure OpenAi chat client
* Adding the user-agent header to the search client in Azure vector store


markpollack:
merged in e644bf77da64975e6ea3bd9a5cbb011d0f3881e5

PR #1390, opened 2024-09-21
Minor update to  RelevancyEvaluator, simplify checks and filtering

alexcheng1982:
Return type of `Content.getContent` is already `String`, no need for type checking and `toString`. Use `StringUtils.hasText` to filter out null or empty strings.


markpollack:
thanks!  merged in 2ecffc10c73404e7d1512c12d7fdaa30443280b5

PR #1392, opened 2024-09-22
Handle HttpClientErrorException on getCollection method

ricken07:
I'm using chroma as a vector database for a demo with spring AI, and I've found that retrieving the collection name not only throws an **HttpServerErrorException**, but also an **HttpClientErrorException**.

I've pushed a fix to additionally capture HttpClientErrorException and added a Pattern to retrieve the error message.

Tests are green


markpollack:
thanks!

PR #1399, opened 2024-09-23
Fix host in ChromaDockerComposeConnectionDetailsFactory

### LINKED ISSUE #1395: Invalid URL scheme when using Chroma and Docker Compose
It's unclear if this is a bug or not (or what is causing it), but it sure feels like a bug.

If I create a brand new Spring Boot app with Docker Compose Support and Chroma, I get a build file that includes the following dependencies:

```
    developmentOnly 'org.springframework.boot:spring-boot-docker-compose'
    developmentOnly "org.springframework.ai:spring-ai-spring-boot-docker-compose:${springAiVersion}"
```

And it gives me a nice compose.yml file with all that I need for Docker to start Chroma.

However, once it actually tries to hit the Chroma vector store, I get an exception that is ultimately caused by this:

```
Caused by: java.net.URISyntaxException: Illegal character in scheme name at index 0: 127.0.0.1:/api/v1/collections/SpringAiCollection
```

I did a little debugging and found that in `ChromaVectorStoreAutoConfiguration` when the `ChromaApi` bean is being created (at https://github.com/spring-projects/spring-ai/blob/main/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/chroma/ChromaVectorStoreAutoConfiguration.java#L66), the `chromaUrl` is set to "127.0.0.1:8000". But when that is given to the `RestClient.Builder` in `ChromaApi`, the "127.0.0.1" is interpreted to be the schema and the "8000" is assigned to the `ssp`. The `host` and `port` are both null. 

HOWEVER, if I remove the following dependency from my build, everything works fine and as expected:

```
developmentOnly "org.springframework.ai:spring-ai-spring-boot-docker-compose:${springAiVersion}"
```

Clearly something in that dependency is ganking things up. And given that it works fine without it, I wonder why it's put there by the Initializr?

eddumelendez:
Host should start with `http://`

Fixes #1395

### DIFF
--- a/spring-ai-spring-boot-docker-compose/src/main/java/org/springframework/ai/docker/compose/service/connection/chroma/ChromaDockerComposeConnectionDetailsFactory.java
+++ b/spring-ai-spring-boot-docker-compose/src/main/java/org/springframework/ai/docker/compose/service/connection/chroma/ChromaDockerComposeConnectionDetailsFactory.java
@@ -60,7 +60,7 @@ class ChromaDockerComposeConnectionDetailsFactory
 
 		@Override
 		public String getHost() {
-			return this.host;
+			return "http://%s".formatted(this.host);
 		}
 
 		@Override


markpollack:
Thanks Eddu!  Hope you are well.  merged in ee00c620c62fb5f86d6ba418b35869c260fced25

Issue #1403, opened 2024-09-24
Add suport to anthropic prompt caching


Claudio-code:
Please do a quick search on GitHub issues first, the feature you are about to request might have already been requested.

**Expected Behavior**

Integration uses cache for many contexts, one of which is cost reduction as shown in the [documentation](https://www.anthropic.com/news/prompt-caching)

**Current Behavior**

**Context**

The cache is in beta but can already be added to the integration

aleksabl:
Should probalby look at Geminis context caching as well: https://ai.google.dev/gemini-api/docs/caching

sobychacko:
New PR: https://github.com/spring-projects/spring-ai/pull/4199

PR #1413, opened 2024-09-26
Added suport to anthropic prompt cache

Claudio-code:

issue https://github.com/spring-projects/spring-ai/issues/1403

I added support to user prompt use cache, added test only to api because Chatmodel/AnthropicChatOptions not can set header beta to works cache


markpollack:
thanks, can you explain the comment "Chatmodel/AnthropicChatOptions not can set header beta to works cache" please?

Claudio-code:
Anthropic models added caching as a beta feature that requires a custom header to be used

I couldn't find a way to send a custom header using the options in the Chatmodel/AnthropicChatOptions classes.

That's why I only added integration tests to the AnthropicApiIT.java class

It would be nice if in the future there was an option to send a custom header to test beta features like langchain4j https://github.com/langchain4j/langchain4j/issues/1591#issuecomment-2357792634

markpollack:
Hi.  Thanks for the clarification.  I think it is a bit too much to get into M3 only because it modifies the AbstractMessage class and I'd like to think through the implications.  I guess much like function calling, this is going to "table stakes" for leading model providers.  I'll schedule it for M4 with the hopes we have a 5-ish week release cadence on our march to GA.

thanks for the nice PR.

Claudio-code:
@markpollack  I can change the implementation and add the cache in the AnthropicChatModel build so I don't change the common APIs. What do you think?

I had added the cache in each message to follow the approach proposed by Anthropic, but this change in the common APIs won't work for other models, for example, Gemini, which uses the cache by time (e.g. 2m) and the implementation I made uses it by label: (e.g. type=ephemeral)

RRajdev:
@Claudio-code @markpollack @tzolov  any chance this be merged any sooner than March?

RRajdev:
Hi @markpollack @tzolov @Claudio-code 

After having looked at a number of frameworks, I settled on Spring AI and I think its awesome. For my use case, I have identified anthropic as the best LLM to use. The use case is a tricky one as I have a large system message and potentially an image that I have to send to the LLM for each turn in the conversation (this is unavoidable). I am running into rate limits fairly early in the process and would benefit from prompt caching. 

Is it possible to merge this change in  before March? As this area is evolving rapidly, is there no way to get new features out sooner? I realise that you would want to maintain the quality of the codebase but in this case it is essential to the viability of my use case.

Claudio-code:
I saw that the cache left beta, now I can add more complete tests to validate the creation of token caches and the reading of these caches.

markpollack:
Thanks, just going through the issues, will aim to merge for m7.  thanks for you patience.

tzolov:
@Claudio-code , thanks for your patience.
I've rebased your PR with recent changes, cherry-picking some parts and creating a new updated PR: https://github.com/spring-projects/spring-ai/pull/2669
You remain the author. Could you please review to ensure nothing was broken during this process?

tzolov:
Replacing this PR by https://github.com/spring-projects/spring-ai/pull/2669

Issue #1417, opened 2024-09-26
[AzureOpenAiChatOptions] Newly added options are null checked in .with* builder methods and are made indirectly mandatory


pradu3:
**Bug description**
The ``Assert.notNull`` checks in ``AzureOpenAiChatOptions.Builder.with*`` methods make the options actually mandatory. ``AzureOpenAiChatOptions.fromOptions`` is called from ``AzureOpenAiChatModel.getDefaultOptions``. When the ``AzureOpenAiChatModel.defaultOption`` attribute doesn't have all the checked fields set (e.g. seed, enhancements), there will be an exception when ``AzureOpenAiChatModel.getDefaultOptions``. This will cause, for example, ``ChatClient.create`` to fail. This happens even when the ``AzureOpenAiChatModel( OpenAIClient microsoftOpenAiClient)`` constructor is used.

New options introduced by the fix of https://github.com/spring-projects/spring-ai/issues/889 are the one causing this.

**Environment**
1.0.0-SNAPSHOT

**Steps to reproduce**
call ``ChatClient.create(new AzureOpenAiChatModel(microsoftOpenAiClient))`` and it will throw an exception:
```
Caused by: java.lang.IllegalArgumentException: responseFormat must not be null
	at org.springframework.util.Assert.notNull(Assert.java:172) ~[spring-core-6.1.13.jar:6.1.13]
	at org.springframework.ai.azure.openai.AzureOpenAiChatOptions$Builder.withSeed(AzureOpenAiChatOptions.java:297) ~[spring-ai-azure-openai-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.springframework.ai.azure.openai.AzureOpenAiChatOptions.fromOptions(AzureOpenAiChatOptions.java:525) ~[spring-ai-azure-openai-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.springframework.ai.azure.openai.AzureOpenAiChatModel.getDefaultOptions(AzureOpenAiChatModel.java:143) ~[spring-ai-azure-openai-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.springframework.ai.azure.openai.AzureOpenAiChatModel.getDefaultOptions(AzureOpenAiChatModel.java:100) ~[spring-ai-azure-openai-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
[...]

PR #1418, opened 2024-09-26
remove unnecessary Assert.notNull which caused ChatClient.create to fail

pradu3:
Fixes for this reported issue: https://github.com/spring-projects/spring-ai/issues/1417


pradu3:
obsolete. fixed with https://github.com/spring-projects/spring-ai/pull/1425.

Issue #1282, opened 2024-08-28
AWS Bedrock - Add support for cross-region inference new feature


DEG-7:
Hey team,
First of all, thanks for the effort you are doing for this amazing project.
I would like to ask for the support of a recently and very important addition in AWS Bedrock, that's the cross-region inference support.
You can find all the tech information in AWS doc: 
https://aws.amazon.com/blogs/machine-learning/getting-started-with-cross-region-inference-in-amazon-bedrock/

Thanks in advance

dafriz:
You should be able to use withought any code changes by adding a "us." or "eu." prefix to the modelId.

Some extra enums could be added but wondering if the extra code is useful enough to add - something like this 

```java
/**
 * Cross Region Inference mapping
 * https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference-support.html
 */
public enum CrossRegion {

	US("us", Set.of(Region.US_EAST_1, Region.US_WEST_2)),
	EU("eu", Set.of(Region.EU_CENTRAL_1, Region.EU_WEST_1, Region.EU_WEST_3));

	private final String prefix;
	private final Set<Region> regions;

	public String prefix() {
		return prefix;
	}

	public Set<Region> regions() {
		return regions;
	}

	CrossRegion(String prefix, Set<Region> regions) {
		this.prefix = prefix;
		this.regions = regions;
	}

}
```

and then adding more enums for Llama and Anthropic like below 

```java
/**
     * Cross Region Inference Llama models version.
	 * https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference-support.html
     */
	public enum CrossRegionLlamaChatModel implements ChatModelDescription {

		US_LLAMA3_2_1B_INSTRUCT_V1(CrossRegion.US, LlamaChatModel.LLAMA3_2_1B_INSTRUCT_V1),
		US_LLAMA3_2_3B_INSTRUCT_V1(CrossRegion.US, LlamaChatModel.LLAMA3_2_3B_INSTRUCT_V1),
		US_LLAMA3_2_11B_INSTRUCT_V1(CrossRegion.US, LlamaChatModel.LLAMA3_2_11B_INSTRUCT_V1),
		US_LLAMA3_2_90B_INSTRUCT_V1(CrossRegion.US, LlamaChatModel.LLAMA3_2_90B_INSTRUCT_V1),
[...]

PR #1419, opened 2024-09-26
Add Meta Llama 3.1, 3.2 instruct model enums to Bedrock

dafriz:
Add Meta Llama 3.1, 3.2 instruct model enums to Bedrock

Reference for model ids - https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html

### DIFF
--- a/models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/llama/api/LlamaChatBedrockApi.java
+++ b/models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/llama/api/LlamaChatBedrockApi.java
@@ -226,7 +226,42 @@ public class LlamaChatBedrockApi extends
 		/**
 		 * meta.llama3-70b-instruct-v1:0
 		 */
-		LLAMA3_70B_INSTRUCT_V1("meta.llama3-70b-instruct-v1:0");
+		LLAMA3_70B_INSTRUCT_V1("meta.llama3-70b-instruct-v1:0"),
+
+		/**
+		 * meta.llama3-1-8b-instruct-v1:0
+		 */
+		LLAMA3_1_8B_INSTRUCT_V1("meta.llama3-1-8b-instruct-v1:0"),
+
+		/**
+		 * meta.llama3-1-70b-instruct-v1:0
+		 */
+		LLAMA3_1_70B_INSTRUCT_V1("meta.llama3-1-70b-instruct-v1:0"),
+
+		/**
+		 * meta.llama3-1-405b-instruct-v1:0
+		 */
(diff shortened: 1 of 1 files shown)


models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/llama/api/LlamaChatBedrockApi.java
```diff
@@ -226,7 +226,42 @@ public enum LlamaChatModel implements ChatModelDescription {
...
+
+		/**
+		 * meta.llama3-1-405b-instruct-v1:0
+		 */
+		LLAMA3_1_405B_INSTRUCT_V1("meta.llama3-1-405b-instruct-v1:0"),
+
+		/**
+		 * meta.llama3-2-1b-instruct-v1:0
+		 */
+		LLAMA3_2_1B_INSTRUCT_V1("meta.llama3-2-1b-instruct-v1:0"),
```
markpollack, line 249: it's hard to find these strings from the docs.  I went [here](https://github.com/meta-llama/llama-stack/blob/8d41e6caa90c0e7bb44ec1755302a943bdf654f1/llama_stack/providers/adapters/inference/bedrock/bedrock.py#L23) to find the previously listed `llama3-1-x` models but dont' see the `llama-2-x` model values in the same code base.

markpollack: There is also cross-region ids, see - https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference-support.html

markpollack: ah, found the model listing here - https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html  added to the javadoc for future reference.

markpollack:
thanks, merged in 250d2afbb63627359a7a1fb2c4d9847db7ac1e26

markpollack:
merged in 250d2afbb63627359a7a1fb2c4d9847db7ac1e26

models/spring-ai-bedrock/src/main/java/org/springframework/ai/bedrock/llama/api/LlamaChatBedrockApi.java
dafriz: I see there is an issue raised about support for cross-region inference - https://github.com/spring-projects/spring-ai/issues/1282 Let's discuss there as this issue is closed.

Issue #889, opened 2024-06-18
[AzureOpenAiChatOptions] Unable to set all possible options when calling AzureOpenAiChatModel


RikJux:
**Bug description**
Consider the attributes that can be set through `AzureOpenAiChatOptions`:  
- maxTokens
- temperature
- topP
- logitBias
- user
- n
- stop
- presencePenalty
- frequencyPenalty
- deploymentName
- responseFormat
- functionCallbacks
- functions

The `AzureOpenAiChatOptions` object is used in the `AzureOpenAiChatModel` to create the instance of `ChatCompletionsOptions`.
`ChatCompletionsOptions` has the following attributes:
- messages
- maxTokens
- temperature
- topP
- logitBias
- user
- n
- stop
- presencePenalty
- frequencyPenalty
- stream
- model			   
- functions
- functionCall
- functionCallConfig
- dataSources
- enhancements
- seed
- responseFormat
- tools
- toolChoice
- logprobs
- topLogprobs

In code it seems like there's no way to set options like **logprobs**, **enhancements** or **seed**.

Furthermore, in the method that should merge two `ChatCompletionsOptions`:
```java
private ChatCompletionsOptions merge(ChatCompletionsOptions fromOptions, ChatCompletionsOptions toOptions) 
```
such options are not taken into account

**Expected behavior**
I'm expecting the `AzureOpenAiChatOptions` to be aligned with the currently available options for Azure OpenAi

pradu3:
Hi All!

This change is not working and is breaking things. It is adding the new options but makes more of them indirectly mandatory. 

The ``Assert.notNull`` checks in ``AzureOpenAiChatOptions.Builder.with*`` make the attributes actually mandatory.  ``AzureOpenAiChatOptions.fromOptions`` is called from ``AzureOpenAiChatModel.getDefaultOptions``. When the ``AzureOpenAiChatModel.defaultOption`` attribute doesn't have all the checked fields set (e.g. ``seed``), there will be an exception when ``AzureOpenAiChatModel.getDefaultOptions``. This will cause, for example, ``ChatClient.create`` to fail. This happen even when the ``AzureOpenAiChatModel( OpenAIClient microsoftOpenAiClient)`` constructor is used.

```
Caused by: java.lang.IllegalArgumentException: seed must not be null
	at org.springframework.util.Assert.notNull(Assert.java:172) ~[spring-core-6.1.13.jar:6.1.13]
[...]

PR #1425, opened 2024-09-26
Remove unnecessary null checks in AzureOpenAiChatOptions.Builder

sobychacko:
This is related to https://github.com/spring-projects/spring-ai/issues/889

This commit addresses a bug where certain fields were being made indirectly mandatory due to Assert.notNull checks in the Builder's with* methods. Specifically:

- Removed Assert.notNull checks from withResponseFormat, withSeed, withLogprobs, withTopLogprobs, and withEnhancements methods.

These checks were causing exceptions in AzureOpenAiChatModel.getDefaultOptions when not all fields were set, leading to failures in methods like ChatClient.create, even when using the AzureOpenAiChatModel constructor with OpenAIClient.

The removal of these checks aligns with the @JsonInclude(Include.NON_NULL) annotation on AzureOpenAiChatOptions, which already ignores null options. This change maintains the intended flexibility while preventing unintended mandatory requirements.


markpollack:
merged e29d38d987d9aabdb35cf87321b13bb33dbdb871

PR #1430, opened 2024-09-28
Make VectorStore extend from DocumentRetriever

alexcheng1982:
The new `DocumentRetriever` should be the interface for all implementations that can retrieve documents, including vector stores. So I think it's reasonable for `VectorStore` to extend from `DocumentRetriever`.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/VectorStore.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/VectorStore.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.Optional;
 
 import org.springframework.ai.document.Document;
+import org.springframework.ai.document.DocumentRetriever;
 import org.springframework.ai.document.DocumentWriter;
 
 /**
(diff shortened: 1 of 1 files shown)


markpollack:
yes, to accept a `SearchRequest` would enable to strategize how the document is retrieved vs passing in just a plain string and then hardcoding the search algorithm to be similarity search.  We wil address a rewrite/review of the VectorStore interface for M4.  See https://github.com/spring-projects/spring-ai/pull/1227/ for related work (I felt that adding single method shortcuts was not a good idea in that PR but otherwise ,adding additional common  built in search options. 

Once that is in the core vector store interface, a ReRanking advisor can take advantage of it.

ThomasVitale:
It might even help decoupling the `DocumentRetriever` from the `VectorStore` object, introducing a `VectorStoreDocumentRetriever` (which internally would delegate to a `VectorStore`). That would be similar to the Indexes defined in LlamaIndex and allow to have `Retriever` objects for specific contexts rather than having to pass metadata filters on every call. I'm working on some experiments to research possible designs that would help building advanced RAG workflows. I'll share them soon.

ThomasVitale:
We are introducing a new modular RAG architecture in Spring AI. As part of that work, the `DocumentRetriever` API has been revamped and it's now the main entry point for retrieving similar documents in a RAG workflow. A `VectorStoreDocumentRetriever` implementation has been introduced in https://github.com/spring-projects/spring-ai/pull/1604.

markpollack:
this has been done in the commit https://github.com/spring-projects/spring-ai/commit/5d8c032bb706a04876687cb801f98b3f84417e50

PR #1442, opened 2024-09-30
Correctly set Chroma collection id when initializeSchema set to false

alexcheng1982:
In the current implementation of `ChromaVectorStore`, when `initializeSchema` is set to `false`, `collectionId` will be set to `null`, all Chroma API calls will fail with 404 error due to empty collection id.

I updated the logic of handling `initializeSchema`. When the collection doesn't exist, if `initializeSchema` is set to `true`, the collection will be created. Otherwise, an exception is thrown. This makes sure that `collectionId` will be correctly set.


tzolov:
Thank you @alexcheng1982 

rebased, squashed and merged at ae743f98a54300bd5053e43c4b58fe492c76ece9

PR #1448, opened 2024-10-02
Enable setting json root via json pointer

ktm:
I found some JSON in the wild which I want to use, but the doc is not currently ingestable via the existing JsonReader.  events.json in the test resources is a simplified example of the wild json.


markpollack:
Wild json!  I love it ;)  Thanks

markpollack:
thanks!  added docs.  merged in 2babb2a87867de8f643c7a544cb16dcffa26ac4a

PR #1461, opened 2024-10-04
[WIP] Introduce a new evaluator to detect hallucinations

eddumelendez:
See https://ollama.com/blog/reduce-hallucinations-with-bespoke-minicheck


markpollack:
I quite like this, i've done a bit of research on the topic and changed the name to FactCheckingEvaluator and added some javadoc.  I will write some docs as well with code samples.  Thanks.  

merged in a987e0305a01e951992964f15096f4eecb800cc2

Issue #1462, opened 2024-10-05
The ElasticsearchVectorStoreAutoConfiguration is not public as all the others


jcgouveia:
I want to exclude ElasticsearchVectorStoreAutoConfiguration from my spring boot application to define a custom configuration, but the class is not public , therefore I cannot add it to the exclusions list.

All the other VectorStoreAutoConfiguration classes are public.

markpollack:
This has been fixed in https://github.com/spring-projects/spring-ai/commit/b1377d194dc1ebceb7032431b7ac362358a7e82d and will be available in 1.0 M3

PR #1463, opened 2024-10-05
Add the public access modifier to ElasticsearchVectorStoreAutoConfiguration

dafriz:
Add the `public` access modifier to `ElasticsearchVectorStoreAutoConfiguration` class, so that it can be excluded and a custom configuration can be defined. 

This makes it consistent with all the other vector store auto configurations classes.

Fixes https://github.com/spring-projects/spring-ai/issues/1462


markpollack:
thanks, merged in b1377d194dc1ebceb7032431b7ac362358a7e82d

PR #1466, opened 2024-10-05
Fix chat model autoconfig when no Micrometer Tracing in context

ThomasVitale:
When Micrometer Tracing is not in the classpath, Spring AI makes Spring Boot fail the startup.


tzolov:
Thanks Thomas

PR #1467, opened 2024-10-05
Improve optional values handling in chat model observations

ThomasVitale:
Chat model observations support several key-value pairs, coming from the Spring AI abstractions. Currently, whenever a value is not available (either because not configured by the user or not supported by the model provider), span/metrics attributes are generated anyway with value "none". 

That causes several issues, including an unneeded increase in time series, challenges in alerting/monitoring (especially for integer/double attributes that suddenly are populated with a string), and non-compliance with the OpenTelemetry Semantic Conventions (according to which, attributes should be excluded altogether if there's no value). 

This pull request changes the conventions for chat model observation to exclude the generation of span/metrics attributes for optional values which don't have any value.

Before:

Screenshot 2024-10-05 at 22 29 53

After:

Screenshot 2024-10-05 at 23 20 19


ThomasVitale:
Additional background: initially, we went with the "none" fallback since that was the main strategy adopted in other Spring projects. But in this case, it turned out not to be efficient for the reasons I mentioned above.

tzolov:
Very useful. Thanks

PR #1469, opened 2024-10-06
Improve optional values handling in vector store observations

ThomasVitale:
Vector store observations support several key-value pairs, coming from the Spring AI abstractions. Currently, whenever a value is not available (either because not configured by the user or not supported by the vector store provider), span/metrics attributes are generated anyway with value none.

That causes several issues, including an unneeded increase in time series, challenges in alerting/monitoring (especially for integer/double attributes that suddenly are populated with a string), and non-compliance with the OpenTelemetry Semantic Conventions (according to which, attributes should be excluded altogether if there's no value).

This pull request changes the conventions for vector store observations to exclude the generation of span/metrics attributes for optional values which don't have any value.


ThomasVitale:
I couldn't run the tests for Azure, Hana and Pinecone since I don't have an account for those services. The rest I tested thanks to Testcontainers.

tzolov:
Thank you @ThomasVitale

PR #1470, opened 2024-10-06
Keep model metric attributes fixed for Prometheus

ThomasVitale:
Unlike other metrics implementations supported by Micrometer, when using the Prometheus integration, all metric attributes are supposed to have a value, or else the related metrics are dropped.


tzolov:
nice learning curve :)

PR #1472, opened 2024-10-06
Improve chat client and advisor observations

ThomasVitale:
* Make ChatClient and Advisor observation logic null-safe
* Simplify naming for Advisor observations
* Include high-cardinality attributes only if a value is present
* Fix condition to include system test to chat client observations
* Add Advisor order information to context
* Streamline usage of enums and utils to reduce hard-coded/duplications
* Fix pending Chroma integration test

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/DefaultAroundAdvisorChain.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/DefaultAroundAdvisorChain.java
@@ -85,6 +85,7 @@ public class DefaultAroundAdvisorChain implements CallAroundAdvisorChain, Stream
 			.withAdvisorType(AdvisorObservationContext.Type.AROUND)
 			.withAdvisedRequest(advisedRequest)
 			.withAdvisorRequestContext(advisedRequest.adviseContext())
+			.withOrder(advisor.getOrder())
 			.build();
 
 		return AdvisorObservationDocumentation.AI_ADVISOR
@@ -106,6 +107,7 @@ public class DefaultAroundAdvisorChain implements CallAroundAdvisorChain, Stream
 				.withAdvisorType(AdvisorObservationContext.Type.AROUND)
 				.withAdvisedRequest(advisedRequest)
 				.withAdvisorRequestContext(advisedRequest.adviseContext())
+				.withOrder(advisor.getOrder())
 				.build();
 
 			var observation = AdvisorObservationDocumentation.AI_ADVISOR.observation(null,
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/observation/AdvisorObservationDocumentation.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/observation/AdvisorObservationDocumentation.java
@@ -27,7 +27,7 @@ import io.micrometer.observation.docs.ObservationDocumentation;
 public enum AdvisorObservationDocumentation implements ObservationDocumentation {
 
 	/**
-	 * AI Chat Client observations
+	 * AI Advisor observations
 	 */
 	AI_ADVISOR {
 		@Override
@@ -65,7 +65,7 @@ public enum AdvisorObservationDocumentation implements ObservationDocumentation 
 		ADVISOR_TYPE {
 			@Override
 			public String asString() {
-				return "spring.ai.chat.client.advisor.type";
+				return "spring.ai.advisor.type";
 			}
 		}
 
@@ -74,12 +74,12 @@ public enum AdvisorObservationDocumentation implements ObservationDocumentation 
 	public enum HighCardinalityKeyNames implements KeyName {
 
 		/**
-		 * Chat Model name.
+		 * Advisor name.
 		 */
 		ADVISOR_NAME {
 			@Override
 			public String asString() {
-				return "spring.ai.chat.client.advisor.name";
+				return "spring.ai.advisor.name";
 			}
 		},
 		/**
@@ -88,7 +88,7 @@ public enum AdvisorObservationDocumentation implements ObservationDocumentation 
 		ADVISOR_ORDER {
 			@Override
 			public String asString() {
-				return "spring.ai.chat.client.advisor.order";
+				return "spring.ai.advisor.order";
 			}
 		}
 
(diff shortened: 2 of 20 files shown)


spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/DefaultAroundAdvisorChain.java
```diff
@@ -85,6 +85,7 @@ public AdvisedResponse nextAroundCall(AdvisedRequest advisedRequest) {
 			.withAdvisorType(AdvisorObservationContext.Type.AROUND)
 			.withAdvisedRequest(advisedRequest)
 			.withAdvisorRequestContext(advisedRequest.adviseContext())
+			.withOrder(advisor.getOrder())
```
ThomasVitale, line 88: I noticed we were not passing the "order" field to the observation context

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/observation/AdvisorObservationDocumentation.java
```diff
@@ -65,7 +65,7 @@ public String asString() {
 		ADVISOR_TYPE {
 			@Override
 			public String asString() {
-				return "spring.ai.chat.client.advisor.type";
+				return "spring.ai.advisor.type";
```
ThomasVitale, line 68: @tzolov what do you think about this simplification of the naming for Advisor attributes?

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientInputContentObservationFilter.java
```diff
@@ -32,72 +31,54 @@
...
+		chatClientUserParams(chatClientObservationContext);
 
 		return chatClientObservationContext;
 	}
 
-	protected KeyValue chatClientSystemText(ChatClientObservationContext context) {
-		if (!StringUtils.hasText(context.getRequest().getUserText())) {
-			return CHAT_CLIENT_SYSTEM_TEXT_NONE;
+	protected void chatClientSystemText(ChatClientObservationContext context) {
+		if (!StringUtils.hasText(context.getRequest().getSystemText())) {
```
ThomasVitale, line 49: Here the condition was checking the user text instead of the system text, I fixed it now

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
```diff
@@ -88,7 +90,7 @@ public String asString() {
 		CHAT_CLIENT_TOOL_FUNCTION_CALLBACKS {
 			@Override
 			public String asString() {
-				return "spring.ai.chat.client.tool.functioncallbacks";
+				return "spring.ai.chat.client.tool.function.callbacks";
```
ThomasVitale, line 93: @tzolov I made this change to align with the other attributes called "*.function.names". Is this ok?

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/observation/AdvisorObservationDocumentation.java
tzolov: The advisors are only applicable in the context of ChatClient. 
But i guess it is enough to reflect this int the observation's name (e.g.  spring.ai.chat.client.advisor) and not in the attribute names?

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientInputContentObservationFilter.java
tzolov: good catch

spring-ai-core/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationDocumentation.java
tzolov: looks good

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/DefaultAroundAdvisorChain.java
tzolov: I've added the order to the context but apparently missed to populate it

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/observation/AdvisorObservationDocumentation.java
ThomasVitale: The related span is also nested under a "chat_client" parent span, so that might be enough. But if we should have it in the attribute name as well, I can revert this

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/observation/DefaultAdvisorObservationConvention.java
```diff
@@ -27,18 +28,9 @@
...
-	public static final String DEFAULT_NAME = "spring.ai.chat.client.advisor";
-
-	private static final String CHAT_CLIENT_ADVISOR_SPRING_AI_KIND = "chat_client_advisor";
-
-	private static final KeyValue ADVISOR_TYPE_NONE = KeyValue.of(LowCardinalityKeyNames.ADVISOR_TYPE,
-			KeyValue.NONE_VALUE);
-
-	private static final KeyValue ADVISOR_NAME_NONE = KeyValue.of(HighCardinalityKeyNames.ADVISOR_NAME,
-			KeyValue.NONE_VALUE);
+	public static final String DEFAULT_NAME = "spring.ai.advisor";
```
ThomasVitale, line 33: @tzolov this name is only used for metrics. Since Micrometer generates several timer-related metrics with long names, I thought of making this base name shorter as well.

So, instead of having metrics with name "spring_ai_chat_client_advisor_active_milliseconds_count" we would have "spring_ai_advisor_active_milliseconds_count".

Thoughts? Since Advisors are not used anywhere else, we might not need the "chat.client" part in the base name.

tzolov:
LGTM, thanks @ThomasVitale

PR #1473, opened 2024-10-06
Optimize Testcontainers config for VectorStores

ThomasVitale:
* Unify image definition for vector stores in vector-store modules
* Unify image definition for vector stores in spring-ai-testcontainers module


tzolov:
Like this consolidation. Thanks

Issue #1478, opened 2024-10-06
Introduce TranscriptionModel interface


habuma:
Even though only OpenAI and Azure OpenAI current offer transcription support, they are each represented by distinct transcription model types: `OpenAiAudioTranscriptionModel` and `AzureOpenAiAudioTranscriptionModel`. There is no common interface between these two, which means that you have to explicitly inject the one you are working with rather than inject a common interface.

They both implement `Model<AudioTranscriptionPrompt, AudioTranscriptionResponse>`, but it would be more handy to have a common interface like this that they each could implement:

```
public interface TranscriptionModel extends Model<AudioTranscriptionPrompt, AudioTranscriptionResponse> {

  AudioTranscriptionResponse call(AudioTranscriptionPrompt transcriptionPrompt);

}
```

Similar to `ChatModel`, there may be opportunity for some additional default convenience methods, as well. Perhaps one that accepts a `Resource` and returns a `String` and another that accepts a `Resource` and `AudioTranscriptionOptions` and produces a `String`.

habuma:
Note that I don't have the time right now to work on this and submit a PR, but don't mind doing so when I get a moment. But if someone wants to beat me to it, then by all means go for it.

mudabirhussain:
I will do it....

markpollack:
We won't have time to do this before GA, so postponing until after.

PR #1484, opened 2024-10-07
feature: Add common TranscriptionModel interface for audio transcription

mudabirhussain:
- Created TranscriptionModel interface that extends Model<AudioTranscriptionPrompt, AudioTranscriptionResponse>
- Implemented `call(AudioTranscriptionPrompt)` method for better compatibility between OpenAI and Azure OpenAI transcription models
- Added default convenience methods for handling Resource and AudioTranscriptionOptions to return transcription as a String

Resolution of this opened issue: https://github.com/spring-projects/spring-ai/issues/1478

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/model/TranscriptionModel.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/model/TranscriptionModel.java
@@ -1,0 +1,21 @@
+package org.springframework.ai.model;
+
+import org.springframework.ai.audio.transcription.AudioTranscriptionOptions;
+import org.springframework.ai.audio.transcription.AudioTranscriptionPrompt;
+import org.springframework.ai.audio.transcription.AudioTranscriptionResponse;
+import org.springframework.core.io.Resource;
+
+public interface TranscriptionModel extends Model<AudioTranscriptionPrompt, AudioTranscriptionResponse> {
+
+	AudioTranscriptionResponse call(AudioTranscriptionPrompt transcriptionPrompt);
+
+	default String transcribe(Resource resource) {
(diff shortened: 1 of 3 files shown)


piotrooo:
Meanwhile, as an enrichment to transcription take a look at https://github.com/spring-projects/spring-ai/pull/1278

spring-ai-core/src/main/java/org/springframework/ai/model/TranscriptionModel.java
```diff
@@ -0,0 +1,22 @@
+package org.springframework.ai.model;
```
ThomasVitale, line 1: Should this interface be in the package `org.springframework.ai.model.audio.transcription`?

markpollack:
Is there another transcription model we can use to verify the abstraction?

spring-ai-core/src/main/java/org/springframework/ai/model/TranscriptionModel.java
markpollack: yes, it should.

Model interfaces should be placed in packages that reflect their **functional domain**:

- For single-level domains:  
  `org.springframework.ai.<domain>`

- For hierarchical domains:  
  `org.springframework.ai.<category>.<subdomain>`
  
| Model Interface       | Package Location                          |
|-----------------------|--------------------------------------------|
| `EmbeddingModel`      | `org.springframework.ai.embedding`         |
| `ModerationModel`     | `org.springframework.ai.moderation`        |
| `TextToSpeechModel`   | `org.springframework.ai.audio.tts`         |

markpollack:
I've added the additional tests to check the default methods etc.

Merged in 4cf23773ab01f147c70053e732999d8d85ed9da0

Thanks @mudabirhussain and others.

PR #1485, opened 2024-10-07
Refactor FunctionCallback API to use ToolContext instead of Map

tzolov:
- Replaced `Map<String, Object>` with `ToolContext` in the `FunctionCallback`, `AbstractFunctionCallback`, and related classes.
 - Updated all BiFunction definitions to use `ToolContext` as the second parameter, enhancing the clarity and structure of the tool context management.
 - Modified `FunctionCallbackWrapper` and `FunctionCallbackContext` to adapt to the new `ToolContext` parameter.
 - Adjusted the handling of tool context in the documentation and test classes.
 - Updated relevant test cases to reflect the API changes and modified the function handling logic to ensure consistency.


markpollack:
minor fixes, merged in 38082de0f8e4ac4533469e81f11d1cddae17754f

PR #1492, opened 2024-10-07
Observation for Vertex gemini chat model streaming operation

sobychacko:
- Adding observability for Vertex Gemini chat model streaming call
- Updating auto-configuration for gemini chat model with the observability components
- Integration test for streaming observability


markpollack:
merged in bac002bf8356be99ff0fca5eac9e855ca1941c01

Issue #1496, opened 2024-10-08
Common types for text-to-speech


habuma:
Similar to what I suggested in #1478, it would be great if text-to-speech had a set of common types. The `SpeechModel` interface, as well as `SpeechPrompt`, `SpeechResponse`, and `StreamingSpeechModel` feel like what I'd expect with such types, but they are currently delivered in the OpenAI module. Even though OpenAI is the only implementation, it feels like those types should be in core with the implementations and OpenAI-specific extensions in the OpenAI module.

Moreover, while `SpeechPrompt` feels like it should be in core, it carries `OpenAiAudioSpeechOptions`. Perhaps there should be a more generic `SpeechOptions` that is carried by `SpeechPrompt`, with `OpenAiAudioSpeechOptions` being an extension of `SpeechOptions`. 

Altogether, this would not only make the types more consistent with how the types for chat and other models are  structured, it also sets the stage for additional text-to-speech implementations should more APIs that offer that be added to Spring AI.

ThomasVitale:
I like the suggestion and I'm available to work on this. I'll have a PR ready soon.

mudabirhussain:
> Similar to what I suggested in #1478, it would be great if text-to-speech had a set of common types. The `SpeechModel` interface, as well as `SpeechPrompt`, `SpeechResponse`, and `StreamingSpeechModel` feel like what I'd expect with such types, but they are currently delivered in the OpenAI module. Even though OpenAI is the only implementation, it feels like those types should be in core with the implementations and OpenAI-specific extensions in the OpenAI module.
> 
> Moreover, while `SpeechPrompt` feels like it should be in core, it carries `OpenAiAudioSpeechOptions`. Perhaps there should be a more generic `SpeechOptions` that is carried by `SpeechPrompt`, with `OpenAiAudioSpeechOptions` being an extension of `SpeechOptions`.
> 
> Altogether, this would not only make the types more consistent with how the types for chat and other models are structured, it also sets the stage for additional text-to-speech implementations should more APIs that offer that be added to Spring AI.

What is the reason for having **SpeechResponse** in the core, since it is already an implementation of **ModelResponse**, which belongs to the core family? Most of the classes or interfaces mentioned above are already implementations or extensions of existing core family classes or interfaces.

PR #1518, opened 2024-10-09
Consolidate SpeechModel APIs

ThomasVitale:
* Consolidate SpeechModel APIs into the spring-ai-core module, make it null-safe and covered by unit tests.
* Refactor OpenAiSpeechModel APIs to implement the new consolidated APIs.
* Delete leftover ImageResponseMetadata class in the spring-ai-openai module.

Fixes gh-1496


ThomasVitale:
Once this PR is merged, I'll have 2 followups PRs: one to update the documentation and one to add observability instrumentation to the OpenAiSpeechModel.

markpollack:
The biggest concern for me is how portable are the abstractions, we need at least two impls to ensure that we have the correct abstraction before moving into main.  Also, most of the changes seem sort of "shallow", e.g. just changing then name of the class and the types, but aside from the name change they have the same functionality and the 'non speach' core api.

Thoughts?

ThomasVitale:
I also thought about that, so I was conservative with the changes. The main benefit I see in having the Speech APIs in Core is that it's simpler to demonstrate and try out new implementations, or even for users to implement their own custom ones. Mainly for the semantics given by the interfaces and the few conventions they have. 

But I don't have strong feeling about it, so I'd be fine parking this for now. I was looking into adding observability for speech models, but we can wait for that (I wouldn't do it inside the OpenAI module so to keep consistency with the rest of the observability implementation).

@habuma do you have any thoughts about this change?

PR #1549, opened 2024-10-16
fix chroma api rest api bug

CodingNowPls:
fix chroma rest api request param bug


markpollack:
Thanks, I've added a test as well.  Merged in f461bd603d554e15139add55c9029e551ec1c828

CodingNowPls:
ok

PR #1566, opened 2024-10-19
Enhance Ollama model auto-pull feature

ThomasVitale:
* Fix configuration inheritance issue when default value is not specified.
* Make it possible to enable the auto-pull feature only for specific model types (e.g. for chat models only).
* Add the possibility to list explicit models to auto-pull at startup time.


tzolov:
Thanks @ThomasVitale 
Rebased, extended with default chat/embedding model config and merged at: 5e8cecd8b196985284ed5b59be2b10dca3f8dd8f

PR #1570, opened 2024-10-20
Normalize Ollama model names in auto-pull feature

ThomasVitale:
In order to support edge cases due to different naming formats, this PR introduced an explicit normalization logic to ensure the correct matching when checking for the availability of a certain model.


tzolov:
Thanks @ThomasVitale 
Rebased, fixed the useTestcontainer logic and squashed at ef2e39a8adad01257ca05b0abf93a086589313a5

Issue #1571, opened 2024-10-20
bug: NPE when processing ByteArrayResource


ogbozoyan:
I've also created proposal how to fix it https://github.com/spring-projects/spring-ai/pull/1572
**Bug description**
 If processing Resource doesn't have getFileName() at org.springframework.ai.transformer.splitter.TextSplitter#createDocuments will throw NPE

**Environment**
Please provide as many details as possible: springAiVersion = "1.0.0-M3", Java 21, PG vector store

PgVectorStore [compose file](https://github.com/ogbozoyan/cron-validator/blob/main/pgvector/docker-compose.yaml)
Also i have [init sql](https://github.com/ogbozoyan/cron-validator/blob/main/src/main/resources/db/migration/V1__init.sql)

**Steps to reproduce**
I've endpoint [CronController](https://github.com/ogbozoyan/cron-validator/blob/main/src/main/kotlin/ru/ogbozoyan/cron/web/controller/CronController.kt) which produce event to [PgEventListener](https://github.com/ogbozoyan/cron-validator/blob/main/src/main/kotlin/ru/ogbozoyan/cron/service/pg/PgEventListener.kt)
in listener asynchronously calls [PGVectorStoreService](https://github.com/ogbozoyan/cron-validator/blob/main/src/main/kotlin/ru/ogbozoyan/cron/service/pg/PGVectorStoreService.kt) i've wrote wrapper method for each document enrich with filename, if i pass documents to `org.springframework.ai.transformer.splitter.TextSplitter#createDocuments` 
without filenames in metadata part, when collector try to get `e.getValue()` will throw unclear NPE

```java
Map<String, Object> metadataCopy = metadata.entrySet()
					.stream()
					.collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()));
``` 

**Expected behavior**
Safely Collectors.toMap calling 

**Minimal Complete Reproducible example**

```log
20-10-2024 23:06:50.018  -  INFO 52440 [Async-1]  r.o.cron.service.pg.PgEventListener:16  : Processing event: PgEvent(resource=Byte array resource [resource loaded from byte array], type=PDF, fileName=Cloud_Architecture_Demystified_Understand_how_to_design_sustainable.pdf)
[...]

PR #1572, opened 2024-10-20
fix: fixed NPE while splitting document chunks

ogbozoyan:
Fixed NPE when processing splitted shards of text,  If metadata has entry 
"key" with null value will cause NPE 
More detailed information provided in https://github.com/spring-projects/spring-ai/issues/1571


Mikhail-Lytov (approved):

serzhe1 (approved):

tzolov:
Thank you @ogbozoyan 
rebased, squashed and merged at 8b1882b2244953ef0735227d4a5525b1b5479eba

PR #1587, opened 2024-10-23
Refine Jackson `ObjectMapper` handling

sdeleuze:
ObjectMapper instantiation is costly, so unless its usage is one-shot, it is better to create a reusable instance for upcoming usage.

Also, before this commit, serialization of most Kotlin classes was not supported due to the lack of proper Jackson KotlinModule detection.

This commit:
 - Avoids per invocation `ObjectMapper` instantiation when relevant
 - Automatically detects and enables well-known Jackson modules including the Kotlin one
 - Removes `org.springframework.ai.vectorstore.JsonUtils` which looks not needed anymore

More optimizations are possible like reusing more `ObjectMapper` instances, but this could introduce more breaking changes so this commit intends to be a good first step.

Kotlin tests will be provided in a follow-up PR.


tzolov:
Thanks @sdeleuze , 
LGTM. there few small changes I will apply with the merge.

tzolov:
Few additional changes (below) where applied, then rebased, squashed and merged at: f21b8a42b521f7a345b3651065fb88613c3add58 

Additional changes:
- Update ModelOptionsUtils to use JacksonUtils.instantiateAvailableModules()
- Add missing license headers
- Add missing author Javadoc comments

PR #1602, opened 2024-10-26
Merge runtime options for Stability AI image gen

habuma:
This PR fixes two problems with Stability AI image generation:

- For Stability AI-specific options, the `mergeOptions()` method was only copying the default options (which were null unless otherwise specified in configuration properties). Consequently, setting those options (such as style preset) in Java code achieved nothing. The main part of this PR fixes that to merge the Stability default properties with runtime properties the same as for portable options.
- There was a typo in the JSON property name for style_preset (it was "style_present"), preventing that property from being serialized into JSON correctly.


markpollack:
Thanks.  I asked Claude to write a test for the PR and low and behold, it found an bug.  So neat.  We maybe able to use `ModelOptionsUtils` here, but I didn't try.  

Merged in ac4571ea522320ca5c1c17e03b668e8070e69c5b

Issue #1603, opened 2024-10-26
Modular RAG - Part 1


ThomasVitale:
This issue will track the first part of the adoption of the Modular RAG architecture in Spring AI.

### Context

The work can be split into two categories.

1. The design and implementation of the building blocks for RAG modules that will be done in the `org.springframework.ai.rag` package. Each component is well encapsulated and can be used by itself to compose any kind of RAG flow. For example, you can use such components to build your own RAG flows using Spring State Machine, Spring Cloud Function or Spring Cloud Data Flow.
2. Using the Advisor API, a `RetrievalAugmentationAdvisor` will be implemented to provide some out-of-the-box RAG flows using the building blocks defined in the previous category.

For background information, please refer to the following sources:

- [Modular RAG: Transforming RAG Systems into LEGO-like Reconfigurable Frameworks](http://export.arxiv.org/abs/2407.21059)
- [Retrieval-Augmented Generation for Large Language Models: A Survey](https://export.arxiv.org/abs/2312.10997)
- [LangChain Retrieval Conceptual Guide](https://python.langchain.com/docs/concepts/retrieval/)
- [LlamaIndex Advanced Retrieval Strategies](https://docs.llamaindex.ai/en/stable/optimizing/advanced_retrieval/advanced_retrieval/)

#### Modules

In this first part, the focus will be on establishing the following modules and submodules:

- Pre-Retrieval
  - Query Transformation https://github.com/spring-projects/spring-ai/pull/1703
  - Query Expansion https://github.com/spring-projects/spring-ai/pull/1703
- Retrieval
  - Document Search https://github.com/spring-projects/spring-ai/pull/1604
  - Document Join https://github.com/spring-projects/spring-ai/pull/1767
- Post-Retrieval (only interfaces)
  - Document Ranking https://github.com/spring-projects/spring-ai/pull/1767 (a)
  - Document Selection https://github.com/spring-projects/spring-ai/pull/1767 (a)
  - Document Compression https://github.com/spring-projects/spring-ai/pull/1767 (a)
- Augmentation
  - Content https://github.com/spring-projects/spring-ai/pull/1644
- Orchestration (experimental)
  - Query Routing https://github.com/spring-projects/spring-ai/pull/1767

This issue will be updated as we progress with the design and implementation work.

#### Advisor

In this first part, the focus will be on:

- Establishing a `RetrievalAugmentationAdvisor`.
- Adopting the available sub-modules to support Naive and Advanced RAG flows.
[...]

Issue #1290, opened 2024-08-29
VectorStore interface should be split, with readonly operations becoming a super interface


johnsonr:
The `VectorStore` interfaces covers two things: similarity search and `DocumentWriter`/add/accept/delete.

These are different concepts and a `SimilaritySearchService` or similarly named interface should be split out.

I currently have something I would like to expose as a `VectorStore` that is readonly and does not support the ingestion operations.

markpollack:
Yes, I agree as well that we need some redesign here and also te ability to pass options to add/delete operations.  There are also other operations besides similarity search that need to be exposed.

rjrudin:
@markpollack Let me know if this worth a separate ticket, happy to open one - I'd also like to provide additional context to a `SearchRequest`, such as additional query context that isn't part of the user's query. For example, a user may wish to engage a chatbot about crimes in a particular area defined by a bounding box. Under the hood, I want to use that bounding box to select a subset of records in a database and then do a similarity search on those records. That additional query context doesn't fit into a `Filter.Expression` - perhaps a metadata map on `SearchRequest` would be a reasonable extension point?

ThomasVitale:
As part of the new (experimental) Modular RAG features, a VectorStoreDocumentRetriever has been introduced to support search operations decoupled from the generic CRUD operations in the VectorStore API.

@johnsonr does that help with your use case?

Docs: https://docs.spring.io/spring-ai/reference/api/retrieval-augmented-generation.html#_retrieval

markpollack:
@sobychacko thoughts on this.

markpollack:
Yep.  We can retrofit a `SimilaritySearchOperations` interface (name TBD) to that it can be passed around to code without fear of that code adding anything.

markpollack:
@rjrudin The `SearchRequest` is specifically to feed in the three  things that are required for a similarity search - the query string, topk to return and filter expression as a passthrough to execution via a vector db that supports similarity search. Your query that you mention seems orthogonal to that execution path/use case.  It sounds more like a geo-query (great with mongodb) in a non-vector-store database and then some sort of custom ranking of those records (perhaps by closest to a point).

Issue #1592, opened 2024-10-23
Consistent design of Options builder classes


markpollack:
In some cases we have a dedicated top level object as a builder and in some cases it in an inner class.  Also, most options builders do not implement an interface.  We should adhere to the design that has been used pretty consistently in spring-framework which has the builder as an inner object and also an interface for the builder.

We should also add tests for the builders.

This came out of the PR discussion #414

youngmoneee:
I still think that ensuring the immutability of objects is crucial to prevent potential concurrency issues. The reason I initially designed the builder as a separate object outside the option object was not only to facilitate simple inheritance but also to allow future extensions such as fromDefaults() and fromConfig().

However, I now think that separation of concerns and consistency are more important. Based on this, my current considerations are as follows:
1. Operate in a consistent manner with other Spring projects through a simple builder.
2. The object itself must be immutable, and modifications should only be allowed through the builder.
3. Minimize inconvenience from this approach (e.g., by passing existing objects as builder parameters).
4. A separate factory or converter that handles the responsibility of object creation and transformation when relying on Spring Boot.
5. Testing for the implementation of the above points.

Do you have any additional ideas or considerations?

markpollack:
Hi!

I agree with these points

1. Operate in a consistent manner with other Spring projects through a simple builder.
2. The object itself must be immutable, and modifications should only be allowed through the builder.
3. Minimize inconvenience from this approach (e.g., by passing existing objects as builder parameters).

As for 4. I want to avoid the code duplication (which is quite large and boilerplate, prone to mistakes) between the 'core options' class and what is exposed to boot.  We have figured out a way around that, basically anything that requires a @NestedConfigurationProperty annotation will be added to the file `META-INF/additional-spring-configuration-metadata.json` in the appropriate autoconfig package.  See [here](META-INF/additional-spring-configuration-metadata.json), for the docs.  

I didn't understand the comment 

> The reason I initially designed the builder as a separate object outside the option object was not only to facilitate simple inheritance but also to allow future extensions such as fromDefaults() and fromConfig().

but perhaps it isn't relevant if we follow the 'simple builder' approach in other spring projects.

One thing that I'm not sure about is to have an interface for the builder.  It is good design, but does feel like a bit overkill.

Thoughts?  Thanks for your feedback and interest!

youngmoneee:
@markpollack

Thank you for your input! The link you provided is currently inaccessible 😞

[...]

PR #1604, opened 2024-10-26
Modular RAG: Retrieval with Vector Stores

ThomasVitale:
* Establish new package for Modular RAG components.
* Add new Query API, representing a query in the context of a RAG flow.
* Define Retrieval package for the RAG building blocks handling the data retrieval operations.
* Relocate DocumentRetriever to Retrieval package and implement VectorStoreDocumentRetriever.
* Introduce RetrievalAugmentationAdvisor as the successor of QuestionAnswerAdvisor. It uses the Retrieval building blocks described in the previous point.
* Make Advisor APIs null-safe and update tests accordingly.

Relates to gh-https://github.com/spring-projects/spring-ai/issues/1603

It might also address the requests from https://github.com/spring-projects/spring-ai/issues/1290

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
@@ -1,0 +1,224 @@
+/*
+ * Copyright 2023-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.chat.client.advisor;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+
+import org.springframework.ai.chat.client.advisor.api.AdvisedRequest;
+import org.springframework.ai.chat.client.advisor.api.AdvisedResponse;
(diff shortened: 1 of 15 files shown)


spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/api/AdvisedRequest.java
```diff
@@ -57,13 +57,53 @@
...
+ * @since 1.0.0
  */
-public record AdvisedRequest(ChatModel chatModel, String userText, String systemText, ChatOptions chatOptions,
-		List<Media> media, List<String> functionNames, List<FunctionCallback> functionCallbacks, List<Message> messages,
-		Map<String, Object> userParams, Map<String, Object> systemParams, List<Advisor> advisors,
-		Map<String, Object> advisorParams, Map<String, Object> adviseContext, Map<String, Object> toolContext) {
+public record AdvisedRequest(
+// @formatter:off
+		ChatModel chatModel,
+		String userText,
```
markpollack, line 67: may be an odd case, but maybe should allow userText to be null?

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
```diff
@@ -0,0 +1,224 @@
...
+			.map(Content::getContent)
+			.collect(Collectors.joining(System.lineSeparator()));
+
+		// 3. Define augmentation prompt parameters.
+		Map<String, Object> promptParameters = Map.of("query", query.text(), "context", documentContext);
+
+		// 4. Augment user prompt with the context data.
+		UserMessage augmentedUserMessage = (UserMessage) this.promptTemplate.createMessage(promptParameters);
+
+		return AdvisedRequest.from(request)
```
markpollack, line 154: We have an issue, #1592 related to how we want to consistently implement builders.  Does `from` go in the 'product' class or in the 'builder' class?

markpollack:
merged in 5d8c032bb706a04876687cb801f98b3f84417e50

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/api/AdvisedRequest.java
ThomasVitale: I thought a lot about a possible use case where I would want that and didn't really come up with anything. For example, OpenAI requires a message to be included in each request to a chat model. It fails if the user message is null. Before this change, we supported the user message to be null, and in that case set the "userText" to be an empty string as a workaround. That could be dangerous because as a developer I wouldn't catch scenarios where I'm not passing a user message by mistake, but the model would still be called and the outcome would be unpredictable.

PR #1605, opened 2024-10-26
Refine FactCheckingEvaluator prompt to specify yes/no answer.

habuma:
This PR addresses what I believe is a shortcoming in the existing prompt used in `FactCheckingEvaluator`.

Specifically: No matter what I provided to `FactCheckingEvaluator`, `isPass()` was always returning `false`. I did a little debugging and noticed that the evaluation response was a long-winded explanation of how the response content either aligned or didn't align with the given context. But the decision on whether the fact-checking passes was based on the much simpler expectation that the evaluation response was either "yes" or "no" (case-insensitive).

This change refines the prompt used by `FactCheckingEvaluator` to specify a yes/no answer, which seems to have resulted in good `isPass()` values in my testing (at least when evaluated against the default OpenAI GPT-4o model).


markpollack:
I've updated the class so that there are two prompt styles, one for general llms and one for bespoke mini.  I made the general llm prompt the default and it is likely the more common - though perhaps less accurate - usage.

merged in f92a3f0fcb5710d0b3e611c88da5a40d83de02c5

PR #1613, opened 2024-10-29
Addressing checkstyle

sobychacko:
Enable checkstyle on more modules and address violations.


markpollack:
I ran into a few issues I couldn't resolve with some new files since the PR.  disabled checkstyle for autconf and ollama.

Can fix in a subsequent pass.

merged in e72ab6ba250adeebec0714f6a781d82c335e93a5

PR #1621, opened 2024-10-29
Fixed an exception where the collectionName could not be found when MilvusVectorStore executed the doDelete method

leevh0908:
When MilvusVectorStore executes doDelete method, if we do not use the default database (default), DeleteParam does not pass the specified DatabaseName, in the actual query will be searched according to the default. There is no collection in the default library, and the collectionName cannot be found


sobychacko:
Merged upstream.

Issue #1628, opened 2024-10-30
Azure Vector Store is not correctly Initilazied


asaikali:
I am using Azure Vector Store with the following settings in my application.yaml 

```yaml
spring:
  ai:
    vectorstore:
      azure:
        initialize-schema: true
        index-name: spring-ai-document-index
```

Loading data into the vector store works just fine, however querying the vector store produces an exception 
```
java.lang.NullPointerException: Cannot invoke "java.util.List.size()" because "floats" is null
	at org.springframework.ai.model.EmbeddingUtils.toPrimitive(EmbeddingUtils.java:40) ~[spring-ai-core-1.0.0-20241029.213159-795.jar:1.0.0-SNAPSHOT]
	at org.springframework.ai.vectorstore.azure.AzureVectorStore.lambda$doSimilaritySearch$3(AzureVectorStore.java:327) ~[spring-ai-azure-store-1.0.0-20241029.180532-549.jar:1.0.0-SNAPSHOT]
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[na:na]
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179) ~[na:na]
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133) ~[na:na]
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1939) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[na:na]
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[na:na]
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682) ~[na:na]
	at org.springframework.ai.vectorstore.azure.AzureVectorStore.doSimilaritySearch(AzureVectorStore.java:332) ~[spring-ai-azure-store-1.0.0-20241029.180532-549.jar:1.0.0-SNAPSHOT]
[...]

PR #1630, opened 2024-10-30
Fixed Azure Vector Store initialization where the embedding field was not set as Retrievable.

zhuyuy:
In Azure Vector Store, the _embedding_ field need to set as _retrievable_.
_com.azure.search.documents.indexes.models.SearchField_ class does not have a _retrievable_ field, but the "_hidden_" field is equivalent to "_retrievable_". For details, refer to the "toJson" and "fromJson" methods.

fix https://github.com/spring-projects/spring-ai/issues/1628


sobychacko:
Closing via https://github.com/spring-projects/spring-ai/commit/382fbaca48448caeaddf78f5fc7bed83c52ed90c.

PR #1632, opened 2024-10-30
Fix JsonReader example

zzzadruga:


ilayaperumalg:
Hi @zzzadruga, thanks for the fixing the sample! LGTM, merging.

PR #1639, opened 2024-10-31
Fix Vector Database providers list

JM-Lab:
index.adoc

Added missing Vector Database providers to the list:
- Included Elasticsearch, OpenSearch, SAP Hana, and Typesense


ilayaperumalg:
@JM-Lab Thanks for fixing the list. I will add the recent additions Azure Cosmos DB and GemFire to the list as well.

PR #1641, opened 2024-10-31
Update property precedence in documentation

jitokim:
Update documentation for property override precedence in ChatModel and EmbeddingModel configurations.

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/qianfan-chat.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/qianfan-chat.adoc
@@ -13,7 +13,8 @@ Exporting an environment variable is one way to set that configuration property:
 
 [source,shell]
 ----
-export SPRING_AI_QIANFAN_API_KEY=<INSERT KEY HERE>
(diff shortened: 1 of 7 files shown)


tzolov:
Thanks @jitokim

markpollack:
The changes to not seem correct, there are dedicated configuration prefix for the embedding models, see `OpenAiEmbeddingProperties`.  What is the intention here?  Embedding can have a different URL than chat, they have separate `@ConfigurationProperties` classes.

jitokim:
> The changes to not seem correct, there are dedicated configuration prefix for the embedding models, see `OpenAiEmbeddingProperties`. What is the intention here? Embedding can have a different URL than chat, they have separate `@ConfigurationProperties` classes.

It seems to be understood as a comment on this note. 

https://github.com/spring-projects/spring-ai/blame/560315ccfc4ac823d2d7c3cc2a1983e22e58992e/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/embeddings/openai-embeddings.adoc#L110

The intention is that when using both the Embedding Model and the Chat Model together, the properties can be overridden individually.

Please let me know if there is any misunderstanding. Thank you.

jitokim:
@markpollack  
Thanks for review.
I re-read the Chat Model documentation and understood your comment. I'll revise it soon.

Issue #1294, opened 2024-08-29
QuestionAnswerAdvisor should be able to inject the document id and metadata into the prompt.


BillFarber:
QuestionAnswerAdvisor only inserts the contents of the Documents into the prompt. It would be helpful to be able to include the document id and metadata.

When inserting Documents into a PromptTemplate, the object's `.toString()` method is automatically, used which includes the id and metadata.
```java
return "Document{id='" + this.id + "', metadata=" + this.metadata + ", content='" + this.content + "', media=" + this.media + "}";
```

However, with QuestionAnswerAdvisor, it's only possible to include the content:
```java
String documentContext =(String)documents.stream().map(Content::getContent).collect(Collectors.joining(System.lineSeparator()));
```

It is frequently useful to include the Document id as well as the Document metadata in the prompt so the LLM can use that information. So, it would be helpful if there were a way to tell QAA to include that information.

For the time-being, I'm using a simple hack to make it happen:
```java
Document tempDoc = new Document(uri, text, metadata);
return new Document(uri, tempDoc.toString(), metadata);
```

ThomasVitale:
Using the new (experimental) Modular RAG features, it's possible to customise the way context is added to the final prompt in a RAG flow. The `QueryAugmenter` interface can be used to implement a custom strategy. More info: https://docs.spring.io/spring-ai/reference/api/retrieval-augmented-generation.html#_query_augmentation

You can then pass your custom implementation when building the `RetrievalAugmentationAdvisor`:

 ```
Advisor retrievalAugmentationAdvisor = RetrievalAugmentationAdvisor.builder()
        .documentRetriever(VectorStoreDocumentRetriever.builder()
                .vectorStore(vectorStore)
                .build())
        .queryAugmenter(new MyCustomQueryAugmenter())
        .build();

String answer = chatClient.prompt()
        .advisors(retrievalAugmentationAdvisor)
        .user(question)
        .call()
        .content();
```

The default `ContextualQueryAugmenter` will soon support customising the context with document IDs and metadata, so that it's not necessary to implement an entirely new `QueryAugmenter`. I'll comment down here when that's delivered.

ilayaperumalg:
Closing this issue as resolved per the PR #2321 and the suggestion above from @ThomasVitale addresses this issue.

PR #1644, opened 2024-10-31
Modular RAG: Query Augmentor

ThomasVitale:
* Add new QueryAugmentor API, a component for augmenting a user query with contextual data.
* Implement ContextualQueryAugmentor that combines the content of each document and add it to the original user prompt, with support for the scenario where the context is empty.
* Extend RetrievalAugmentationAdvisor to use the new augmentation building block.
* Introduce utility to assist in validating arguments for prompt-related operations.

Relates to gh-#1603

It also makes possible the customization reported in https://github.com/spring-projects/spring-ai/issues/1294

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
@@ -20,7 +20,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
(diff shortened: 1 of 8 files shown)


tzolov:
Thanks @ThomasVitale 
Changes look good.  
- It reminded me that we need to make the Builder methods consistent. Some use `with` prefix other don't. But it seems the Spring's convention is not to have prefix (https://github.com/spring-projects/spring-ai/issues/1592#issuecomment-2453389119) 
- After we merge this PR, we can refactor the QuedstionAnswerAdvisor as a wrapper around the RetrieverAugmentor one. Not sure if we should mark it as deprecated though.

tzolov:
Rebased and merged at bc50d28f2db64e8738a7f8b7470f4c053e6de56e

ThomasVitale:
@tzolov thanks for the review! For the builder naming convention, I followed what was used already for `ChatClient`, but indeed both strategies are used throughout the project. It would be nice to converge towards one single naming strategy. I like the Spring convention of not having the prefix.

ThomasVitale:
About the `QuestionAnswerAdvisor`, it's used a lot by developers. To protect the users from breaking changes, I would suggest keeping as-is, mark it as deprecated, and document how to migrate to `RetrievalAugmentedGeneration`. We could even provide a simple OpenRewrite recipe to automate the migration.

tzolov:
Makes sense. 
Have not used the OpenRewrite yet and curious to see how it looks like :)

PR #1646, opened 2024-11-01
refactor: Use Neo4j transactional functions.

michael-simons:
This change uses `session.executeWrite` and `session.executeRead` to interact with the database.
Those methods use build-in retries for several Neo4j specific error states.
Doing will improve the overall user experience in this case as the no other external retry mechanism should be used.
Doing so is fine, as the Neo4j vector store is not subject to Springs general transaction management.


tzolov:
Looks like an useful improvement thank you @michael-simons

tzolov:
Rebased and merged at 560315ccfc4ac823d2d7c3cc2a1983e22e58992e

michael-simons:
Thank you, @tzolov !

PR #1649, opened 2024-11-01
Fix: convert API records to POJOs and fix related tests

sobychacko:
Following removal of Spring Boot's ConstructorBinding, convert record types to regular POJOs to maintain JSON serialization. Update API classes:

- Convert FunctionTool and Function records to standard Java classes with getters/setters
- Update test assertions to use getter methods instead of record accessors
- Fix failing tests in MiniMax, OpenAi and ZhiPu API implementations

This keeps core API models independent of Spring Boot while ensuring proper serialization through standard Java beans.


markpollack:
There are failing tests related to the changes.  I suspect that the OPEN_AI_KEY may not have been set or the tests in `spring-ai-openai` weren't run before the PR was submitted.  There is an issue with converting the records to classes and the calls to the openai model fail as `response_format` is null.

Issue #802, opened 2024-05-31
AWS Bedrock models tools support


DEG-7:
Hi team, 
Could you extend the tools support to AWS Bedrock models?

Amazon Bedrock supports tool calling with the following models.
    Anthropic Claude 3 models
    Mistral AI Large
    Cohere Command R and Command R+

AWS Bedrock documentation: https://docs.aws.amazon.com/bedrock/latest/userguide/tool-use.html

Thanks for the great work !!

maxjiang153:
I'm working on this, but first one is Anthropic Claude 3 models.

Mistral AI models and Cohere Command R models have not been merged yet:
https://github.com/spring-projects/spring-ai/pull/808
https://github.com/spring-projects/spring-ai/pull/667

maxjiang153:
@tzolov plz take a look at recent AWS Bedrock PRs, I wonder if creating a PR includes all supported tool models instead of once a time.

markpollack:
Thanks, this is a large change, so will take a bit to digest.

piratovi:
@tzolov Hi guys! Any updates about Anthropic Claude 3 tools support ?

tzolov:
@piratovi we are working on it PR #1650 
The main parts hopefully will be released with the forthcoming M4 milestone.

tzolov:
Resolved by the [Amazon Bedrock Converse API](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html). 

> Following the Bedrock recommendations, Spring AI is transitioning to using Amazon Bedrock’s Converse API for all chat conversation implementations in Spring AI. While the existing [InvokeModel API](https://docs.spring.io/spring-ai/reference/api/bedrock-chat.html) supports conversation applications, we strongly recommend adopting the Converse API for all Char conversation models.
> The Converse API does not support embedding operations, so these will remain in the current API and the embedding model functionality in the existing InvokeModel API will be maintained

> Bedrock Converse provides a unified interface for conversational AI models with enhanced capabilities including tool calling, multimodal inputs, and streaming responses.

PR #1650, opened 2024-11-02
Add Bedrock Converse API chat model support

### LINKED ISSUE #809: Add support for the Amazon Bedrock Converse API
On 30 May, Amazon announced the Converse API, which is intended to create an abstraction for accessing different models. 

https://aws.amazon.com/about-aws/whats-new/2024/05/amazon-bedrock-new-converse-api/

For non-streaming responses, the [Converse API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) can be used, while for streaming, the [Converse Stream API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStream.html) can be used.

I would love to use this API with Spring AI.

tzolov:
Introduces support for Amazon Bedrock Converse API through a new BedrockProxyChatModel implementation. This enables integration with Bedrock's conversation models with features including:
- Support for sync/async chat completions
- Stream response handling
- Tool/function calling capabilities
- System message support
- Image input support
- Observation and metrics integration
- Configurable model parameters and AWS credentials

Adds core support classes:
- BedrockUsage: Implements Usage interface for token tracking
- ConverseApiUtils: Utility class for handling Bedrock API responses including:
  - Tool use event aggregation and processing
  - Chat response transformation from stream outputs
  - Model options conversion
  - Support for metadata aggregation
- URLValidator: Utility for URL validation and normalization with support for:
  - Basic and strict URL validation
  - URL normalization
  - Multimodal input handling
- Enhanced FunctionCallingOptionsBuilder with merge capabilities for both ChatOptions and FunctionCallingOptions
- Added BEDROCK_CONVERSE to AiProvider enum for metrics tracking
- Extended AWS credentials support with session token capability
- Added configurable session token property to BedrockAwsConnectionProperties

Adds new auto-configuration support:
- BedrockConverseProxyChatAutoConfiguration for automatic setup of the Bedrock Converse chat model
- BedrockConverseProxyChatProperties for configuration including:
  - Model selection (defaults to Claude 3 Sonnet)
  - Timeout settings (defaults to 5 minutes)
  - Temperature and token control
  - Top-K and Top-P sampling parameters
- Integration with existing BedrockAwsConnectionConfiguration for AWS credentials

Updates to testing infrastructure:
- Adds comprehensive test suite for Bedrock Converse properties and auto-configuration
- Integration tests for chat completion and streaming scenarios
- Property validation tests for configuration options
- Temporarily disabled other Bedrock tests due to AWS quota limitations
- Added ObjectMapper configuration for proper JSON handling

Added new spring-ai-bedrock-converse-spring-boot-starter module

Updates module configuration in parent POM and BOM to include new bedrock-converse modules and starters. Adds necessary auto-configuration imports for seamless integration with Spring Boot applications.

Unrelated changes:
- Disabled several Bedrock model tests (Jurassic2, Llama, Titan) due to AWS quota limitations
- Disabled PaLM2 tests due to API decommissioning by Google

Resolves #809, #802

Bedrock-Converse-API v2


models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
```diff
@@ -0,0 +1,710 @@
...
+			.provider(AiProvider.BEDROCK_CONVERSE.value())
+			.requestOptions(buildRequestOptions(converseRequest))
+			.build();
+
+		ChatResponse chatResponse = ChatModelObservationDocumentation.CHAT_MODEL_OPERATION
+			.observation(this.observationConvention, DEFAULT_OBSERVATION_CONVENTION, () -> observationContext,
+					this.observationRegistry)
+			.observe(() -> {
+
+				ConverseResponse converseResponse = this.bedrockRuntimeClient.converse(converseRequest);
```
maxjiang153, line 187: Should we add a retry mechanism via `RetryTemplate` in case something might happen?

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/bedrock/converse/BedrockConverseProxyChatProperties.java
```diff
@@ -0,0 +1,94 @@
...
+	private boolean enabled = true;
+
+	/**
+	 * The generative id to use. See the {@link BedrockProxyChatModel} for the supported
+	 * models.
+	 */
+	private String model = "anthropic.claude-3-5-sonnet-20240620-v1:0";
+
+	@NestedConfigurationProperty
+	private PortableFunctionCallingOptions options = PortableFunctionCallingOptions.builder()
```
maxjiang153, line 55: With the Bedrock Converse API, different models have different options such as `stop sequences` etc. I wonder if we should add something to set default options for models.

Maybe we could add a new class `BedrockConversePortableFunctionCallingOptions` with a key-value map dynamic field and convert them to Bedrock Converse API's additionalModelRequestFields?

models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
tzolov: I've noticed that the AWS SDK has a built-in retry utilities. 
But we need to expose and configure them hopefully they will support the async model too.

tzolov: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/retry-strategy.html

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/bedrock/converse/BedrockConverseProxyChatProperties.java
tzolov: Currently the PortableChatOptions and the PortableFunctionCallingOptions allow setting stopSequences and common chat and tool calling options that should work with all models. 

The BedrockConversePortableFunctionCallingOptions with a key-value map dynamic field could be easiest, dynamic,  approach to allow adding additionalModelRequestFields. 
We can add it a follow up PR. 

And in the future we can still add a type-safe options (e.g. dedicated per-model option class).

tzolov: Also we should explore if we can :
- implement additional metrics handler for the Bedrock Converse metrics.
- provide support for the guard features.
- review the cumbersome  streaming tools calling impl.

models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
```diff
@@ -0,0 +1,710 @@
...
+			.map(message -> {
+				if (message.getMessageType() == MessageType.USER) {
+					List<ContentBlock> contents = new ArrayList<>();
+					if (message instanceof UserMessage) {
+						var userMessage = (UserMessage) message;
+						contents.add(ContentBlock.fromText(userMessage.getContent()));
+
+						if (!CollectionUtils.isEmpty(userMessage.getMedia())) {
+							List<ContentBlock> mediaContent = userMessage.getMedia().stream().map(media -> {
+								ContentBlock cb = ContentBlock.fromImage(ImageBlock.builder()
```
kalcifield, line 250: Bedrock also supports documents, it would be nice to have this feature implemented too. In my use-case I solved it with a convertMediaToContentBlock method:
```
        if (IMAGE.equals(mimeType.getType())) {
            return ContentBlock.fromImage(image -> image.format(
                            media.getMimeType().getSubtype())
                    .source(ImageSource.fromBytes(SdkBytes.fromByteArray(getContentMediaData(media.getData())))));
        } else {
            String fileName = mimeType.getParameter(FILE_NAME_KEY);
            return ContentBlock.fromDocument(document -> document.format(
                            FileNameUtils.detectDocFormatByFileName(fileName))
                    .name(FilenameUtils.getBaseName(fileName))
                    .source(DocumentSource.fromBytes(SdkBytes.fromByteArray(getContentMediaData(media.getData())))));
        }
```
possibly passing the filename in the mimeType is not general enough

models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockConverseChatClientIT.java
```diff
@@ -0,0 +1,341 @@
...
+		logger.info(response);
+		assertThat(response).contains("bananas", "apple");
+		assertThat(response).containsAnyOf("bowl", "basket");
+	}
+
+	@ParameterizedTest(name = "{0} : {displayName} ")
+	@ValueSource(strings = { "anthropic.claude-3-5-sonnet-20240620-v1:0" })
+	void multiModalityImageUrl(String modelName) throws IOException {
+
+		// TODO: add url method that wrapps the checked exception.
```
markpollack, line 299: create issues for TODOs

models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/api/ConverseApiUtils.java
```diff
@@ -0,0 +1,503 @@
...
+					.withTokenUsage(metadataEvent.usage())
+					.withMetrics(metadataEvent.metrics())
+					.withTrace(metadataEvent.trace())
+					.build();
+
+				DefaultUsage usage = new DefaultUsage(metadataEvent.usage().inputTokens().longValue(),
+						metadataEvent.usage().outputTokens().longValue(),
+						metadataEvent.usage().totalTokens().longValue());
+
+				// TODO
```
markpollack, line 323: Create an issue on what 'TODO'

markpollack:
```
 * The Converse API doesn't support any embedding models (such as Titan Embeddings G1 -
 * Text) or image generation models (such as Stability AI).
 *
```

I suppose then we keep the delta of converse and the existing bedrock support in the project and deprecate and eventually remove the areas of overlap?

markpollack:
What about `BedrockConverseChatModel` instead of `BedrockProxyChatModel` and similar changes for where `Proxy` appears in the name, replace with `Converse`

tzolov:
rebased, squashed and merged at 0d2d4b7385a23ddf4987e2a144ce11c3a66ad856

PR #1651, opened 2024-11-02
Make ChatClient APIs null-safe and predictable

ThomasVitale:
* Introduced null-safety for all ChatClient APIs and verified via extensive auto-tests.
* Made the final message list computed by ChatClient consistent and predictable across the different options for providing messages (`prompt()`, `messages()`, `user()`, `system()`) and verified via extensive auto-tests.
* Added even more auto-tests to cover as many scenarios and edge cases as possible.


spring-ai-core/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
```diff
@@ -88,18 +90,44 @@ public class DefaultChatClient implements ChatClient {
...
+
 		Map<String, Object> advisorContext = new ConcurrentHashMap<>(inputRequest.getAdvisorParams());
 		if (StringUtils.hasText(formatParam)) {
 			advisorContext.put("formatParam", formatParam);
 		}
 
-		return new AdvisedRequest(inputRequest.chatModel, inputRequest.userText, inputRequest.systemText,
-				inputRequest.chatOptions, inputRequest.media, inputRequest.functionNames,
-				inputRequest.functionCallbacks, inputRequest.messages, inputRequest.userParams,
+		// Process userText, media and messages before creating the AdvisedRequest.
```
ThomasVitale, line 106: This piece of logic used to be executed only in one possible branch of execution (when using `.prompt(Prompt prompt)`). I moved it here so to ensure consistent processing of the messages before creating the `AdvisedRequest`, no matter how we get here (for example, when using `.messages(List<Message> messages)`, which was previously behaving inconsistently).

tzolov: yeh, good catch. 
Initially the ChatClinet from Prompt used to have a very different path, excluded from the advisors. 
I've tried to unify it with there unsealed path (by unsealing the prompt) but have missed the ordering of this section.

tzolov:
Very useful improvements for reliability and consistency.  Thanks @ThomasVitale

tzolov:
Rebased and merged at 048d54b82bcb3fb984aafacc5f28efd194342ab4

PR #1653, opened 2024-11-02
Simplify and remove duplicate logic in AI model observation components

jxblum:
See title.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/observation/ChatModelObservationContentProcessor.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/observation/ChatModelObservationContentProcessor.java
@@ -16,9 +16,15 @@
 
 package org.springframework.ai.chat.observation;
 
+import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
+import org.springframework.ai.chat.messages.Message;
+import org.springframework.ai.chat.model.ChatResponse;
+import org.springframework.ai.chat.model.Generation;
 import org.springframework.ai.model.Content;
+import org.springframework.lang.Nullable;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 
@@ -26,36 +32,29 @@ import org.springframework.util.StringUtils;
  * Utilities to process the prompt and completion content in observations for chat models.
  *
  * @author Thomas Vitale
+ * @author John Blum
+ * @since 1.0.0
  */
-public final class ChatModelObservationContentProcessor {
-
-	private ChatModelObservationContentProcessor() {
-	}
+public abstract class ChatModelObservationContentProcessor {
 
 	public static List<String> prompt(ChatModelObservationContext context) {
-		if (CollectionUtils.isEmpty(context.getRequest().getInstructions())) {
-			return List.of();
-		}
 
-		return context.getRequest().getInstructions().stream().map(Content::getContent).toList();
+		List<Message> instructions = context.getRequest().getInstructions();
+
+		return CollectionUtils.isEmpty(instructions) ? Collections.emptyList()
+				: instructions.stream().map(Content::getContent).toList();
 	}
 
-	public static List<String> completion(ChatModelObservationContext context) {
-		if (context == null || context.getResponse() == null || context.getResponse().getResults() == null
-				|| CollectionUtils.isEmpty(context.getResponse().getResults())) {
-			return List.of();
-		}
+	public static List<String> completion(@Nullable ChatModelObservationContext context) {
 
-		if (!StringUtils.hasText(context.getResponse().getResult().getOutput().getContent())) {
-			return List.of();
-		}
-
-		return context.getResponse()
-			.getResults()
+		return Optional.ofNullable(context)
+			.map(ChatModelObservationContext::getResponse)
+			.map(ChatResponse::getResults)
+			.orElseGet(Collections::emptyList)
 			.stream()
-			.filter(generation -> generation.getOutput() != null
-					&& StringUtils.hasText(generation.getOutput().getContent()))
-			.map(generation -> generation.getOutput().getContent())
+			.map(Generation::getOutput)
+			.map(Message::getContent)
+			.filter(StringUtils::hasText)
 			.toList();
 	}
 
--- a/spring-ai-core/src/main/java/org/springframework/ai/model/observation/ErrorLoggingObservationHandler.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/model/observation/ErrorLoggingObservationHandler.java
@@ -31,6 +31,7 @@ import org.springframework.util.Assert;
 
 /**
  * @author Christian Tzolov
+ * @author John Blum
  * @since 1.0.0
  */
 @SuppressWarnings({ "rawtypes", "null" })
@@ -62,18 +63,22 @@ public class ErrorLoggingObservationHandler implements ObservationHandler {
 	}
 
 	@Override
+	@SuppressWarnings("all")
 	public boolean supportsContext(Context context) {
-		return (context == null) ? false : this.supportedContextTypes.stream().anyMatch(clz -> clz.isInstance(context));
+		return context != null && this.supportedContextTypes.stream().anyMatch(clz -> clz.isInstance(context));
 	}
 
 	@Override
+	@SuppressWarnings("unused")
 	public void onError(Context context) {
-		if (context != null) {
-			TracingContext tracingContext = context.get(TracingContext.class);
-			if (tracingContext != null) {
-				try (var val = this.tracer.withSpan(tracingContext.getSpan())) {
-					this.errorConsumer.accept(context);
-				}
+
+		Assert.notNull(context, "Context must not be null");
+
+		TracingContext tracingContext = context.get(TracingContext.class);
+
+		if (tracingContext != null) {
+			try (var val = this.tracer.withSpan(tracingContext.getSpan())) {
+				this.errorConsumer.accept(context);
 			}
 		}
 	}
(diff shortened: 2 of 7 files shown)


spring-ai-core/src/main/java/org/springframework/ai/model/observation/ErrorLoggingObservationHandler.java
```diff
@@ -62,12 +64,13 @@ public ErrorLoggingObservationHandler(Tracer tracer,
 	}
 
 	@Override
-	public boolean supportsContext(Context context) {
-		return (context == null) ? false : this.supportedContextTypes.stream().anyMatch(clz -> clz.isInstance(context));
+	public boolean supportsContext(@Nullable Context context) {
```
ThomasVitale, line 67: The `Context` is never null as per the Micrometer infrastructure. Is there a reason for making it nullable?

spring-ai-core/src/main/java/org/springframework/ai/model/observation/ErrorLoggingObservationHandler.java
```diff
@@ -62,12 +64,13 @@ public ErrorLoggingObservationHandler(Tracer tracer,
...
+		return context != null && this.supportedContextTypes.stream().anyMatch(clz -> clz.isInstance(context));
 	}
 
 	@Override
-	public void onError(Context context) {
+	@SuppressWarnings("unused")
+	public void onError(@Nullable Context context) {
```
ThomasVitale, line 73: Same thing here, the `Context` is never null as per the Micrometer infrastructure.

ThomasVitale:
General comment about the added `@Nullable` annotations in the observation classes. The `Observation.Context` passed from the Micrometer infrastructure is never null, it's actually marked as non-null upstream. Is there a reason for making it nullable?

spring-ai-core/src/main/java/org/springframework/ai/model/observation/ErrorLoggingObservationHandler.java
jxblum: I am going to remove the @Nullable annotation so it generates the appropriate warnings in the supported IDEs.

spring-ai-core/src/main/java/org/springframework/ai/model/observation/ErrorLoggingObservationHandler.java
jxblum: Same comment as above.

jxblum:
> General comment about the added `@Nullable` annotations in the observation classes. The `Observation.Context` passed from the Micrometer infrastructure is never null, it's actually marked as non-null upstream. Is there a reason for making it nullable?

Well, [this](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/model/observation/ErrorLoggingObservationHandler.java#L65) and [this](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/model/observation/ErrorLoggingObservationHandler.java#L70), contradicts the contract.

I would argue that is never a good think for the code to throw a NPE, especially in boolean method that should simply return `true` or `false` the majority of the time.

So, I think the boolean method should be "null-safe" and the onError(:Context) method should have an appropriate validation.

In both cases, I have made changes.

ThomasVitale:
> Well, [this](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/model/observation/ErrorLoggingObservationHandler.java#L65) and [this](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/model/observation/ErrorLoggingObservationHandler.java#L70), contradicts the contract.

Thanks for mentioning those two points. I didn't actually notice we were performing a null check there. It might be the only Micrometer API-related implementation where we do that, because all other handlers, predicates, and filters I know of don't do that. I would say those two checks shouldn't be there since the interface is marked with `@NonNullApi`, establishing the input `Observation.Context` to be non-null. If we allow null values, we are changing the behaviour of the API.

> I would argue that it is never good for any code to throw a NPE, especially in boolean methods that should simply return `true` or `false` the majority of the time.

Totally agree with you. But then we should add a check to verify the compliance with the contract, which is a non-null contract, rather than allowing null. And that check is indeed missing. We could add the following. What do you think?

`Assert.notNull(context, "context cannot be null);`

Micrometer itself ensures that handlers, predicates, and filters are called with a non-null context, but if at some point a null value is passed, then we get an exception thrown since that should never happen and we entered a faulty state of the application.

spring-ai-core/src/main/java/org/springframework/ai/chat/observation/ChatModelObservationContentProcessor.java
```diff
@@ -16,46 +16,45 @@
...
  *
  * @author Thomas Vitale
+ * @author John Blum
+ * @since 1.0.0
  */
-public final class ChatModelObservationContentProcessor {
-
-	private ChatModelObservationContentProcessor() {
-	}
+public abstract class ChatModelObservationContentProcessor {
```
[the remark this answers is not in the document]
jxblum: Great question!

For extensibility purposes, I often prefer classes marked as `abstract`, rather than `final` with a `private` constructor, even if they mostly, or entirely, contain static methods (e.g. utility classes).

This prevents the class from being immediately instantiated (without extra effort) and allows the class to be extended with another abstract (or concrete) class that may be context-specific. Additionally this helps to reduce the responsibility and footprint of the parent class.

For example, [`FileSystemUtils`](https://github.com/codeprimate-software/cp-elements/blob/v2.0.0-M5/src/main/java/org/cp/elements/io/FileSystemUtils.java#L44) to [`FileUtils`](https://github.com/codeprimate-software/cp-elements/blob/v2.0.0-M5/src/main/java/org/cp/elements/io/FileUtils.java#L61) to [`IOUtils`](https://github.com/codeprimate-software/cp-elements/blob/v2.0.0-M5/src/main/java/org/cp/elements/io/IOUtils.java#L49) (extending nothing).

Of course this is debatable and so I am not a huge stickler on this.

I believe one of Spring's primary strengths is its excellent implementation of the _Open/Closed principle_. This upholds Spring's principle of choice and gives developers options OOTB if the base functionality does not quite meet their needs.

In fact, in this particular case, I'd probably even prefer that the `ChatModelObservationContentProcessor` class not contain `static` methods, and that instance could even be treated as a proper bean.

Again, I am flexible here.

jxblum:
> But then we should add a check to verify the compliance with the contract, which is a non-null contract, rather than allowing null. And that check is indeed missing. We could add the following. What do you think?

Agreed. And, I already [changed the PR](https://github.com/spring-projects/spring-ai/pull/1653/files#diff-b475d66ccfd7f82671c8ef85feadc98df8a5802a1ebe42d618f0c1999aa9eef9R75) to do exactly what you suggested.

jxblum:
Hi @tzolov & @ThomasVitale (CC'ing @jonatan-ivanov),

I did have a question (or 2) about Spring AI's Observability story. 

While my questions focus on `VectorStore`, it is equally applicable to chat as well as other observable bits in Spring AI.

1) I am curious how it was decided what was considered a _high-cardinality key_ vs. _low-cardinality key_?

For instance, in `VectorStoreObservationDocumentation`, we see [_low-cardinality keys_](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java#L54-L90) and [_high-cardinality keys_](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java#L92-L205) defined. I understand and can reason about the _low-cardinality keys_, although, `DB_OPERATION_NAME` seems debatable to me, which could be quite "high", depending on the application use-cases.

However, some of the _high-cardinality keys_ do not seem as though they would have an "unbounded range of values", for example: `[DB_NAMESPACE, DB_VECTOR_FIELD_NAME, DB_SEARCH_SIMILARITY_METRIC]`. Whereas, other keys would definitely have an unbounded range, for example: `[DB_VECTOR_QUERY_CONTENT, DB_VECTOR_QUERY_FILTER, DB_VECTOR_QUERY_RESPONSE_DOCUMENTS, DB_VECTOR_QUERY_SIMILARITY_THRESHOLD, DB_VECTOR_QUERY_TOP_K]`, which naturally [vary by request](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/SearchRequest.java#L47-L53)).

I even debated about `DB_COLLECTION_NAME`, but I can see that the "collection" (e.g. table) could vary quite a lot given a properly organized (or normalized) data store.

On the other hand, the `DB_SEARCH_SIMILARITY_METRIC` is definitely bounded, presently to a pre-defined [enumeration](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/observation/conventions/VectorStoreSimilarityMetric.java#L33-L51).

2) This is probably more of a **Micrometer** question, but does the [KeyNames](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationDocumentation.java#L50) and [KeyValues](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/observation/DefaultVectorStoreObservationConvention.java#L79-L87) need to match for each of the _high-cardinality_ and _low-cardinality keys_?

This seems logical to me that they nicely align in this way.

> NOTE: I only referred to the _high-cardinality keys_ and _value_ in the references above.

markpollack:
similar to the other PR, the ocde base has changed quite alot.  need to close.  sorry it didn't work out.

PR #1654, opened 2024-11-03
Support case with missing system text in VectorStoreChatMemoryAdvisor

ThomasVitale:
The system text advise prompt is only joined with the value of the system text if it's non-null and non-empty.


tzolov:
Yep, the PgVectorStoreWithChatMemoryAdvisorIT caught this  :)
Thanks Thomas

Issue #1622, opened 2024-10-29
App fails to boot on Cloud Foundry when using Spring Cloud Services Service Discovery


odedia:
When running an application with both Spring AI and `spring-cloud-services-starter-config-client` as dependencies, the application fails to boot with the error below:

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'retryTemplate' defined in class path resource [org/springframework/ai/autoconfigure/retry/SpringAiRetryAutoConfiguration.class]: Failed to instantiate [org.springframework.retry.support.RetryTemplate]: Factory method 'retryTemplate' threw exception with message: org/springframework/ai/retry/TransientAiException
	at org.springframework.beans.factory.support.ConstructorResolver.instantiate(ConstructorResolver.java:648)
	at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:636)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1355)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1185)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:562)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:522)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:337)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:234)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:335)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:200)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:975)
	at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:971)
[...]

PR #1657, opened 2024-11-03
Fixed the incorrect @ConditionalOnClass condition in SpringAiRetryAutoConfiguration

zhuyuy:
fix https://github.com/spring-projects/spring-ai/issues/1622

### DIFF
--- a/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/retry/SpringAiRetryAutoConfiguration.java
+++ b/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/retry/SpringAiRetryAutoConfiguration.java
@@ -43,7 +43,7 @@ import org.springframework.web.client.ResponseErrorHandler;
  * @author Christian Tzolov
  */
 @AutoConfiguration
-@ConditionalOnClass(RetryTemplate.class)
+@ConditionalOnClass(TransientAiException.class)
 @EnableConfigurationProperties({ SpringAiRetryProperties.class })
 public class SpringAiRetryAutoConfiguration {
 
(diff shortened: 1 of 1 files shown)


spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/retry/SpringAiRetryAutoConfiguration.java
```diff
@@ -43,7 +43,7 @@
  * @author Christian Tzolov
  */
 @AutoConfiguration
-@ConditionalOnClass(RetryTemplate.class)
+@ConditionalOnClass(TransientAiException.class)
```
jitokim, line 46: The condition for retryTemplate should be maintained. Therefore, I suggest the following

```suggestion
@ConditionalOnClass({ RetryTemplate.class, TransientAiException.class })
```

jitokim, line 46: Wasn't the intention of author to use RetryUtils instead of RetryTemplate? 
The names are similar, and I got confused while reviewing the PR as well 😃

jitokim (approved):

jitokim (approved):
👍

markpollack:
This condition will always be true.  In the issue described it looks to be a matter of providing the spring retry library into the project.  It is defined in `spring-ai-openai` module

```
		<dependency>
			<artifactId>spring-ai-retry</artifactId>
			<version>${project.parent.version}</version>
		</dependency>
```

in short, I don't think PR is solving the problem.

jitokim:
In the issue https://github.com/spring-projects/spring-ai/issues/1622 , azure-openai model is used, but it has not retry module. 
(it is same in some other modules)

https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-azure-openai/pom.xml

zhuyuy:
Yes, the issue uses `spring-ai-azure-openai`, not `spring-ai-openai`, it hasn't `spring-ai-retry`.
Expected `@ConditionalOnClass` is false, but actually is true, because the condition is `org.springframework.retry.support.RetryTemplate`, a class in `spring-retry`, not `spring-ai-retry`.
I believe that this condition is indeed unreasonable. PR is fine

markpollack:
Thanks for the clarification.  indeed, for those implementations that don't use Spring Retry (as they already provide it in the SDK or we just haven't gotten to it yet) it would pose a problem.  Testing for a fix in `spring-ai-retry` is the way to go.

markpollack:
merged in 99d9864972c0fd69951dafd66cdfade70333ed5f

PR #1663, opened 2024-11-04
Simplify the implementation of ChatClient-based interfaces

jxblum:
Provide sensible defaults for overloaded methods in PromptUserSpec, PromptSystemSpec, AdvisorSpec, CallResponseSpec, ChatClientRequestSpec, and Builder interfaces.


markpollack:
I'd like to keep the interfaces without any defaults as in WebClient and other spring projects.  This PR also introduces a conflict in that the single implementation of the existing spec classes have different behavior it seems at first glance.  I took a look at `PromptUserSpec` for example.

```
@Override
public PromptUserSpec text(String text) {
	Assert.hasText(text, "text cannot be null or empty");
	this.text = text;
	return this;
}
```

vs.
```
default PromptUserSpec text(String text) {
	Charset defaultCharset = Charset.defaultCharset();
	return text(new ByteArrayResource(text.getBytes(defaultCharset)), defaultCharset);
}

What is wrong with using the string as supplied?  In anycase, these impls are used in one place, so if there are changes to make in the impl, i'd do it there.

```

jxblum:
Keeping the interfaces inside `ChatClient` (and other areas of Spring AI) void of implementation (i.e. without defaults) is reasonable. I actually considered this before submitting the PR and wondered whether I might add abstract base classes instead (Java's own Collection framework is a really good example of providing many abstract base classes supporting implementations.

I tend to provide defaults when:

1) The default implementation is very simple to implement (only a few lines)
2) The functionality can be easily and reliably implemented in terms of another method
3) Reduces effort and promotes consistency across implementations
4) Enables the use of functional interfaces (aka `@FunctionalInterface`) as _Lambda_ expressions 

> NOTER: #4 is not as applicable in this case since the `ChatClient` interfaces provides several related but distinct methods. For example, in the `PromptUserSpec` interface (and similar) there is the `text(:Resource, :Charset)` ([here](https://github.com/spring-projects/spring-ai/pull/1663/files#diff-e66d3344a9081dedb084e05cfbbcc20901d79e1caecae67d186e2bebcc3d7fa1R114)) and `params(:Map)` methods... related but distinct ops with a couple of "overloaded" variants that do not differ much in implementation and therefore can follow the most complete method signature.


Regarding...

> _What is wrong with using the string as supplied i the current impl?_

Nothing, other than using the given `String` directly is an implementation detail.

Any implementation is free to still override a default implementation, necessarily based on internal data structure, performance, or all sorts of reasons. But the key is, implementations do not need to override and explicitly implement every overloaded variant, keeping the implementations focused on the key details that matter.

A lot of the methods in the `ChatClient` interfaces are for "convenience". They are not functionality independent.

Finally, Spring is an excellent example of the _Open/Close_ principle, and making the task of implementation by downstream frameworks or libraries in addition to application's occasional need for extension simpler is generally a good thing.

Food for thought.

jxblum:
FYI, in the process of writing my book, I have thought of many extensions and areas for improvement in Spring AI, while considering all angles, other: libraries, frameworks, applications and tooling implementations.

jxblum:
@markpollack 

I should probably add that, in several cases, Spring AI does use default methods on interfaces.

For example, [`ChatModel`](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/chat/model/ChatModel.java#L31-L41) and [`StreamingChatModel`](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/chat/model/StreamingChatModel.java#L30-L42) both use default methods, as does [`ChatMemory`](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/chat/memory/ChatMemory.java#L35-L37).

In fact, this is quite common through Spring AI's codebase.

* [`StructuredOutputConverter`](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/converter/StructuredOutputConverter.java#L33-L38)
* [`DocumentReader`](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/document/DocumentReader.java#L24-L26) & [`DocumentWriter`](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/document/DocumentWriter.java#L27-L29)
* [`Evaluator`](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/evaluation/Evaluator.java#L30-L36)
* And on and on...

jxblum:
1 last comment...

Declaring default methods in interfaces does not change the default implementation/behavior already provided by Spring AI in the DefaultXyz classes, for example `DefaultChatClient`.

In particular, the `text(:String):PromptUserSpec` method ([source](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java#L218-L223)) that you referred to [above](https://github.com/spring-projects/spring-ai/pull/1663#issuecomment-2460847374) is still used over the [default method](https://github.com/spring-projects/spring-ai/pull/1663/files#diff-e66d3344a9081dedb084e05cfbbcc20901d79e1caecae67d186e2bebcc3d7fa1R132-R134) provided in the interface.

Finally, the `ChatClientBuilder` used to build and construct a `ChatClient` will still [build](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/chat/client/DefaultChatClientBuilder.java#L71) a `DefaultChatClient`, for which all the "default" [sub-interface implementations](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java#L185-L1056) are maintained.

So, this PR changes nothing, and only provides convenience.

PR #1666, opened 2024-11-04
Add support for Kotlin functions

sdeleuze:
This PR adds:
 - Support for Kotlin functions
 - A Kotlin extension to `FunctionCallbackWrapper.Builder`
 - Kotlin documentation for language specifics

It also removes the dependency on Spring Cloud Function and `net.jodah:typetools` which are replaced by leveraging
Spring Framework `RootBeanDefinition` and `ResolvableType` capabilities.

### DIFF
--- a/spring-ai-core/pom.xml
+++ b/spring-ai-core/pom.xml
@@ -54,18 +54,6 @@
 			<version>${jsonschema.version}</version>
 		</dependency>
 
-		<dependency>
-			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-function-context</artifactId>
-			<version>${spring-cloud-function-context.version}</version>
-			<exclusions>
-				<exclusion>
-					<groupId>org.springframework.boot</groupId>
-					<artifactId>spring-boot-autoconfigure</artifactId>
-				</exclusion>
-			</exclusions>
-		</dependency>
-
 		<!-- production dependencies -->
 		<dependency>
 			<groupId>org.antlr</groupId>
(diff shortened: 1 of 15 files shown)


sdeleuze:
@markpollack Forced-pushed to remove `RootBeanDefinition` cast which was not needed and rebased to fix a merge-conflict.

markpollack:
merged in 9cf66333b5f0e36753f1a74931a59b8ace68bdec

sdeleuze:
Reopening the PR after finding a regression and reverting related commit from `main`.

sdeleuze:
I spent some time debugging the initial version of this PR that we had to revert after a regression detailed in #1685 and reproducible with @ThomasVitale [repro](https://github.com/spring-projects/spring-ai/issues/1685#issuecomment-2460183050).

The reason why our tests did not catch this regression is not straightforward. In our tests, `ConfigurationClassParser` is creating `AnnotationMetadata` via reflection (`StandardAnnotationMetadata`) and with that, `beanDefinition.getResolvableType()` works as expected because the function type is resolved via `RootBeanDefinition#factoryMethodToIntrospect` but in the application that reproduces the issue, the `AnnotationMetadata` is a `SimpleAnnotationMetadata` created via ASM where `beanDefinition.getResolvableType()` is unable to resolve the type.

Generally speaking, `BeanDefinition#getResolvableType()` is a sort of best-effort representation of the currently known time information at runtime, so I have added a fallback resolution using the factory method, duplicating (for now) `ConstructorResolver#resolveFactoryMethodIfPossible` logic which is not exposed in the public API. We may expose that in a future version of Spring Framework.

I have added related tests using tests that involve scanning to trigger the ASM code path. Both tests and @ThomasVitale repro are green with that new version I forced pushed, and Open AI IT tests seems to be fine as well.

There still may be some small differences with original SCF algorithm, but I guess we could refine with proper test if something is not supported yet.

sdeleuze:
Notice that suspending functions are not yet supported a discussed in [that comment](https://github.com/spring-projects/spring-ai/issues/1667#issuecomment-2464283753).

tzolov:
rebased, squashed and merged at c3c95a82c16e0e80f9cc34973801a4eb2298a0bd

PR #1670, opened 2024-11-04
Fix TokenCountBatchingStrategy issues.

1993heqiang:
Fix two issues:
1.The constructor parameter thresholdFactor should be changed to reservePercentage. 
2.All input parameters need to be verified.

If this PR is accepted, please help modify the corresponding documents. 😆


markpollack:
thanks, merged in 0825d52224dfb3091c162d9d266c4321dcc9e9f4

PR #1673, opened 2024-11-05
Bugfix milvus similaritySearch Filter Expression condition NIN

leevh0908:
Fix MilvusVectorStore: When executing the similaritySearch method and adding the Filter Expression condition with ExpressionType set to NIN, the concatenated Milvus syntax is incorrect. The current error is "nin", and the correct Milvus syntax is "not in"


markpollack:
Thanks!  https://milvus.io/docs/boolean.md is the ref.  I updated the test and also added some others for operators not covered.

merged in 5235696160da0519b09cc89a97287a707f5e9e17

PR #1683, opened 2024-11-06
Add tests for keyless auth to Azure Search vector store

mattgotteiner:
* Add test validating [DefaultAzureCredential](https://learn.microsoft.com/en-us/java/api/com.azure.identity.defaultazurecredential?view=azure-java-stable) can be used with the Azure Search vector store
* Fix filter tests on Windows due to differences in checked-in newlines vs. platform-specific newline.
* Adds dependencies on azure-core, azure-identity, and azure-json packages

### DIFF
--- a/vector-stores/spring-ai-azure-store/src/test/java/org/springframework/ai/vectorstore/azure/AzureVectorStoreIT.java
+++ b/vector-stores/spring-ai-azure-store/src/test/java/org/springframework/ai/vectorstore/azure/AzureVectorStoreIT.java
@@ -25,7 +25,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Stream;
 
 import com.azure.core.credential.AzureKeyCredential;
 import com.azure.identity.DefaultAzureCredentialBuilder;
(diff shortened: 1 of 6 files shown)


vector-stores/spring-ai-azure-store/src/test/java/org/springframework/ai/vectorstore/azure/AzureVectorStoreIT.java
```diff
@@ -300,6 +301,14 @@ public static class Config {
 
 		@Bean
 		public SearchIndexClient searchIndexClient() {
+			// Only set AZURE_AI_SEARCH_TEST_KEYLESS if role-based authentication is set up correctly on the integration service
+			// https://learn.microsoft.com/azure/search/search-security-rbac
+			if (System.getenv("AZURE_AI_SEARCH_TEST_KEYLESS").equals("true")) {
```
markpollack, line 306: I believe we should add some sort of `useKeyless` or similar name to `AzureVectorStoreProperties` rather than hard coding a search for an environment variable.  Boot will be able to support getting this value from an env-var as well as lots of other places.

markpollack:
much appreciated!  See comment above.

mattgotteiner:
thanks, i'll check it

mattgotteiner:
> thanks, i'll check it

LGTM

PR #1688, opened 2024-11-07
Add spring-ai-coherence-store to bom

eddumelendez:


ilayaperumalg:
@eddumelendez Thanks for the PR. We also need to add the coherence-store-boot-starter to the BOM as well. Will take care of it when merging.

eddumelendez:
Hi @ilayaperumalg, I checked [vector store autoconfigurations](https://github.com/spring-projects/spring-ai/tree/main/spring-ai-spring-boot-autoconfigure/src/test/java/org/springframework/ai/autoconfigure/vectorstore) and couldn't find any for Coherence. So, that's the reason I missed the starter.

PR #1693, opened 2024-11-07
Improve Ollama test container management

markpollack:
This change simplifies how we manage Ollama containers in tests by moving from manual toggles to environment variables for better control. Instead of scattered container configuration, we now have:

- OLLAMA_WITH_REUSE: Toggle reuse of existing containers between tests
- OLLAMA_TESTS_ENABLED: Control test execution globally

The motivation is to make tests more reliable and easier to maintain. Previously, developers had to modify code to run tests locally vs CI. Now they can control this via environment variables.

We also introduce thread-safe API access and consistent default settings across all test classes, removing duplicated configuration and potential resource leaks.

This makes the test infrastructure more maintainable and provides clearer separation between local development and CI environments.


sobychacko:
Closed via https://github.com/spring-projects/spring-ai/commit/21fc65346469ca479feabbd3dab0a456c4bebf7f.

Issue #1335, opened 2024-09-09
BeanOutputConverter.getJsonSchema() should keep the Key ordering of Record


Zhennan-Xu:
**Expected Behavior**

Base on the OpenAI document, when using Structured Outputs, outputs will be produced in the same order as the ordering of keys in the json schema. Thus the BeanOutputConverter.getJsonSchema() should keep the key ordering of the Record class to enforce the key order, which would provide benefit of [Chain of thought](https://platform.openai.com/docs/guides/structured-outputs/chain-of-thought)

**Current Behavior**

The order of key is not preserved, meaning that sometimes it will generate result first, then reasoning. For example:

```
record MathReasoning(@JsonProperty(required = true, value = "steps") Steps steps,
                            @JsonProperty(required = true, value = "final_answer") String finalAnswer) {

    record Steps(@JsonProperty(required = true, value = "items") Items[] items) {

        record Items(@JsonProperty(required = true, value = "explanation") String explanation,
                     @JsonProperty(required = true, value = "output") String output) {
        }
    }
}
```
might produce result 
```
{
        "final_answer": "x = -3.75",
        "steps": {
            "items": [
                {
                    "explanation": "Start with the original equation: 8x + 7 = -23",
                    "output": "8x + 7 = -23"
                },
                {
                    "explanation": "Subtract 7 from both sides to isolate the term with x.",
                    "output": "8x = -23 - 7"
                },
                {
                    "explanation": "Calculate -23 - 7 which equals -30.",
                    "output": "8x = -30"
                },
                {
                    "explanation": "Divide both sides by 8 to solve for x.",
                    "output": "x = -30 / 8"
                },
                {
                    "explanation": "Simplify -30 / 8 to get the final answer.",
                    "output": "x = -3.75"
                }
            ]
        }
    }
```
which is final_answer showing first.

**Context**

In order to have better inference performance, it should be steps show up first, then final_answer. Using string directly for jsonSchema doesn't have the issue, but BeanOutputConverter can't keep the original key order, making it only useful on cases where key order doesn't matter.

christophostertag:
Our current work-around is to change the naming of the variables to alphabetical. Works, but inconsistent, ugly and one more argument in the team to use Python. Please fix so we can continue using Spring-AI.

markpollack:
Will investigate, thanks for reporting.

PR #1695, opened 2024-11-07
GH-1335: Add JSON schema property order support

sobychacko:
Fixes: #1335

https://github.com/spring-projects/spring-ai/issues/1335

Add support for maintaining JSON property order in generated schemas using @JsonPropertyOrder. Users can now control the order of properties in their JSON schemas by annotating their classes/records with @JsonPropertyOrder.

- Add JacksonOption.RESPECT_JSONPROPERTY_ORDER to BeanOutputConverter
- Add test to verify schema property ordering
- Update documentation with property ordering example

### DIFF
--- a/spring-ai-core/src/test/java/org/springframework/ai/converter/BeanOutputConverterTest.java
+++ b/spring-ai-core/src/test/java/org/springframework/ai/converter/BeanOutputConverterTest.java
@@ -17,11 +17,14 @@
 package org.springframework.ai.converter;
 
 import java.time.LocalDate;
+import java.util.ArrayList;
 import java.util.List;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.databind.DeserializationFeature;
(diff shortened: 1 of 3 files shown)


spring-ai-core/src/test/java/org/springframework/ai/converter/BeanOutputConverterTest.java
```diff
@@ -112,6 +116,15 @@ String getSomeString() {
 
 	}
 
+	@JsonPropertyOrder({ "string_property", "foo_property", "bar_property" })
+	record TestClassWithJsonPropertyOrder(
+			@JsonProperty("string_property") @JsonPropertyDescription("string_property_description") String someString,
+
+			@JsonProperty(required = true, value = "foo_property") String foo,
```
ilayaperumalg, line 123: Just as a validation, we could perhaps change the order here to make sure the actual order from "JsonPropertyOrder" dictates the ordering. will change it when merging.

ilayaperumalg:
The changes LGTM. We also can add/update test at the OpenAiChatModelResponseFormatIT as quoted from the issue. Will update this when merging.

ilayaperumalg:
Rebased and merged as 79266be9700daeb894e265b1dae17e1257194769 along with the tests update c1fc687730eaeeed2e4727bbab8b22f27a96dbb6

Issue #873, opened 2024-06-16
[ChatClient] Inconsistent handling of system messages


ThomasVitale:
**Bug description**
The `ChatClient` API makes it possible to pass system and user messages via the `system()` and `user()` clauses.
In that case, the final `List<Message>` passed to the `Prompt` contains the following:

* `SystemMessage` built from `system()`
* `UserMessage` built from `user()`

The API also allows passing a list of messages directly via `messages()`.
In that case, the final `List<Message>` passed to the `Prompt` contains the following:

* `List<Message>` built from `messages()`
* `SystemMessage` built from `system()`
* `UserMessage` built from `user()`.

There is inconsistency in the two scenarios about where will the `SystemMessage` built from `system()` end up in the chat history.

Now, imagine using the few-shot prompting strategy and using the `messages()` clause to pass the few-shot examples (list of `UserMessage` and `AssistantMessage` pairs). In that case, the `SystemMessage` built from `system()` ends up at the bottom of the list, which makes the few-shot prompting strategy not working in many cases due to the wrong position of the `SystemMessage`.

A possible workaround is to pass the desired `SystemMessage` via `messages()` together with the few-shot examples, perhaps even failing the ChatClient call request if both `messages()` and `system()`+`user()` are defined, but that would limit the convenience of the API.

**Environment**
* Spring AI 1.0.0-SNAPSHOT
* Java 22

**Steps to reproduce**

Single system message:

```
var content = chatClient.prompt()
				.system("System text")
				.messages(
						new UserMessage("My question"),
						new AssistantMessage("Your answer")
				)
				.call().content();
```

Multiple system messages:

```
var content = chatClient.prompt()
				.system("System text")
				.messages(
						new SystemMessage("Historical system text"),
						new UserMessage("My question"),
						new AssistantMessage("Your answer")
				)
				.call().content();
```

**Expected behavior**

I would expect the use of `system()` to result in a `SystemMessage` always placed on the top of the message list, unless another one already exists passed via `messages()`.

When `messages()` is used and it does NOT include any `SystemMessage`, I expect the following `List<Message>` passed to the `Prompt`:

* `SystemMessage` built from `system()`
* `List<Message>` built from `messages()`
* `UserMessage` built from `user()`.

When `messages()` is used and it DOES include any `SystemMessage`, I expect the following `List<Message>` passed to the `Prompt`:

* `List<Message>` built from `messages()` (including one or more `SystemMessage`)
* `SystemMessage` built from `system()`
[...]

PR #1697, opened 2024-11-08
Update AdvisedRequest.java

StudiousXiaoYu:
systemMessage should come first index


ThomasVitale:
Thanks for your PR! We have an issue tracking how to improve the way system messages are handled in ChatClient: https://github.com/spring-projects/spring-ai/issues/873 I would suggest having a conversation there before making any change to make sure all use cases that have been identified are covered.

PR #1700, opened 2024-11-08
Fix Ollama Kotlin function callback integration tests

sobychacko:
- Add default values to KotlinRequest data class to fix Jackson deserialization
- Replace hardcoded model name with OllamaModel.LLAMA3_2 for consistency
- Extract common test setup into companion objects for Ollama tests
- Replace var with val using when expression in weather service
- Clean up imports and remove unused annotations
- Standardize property access using Kotlin conventions


markpollack:
merged in 5e1f681da25ee4895b3ef5749698abbdc1da8538

PR #1701, opened 2024-11-09
Optimize PromptTemplate constructor.

1993heqiang:
Polishing: Extract repeated code into a method to make the code more concise.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java
@@ -19,6 +19,7 @@ package org.springframework.ai.chat.prompt;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.Charset;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


markpollack:
I would chain the constructors instead and not call out to a utility method.

1993heqiang:
> I would chain the constructors instead and not call out to a utility method.

The chained constructor must be first statement in constructor body.

1993heqiang:
> I would chain the constructors instead and not call out to a utility method.

I have modified it according to your suggestion. Please help me review it again.

1993heqiang:
@markpollack 
I saw a test case is disabled, do you think this modification can solve the problem?

**Test Case:**
```java
	@Disabled("Need to improve PromptTemplate to better handle Resource toString and tracking with 'dynamicModel' for underlying StringTemplate")
	@Test
	public void testRenderResourceAsValue() throws Exception {
		...
	}
```

**Fix code:**
```java
	public PromptTemplate(String template, Map<String, Object> model) {
		Assert.notNull(template, "template must not be null");
		Assert.notNull(model, "model must not be null");
		this.template = template;
		// If the template string is not valid, an exception will be thrown
		try {
			STGroup stGroup = new STGroup('{', '}');
			// Register a custom renderer for Resource objects
			stGroup.registerRenderer(Resource.class, (value, formatString, locale) -> readContentFromResource(value));
			this.st = new ST(stGroup, this.template);
			for (Entry<String, Object> entry : model.entrySet()) {
				add(entry.getKey(), entry.getValue());
			}
		}
		catch (Exception ex) {
			throw new IllegalArgumentException("The template string is not valid.", ex);
		}
	}

	private static String readContentFromResource(Resource resource) {
		Assert.notNull(resource, "resource must not be null");
		try {
			return resource.getContentAsString(Charset.defaultCharset());
		}
		catch (IOException ex) {
			throw new RuntimeException("Failed to read resource", ex);
		}
	}
```

1993heqiang:
@markpollack @ilayaperumalg @sobychacko This PR has been open for a long time. If you think this is not an issue, I will close this PR.

PR #1703, opened 2024-11-09
Modular RAG - Query Analysis

ThomasVitale:
Query Analysis
* Introduce Query Analysis Module
* Define QueryTransformer API and TranslationQueryTransformer implementation
* Define QueryExpander API and MultiQueryExpander implementation
* Support QueryTransformer in RetrievalAugmentationAdvisor (support for QueryExpander will be in the next PR together with the needed DocumentFuser API).

Improvements
* Refine Retrieval and Augmentation Modules for increased robustness
* Expand test coverage for both modules
* Define clone() method for ChatClient.Builder

Tests
* Introduce “spring-ai-integration-tests” for full-fledged integration tests
* Add integration tests for RAG modules
* Add integration tests for RAG advisor

Relates to #gh-1603

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/DefaultChatClientBuilder.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/DefaultChatClientBuilder.java
@@ -71,6 +71,10 @@ public class DefaultChatClientBuilder implements Builder {
 		return new DefaultChatClient(this.defaultRequest);
 	}
 
+	public Builder clone() {
+		return this.defaultRequest.mutate();
+	}
+
 	public Builder defaultAdvisors(Advisor... advisors) {
 		this.defaultRequest.advisors(advisors);
 		return this;
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
@@ -16,11 +16,14 @@
 
 package org.springframework.ai.chat.client.advisor;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Predicate;
 
+import org.springframework.ai.rag.analysis.query.transformation.QueryTransformer;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
@@ -37,13 +40,13 @@ import org.springframework.ai.document.Document;
 import org.springframework.ai.rag.Query;
 import org.springframework.ai.rag.augmentation.ContextualQueryAugmentor;
 import org.springframework.ai.rag.augmentation.QueryAugmentor;
-import org.springframework.ai.rag.retrieval.source.DocumentRetriever;
+import org.springframework.ai.rag.retrieval.search.DocumentRetriever;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 /**
- * This advisor implements common Retrieval Augmented Generation (RAG) flows using the
+ * Advisor that implements common Retrieval Augmented Generation (RAG) flows using the
  * building blocks defined in the {@link org.springframework.ai.rag} package and following
  * the Modular RAG Architecture.
  * <p>
@@ -55,9 +58,11 @@ import org.springframework.util.StringUtils;
  * @see <a href="http://export.arxiv.org/abs/2407.21059">arXiv:2407.21059</a>
  * @see <a href="https://export.arxiv.org/abs/2312.10997">arXiv:2312.10997</a>
  */
-public class RetrievalAugmentationAdvisor implements CallAroundAdvisor, StreamAroundAdvisor {
+public final class RetrievalAugmentationAdvisor implements CallAroundAdvisor, StreamAroundAdvisor {
 
 	public static final String DOCUMENT_CONTEXT = "rag_document_context";
+
+	private final List<QueryTransformer> queryTransformers;
 
 	private final DocumentRetriever documentRetriever;
 
@@ -67,9 +72,12 @@ public class RetrievalAugmentationAdvisor implements CallAroundAdvisor, StreamAr
 
 	private final int order;
 
-	public RetrievalAugmentationAdvisor(DocumentRetriever documentRetriever, @Nullable QueryAugmentor queryAugmentor,
-			@Nullable Boolean protectFromBlocking, @Nullable Integer order) {
+	public RetrievalAugmentationAdvisor(List<QueryTransformer> queryTransformers, DocumentRetriever documentRetriever,
+			@Nullable QueryAugmentor queryAugmentor, @Nullable Boolean protectFromBlocking, @Nullable Integer order) {
+		Assert.notNull(queryTransformers, "queryTransformers cannot be null");
+		Assert.noNullElements(queryTransformers, "queryTransformers cannot contain null elements");
 		Assert.notNull(documentRetriever, "documentRetriever cannot be null");
+		this.queryTransformers = queryTransformers;
 		this.documentRetriever = documentRetriever;
 		this.queryAugmentor = queryAugmentor != null ? queryAugmentor : ContextualQueryAugmentor.builder().build();
 		this.protectFromBlocking = protectFromBlocking != null ? protectFromBlocking : false;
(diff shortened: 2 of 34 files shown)


spring-ai-core/src/main/java/org/springframework/ai/chat/client/DefaultChatClientBuilder.java
```diff
@@ -71,6 +71,10 @@ public ChatClient build() {
 		return new DefaultChatClient(this.defaultRequest);
 	}
 
+	public Builder clone() {
```
markpollack, line 74: good catch

markpollack:
Just some general thoughts, not necessary to resolve for this PR.

* The Query record. This is essentially the user text.  What is this was instead the UserMessage so that it can take into account multi-modality (media)  A QueryTransformer could for example retrieve images related to the text.
* The QueryAugmentor - what if we had as the signature `QueryAugmentor extends BiFunction<Query, RetrievalContext, Query>` where `RetrievalContext` is
```public class RetrievalContext {
      private final List<Document> documents;
      private final Map<String, Object> metadata;
   ...
``` this way additional information that can be used by an implementing class can be taken into account.  The document is the core part, but there could be additional information to influence how the query gets augmented.  Perhaps that would be passed in the constrcutor to the QueryAugmentor impl.  Just looking to move away from types that don't have extensibility in the interface.

File: spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java [hunk 2]
```diff
@@ -119,30 +127,45 @@ private AdvisedRequest before(AdvisedRequest request) {
...
+		// 2. Retrieve similar documents for the original query.
+		List<Document> documents = this.documentRetriever.retrieve(transformedQuery);
 		context.put(DOCUMENT_CONTEXT, documents);
 
-		// 2. Augment user query with the contextual data.
-		Query augmentedQuery = this.queryAugmentor.augment(query, documents);
+		// 3. Augment user query with the document contextual data.
+		Query augmentedQuery = this.queryAugmentor.augment(transformedQuery, documents);
 
 		return AdvisedRequest.from(request).withUserText(augmentedQuery.text()).withAdviseContext(context).build();
```
markpollack, line 145: wondering about the multimodal/media aspects of this as it focused on text and any media for the request is lost.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
```diff
@@ -119,30 +127,45 @@ private AdvisedRequest before(AdvisedRequest request) {
...
+		Query transformedQuery = originalQuery;
+		for (var queryTransformer : queryTransformers) {
+			transformedQuery = queryTransformer.apply(transformedQuery);
+		}
 
-		// 1. Retrieve similar documents for the original query.
-		List<Document> documents = this.documentRetriever.retrieve(query);
+		// 2. Retrieve similar documents for the original query.
+		List<Document> documents = this.documentRetriever.retrieve(transformedQuery);
 		context.put(DOCUMENT_CONTEXT, documents);
```
markpollack, line 140: maybe a RAGContext that in turn contains the documents so that for the future when additional things are added to the context for RAG, it is all in a central spot?

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
```diff
@@ -67,9 +72,12 @@ public class RetrievalAugmentationAdvisor implements CallAroundAdvisor, StreamAr
...
-			@Nullable Boolean protectFromBlocking, @Nullable Integer order) {
+	public RetrievalAugmentationAdvisor(List<QueryTransformer> queryTransformers, DocumentRetriever documentRetriever,
+			@Nullable QueryAugmentor queryAugmentor, @Nullable Boolean protectFromBlocking, @Nullable Integer order) {
+		Assert.notNull(queryTransformers, "queryTransformers cannot be null");
+		Assert.noNullElements(queryTransformers, "queryTransformers cannot contain null elements");
 		Assert.notNull(documentRetriever, "documentRetriever cannot be null");
+		this.queryTransformers = queryTransformers;
 		this.documentRetriever = documentRetriever;
 		this.queryAugmentor = queryAugmentor != null ? queryAugmentor : ContextualQueryAugmentor.builder().build();
 		this.protectFromBlocking = protectFromBlocking != null ? protectFromBlocking : false;
```
markpollack, line 83: I think `protectFromBlocking` being true by default is a more conservative options, this way it works out of the box in a reactive environment and for a non-reactive environment there isn't much of an impact.

spring-ai-core/src/main/java/org/springframework/ai/rag/analysis/query/expansion/QueryExpander.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * A component responsible for expanding the input query into a list of related queries
+ * based on a specified strategy. These expansions can be used to capture different
+ * perspectives or to break down complex queries into simpler, more manageable
+ * sub-queries, thereby improving the retrieval process.
+ *
+ * @author Thomas Vitale
+ * @since 1.0.0
+ */
+public interface QueryExpander extends Function<Query, List<Query>> {
```
markpollack, line 33: i'm wondering if this is something that should be used inside `RetrievalAugmentationAdvisor` or it is used in code at higher level.

spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/client/advisor/RetrievalAugmentationAdvisorIT.java
```diff
@@ -0,0 +1,140 @@
...
+ * @author Thomas Vitale
+ */
+@SpringBootTest(classes = TestApplication.class)
+@EnabledIfEnvironmentVariable(named = "OPENAI_API_KEY", matches = ".*")
+class RetrievalAugmentationAdvisorIT {
+
+	private List<Document> knowledgeBaseDocuments;
+
+	@Autowired
+	OpenAiChatModel openAiChatModel;
```
markpollack, line 58: I'm getting a test failure 

```
Caused by: java.lang.IllegalArgumentException: OpenAI API key must be set. Use the connection property: spring.ai.openai.api-key or spring.ai.openai.chat.api-key property.
	at org.springframework.util.Assert.hasText(Assert.java:240) ~[spring-core-6.1.13.jar:6.1.13]
	at org.springframework.ai.autoconfigure.openai.OpenAiAutoConfiguration.resolveConnectionProperties(OpenAiAutoConfiguration.java:100) ~[classes/:na]
```

I think some additional context config is needed.

markpollack:
I get test failures for tests  `org.springframework.ai.integration.tests` due to the api-key not being set.  Prob works in your environment but in my I don't have any `spring.ai.openai.*`properties set.

spring-ai-core/src/main/java/org/springframework/ai/rag/analysis/query/expansion/QueryExpander.java
ThomasVitale: I haven't adopted that in `RetrievalAugmentationAdvisor` yet, but I'll do so in a next PR where I introduce the Router and Fuser components that are needed to handle multiple queries/retrievals.

spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/client/advisor/RetrievalAugmentationAdvisorIT.java
ThomasVitale: I'll look into that. I wonder if it's due to the new APIKey concept introduced after I opened this PR

ThomasVitale: Oh yeah, it's actually the missing property you mentioned in https://github.com/spring-projects/spring-ai/pull/1703#issuecomment-2472106234 I'll fix that

spring-ai-core/src/main/java/org/springframework/ai/chat/client/DefaultChatClientBuilder.java
ThomasVitale: As discussed, we figured the mystery :) The Builder has the "clone()" method and the Client APIs the "mutate()" method.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
```diff
@@ -119,30 +127,45 @@ private AdvisedRequest before(AdvisedRequest request) {
...
 
-		// 2. Augment user query with the contextual data.
-		Query augmentedQuery = this.queryAugmentor.augment(query, documents);
+		// 3. Augment user query with the document contextual data.
+		Query augmentedQuery = this.queryAugmentor.augment(transformedQuery, documents);
 
 		return AdvisedRequest.from(request).withUserText(augmentedQuery.text()).withAdviseContext(context).build();
 	}
 
 	private AdvisedResponse after(AdvisedResponse advisedResponse) {
```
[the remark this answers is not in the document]
ThomasVitale: As discussed, the customisation approach will be based on implementing a custom Advisor and using the available RAG components to build a custom RAG flow, whereas all these out-of-the-box implementations will be "final" following the Spring Security approach.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
ThomasVitale: The AdvisedRequest has a separate Media property for the media elements, so they are currently propagated correctly.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
ThomasVitale: We agreed to talk more about this when we reconsider the design of ChatResponse and how to propagate back evidence from different intermediate steps in a ChatClient, such as retrieved documents and called functions.

markpollack:
merged in 263fe2fba7b6e940ac802e996db474d8909fe5d2

Issue #1601, opened 2024-10-25
MessageChatMemoryAdvisor adds empty user message


fmunch:
**Bug description**
When calling

```java
ChatClient.builder(chatModel)
        .defaultAdvisors(new MessageChatMemoryAdvisor(chatMemory))
        .build()
        .prompt()
        .call()
        .chatResponse()
```

`MessageChatMemoryAdvisor` adds an empty `UserMessage` in the chat memory (see [line 94](https://github.com/spring-projects/spring-ai/blob/v1.0.0-M3/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java#L94)).

In our use case we send the user message in a POST, store it in the conversation, then send a GET to benefit from `SseEmitter`. Hence the empty call to `prompt()`.

**Environment**
1.0.0-M3

**Expected behavior**
When using an empty prompt `MessageChatMemoryAdvisor` should probably not add an empty `UserMessage` in the chat memory.

DHKIM-0511:
Fix PR https://github.com/spring-projects/spring-ai/pull/1708

ThomasVitale:
@fmunch thanks for reporting this. Could you elaborate a bit more on this use case? I'm asking because I see the empty prompt and, in particular, the absence of a user message in `ChatClient` as an incorrect state (unless it's part of a function calling special flow). Indeed, the current snapshot version of Spring AI throws an exception if there's no user message (whereas in M3 an empty message was created, as you noticed) after adding null-safety to the ChatClient and Advisor APIs. Supporting a null/empty user message would make the behaviour of `ChatClient` unpredictable and hide pesky errors from developers.

If I understood correctly, your application is adding user messages to the `ChatMemory` object explicitly, outside the `ChatClient` workflow.  The `MessageChatMemoryAdvisor` is designed to handle the entire lifecycle of a chat memory for each conversation using an internal `ChatMemory` object. It might not be the best candidate for handling only parts of the lifecycle.

Since you have already a `ChatMemory` object where you're storing a user message explicitly, you could pass directly the messages from the memory via the `.prompt()` clause. And then save the response back into the memory.

Something like this:

```java
var chatResponse = ChatClient.builder(chatModel).build()
                .prompt(new Prompt(chatMemory.get(conversationId, lastN)))
                .call()
                .chatResponse();
chatMemory.add(chatResponse.getResult().getOutput());
```

[...]

Issue #1665, opened 2024-11-04
 userText cannot be null or empty caused by incomplete/incorrect check for messages in DefaultChatClientClass


github2atauhiq:
DefaultChatClientClass is assuming that the last message in the messages array is of type "USER". 
If the last message is not of type USER, userText is not being set and the subsequent null check will fail, although a USER message has been provided:

Pseudo-Code that causes the problem:

`chatClient
                .prompt(
                        new Prompt(
                                List.of(
                                        createUserMessage(request),
                                        createSystemMessage()
                                )))
                .call()
                .chatResponse();`

will result in :

java.lang.IllegalArgumentException: userText cannot be null or empty
	at org.springframework.util.Assert.hasText(Assert.java:240) ~[spring-core-6.1.10.jar:6.1.10]
	at org.springframework.ai.chat.client.advisor.api.AdvisedRequest.<init>(AdvisedRequest.java:87) ~[spring-ai-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.springframework.ai.chat.client.DefaultChatClient.toAdvisedRequest(DefaultChatClient.java:129) ~[spring-ai-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.springframework.ai.chat.client.DefaultChatClient$DefaultCallResponseSpec.doGetChatResponse(DefaultChatClient.java:483) ~[spring-ai-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.springframework.ai.chat.client.DefaultChatClient$DefaultCallResponseSpec.lambda$doGetObservableChatResponse$1(DefaultChatClient.java:477) ~[spring-ai-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at io.micrometer.observation.Observation.observe(Observation.java:565) ~[micrometer-observation-1.13.1.jar:1.13.1]
	at org.springframework.ai.chat.client.DefaultChatClient$DefaultCallResponseSpec.doGetObservableChatResponse(DefaultChatClient.java:477) ~[spring-ai-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
[...]

PR #1708, opened 2024-11-10
Do not add empty user text on chat memory

DHKIM-0511:
close https://github.com/spring-projects/spring-ai/issues/1601

someone did not want add empty user text on chat memory

so, I fix chat advisor to not add message when user text is empty

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
@@ -19,6 +19,7 @@ package org.springframework.ai.chat.client.advisor;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.springframework.util.StringUtils;
 import reactor.core.publisher.Flux;
 
 import org.springframework.ai.chat.client.advisor.api.AdvisedRequest;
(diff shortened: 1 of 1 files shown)


ThomasVitale:
Thanks for submitting this PR! This shouldn't be needed anymore since we introduced null-safety for ChatClient and Advisor APIs, and removed the default empty user text. The user message is now required and it will never be null or empty in the context of an Advisor.

DHKIM-0511:
> Thanks for submitting this PR! This shouldn't be needed anymore since we introduced null-safety for ChatClient and Advisor APIs, and removed the default empty user text. The user message is now required and it will never be null or empty in the context of an Advisor.

Oh, I see. I'm also curious about other possible solutions—would it be possible to check the commit history to see how it was implemented? Thanks for your kind response!

ThomasVitale:
Yes, of course! Most changes are in this PR: https://github.com/spring-projects/spring-ai/pull/1651

ThomasVitale:
@DHKIM-0511 we have an open discussion about making the logic of identifying a user message more flexible, while still ensuring internal consistency. If you have some use cases or ideas related to that, feel free to share them in https://github.com/spring-projects/spring-ai/issues/1665 Thanks!

DHKIM-0511:
Looking at the PR, I realized I missed a few lines of code. I understand now that the feature is no longer needed. I’ll also consider the issue you mentioned. Thanks!

PR #1721, opened 2024-11-11
Fix AnthropicApi Error Message

DHKIM-0511:
close https://github.com/spring-projects/spring-ai/issues/1665

I have modified the error message that is output by the AnthropicApi when an error occurs.


markpollack:
Sorry, we should have closed the issue last friday, the fix is already part of this commit which is in main branch

9fa89a2fd65dba03bd27b26ac49fd32a9fab558a

Issue #1240, opened 2024-08-18
Newly created ChromaDB collection is missing collectionId and failing on insert


novakma2:
**Bug description**
When trying to add documents to ChromaDB, client returns `404 Not Found: "{"detail":"Not Found"}"` with my multistore setup on newly added collection. 

**Environment**
Java 21; 
Springboot 3.3.2;
SpringAI 1.0.0 M1;
ChromaDB latest;

**Steps to reproduce**
Create new Springboot project with ChromaDB, create a collection using chromaAPI, and try to add a document to that collection.

**Expected behavior**
When new ChromaDB collection is created, id should be populated and all actions should be working.

**Minimal Complete Reproducible example**
In my current multi collection setup, I am creating new collection per customer and adding the client to map, as that seems to be the only way to add data to different collections.
```java
public void synchronizeCollections() {
        customerService.getCustomers().stream()
                .filter(customer -> !storeByHost.containsKey(customer.host()))
                .forEach(customer -> {
                    final var collection = chromaApi.createCollection(new ChromaApi.CreateCollectionRequest(customer.host()));
                    logger.info("Inserting new ChromaDB collection {} with name {}", collection.id(), collection.name());
                    final var store = new ChromaVectorStore(embeddingModel, chromaApi, collection.name(), true);
                    storeByHost.put(collection.name(), store);
                });
    }
```
Then when running the add for the collection of given host
```java
    public void insertToStore(String host, List<Document> documents) {
        if (!storeByHost.containsKey(host)){
            logger.error("Host is not present in ChromaDB collection, host:{}", host);
            return;
        }
        storeByHost.get(host).add(documents);
    }
```
I am receiving following exception:
```
org.springframework.web.client.HttpClientErrorException$NotFound: 404 Not Found: "{"detail":"Not Found"}"
	at org.springframework.web.client.HttpClientErrorException.create(HttpClientErrorException.java:112) ~[spring-web-6.1.11.jar:6.1.11]
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:183) ~[spring-web-6.1.11.jar:6.1.11]
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:137) ~[spring-web-6.1.11.jar:6.1.11]
[...]

PR #1723, opened 2024-11-12
GH-1240: Add builder pattern to ChromaVectorStore for better initiali…

sobychacko:
…zation control

Fixes: #1240

Issue: https://github.com/spring-projects/spring-ai/issues/1240

The change addresses initialization issues when ChromaVectorStore is used outside Spring context, particularly in scenarios where collections are created manually before store instantiation. Previously, collection ID wasn't properly populated when afterPropertiesSet() wasn't called by Spring container.

- Add builder pattern to ChromaVectorStore for better initialization control
- Add initialization flag to prevent multiple collection creation calls
- Add integration tests for builder pattern usage scenarios
- Add spring-ai-transformers dependency for testing
- Remove unused constants (SIMILARITY_THRESHOLD_ALL, DEFAULT_TOP_K)

Collection ID is now properly set regardless of whether the store is managed by Spring or created manually, solving the 404 Not Found errors during document insertion.

### DIFF
--- a/vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/vectorstore/ChromaVectorStore.java
+++ b/vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/vectorstore/ChromaVectorStore.java
@@ -41,7 +41,6 @@ import org.springframework.ai.util.JacksonUtils;
 import org.springframework.ai.vectorstore.filter.FilterExpressionConverter;
 import org.springframework.ai.vectorstore.observation.AbstractObservationVectorStore;
 import org.springframework.ai.vectorstore.observation.VectorStoreObservationContext;
-import org.springframework.ai.vectorstore.observation.VectorStoreObservationContext.Builder;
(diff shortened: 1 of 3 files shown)


vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/vectorstore/ChromaVectorStore.java
```diff
@@ -111,6 +108,26 @@ public ChromaVectorStore(EmbeddingModel embeddingModel, ChromaApi chromaApi, Str
 		this.objectMapper = JsonMapper.builder().addModules(JacksonUtils.instantiateAvailableModules()).build();
 	}
 
+	private ChromaVectorStore(Builder builder) {
```
markpollack, line 111: I think this is yet another variation of the builder pattern in our docs, any reason not to have this code inside the builder's build method?

markpollack:
merged in c8a9b1640a66d022c02323dfe600b0fabb3c5cb5

PR #1725, opened 2024-11-12
GH-1589 Fix OpenSearch client indices mapping

### LINKED ISSUE #1589: Opensearch vector store bugs
Following the walkthrough in spring ai documenation for opensearch vector store on aws, the following properties are incorrectly located:

spring.opensearch.aws.host
spring.opensearch.aws.service-name 
spring.opensearch.aws.access-key
spring.opensearch.aws.secret-key 
spring.opensearch.aws.region

these only work if located on
spring.ai.vectorstore.opensearch.*

else causes error:
Caused by: java.lang.NullPointerException: Access key ID cannot be blank.

Also similarity search for index with the following json mapping gives:
Similarity search results with error:
org.opensearch.client.opensearch._types.OpenSearchException: Request failed: [search_phase_execution_exception] all shards failed

for mapping json:
mappingJson: |
              "properties": {
                "embedding": {
                  "type": "knn_vector",
                  "dimension": 1536
                }
              }

ilayaperumalg:
- Fix the embedding dimension configuration for opensearch client indices mapping
    - The dimension config is obtained by the underlying embedding model's dimension

Resolves #1589


markpollack:
Add docs.  Also test case when the properties string is provided by the user but there isn't a `%s` in the string, does it fail?

tzolov:
rebased, squashed and merged at de53e64bfd73da786c1e4ab0d2514f5941a7caa7

PR #1726, opened 2024-11-12
feat: Move the advisor name to low cardinality key values

### LINKED ISSUE #1716: Consider making spring.ai.advisor.name a low-cardinality key for spring.ai.advisor metric
Of the existing low-cardinality keys for spring.ai.advisor, none of them vary much...

- gen_ai.operation.name is always "framework"
- gen_ai.system is always "spring_ai"
- spring.ai.kind is always "advisor"
- spring.ai.advisor.type could be BEFORE, AFTER, or AROUND, but in reality it is always AROUND because that's the only kind of advisor that currently exists in Spring AI.

As a result, none of these keys are useful for filtering the metric. 

But what might be interesting is to be able to filter the metric based on what advisor is in play. That is, if the spring.ai.advisor.name were a low-cardinality key, then you could do this.

It is conceptually possible that there could be several advisors in play, which is likely the reason it is a high-cardinality key. But I'd think that in most application, there wouldn't be that many advisors...likely less than 10...so while it has potential for being high-cardinality, in practice it will probably be low-cardinality. 

Therefore, I would like to consider making it a low-cardinality key so that it can be used to filter the spring.ai.advisor metric.

tzolov:
- Adviser name is moved from hight to low cardinality key values.
- Updated tests in  to reflect these changes, ensuring the new low cardinality key is correctly captured.

Resolves #1716

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/observation/AdvisorObservationDocumentation.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/observation/AdvisorObservationDocumentation.java
@@ -68,11 +68,7 @@ public enum AdvisorObservationDocumentation implements ObservationDocumentation 
 			public String asString() {
 				return "spring.ai.advisor.type";
 			}
-		}
-
-	}
-
-	public enum HighCardinalityKeyNames implements KeyName {
+		},
 
 		/**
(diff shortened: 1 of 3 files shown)


markpollack:
merged in 7474852b58fd30829924beb8aea5aeb65cd2d586

PR #1729, opened 2024-11-12
extension function for Chatclient.CalLResponseSpec.entity(Class<?>) a…

joshlong:
this adds kotlin support for using the `ChatClient` to get typesafe responses back

```kotlin
val joke:Joke = cc
      .prompt("tell me a joke")
      .call()
      .entity<Joke>()
println(joke)
```

this is as opposed to the more verbose and less idiomatic Kotlin code without the extension function

````
val joke : Joke = cc
    .prompt("tell me a joke")
    .call()
    .entity(Joke::class.java)
```

### DIFF
--- a/spring-ai-core/src/main/kotlin/org/springframework/ai/chat/client/ChatClientExtensions.kt
+++ b/spring-ai-core/src/main/kotlin/org/springframework/ai/chat/client/ChatClientExtensions.kt
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2023-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


spring-ai-core/src/main/kotlin/org/springframework/ai/chat/client/ChatClientExtensions.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package org.springframework.ai.chat.client
+
+/**
+ * Extension for [ChatClient] providing a `entity<Foo>()` variant.
+ *
+ * @author Josh Long
+ */
+
+inline fun <reified T> ChatClient.CallResponseSpec.entity(): T = entity(T::class.java) as T
```
sdeleuze, line 25: @joshlong Could you please use `ParameterizedTypeReference` instead of `Class` in order to avoid type erasure?

sdeleuze (changes requested):

joshlong:
ive updated the PR accordingly. thanks for the great feedback

PR #1730, opened 2024-11-13
GH-1529 Fix ChromaVectorStore delete Ids operation

### LINKED ISSUE #1529: ChromaVectorStore doDelete  The return value is null
image

image

deletedIds The return value is null

ilayaperumalg:
- As a result of the the change:https://github.com/chroma-core/chroma/pull/2880  introduced in chroma:0.5.13,the delete operation doesn't return any values. Hence, we need to check the Http Client Response with the status code instead of return value.
   - Check the status code for successful delete operation
   - Update test

Resolves #1529


markpollack:
changed to return int and merged as a98c04296002932d88a12b3a2e3c4d797442baf1

Issue #1727, opened 2024-11-12
Huggingface /generate in path to model is not expected


nastyabakhshieva:
**Bug description**
I was experimenting on Spring AI and was looking for a free model for ChatApi implementation. Upon investigation I found I can use HuggingFace `microsoft/Phi-3-mini-4k-instruct` model. So, I added the following dependency in my project and started to investigate:
```groovy
// POM version is 1.0.0-M3
implementation 'org.springframework.ai:spring-ai-huggingface-spring-boot-starter'
```
Similarly, I added in application yaml:
```yaml
spring:
  ai:
    huggingface:
      chat:
        api-key: hf_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
        url: https://api-inference.huggingface.co/models/microsoft/Phi-3-mini-4k-instruct
```
Once done, I called `HuggingfaceChatModel`

Unfortunately, The following exception was received:
```
org.springframework.web.client.HttpClientErrorException$NotFound: 404 Not Found: "{"error":"Model microsoft/Phi-3-mini-4k-instruct/generate does not exist"}"
```
What I found is that, /generate postfix is appended in the code & hardcoded in org.springframework.ai.huggingface.api.TextGenerationInferenceApi#generateWithHttpInfo, line 83
Whereas, with postman everything works as expected as per hugging face documentation - https://huggingface.co/docs/api-inference/tasks/text-generation?code=curl

**Environment**
Java version - 21
Build tool - gradle 8.3
Spring AI version - 1.0.0-M3

**Steps to reproduce**
Just a simple ChatClient setup with hugging face model provided along with access key generated in hugging face account

**Expected behavior**
/generate is not hardcoded hence can put whatever url is needed

**Minimal Complete Reproducible example**
I can't put my access key here, but all you need to do is
1. Create a simple project
2. Add huggingface-ai dependency
3. Define hugging face url & access key in application properties file
4. Call the ChatClient

jitokim:
I removed the path from the generated client file using the openapi.json and ran it because of this issue. 

It seems like I might need to get the latest version of the OpenAPI file from Hugging Face. (I haven’t fully analyzed it yet.)


As a temporary workaround, it works if I remove the path in the URI component builder for the path.

jitokim:
@nastyabakhshieva  Hi, thanks for reporting issue.
[...]

PR #1733, opened 2024-11-13
fix huggingface generate text

jitokim:
fix issue #1727  when using text-generation-inference models

I think the [openapi.json ](https://github.com/huggingface/text-generation-inference/blob/v2.4.0/docs/openapi.json) needs to be fixed.
you can find the response type is an array [here](https://github.com/huggingface/text-generation-inference/blob/a7850008429c4c1c4a2ded7bbed4c1b12d22d287/router/src/server.rs#L145)

test models:
microsoft/Phi-3-mini-4k-instruct
mistralai/Mistral-7B-Instruct-v0.3
Qwen/Qwen2.5-Coder-32B-Instruct

microsoft/Phi-3-mini-4k-instruct
image

mistralai/Mistral-7B-Instruct-v0.3
image

Qwen/Qwen2.5-Coder-32B-Instruct
image


ilayaperumalg:
Hi @jitokim, Thanks for the fix! LGTM, merged as 3c14fa63 after updating the ClientIT's prompt message to insist on providing strict JSON output so that the test assertion has more probability of passing.

ilayaperumalg:
@jitokim Thank you for the kind words! It was the suggestion from @markpollack!

Issue #1728, opened 2024-11-12
Misleading Cassandra TTL property name


tusharleek:
https://github.com/spring-projects/spring-ai/blob/6acd202c12220b27d39e3d3dd25cfe6a527fa2a8/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/memory/cassandra/CassandraChatMemoryProperties.java#L48

Can we please change name or its implementation as the value I am assigning this using properties is in milliseconds and field name is time-to-live-seconds and can we add any provision that if I add or update this field value then after application restart changes will reflect as of now it only works if table is not there previously

jitokim:
https://github.com/spring-projects/spring-ai/blob/420bf5ed78559a0c0f40006b804d3edda6efce2e/vector-stores/spring-ai-cassandra-store/src/main/java/org/springframework/ai/chat/memory/CassandraChatMemoryConfig.java#L168

tusharleek:
Yes But when in properties I am providing 
spring:
  ai:
    chat:
      memory:
        cassandra:
          keyspace: conversation
          time-to-live-seconds: 120000
          
 then my rows are deleting after 120 seconds
 
 if I am using above property in logs you can see that defaultTimetoLive is 120 even after setting 120000 in properties
 
<img width="1383" alt="Screenshot 2024-11-14 at 10 20 30 PM" src="https://github.com/user-attachments/assets/83c372c1-302f-4a86-a528-d58b95789db9">

jitokim:
@tusharleek  
I have created a PR. Please review it 😃

PR #1737, opened 2024-11-14
rename timeToLiveSeconds of cassandra property

jitokim:
fix issue #1728

rename `timeToLiveSeconds` to `timeToLive` because the type is Duration.

add test code for ISO-8601 Format and also numeric type

image

image


jitokim:
The reason I changed the name of timeToLiveSeconds without changing its type is,

Using the Duration type allows us to preserve the ISO-8601 format, but its default unit is milliseconds.

To keep the function name (`withTimeToLive`) and its parameter type (Duration) consistent in the `CassandraChatMemoryConfig` class.

tusharleek (approved):
Can we add any comments in documentation or here that specify we are expecting this value in milliseconds

markpollack:
Thanks, merged in a7a1804fac2987dd2f34640733a29f94e1b46a9c

PR #1740, opened 2024-11-15
docs: correct code in vertex chat functions doc

garethjevans:


ilayaperumalg:
@garethjevans Thanks for the PR! There is one other place in the doc which requires fixing. I will fix when merging.

PR #1741, opened 2024-11-15
docs: document how to programatically configure vertex embeddings to use a service account

garethjevans:


tzolov:
Thanks @garethjevans

PR #1751, opened 2024-11-18
chore(deps): bump azure-openai to beta.12

garethjevans:
This is needed so that we can enable the "later" service versions (2024-08-01-preview and 2024-06-01-preview) in order to access the usage metrics when using streaming chat calls.


markpollack:
This requires IT testing and checking that nothing blows up in the autoconf module due to conflicting libs.

markpollack:
For visibility

Core dependency upgrades:

azure-core: 1.49.1 → 1.53.0
azure-json: 1.1.0 → 1.3.0
azure-xml: 1.0.0 → 1.1.0
azure-core-http-netty: 1.15.1 → 1.15.5

All other dependencies remain unchanged between the two versions.

markpollack:
merged in a7b5652790cd14efa19d59b3cabf6a54b28aa215  IT test passed locally.

PR #1752, opened 2024-11-18
feat(tools): Add conversation history to ToolContext

### LINKED ISSUE #1202: Possibility to connect function calls with messages they are related to
**Expected Behavior**

It would be good to be able to easily connect function calls with user prompts that trigger them or with assistant responses that were generated based on these function calls. It would be especially beneficial for chats with memory.

**Current Behavior**

Currently some implementations deep down use history of function calls (a.k.a. tool calls) that were used in process of answering user's question, but it is hard to get these with IDs of messages they are connected to.

**Context**

Use-case for this feature: Often it would be beneficial if end-users of chat tools that are using Spring AI in the backend could validate what function calls (and with what parameters) were used in order to answer their questions instead of accepting answers blindly as if they were coming from some black box.

I believe that currently this is not easy to implement this feature in Spring AI (in case of integration with OpenAI I had to use AOP, which I consider kind of last resort solution).

tzolov:
- Add TOOL_CONVERSATION_KEY constant to store conversation history
- Extend ToolContext with getToolConversationHistory method
- Include conversation history in tool context during function execution
- Add test coverage for tool conversation history verification

Resolves #1202

Decisions to make: 

- [ ] Should we have a dedicated `FunctionCallingOption` property to control the addition of the current tool conversation history to the TooContext.  Or assume that any ToolContext will provide this information? 
- [ ] Should we have a dedicated `ToolContext#getToolConversationHistory` or use the raw map with predefined key `TOOL_CONVERSATION_KEY`.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/model/AbstractToolCallSupport.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/model/AbstractToolCallSupport.java
@@ -146,10 +146,11 @@ public abstract class AbstractToolCallSupport {
 
 			toolContextMap = new HashMap<>(functionCallOptions.getToolContext());
 
-			List<Message> beforeFunctionCallConversationHistory = new ArrayList<>(prompt.copy().getInstructions());
-			beforeFunctionCallConversationHistory.add(new AssistantMessage(assistantMessage.getContent(),
-					assistantMessage.getMetadata(), assistantMessage.getToolCalls()));
-			toolContextMap.put(ToolContext.TOOL_CONVERSATION_KEY, beforeFunctionCallConversationHistory);
+			List<Message> toolCallHistory = new ArrayList<>(prompt.copy().getInstructions());
+			toolCallHistory.add(new AssistantMessage(assistantMessage.getContent(), assistantMessage.getMetadata(),
+					assistantMessage.getToolCalls()));
+
+			toolContextMap.put(ToolContext.TOOL_CALL_HISTORY, toolCallHistory);
 		}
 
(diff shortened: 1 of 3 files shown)


spring-ai-core/src/main/java/org/springframework/ai/chat/model/AbstractToolCallSupport.java
```diff
@@ -141,12 +142,22 @@ protected List<Message> handleToolCalls(Prompt prompt, ChatResponse response) {
 		Map<String, Object> toolContextMap = Map.of();
 		if (prompt.getOptions() instanceof FunctionCallingOptions functionCallOptions
 				&& !CollectionUtils.isEmpty(functionCallOptions.getToolContext())) {
```
ilayaperumalg, line 144: This logic would prevent the conversion history to be inserted if the tool context was empty from chat client request spec.

spring-ai-core/src/main/java/org/springframework/ai/chat/model/ToolContext.java
```diff
@@ -57,4 +69,13 @@ public Map<String, Object> getContext() {
 		return this.context;
 	}
 
+	/**
+	 * Returns the tool conversation history from the context map.
```
ilayaperumalg, line 73: To be consistent, we can update the javadoc to reflect "tool call history"

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased, updated the javadoc, verified the changes and merged as 0eacc919

PR #1754, opened 2024-11-18
Refactor: Replace instanceof checks and switch expression

delvering17:
I refactored some code to improve readability:
- Extracted the result of `instanceof` checks into variables
- Refactored to use switch expressions instead of `if-else` blocks


jitokim:
you could include your name in the "author" field.  
also, adding test cases to verify the functionality would make it even better.
👍

delvering17:
Thank you for the feedback!

I also refactored the `BedrockProxyChatModel`. There was an unnecessary `instanceof` check, so I removed it. 
Additionally, I updated the previously refactored OpenAiChatModel to use `swtich expression` instead of `switch-case statement`.

PR #1755, opened 2024-11-18
fix callback merging in AzureOpenAiChatModel constructor

jitokim:
and minor fixes ( typo, unused code ).

Function callbacks from `chatOptions` and constructor parameters were not properly merged because the `toolFunctionCallbacks` field in `AzureOpenAiChatModel` was hardcoded to an empty list (`List.of()`).

Ensuring both sources of function callbacks are combined into the functionCallbackRegister during initialization.

AS-IS
Only one callback function was registered from `chatOptions`.

image

TO-BE
Both callback functions were merged.

image

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/AnthropicApi.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/AnthropicApi.java
@@ -53,6 +53,7 @@ import org.springframework.web.reactive.function.client.WebClient;
  * @author Christian Tzolov
  * @author Mariusz Bernacki
  * @author Thomas Vitale
+ * @author Jihoon Kim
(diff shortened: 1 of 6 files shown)


tzolov:
This is nasty indeed:
>Function callbacks from chatOptions and constructor parameters were not properly merged because the toolFunctionCallbacks field in AzureOpenAiChatModel was hardcoded to an empty list (List.of()).

Thanks for catching and fixing it. Much appreciated!

P.S. In the future please submit the unrelated changes in a separate PR

tzolov:
Rebased and merged at d8583271c7f28dea885c11c0b96f8a1e0ba983f8

jitokim:
> P.S. In the future please submit the unrelated changes in a separate PR
> 
> 

Got it, I'll do that.
And thanks for the review. :)

Issue #1758, opened 2024-11-19
Improve Spring AI Bedrock Converse API Integration


tzolov:
Extend the [Spring AI Bedrock Converse](https://docs.spring.io/spring-ai/reference/api/bedrock-converse.html) integration with support for the following Bedrock [Converse](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) and [ConverseStream](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStream.html) API features:

- [ ] [additionalModelRequestFields](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html#API_runtime_Converse_RequestSyntax) - Additional inference parameters that the model supports, beyond the base set of inference parameters that Converse and ConverseStream support in the inferenceConfig field.
- [ ] [ConverseMetrics](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseMetrics.html) and [ConverseStreamMetrics](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStreamMetrics.html) - Metrics for a call to Converse and ConverseStream
- [ ] [Guardrail](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_GuardrailConfiguration.html), [Include a guardrail with Converse API](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-use-converse-api.html) - to guard conversational apps that you create with the Converse

starJammer:
I want to propose, in addition to the above, to add support for a 'guardContent' message type specifically for AWS. Guardrails on AWS, by default execute on the entire conversation and all messages. However, this is efficient and drives up cost significantly for ongoing conversations. 

[Refer to this document](https://github.com/aws/agent-toolkit-for-aws/blob/main/plugins/aws-core/skills/amazon-bedrock/references/guardrails.md) for how to apply the 'guardContent' message type. 

Essentially, guardContent message types should be supported so that AWS knows to apply the guardrail only to a limited number of messages in the message stack. This way, guardrails execute more quickly and efficiently and don't re-scan old messages in the conversation.

Example: This sort of global guardrail as suggested by @tzolov  will execute on all messages.
```json
{
  "guardrailConfig": {
    "guardrailIdentifier": "my-guardrail-id",
    "guardrailVersion": "1",
    "trace": "disabled"
  }
}
```

[...]

PR #1760, opened 2024-11-19
feat(spring-ai-bedrock-converse): Introduce BedrockProxyChatOptions

tzolov:
- Add new BedrockProxyChatOptions class to replace FunctionCallingOptions
- Create BedrockProxyChatOptionsBuilder for fluent configuration
- Update BedrockProxyChatModel and related classes to use new options
- Add support for additional model-specific parameters via 'additional' map
- Improve options merging and handling of chat configuration

Related to #1758

### DIFF
--- a/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatOptions.java
+++ b/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatOptions.java
@@ -1,0 +1,277 @@
+/*
+* Copyright 2024 - 2024 the original author or authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* https://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package org.springframework.ai.bedrock.converse;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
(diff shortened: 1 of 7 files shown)


models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatOptions.java
```diff
@@ -0,0 +1,277 @@
...
+import org.springframework.ai.model.function.FunctionCallingOptions;
+import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Christian Tzolov
+ * @since 1.0.0
+ */
+public class BedrockProxyChatOptions implements FunctionCallingOptions {
```
ilayaperumalg, line 37: Any reason why `BedrockProxyChatOptions` can not implement `ChatOptions` here as I think we got all of the ChatOptions as well?

ilayaperumalg, line 37: Not specific to this PR, but, do you think it would be good to have a `DefaultFunctionCallingOptions` as the default implementation for `FunctionCallingOptions` so that all the Model Chat options can extend? or, any reason why we didn't have the default implementation at the first place?

tzolov: FunctionCallingOptions implements ChatOptions already

ilayaperumalg: Got it. I got deceived by the explicit declaration of ChatOptions in some of the model option classes. I just pushed a minor PR to fix this: https://github.com/spring-projects/spring-ai/pull/1762

markpollack:
@tzolov still relevant?

tzolov:
Yes it is. 
Amongst others it aims to expose the [additionalModelRequestFields](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html#API_runtime_Converse_RequestSyntax) - Additional inference parameters that the model supports, beyond the base set of inference parameters that Converse and ConverseStream support in the inferenceConfig field. (part of https://github.com/spring-projects/spring-ai/issues/1758)

But perhaps we can move it for after the GA

tzolov:
won't fix

PR #1762, opened 2024-11-19
Remove explicit ChatOptions for the models using FunctionCallingOptions

ilayaperumalg:
- Since FunctionCallingOptions already extends ChatOptions, removing explicit usage of ChatOptions in the Model Options classes


sobychacko:
Merged via https://github.com/spring-projects/spring-ai/commit/3812d2e6d89bdf1079f63013154cc6a6c5e5d466.

PR #1763, opened 2024-11-19
Fix Bedrock Converse streaming and token handling

### LINKED ISSUE #1743: Using BedrockProxyChatModel with tools, chatResponse.metadata.usage is returning ONLY the last model invocation usage
Using BedrockProxyChatModel with tools, the chatResponse.metadata.usage only contains the token usage of the last invocation to the model instead of the sum of all the invocations when tools were invoked

tzolov:
- Modify stream method to support recursive tool call handling
- Update token tracking and metadata merging for streamed responses
- Improve token usage calculation for tool use events
- Update test cases to handle new response processing

Resolves #1743

### DIFF
--- a/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
+++ b/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
@@ -473,13 +473,15 @@ public class BedrockProxyChatModel extends AbstractToolCallSupport implements Ch
 	 */
 	@Override
 	public Flux<ChatResponse> stream(Prompt prompt) {
+		return this.internalStream(prompt, null);
+	}
+
+	private Flux<ChatResponse> internalStream(Prompt prompt, ChatResponse perviousChatResponse) {
 		Assert.notNull(prompt, "'prompt' must not be null");
 
 		return Flux.deferContextual(contextView -> {
 
 			ConverseRequest converseRequest = this.createRequest(prompt);
-
-			// System.out.println(">>>>> CONVERSE REQUEST: " + converseRequest);
 
 			ChatModelObservationContext observationContext = ChatModelObservationContext.builder()
 				.prompt(prompt)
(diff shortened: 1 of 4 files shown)


models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
```diff
@@ -195,7 +199,7 @@ public ChatResponse call(Prompt prompt) {
 		if (!this.isProxyToolCalls(prompt, this.defaultOptions) && chatResponse != null
 				&& this.isToolCall(chatResponse, Set.of("tool_use"))) {
 			var toolCallConversation = this.handleToolCalls(prompt, chatResponse);
-			return this.call(new Prompt(toolCallConversation, prompt.getOptions()));
+			return this.internalCall(new Prompt(toolCallConversation, prompt.getOptions()), chatResponse);
```
ilayaperumalg, line 202: We need test coverage to test this flow - especially to verify the final ChatResponse containing the preceding one.

models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
```diff
@@ -448,8 +452,19 @@ private ChatResponse toChatResponse(ConverseResponse response) {
...
-		DefaultUsage usage = new DefaultUsage(response.usage().inputTokens().longValue(),
-				response.usage().outputTokens().longValue(), response.usage().totalTokens().longValue());
+		Long promptTokens = response.usage().inputTokens().longValue();
+		Long generationTokens = response.usage().outputTokens().longValue();
+		Long totalTokens = response.usage().totalTokens().longValue();
+
+		if (perviousChatResponse != null && perviousChatResponse.getMetadata() != null
+				&& perviousChatResponse.getMetadata().getUsage() != null) {
+
+			promptTokens += perviousChatResponse.getMetadata().getUsage().getPromptTokens();
```
ilayaperumalg, line 462: I have manually verified this step. Is there an effective way we can add a test to validate this additive operation in our tests?

ilayaperumalg:
Thanks for updating the tests! LGTM, merging.

ilayaperumalg:
Rebased, squashed, fixed check style error and merged as 00ede3f5

PR #1767, opened 2024-11-19
Modular RAG: Orchestration and Post-Retrieval

ThomasVitale:
Pre-Retrieval:
* Consolidated naming and documentation

Retrieval:
* Consolidated naming and documentation
* Introduced DocumentJoiner sub-module and CompositionDocumentJoiner operator

Post-Retrieval:
* Introduced main interfaces for sub-modules. Implementation waiting for missing features in Document APIs

Orchestration:
* Introduced QueryRouter sub-module and AllDocumentRetrieversQueryRouter operator

Generation:
* Consolidated naming and documentation

Advisor:
* Introduced BaseAdvisor to reduce boilerplate when implementing Advisors
* Extended RetrievalAugmentationAdvisor to include the new sub-modules

Relates to #gh-1603


spring-ai-core/pom.xml
```diff
@@ -96,6 +96,11 @@
 			<artifactId>micrometer-core</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>io.micrometer</groupId>
+			<artifactId>context-propagation</artifactId>
```
ThomasVitale, line 101: When using the TaskExecutor to run things asynchronously, no observation data is propagated unless we add this dependency (https://docs.micrometer.io/context-propagation/reference/index.html)

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
```diff
@@ -16,71 +16,81 @@
...
+
+	private final DocumentJoiner documentJoiner;
+
+	private final QueryAugmenter queryAugmenter;
 
-	private final QueryAugmentor queryAugmentor;
+	private final TaskExecutor taskExecutor;
 
-	private final boolean protectFromBlocking;
+	private final Scheduler scheduler;
```
ThomasVitale, line 77: Instead of passing a `protectFromBlocking` parameter, users can now pass a `Scheduler` implementation if they want to customise the streaming behavior. This has two main benefits: more clarity about what is happening under the hood and also the possibility of using different schedulers than the elastic one.

The case where we don't need the protection from blocking is supported when passing `Schedulers.immediate()`. The default, as we discussed last time, is running with `Schedulers.boundedElastic()`.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
```diff
@@ -135,17 +111,47 @@ private AdvisedRequest before(AdvisedRequest request) {
...
 
-		// 2. Retrieve similar documents for the original query.
-		List<Document> documents = this.documentRetriever.retrieve(transformedQuery);
+		// 2. Expand query into one or multiple queries.
+		List<Query> expandedQueries = queryExpander != null ? queryExpander.expand(transformedQuery)
+				: List.of(transformedQuery);
+
+		// 3. Get similar documents for each query.
+		Map<Query, List<List<Document>>> documentsForQuery = expandedQueries.stream()
+			.map(query -> CompletableFuture.supplyAsync(() -> getDocumentsForQuery(query), taskExecutor))
```
ThomasVitale, line 120: Query expansion and/or routing can result in parallel executions that is supported by a `TaskExecutor`, following the common practice in Spring. By default, a `ThreadPoolTaskExecutor` is used, but it's possible to customise it. For example, you can pass the auto-configured `TaskExecutor` from Spring Boot or another implementation, such as `VirtualThreadTaskExecutor`. I'll describe this in detail in the separate PR where I'm writing the documentation.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/api/BaseAdvisor.java
```diff
@@ -0,0 +1,107 @@
...
+ * advisor. It provides default implementations for the
+ * {@link #aroundCall(AdvisedRequest, CallAroundAdvisorChain)} and
+ * {@link #aroundStream(AdvisedRequest, StreamAroundAdvisorChain)} methods, delegating the
+ * actual logic to the {@link #before(AdvisedRequest)} and {@link #after(AdvisedResponse)}
+ * methods.
+ *
+ * @author Thomas Vitale
+ * @since 1.0.0
+ */
+public interface BaseAdvisor extends CallAroundAdvisor, StreamAroundAdvisor {
```
ThomasVitale, line 41: As we talked last time, it's useful to remove some of the boilerplate when creating a new Advisor. I have created this interface which hides most of the boilerplate. And it's not RAG specific, so it can be used for any Advisor.

spring-ai-core/src/main/java/org/springframework/ai/rag/preretrieval/query/expansion/MultiQueryExpander.java
```diff
@@ -70,7 +69,7 @@ public final class MultiQueryExpander implements QueryExpander {
 			Query variants:
 			""");
 
-	private static final Boolean DEFAULT_INCLUDE_ORIGINAL = false;
+	private static final Boolean DEFAULT_INCLUDE_ORIGINAL = true;
```
ThomasVitale, line 72: By default, the original query should be included since that's the most common case. My previous decision here for the default was not great. I fixed that now.

spring-ai-core/src/main/java/org/springframework/ai/rag/generation/augmentation/ContextualQueryAugmenter.java
```diff
@@ -74,15 +74,15 @@ public final class ContextualQueryAugmentor implements QueryAugmentor {
 			Politely inform the user that you can't answer it.
 			""");
 
-	private static final boolean DEFAULT_ALLOW_EMPTY_CONTEXT = true;
+	private static final boolean DEFAULT_ALLOW_EMPTY_CONTEXT = false;
```
ThomasVitale, line 77: This one is very debatable so I'm eager the get some feedback.

When using explicit RAG, I probably want to reduce hallucinations as much as possible. If no similar documents are retrieved from a vector store, I probably want to short-circuit the generation and have the model answering that it doesn't know rather than making stuff up.

I would therefore recommend disallowing having empty context by default. Then the user can decide if they want the model to make stuff up. This should help a lot with troubleshooting and in spotting retrieval errors earlier.

The case where I want to fallback to "standard" generation without context is something that should be handled at the beginning of the RAG flow. There's typically an agentic component that starts a flow by determining whether to send a request directly to the model or go through the retrieval process. It's something I have in my backlog and that will be introduced as a "Scheduling" sub-module.

What do you think about this choice of the default value?

spring-ai-core/pom.xml
markpollack: :+1:

spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/RetrievalAugmentationAdvisor.java
markpollack: I think is a good starting point that is generally useful.  Users can implement their own RAG Advisor to customize how they do scatter gather based on other frameworks and patterns.

spring-ai-core/src/main/java/org/springframework/ai/rag/generation/augmentation/ContextualQueryAugmenter.java
markpollack: Since the main point of advanced rag techniques is better accuracy, i think explicitly dealing with the empty context makes sense, as otherwise it will make up stuff, going against the 'pedal tone' of intent for these classes.

markpollack:
I had to add 

```
<dependency>
    <groupId>io.micrometer</groupId>
    <artifactId>context-propagation</artifactId>
    <scope>test</scope>
</dependency>
```

to the `spring-ai-integration-test` to avoid classdef not found . I would have thought it would come transitively.

sjivan:
@ThomasVitale  where does the QueryRouter sub-module and AllDocumentRetrieversQueryRouter operator reside? Has it made it to any Spring AI release or pre-release build?

PR #1772, opened 2024-11-20
refactor: standardized logger parameter bind type-assured

loveysuby:
* Improving memory occupancy of unnecessary log data by formatting log parameter targets in test logger data dummy.
* The parsed string comes in from the previous row, but the log output specification is readable. :)
  test-passed-snapshot-spring-ai-PR-standardized-logger


ericbottard:
Applied on main via 17c8e9e0cfd2d6a291911d2f24246130d4207596

loveysuby:
Thanks for resolving. @ericbottard

PR #1774, opened 2024-11-20
add ollama model options:llama-vision,qwen2.5

csuwl:
add ollama model options:llama-vision,qwen2.5

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
@@ -30,118 +30,159 @@ public enum OllamaModel implements ChatModelDescription {
 	/**
 	 * Llama 2 is a collection of language models ranging from 7B to 70B parameters.
 	 */
-	LLAMA2("llama2"),
+	LLAMA2("llama2", null),
 
 	/**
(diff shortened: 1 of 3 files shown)


models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
```diff
@@ -30,118 +30,159 @@ public enum OllamaModel implements ChatModelDescription {
...
 	}
 
+	/**
+	 * for example: qwen2.5:7b,llama3.2:1b
+	 * @return model id with size
+	 */
 	@Override
 	public String getName() {
-		return this.id;
+		return this.id + (null == this.size ? "" : ":" + this.size);
```
ogbozoyan, line 177: Missing Null Handling in getName:

The logic in getName assumes size will be null or non-empty but doesn't guard against unexpected input like an empty string.
Add validation to handle this edge case

ogbozoyan, line 177:
```suggestion
		return this.id + (this.size == null || this.size.isEmpty() ? "" : ":" + this.size);
```

csuwl: If there is no difference between id and name, then that's right

ogbozoyan (approved):

PR #1785, opened 2024-11-20
Use TypesenseContainer for Service Connection

eddumelendez:
Testcontainers 1.20.4 provides a new module for typesense with
TypesenseContainer implementation.

It is a breaking change before RC so it should be fine but if not, please LMK and I can add and additional support.


markpollack:
thanks!  merged in 840304955ef297278df1394ff334df212988402e

Issue #517, opened 2024-03-27
Add advanced RAG (Hybrid Search and Semantic Hybrid Search) with Azure AI Search


iAMSagar44:
Azure AI Search has the following feature along with Vector Similarity Search (which is already present in Spring AI).

- Hybrid Search (Vector Search + Text Search)
- Semantic Hybrid Search (Hybrid search + re-ranking)

In the JavaScript Langchain [project](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/azure_aisearch.ts) they basically have the 3 search models in 3 functions next to each other:

1. similaritySearchVectorWithScore - this is the current vector search Spring AI already has
2. hybridSearchVectorWithScore - this is vector search + text search
3. semanticHybridSearchVectorWithScore - this is hybrid search + reranking

I see something similar was recently introduced in the Langchain4j project too. Would be good to have this feature in Spring AI as well.

markpollack:
Hi.  Yes, this is a good idea.

markpollack:
related issue - https://github.com/spring-projects/spring-ai/issues/579

alessiobertazzo88:
Hello everyone, I have developed this feature to enable users to choose their preferred search mode. Within the SearchRequest, it is possible to use similarity search in line with what is currently available (vector-based, enabled by default), full-text search, or both (hybrid search). Additionally, it provides the option to use reranking approaches as well. 

This implementation is applied to Azure SearchAI as first vector store, but new methods defaults can be overridden to enhance the feature for all the others compatible.

iAMSagar44:
Hi @markpollack / @csterwa / @alessiobertazzo88 

I have done some analysis on using the Hybrid Search with Semantic Ranking in Azure AI Search. The existing code in Spring AI for filtering the search results (response from Azure AI Search) based on the result score and similarity threshold **will not** work. This is due to a different algorithm used by AI Search when semantic ranking is used (please refer to the last section named ‘Ranking’ in this link - [Hybrid Search](https://learn.microsoft.com/en-us/azure/search/hybrid-search-how-to-query)).
In short, with the Hybrid Search, the result.getScore() can appear quite low, even with a high similarity match and the following code, which is currently present in the Spring AI Azure Vector Store class will always return false - .`filter(result -> result.getScore() >= request.getSimilarityThreshold())`

[...]

PR #1794, opened 2024-11-21
Support similarity scores in Document API

ThomasVitale:
Document
* Introduced “score” attribute in Document API. It stores the similarity score.
* Consolidate “distance” metadata for Documents. It stores the distance measurement.
* Adopted prefix-less naming convention in Document.Builder and deprecated old methods.
* Deprecated the many overloaded Document constructors in favour of Document.Builder.

Vector Stores
* Every vector store implementation now configures a “score” attribute with the similarity score of the Document embedding. It also includes the “distance” metadata with the distance measurement.
* Fixed error in Elasticsearch where distance and similarity were mixed up.
* Added missing integration tests for SimpleVectorStore.
* The Azure Vector Store and HanaDB Vector Store do not include those measurements because the product documentation do not include information about how the similarity score is returned, and without access to the cloud products I could not verify that via debugging.
* Improved tests to actually assert the result of the similarity search based on the returned score.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
@@ -71,7 +71,7 @@ public class Document implements MediaContent {
 	 * measure.
 	 */
 	@Nullable
-	private Double score;
+	private final Double score;
 
 	/**
 	 * Embedding of the document. Note: ephemeral field.
@@ -90,10 +90,6 @@ public class Document implements MediaContent {
 		this(content, new HashMap<>());
 	}
 
-	/**
-	 * @deprecated Use builder instead: {@link Document#builder()}.
-	 */
-	@Deprecated(since = "1.0.0-M5", forRemoval = true)
 	public Document(String content, Map<String, Object> metadata) {
 		this(content, metadata, new RandomIdGenerator());
 	}
@@ -114,10 +110,6 @@ public class Document implements MediaContent {
 		this(idGenerator.generateId(content, metadata), content, metadata);
 	}
 
-	/**
-	 * @deprecated Use builder instead: {@link Document#builder()}.
-	 */
-	@Deprecated(since = "1.0.0-M5", forRemoval = true)
 	public Document(String id, String content, Map<String, Object> metadata) {
 		this(id, content, List.of(), metadata);
 	}
@@ -194,10 +186,6 @@ public class Document implements MediaContent {
 		return this.score;
 	}
 
-	public void setScore(@Nullable Double score) {
-		this.score = score;
-	}
-
 	/**
 	 * Return the embedding that were calculated.
 	 * @deprecated We are considering getting rid of this, please comment on
@@ -233,20 +221,27 @@ public class Document implements MediaContent {
 		this.contentFormatter = contentFormatter;
 	}
 
-	@Override
-	public int hashCode() {
-		return Objects.hash(id, content, media, metadata);
+	public Builder mutate() {
+		return new Builder().id(this.id)
+			.content(this.content)
+			.media(new ArrayList<>(this.media))
+			.metadata(this.metadata)
+			.score(this.score);
 	}
 
 	@Override
 	public boolean equals(Object o) {
-		if (this == o)
-			return true;
 		if (o == null || getClass() != o.getClass())
 			return false;
 		Document document = (Document) o;
 		return Objects.equals(id, document.id) && Objects.equals(content, document.content)
-				&& Objects.equals(media, document.media) && Objects.equals(metadata, document.metadata);
+				&& Objects.equals(media, document.media) && Objects.equals(metadata, document.metadata)
+				&& Objects.equals(score, document.score);
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(id, content, media, metadata, score);
 	}
 
 	@Override
@@ -267,6 +262,7 @@ public class Document implements MediaContent {
 
 		private float[] embedding = new float[0];
 
+		@Nullable
 		private Double score;
 
 		private IdGenerator idGenerator = new RandomIdGenerator();
@@ -314,7 +310,7 @@ public class Document implements MediaContent {
 			return this;
 		}
 
-		public Builder score(Double score) {
+		public Builder score(@Nullable Double score) {
 			this.score = score;
 			return this;
 		}
--- a/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
@@ -21,6 +21,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
@@ -31,6 +32,7 @@ import org.springframework.ai.document.id.IdGenerator;
 import org.springframework.ai.document.id.RandomIdGenerator;
 import org.springframework.ai.model.Media;
 import org.springframework.ai.model.MediaContent;
+import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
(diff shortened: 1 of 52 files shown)


spring-ai-core/src/main/java/org/springframework/ai/document/DocumentMetadata.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.ai.vectorstore.VectorStore;
+
+/**
+ * Common set of metadata keys used in {@link Document}s by {@link DocumentReader}s and
+ * {@link VectorStore}s.
+ *
+ * @author Thomas Vitale
+ * @since 1.0.0
+ */
+public enum DocumentMetadata {
```
ThomasVitale, line 28: The idea for this enum is to use it for other common metadata used in Documents, such as the "source file" or "page" when using a `DocumentReader`, helping the RAG flow traceability.

spring-ai-core/src/main/java/org/springframework/ai/document/DocumentMetadata.java
```diff
@@ -0,0 +1,50 @@
...
+
+// @formatter:off
+
+	/**
+	 * Measure of distance between the document embedding and the query vector.
+	 * The lower the distance, the more they are similar.
+	 * It's the opposite of the similarity score.
+	 */
+	DISTANCE("distance");
```
ThomasVitale, line 37: I kept this metadata for backward compatibility, but we might consider removing it completely since we now have the "score" field in each Document (and "distance" is always the opposite value of "score").

spring-ai-core/src/main/java/org/springframework/ai/vectorstore/SimpleVectorStore.java
```diff
@@ -120,12 +122,12 @@ public List<Document> doSimilaritySearch(SearchRequest request) {
 		float[] userQueryEmbedding = getUserQueryEmbedding(request.getQuery());
 		return this.store.values()
 			.stream()
-			.map(entry -> new Similarity(entry.getId(),
-					EmbeddingMath.cosineSimilarity(userQueryEmbedding, entry.getEmbedding())))
-			.filter(s -> s.score >= request.getSimilarityThreshold())
-			.sorted(Comparator.<Similarity>comparingDouble(s -> s.score).reversed())
+			.peek(document -> document
+				.setScore(EmbeddingMath.cosineSimilarity(userQueryEmbedding, document.getEmbedding())))
```
ThomasVitale, line 126: If we remove the "embedding" field, the `SimpleVectorStore` will not work

vector-stores/spring-ai-azure-cosmos-db-store/src/test/java/org/springframework/ai/vectorstore/CosmosDbImage.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * @author Thomas Vitale
+ */
+public final class CosmosDbImage {
+
+	// It must always be "latest" or else Azure locks the image after a while. See:
+	// https://github.com/Azure/azure-cosmos-db-emulator-docker/issues/60
+	public static final DockerImageName DEFAULT_IMAGE = DockerImageName
+		.parse("mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:latest");
```
ThomasVitale, line 29: I hope we'll be able to have integration tests for CosmosDB based on Testcontainers in the future. For now, this image includes the vector store-specific features disabled and there's no way to enable them, so it cannot be used.

vector-stores/spring-ai-pinecone-store/src/main/java/org/springframework/ai/vectorstore/PineconeVectorStore.java
```diff
@@ -298,6 +303,8 @@ public static final class PineconeVectorStoreConfig {
 
 		private final String contentFieldName;
 
+		// TODO: Why is this field configurable? Can we remove this after standardizing
```
ThomasVitale, line 306: Would it be ok to remove this and keep the standard "distance" metadata? Having this configurable means we cannot use the metadata reliably across implementations.

vector-stores/spring-ai-redis-store/src/main/java/org/springframework/ai/vectorstore/RedisVectorStore.java
```diff
@@ -240,14 +241,21 @@ public List<Document> doSimilaritySearch(SearchRequest request) {
...
 	private Document toDocument(redis.clients.jedis.search.Document doc) {
 		var id = doc.getId().substring(this.config.prefix.length());
-		var content = doc.hasProperty(this.config.contentFieldName) ? doc.getString(this.config.contentFieldName)
-				: null;
+		var content = doc.hasProperty(this.config.contentFieldName) ? doc.getString(this.config.contentFieldName) : "";
 		Map<String, Object> metadata = this.config.metadataFields.stream()
 			.map(MetadataField::name)
 			.filter(doc::hasProperty)
 			.collect(Collectors.toMap(Function.identity(), doc::getString));
+		// TODO: this seems wrong. The key is named "vector_store", but the value is the
```
ThomasVitale, line 249: Would it be ok to remove this and keep the standard "distance" metadata?

ThomasVitale:
PR updated after https://github.com/spring-projects/spring-ai/pull/1822 was merged

spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
```diff
@@ -149,6 +189,15 @@ public Map<String, Object> getMetadata() {
...
 	}
 
+	@Nullable
+	public Double getScore() {
+		return this.score;
+	}
+
+	public void setScore(@Nullable Double score) {
+		this.score = score;
+	}
```
markpollack, line 199: Should the document object be immutable?  If we want to update a Document with a score we should use a builder that takes the existing document and then call the 'score' builder method?

spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
```diff
@@ -80,31 +90,61 @@ public Document(@JsonProperty("content") String content) {
 		this(content, new HashMap<>());
 	}
 
+	/**
+	 * @deprecated Use builder instead: {@link Document#builder()}.
+	 */
+	@Deprecated(since = "1.0.0-M5", forRemoval = true)
 	public Document(String content, Map<String, Object> metadata) {
```
markpollack, line 97: There are probably many users out there using this constructor and perhaps the one with `String id, String content, Map<String, Object> metadata` args.  Despite the having the builder, maybe we keep these ctors?

spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
ThomasVitale: The post-retrieval steps in a RAG flow would all modify the Documents in some way, including the score, the content and the metadata. I guess we could create new instances on each hop to the next step in the flow. But we'd need several builders based on the type of field we change. Should I go with that?

spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
markpollack: yea, the two are good, once we get to three is when the ambiguity starts so the builder should be preferred

spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
```diff
@@ -249,56 +265,102 @@ public static class Builder {
...
-			Assert.notNull(content, "content must not be null");
+		public Builder content(String content) {
 			this.content = content;
 			return this;
 		}
 
-		public Builder withMedia(List<Media> media) {
-			Assert.notNull(media, "media must not be null");
+		public Builder media(List<Media> media) {
 			this.media = media;
```
markpollack, line 288: i've updated it so that it adds to the existing list vs. replacing it.  There are tests that assume it aggregates.

spring-ai-core/src/main/java/org/springframework/ai/document/DocumentMetadata.java
markpollack: I added a `toString()` method so that `.value()` doesn't need to be used when calling, makes the calling more terse.

markpollack:
I removed usage of deprecated methods and some other minor cleanup.  Sorry it took a while.
We need another pass at improving Document as you noted in #1838

merged in fe58fd30ebd649985d459200fe23469b1077798e

iAMSagar44:
@ThomasVitale - Regarding the scores returned by Azure AI Search, please have a read of my latest comment here -
[517](https://github.com/spring-projects/spring-ai/issues/517).

Depending on the type of search you carry out in Azure AI Search, the score might be different and its usage will be different.

PR #1798, opened 2024-11-22
chore: Update dependencies and fix OpenAPI client generation

tzolov:
- Upgrade Spring Boot to 3.4.0
- Update swagger-codegen-maven-plugin to 3.0.64
- Add custom template for HttpBasicAuth
- Fix mockwebserver version to 4.12.0
- Fix parent version reference in opensearch-store

NOTE: Revert the boot to 3.3.6 before merging!!!

### DIFF
--- a/spring-ai-test/pom.xml
+++ b/spring-ai-test/pom.xml
@@ -70,7 +70,7 @@
 		<dependency>
 			<groupId>com.squareup.okhttp3</groupId>
 			<artifactId>mockwebserver</artifactId>
-			<version>4.12.0</version>
+			<version>${okhttp3.version}</version>
 		</dependency>
 	</dependencies>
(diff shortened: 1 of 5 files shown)


spring-ai-test/pom.xml
```diff
@@ -70,6 +70,7 @@
 		<dependency>
 			<groupId>com.squareup.okhttp3</groupId>
 			<artifactId>mockwebserver</artifactId>
+			<version>4.12.0</version>
```
ilayaperumalg, line 73: We can convert this into a property for better operability

tzolov:
Changed the boot version to 3.3.6, rebased, squashed and merged at c05714881303e9b43ab710ec900cef11f3dbd9e4

Issue #942, opened 2024-06-24
VertexAiGeminiChatModel: Make safetySettings configurable


patrick-zinner:
**Expected Behavior**

It should be possible for each call (and globally) to define the desired safety settings for each harm category.

**Current Behavior**

Right now there is no way at all to configure these settings. The default settings (which I think are medium) are used. For many nonharmful use-cases Gemini refuses to generate a response.

**Context**

The Gemini API provides safety settings where you can configure the level at which certain harmful content (e.g., hate speech) should be blocked. Gemini seems to interpret what is harmful rather strictly, and even nonharmful content will be blocked when using the default settings. The protobuf model that Spring AI uses has fields to configure these settings, but unfortunately, there is no way to configure them in the Spring abstraction.

ddobrin:
The Gemini safety filters are being provided in detail here:
https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/configure-safety-filters 

Assign this issue to me and I will look into it for the M4 milestone or next one

ilayaperumalg:
Safety Settings are currently supported. Thanks @terussell85 for the PR adding support. Closing this as addressed.

More documentation on this can be referred from [here](https://docs.spring.io/spring-ai/reference/api/chat/vertexai-gemini-chat.html#_chat_properties)

PR #1803, opened 2024-11-23
feat(gemini) Support Safety Settings for VertexAiGeminiChatModel

terussell85:
Adds support for SafetySettings configuration to `VertexAiGeminiChatModel` and `VertexAiGeminiChatOptions`

Should fix https://github.com/spring-projects/spring-ai/issues/942

### DIFF
--- a/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatModel.java
+++ b/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatModel.java
@@ -453,7 +453,8 @@ public class VertexAiGeminiChatModel extends AbstractToolCallSupport implements 
 		GenerationConfig generationConfig = this.generationConfig;
 
 		var generativeModelBuilder = new GenerativeModel.Builder().setModelName(this.defaultOptions.getModel())
-			.setVertexAi(this.vertexAI);
+			.setVertexAi(this.vertexAI)
+			.setSafetySettings(this.defaultOptions.getSafetySettings());
 
 		if (prompt.getOptions() != null) {
 			if (prompt.getOptions() instanceof FunctionCallingOptions functionCallingOptions) {
@@ -497,6 +498,10 @@ public class VertexAiGeminiChatModel extends AbstractToolCallSupport implements 
 		}
 		if (!CollectionUtils.isEmpty(tools)) {
 			generativeModelBuilder.setTools(tools);
+		}
+
+		if (prompt.getOptions() instanceof VertexAiGeminiChatOptions options && options.getSafetySettings() != null) {
+			generativeModelBuilder.setSafetySettings(options.getSafetySettings());
 		}
 
 		generativeModelBuilder.setGenerationConfig(generationConfig);
--- a/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatOptions.java
+++ b/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatOptions.java
@@ -28,6 +28,7 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import com.google.cloud.vertexai.api.SafetySetting;
 import org.springframework.ai.model.function.FunctionCallback;
 import org.springframework.ai.model.function.FunctionCallingOptions;
 import org.springframework.ai.vertexai.gemini.VertexAiGeminiChatModel.ChatModel;
@@ -118,6 +119,9 @@ public class VertexAiGeminiChatOptions implements FunctionCallingOptions {
 	private boolean googleSearchRetrieval = false;
 
 	@JsonIgnore
+	private List<SafetySetting> safetySettings = new ArrayList<>();
+
+	@JsonIgnore
 	private Boolean proxyToolCalls;
 
 	@JsonIgnore
@@ -143,6 +147,7 @@ public class VertexAiGeminiChatOptions implements FunctionCallingOptions {
 		options.setFunctions(fromOptions.getFunctions());
 		options.setResponseMimeType(fromOptions.getResponseMimeType());
 		options.setGoogleSearchRetrieval(fromOptions.getGoogleSearchRetrieval());
+		options.setSafetySettings(fromOptions.getSafetySettings());
 		options.setProxyToolCalls(fromOptions.getProxyToolCalls());
 		options.setToolContext(fromOptions.getToolContext());
 		return options;
@@ -267,6 +272,14 @@ public class VertexAiGeminiChatOptions implements FunctionCallingOptions {
 
 	public void setGoogleSearchRetrieval(boolean googleSearchRetrieval) {
 		this.googleSearchRetrieval = googleSearchRetrieval;
+	}
+
+	public List<SafetySetting> getSafetySettings() {
+		return safetySettings;
+	}
+
+	public void setSafetySettings(List<SafetySetting> safetySettings) {
+		this.safetySettings = safetySettings;
 	}
 
 	@Override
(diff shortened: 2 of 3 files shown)


terussell85:
@tzolov I updated the code with updates from the `spring-javaformat` plugin. Apologies that I didn't catch that the first time.

models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatOptions.java
```diff
@@ -28,6 +28,7 @@
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import com.google.cloud.vertexai.api.SafetySetting;
```
tzolov, line 31: Spilling, an underlying, 3rd party code in the Spring AI API, is a design anti-pattern.
But trying to create Spring AI shallow wrappers doesn't make much sense either.
I guess, given that the `VertextAiGeminiChatModelOptions` is meant to allow breaking portability (e.g., offers model-specific configurations), means that we can allow such spilling. Furthermore we've done it with other provides too. I hope it won't bite us later.

On the other hand i'm not sure you can use the `spring.ai.vertex.ai.gemini.chat.options.SafetySettings` to configure the SafetySettings declaratively via properties? have you tried this.

models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatOptions.java
```diff
@@ -269,6 +274,14 @@ public void setGoogleSearchRetrieval(boolean googleSearchRetrieval) {
 		this.googleSearchRetrieval = googleSearchRetrieval;
 	}
 
+	public List<SafetySetting> getSafetySettings() {
+		return safetySettings;
+	}
+
+	public void setSafetySettings(List<SafetySetting> safetySettings) {
```
tzolov, line 281: Please add non-null assertion check.

models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatModel.java
```diff
@@ -499,6 +500,10 @@ GeminiRequest createGeminiRequest(Prompt prompt, VertexAiGeminiChatOptions updat
 			generativeModelBuilder.setTools(tools);
 		}
 
+		if (prompt.getOptions() instanceof VertexAiGeminiChatOptions options && options.getSafetySettings() != null) {
```
tzolov, line 503: You have to check for `!CollectionUtils.isEmpty(options.getSafetySettings())` instead. 
Since the `options.getSafetySettings()` would never be null (you are creating empty set by default) it will always override the default SafetySettings. 
It is anther question as how to set the SafetySettings via the `spring.ai.vertex.ai.gemini.chat.options.SafetySettings`?

models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatOptions.java
```diff
@@ -408,6 +421,11 @@ public Builder withGoogleSearchRetrieval(boolean googleSearch) {
 			return this;
 		}
 
+		public Builder withSafetySettings(List<SafetySetting> safetySettings) {
```
tzolov, line 424: Non-null assertion check

tzolov (commented):
Hi @terussell85 , thank you, for the important contribution! 

I've left couple of comments.  The most important one is to verify if we can use the `spring.ai.vertex.ai.gemini.chat.options.SafetySetting` to configure the SafetySettings declaratively via app. properties. You can check the auto-config ITs [here](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-spring-boot-autoconfigure/src/test/java/org/springframework/ai/autoconfigure/vertexai/gemini/VertexAiGeminiAutoConfigurationIT.java).

Also can you please check if you can update the VertexAIGemini docs [here](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/vertexai-gemini-chat.adoc). 
You can build the docs locally with: `./mvnw -pl spring-ai-docs antora   `.

terussell85:
Great feedback. I'll take a look and get back to you.

models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatOptions.java
markpollack: here is a claude generated POJO class from the protobuff class that would be friendly for declarative configuration in yaml

```
/**
 * A Plain Old Java Object (POJO) representing safety settings for content filtering.
 */
public class SafetySetting {
    
    /**
     * Enum representing different threshold levels for blocking harmful content.
     */
    public enum HarmBlockThreshold {
        UNSPECIFIED,
        BLOCK_LOW_AND_ABOVE,    // Block low threshold and above (more strict)
        BLOCK_MEDIUM_AND_ABOVE, // Block medium threshold and above
        BLOCK_ONLY_HIGH,        // Block only high threshold (less strict)
        BLOCK_NONE,             // Don't block any content
        OFF                     // Turn off safety filter completely
    }
    
    /**
     * Enum representing methods for evaluating harmful content.
     */
    public enum HarmBlockMethod {
        UNSPECIFIED,
        SEVERITY,    // Uses both probability and severity scores
        PROBABILITY  // Uses only probability score
    }
    
    /**
     * Enum representing different categories of harmful content.
     */
    public enum HarmCategory {
        UNSPECIFIED,
        HATE_SPEECH,
        DANGEROUS_CONTENT,
        HARASSMENT,
        SEXUAL_CONTENT
    }
    
    private HarmCategory category;
    private HarmBlockThreshold threshold;
    private HarmBlockMethod method;
    
    // Default constructor
    public SafetySetting() {
        this.category = HarmCategory.UNSPECIFIED;
        this.threshold = HarmBlockThreshold.UNSPECIFIED;
        this.method = HarmBlockMethod.UNSPECIFIED;
    }
    
    // Constructor with all fields
    public SafetySetting(HarmCategory category, HarmBlockThreshold threshold, HarmBlockMethod method) {
        this.category = category;
        this.threshold = threshold;
        this.method = method;
    }
    
    // Getters and setters
    public HarmCategory getCategory() {
        return category;
    }
    
    public void setCategory(HarmCategory category) {
        this.category = category;
    }
    
    public HarmBlockThreshold getThreshold() {
        return threshold;
    }
    
    public void setThreshold(HarmBlockThreshold threshold) {
        this.threshold = threshold;
    }
    
    public HarmBlockMethod getMethod() {
        return method;
    }
    
    public void setMethod(HarmBlockMethod method) {
        this.method = method;
    }
    
    @Override
    public String toString() {
        return "SafetySetting{" +
               "category=" + category +
               ", threshold=" + threshold +
               ", method=" + method +
               '}';
    }
    
    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;
        
        SafetySetting that = (SafetySetting) o;
        
        if (category != that.category) return false;
        if (threshold != that.threshold) return false;
        return method == that.method;
    }
    
    @Override
    public int hashCode() {
        int result = category != null ? category.hashCode() : 0;
        result = 31 * result + (threshold != null ? threshold.hashCode() : 0);
        result = 31 * result + (method != null ? method.hashCode() : 0);
        return result;
    }
}
```

models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatOptions.java
terussell85: @tzolov Other properties have their assertions on the Builder, not the Options itself (see `mimeType` and `functions`). Would you prefer to favor consistency in this case or set it on the Options class?

terussell85: I added the check in both places. Let me know if you'd like to favor consistency and I'll remove it from the class setter.

terussell85:
@tzolov I think we're good to go now. 

Changes since last review:
1. Introduce new `VertexAiGeminiSafetySetting` class to allow for declarative configuration. This is converted to the Vertex AI setting at request-time.
2. Not null assertions for `safetySettings` on both the Builder and the class property.
3. Verified that declarative configuration works
4. Updated docs

I made some changes to the originally suggested SafetySetting class, including renaming it and adding a Builder. I also kept the Spring-defined enums both value- and name- consistent with the Gemini enums so that there won't be accidental confusion of users.

Let me know if there is anything else.

tzolov:
thanks @terussell85

Issue #1796, opened 2024-11-22
Ollama: Duration metrics deserialized to the wrong time unit


jorander:
**Bug description**
Ollama reports duration values (total_duration, load_duration, prompt_eval_duration and eval_duration) in chat and embed responses as [nanoseconds in JSON INTEGER format](https://github.com/ollama/ollama/blob/main/docs/api.md#durations). In Spring AI the Jackson class [DurationDeserializer](https://github.com/FasterXML/jackson-datatype-jsr310/blob/master/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java) from the jsr310 module is used to  deserialize these values into Java Duration objects. In this process the integer value is interpreted as seconds, instead of nanoseconds, making the Duration 10^9 times larger. 

This is because the DurationDeserializer expects [durations with nanosecond precision to be formatted as decimal values with a decimal separator (dot) separating the seconds part from the nanoseconds part](https://github.com/FasterXML/jackson-datatype-jsr310/blob/master/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java#L52-L56). [Durations formatted as integers are, depending on context settings, interpreted as seconds (default) or milliseconds](https://github.com/FasterXML/jackson-datatype-jsr310/blob/master/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java#L58-L62). None of these work for durations reported by Ollama.

**Environment**
Spring AI: 1.0.0-M4
Java: 21
Ollama: 0.4.2

**Steps to reproduce**
Run a chat-request towards an Ollama server.
Compare the durations reported by Ollama with the values found in the OllamaApi.ChatResponse and propagated to the map available at ChatClient.ChatResponse#getMetadata.

**Expected behavior**
Serialization of Duration takes into account that Ollama reports duration values as a JSON INTEGER value in nanoseconds.

**Minimal Complete Reproducible example**
None at the moment.

jorander:
Modified unit tests to reproduce error in connected PR.

PR #1808, opened 2024-11-24
Issue #1796: Bugfix for serialization of Ollama duration metrics

jorander:
…ime unit is used while deserializing the duration values coming from Ollama.

Duration metrics are sent in nano seconds by Ollama, but interpreted as seconds by Spring AI. See issue #1796 for more details.

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
@@ -16,18 +16,9 @@
 
 package org.springframework.ai.ollama.api;
 
-import java.io.IOException;
-import java.time.Duration;
-import java.time.Instant;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.function.Consumer;
-
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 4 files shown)


jorander:
Hi @markpollack,

I've started working on a fix based on annotating attributed of type Duration with @JsonDeserialize and a custom deserializing-class handling the JSON INT as nanoseconds.

For symmetry I've thought about also adding @JsonSerialize that converts the Duration back to a JSON INT of nanoseconds, Event though it is not strictly necessary for the current use-case it will otherwise be surprising results for anyone attempting to serialize a OllamaApi.ChatResponse.

Finally, I'm also thinking that it might be a good thing to change the type of totalDuration and loadDuration in OllamaApi.EmbeddingsResponse from Long to Duration (and add the same serialization/deserialization) to make the two classes more similar.

If you agree on these, or part of them, I'll be happy to work on a PR including those changes.

jorander:
Added changes as outlined in comment above.

Note: For OllamaApi.EmbeddingsResponse this is an api-breaking change since datatypes for duration-attributes are changed from Long to Duration.

jorander:
Closing this PR since it is no longer needed. Issue #1796 is solved in https://github.com/spring-projects/spring-ai/commit/c4ca5826f79db851e466c9513818550cad64639e

Issue #1819, opened 2024-11-25
Add Anthropic PDF input modality support


tzolov:
Latest Sonnet 3.5 offers (beta) support for [PDF input modality](https://docs.anthropic.com/en/docs/build-with-claude/pdf-support).
Extend the Spring AI Anthropic multimodality support to allow support for PDF input types as well.

PR #1821, opened 2024-11-25
Add Anthropic PDF Multimodal Support

tzolov:
- Enhance AnthropicChatModel to support PDF and document content types
- Introduce getContentBlockTypeByMedia method for flexible media type handling
- Update ContentBlock handling to dynamically determine content type for media
- Add multimodal PDF support test case for Claude 3.5 Sonnet
- Update documentation to reflect PDF and multimodal capabilities
- Modify comparison chart to show PDF support for Anthropic Claude

Release #1819


models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -355,6 +357,16 @@ else if (mediaData instanceof String text) {
...
 	}
 
+	private Type getContentBlockTypeByMedia(Media media) {
+		if (media.getMimeType().toString().startsWith("image")) {
+			return Type.IMAGE;
+		}
+		else if (media.getMimeType().toString().contains("pdf")) {
+			return Type.DOCUMENT;
+		}
+		return Type.IMAGE;
```
markpollack, line 367: if it wasn't classified as image, prob better to throw an exception

markpollack:
updated to throw exception if mime type is not know.

merged in e543cd082d86dfb71702924bcff56e16621d1eb4

PR #1822, opened 2024-11-25
Refactor SimpleVectorStore

ilayaperumalg:
- Remove SimpleVectorStore's dependency on deprecated embeddings from Document object
   - Create a custom Content object that represents the SimpleVectorStore's contents and embedding


markpollack:
I started off making `SimpleVectorStoreContent` immutable and then let Claude loose to write more tests.  Pretty impressive!

merged in 9d207e37bed5809f9f4ad1c31ff658f613222d6f

Issue #1823, opened 2024-11-25
VectorStoreChatMemoryAdvisor builder class should have additional methods to set all properties


markpollack:
Currently, `order` and other properties are not able to be set by the builder.  We should deprecate most of the current constructors and use the builder.

ogbozoyan:
Agree with you. Builder a much prettier approach

ogbozoyan:
@markpollack could you review please, may be some constructors shouldn't be deprecate

markpollack:
The ctor is now private.  While reviewing this issue one of the properties was not in the builder, this has been fixed in 6ae1c13aeefe462448c8e795200c09c69f6f8554

PR #1824, opened 2024-11-25
feat: added missed builder methods for VectorStoreChatMemoryAdvisor; 

ogbozoyan:
Issue: https://github.com/spring-projects/spring-ai/issues/1823 

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/VectorStoreChatMemoryAdvisor.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/VectorStoreChatMemoryAdvisor.java
@@ -44,6 +44,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Christian Tzolov
  * @author Thomas Vitale
+ * @author Oganes Bozoyan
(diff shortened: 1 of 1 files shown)


spring-ai-core/src/main/java/org/springframework/ai/chat/client/advisor/VectorStoreChatMemoryAdvisor.java
```diff
@@ -209,10 +219,25 @@ public Builder withSystemTextAdvise(String systemTextAdvise) {
 			return this;
 		}
 
+		public Builder withConversationId(String conversationId) {
```
ilayaperumalg, line 222: @ogbozoyan Thanks for your PR. These methods are already part of the AbstractChatMemoryAdvisor.AbstractBuilder<VectorStore> and hence they don't need to be added here. I will remove these but merge your PR with the deprecated constructors.

ilayaperumalg:
Rebased, removed the builder methods which were part of the AbstractChatMemoryAdvisor's Builder and merged as cd871eadd. @ogbozoyan Thanks for your PR!

PR #1835, opened 2024-11-27
Portable ChatOptions ignored by OpenAiChatModel

ThomasVitale:
When calling the `OpenAiChatModel` with `ChatOptions`, they are always ignored. Options are considered only when using `OpenAiChatOptions` or `FunctionCallingOptions`.

This seems a regression after the recent refactoring of `FunctionCallingOptions`. I checked all the model implementation and it's only OpenAI having this error.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -473,7 +473,7 @@ public class OpenAiChatModel extends AbstractToolCallSupport implements ChatMode
 				updatedRuntimeOptions = ModelOptionsUtils.copyToTarget(((FunctionCallingOptions) prompt.getOptions()),
 						FunctionCallingOptions.class, OpenAiChatOptions.class);
 			}
-			else if (prompt.getOptions() instanceof OpenAiChatOptions) {
+			else {
 				updatedRuntimeOptions = ModelOptionsUtils.copyToTarget(prompt.getOptions(), ChatOptions.class,
 						OpenAiChatOptions.class);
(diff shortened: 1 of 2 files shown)


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -473,7 +473,7 @@ else if (message.getMessageType() == MessageType.TOOL) {
 				updatedRuntimeOptions = ModelOptionsUtils.copyToTarget(((FunctionCallingOptions) prompt.getOptions()),
 						FunctionCallingOptions.class, OpenAiChatOptions.class);
 			}
-			else if (prompt.getOptions() instanceof OpenAiChatOptions) {
+			else {
```
ThomasVitale, line 476: This is the same implementation as all the other model integrations. The problem was only in OpenAI.

tzolov:
I've noticed this issue last week. Initially the `FunctionCallingOptions` where NOT extending from `ChatOptions`. 
So the above implementation was making sense. But since we made FunctionCallingOptions extends ChatOptions this obviously a bogus behavior and IMO we need to fix for all models.

ThomasVitale:
@tzolov I haven't seen the bug in the other chat implementations since they all already have an "else" statement that makes sure the portable options are mapped correctly. Can we get the bug fixed in OpenAI with this PR and then have the refactoring to improve the logic in a separate issue, now that `FunctionCallingOptions` extends `ChatOptions`?

tzolov:
@ThomasVitale i guess you are right. I was concerned about the order, but it correct (e.g. from more specific to the generic type)
Thanks for the fix!

Issue #1826, opened 2024-11-25
Refactor VectoStore's usage on Document.getEmbedding


ilayaperumalg:
As a result of deprecating `embedding` from the `Document` object, refer: https://github.com/spring-projects/spring-ai/issues/1781, some of the vector store operations, specifically the `doSimilaritySearch` method relies on `Document#getEmbedding`. 

This issue addresses the refactoring of this dependency.

PR #1836, opened 2024-11-27
GH-1826 Fix EmbeddingModel's usage on Document#embedding

ilayaperumalg:
- Since the Document object's reference to the `embedding` is deprecated and will be removed, the VectorStore implementations require a way to store the embedding of the corresponding Document objects  
  
    - One way to fix this is, to have the EmbeddingModel#embed to return the embeddings in the same order as that of the Documents passed to it. 
         - Since both the Document and embedding collections use the List object, their iteration operation will make sure to keep them in line with the same order. 
              - A fix is required to preserve the order when batching strategy is applied.
	          - Updated the Javadoc for BatchingStrategy -
	          -  Fixed the Document List order in TokenCountBatchingStrategy

    - Refactored the vector store implementations to update this change

Resolves #GH-1826


tzolov:
rebased, cleaned, squashed and merged at : ebd29e0959086ea950d505a8aefb9e32ca7732a7

PR #1837, opened 2024-11-27
refactor(core): extract common function callback builder functionality

tzolov:
Extracts shared function callback builder functionality into DefaultCommonInvokingSpec base class, reducing code duplication across builder implementations. Makes FunctionInvokingSpec and MethodInvokingSpec extend CommonInvokingSpec for better code organization. Also fixes function/description builder order in Anthropic tests.

- Introduced a common base class for function callback builders to centralize shared logic
- Standardized the order of method chaining for function and description in multiple AI model test classes
- Refactored test cases across various AI model integrations
- Corrected builder method order from .description().function() to .function().description() and .description().method() to .method().description()
- Updated multiple test files to consistently use .function() before .description()
- Updated documentation examples to reflect new builder method order
- Modified DefaultFunctionCallbackResolver to maintain new builder method order
- Updated DefaultChatClient and ChatClient test classes to reflect new builder pattern


spring-ai-core/src/main/java/org/springframework/ai/model/function/DefaultFunctionCallbackBuilder.java
```diff
@@ -50,86 +44,11 @@
  * @author Christian Tzolov
  * @since 1.0.0
  */
-public class DefaultFunctionCallbackBuilder implements FunctionCallback.Builder {
+public class DefaultFunctionCallbackBuilder extends DefaultCommonInvokingSpec<FunctionCallback.Builder>
```
tzolov, line 47: I've made the DefaultFunctionCallbackBuilder extend DefaultCommonInvokingSpec only for backward compatibility. 
But this can be confusing and might be better to remove it before M5?

spring-ai-core/src/main/java/org/springframework/ai/model/function/FunctionCallback.java
```diff
@@ -116,63 +116,67 @@ enum SchemaType {
 	 * <li>{@link MethodInvokingSpec} - The method invoking builder interface.
 	 * </ul>
 	 */
-	interface Builder {
+	interface Builder extends CommonInvokingSpec<Builder> {
```
tzolov, line 119: The Builder extends the CommonInvokingSpec<Builder> only for backward compatibility. 
But this might cause confusion and requires merging logic inside the function() and method() leafs. 
Perhaps we can remove it all together causing a breaking change?

spring-ai-core/src/main/java/org/springframework/ai/model/function/DefaultFunctionCallbackBuilder.java
ilayaperumalg: I vote for removing it before M5 as the current common builder spec is confusing anyway and the users would prefer to switch than stick to it.

spring-ai-core/src/main/java/org/springframework/ai/model/function/FunctionCallback.java
```diff
@@ -116,63 +116,67 @@ enum SchemaType {
...
 		/**
-		 * Custom object mapper for JSON operations.
+		 * Builds a Method invoking {@link FunctionCallback} instance.
 		 */
-		Builder objectMapper(ObjectMapper objectMapper);
+		MethodInvokingSpec method(String methodName, Class<?>... argumentTypes);
+
+	}
+
+	interface CommonInvokingSpec<B extends CommonInvokingSpec<B>> {
```
ilayaperumalg, line 148: Instead of `CommonInvokingSpec`, can we name this as `CommonCallbackInvokingSpec`?

ilayaperumalg:
Rebased, updated minor changes javadoc, fix on PerplexityWithOpenAiChatModelIT and merged as 6a195ee9

PR #1872, opened 2024-12-05
Fix OpenAI ChatResponse usage calculation when toolcalling is used

ilayaperumalg:
- Fix OpenAI ChatModel's call() operation

   - When toolcalling is used, calculate cumulative usage from the preceding ChatResponses

 - Fix OpenAI ChatModel's stream() operation

   - Make sure that cumulative usage is calculated from the ChatResponse which has a valid usage
     - Use overlapping buffer to check and store the usage from the response that holds the usage.

 - Add tests for both call() and stream()


ilayaperumalg:
This was merged as 37cf20dfe4679a17e96759c139b4ba3ec765fe9c

PR #1873, opened 2024-12-05
Ensure matching enum id to value for multilingual cohere embedding model

bruno-oliveira:
This ensures that the enum name matches the actual value from Bedrock as a provider.


bruno-oliveira:
FYI @tzolov :)

markpollack:
tx, sorry it took a while.  merged in 2aaa57fc7abf097d40dc2bedb5c0242620615b03

PR #1884, opened 2024-12-06
Enable tests on PRs

eddumelendez:
It only executes unit tests.

Checkstyle and Spring Java Format are executed.


tzolov:
Thanks @eddumelendez Lets give it a try

PR #1889, opened 2024-12-08
refactor: move PgVectorStore to dedicated package and update builder …

sobychacko:
…pattern

 - Move PgVectorStore and related classes to org.springframework.ai.pg.vectorstore package
 - Update builder pattern to use more idiomatic method names (e.g. withSchemaName -> schemaName)
 - Deprecate existing constructors and old Builder class in favor of new static builder() method
 - Update tests to reflect the new builder style usage

### DIFF
--- a/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/pgvector/PgVectorStoreAutoConfiguration.java
+++ b/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/pgvector/PgVectorStoreAutoConfiguration.java
@@ -62,7 +62,8 @@ public class PgVectorStoreAutoConfiguration {
 
 		var initializeSchema = properties.isInitializeSchema();
 
-		return PgVectorStore.builder(jdbcTemplate)
+		return PgVectorStore.builder()
+			.jdbcTemplate(jdbcTemplate)
 			.embeddingModel(embeddingModel)
 			.schemaName(properties.getSchemaName())
 			.vectorTableName(properties.getTableName())
--- a/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/pgvector/PgVectorStoreAutoConfiguration.java
+++ b/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/pgvector/PgVectorStoreAutoConfiguration.java
@@ -23,7 +23,7 @@ import io.micrometer.observation.ObservationRegistry;
 import org.springframework.ai.embedding.BatchingStrategy;
 import org.springframework.ai.embedding.EmbeddingModel;
 import org.springframework.ai.embedding.TokenCountBatchingStrategy;
-import org.springframework.ai.vectorstore.PgVectorStore;
+import org.springframework.ai.pg.vectorstore.PgVectorStore;
 import org.springframework.ai.vectorstore.observation.VectorStoreObservationConvention;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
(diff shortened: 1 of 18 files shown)


spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/pgvector/PgVectorStoreAutoConfiguration.java
```diff
@@ -62,18 +62,20 @@ public PgVectorStore vectorStore(JdbcTemplate jdbcTemplate, EmbeddingModel embed
 
 		var initializeSchema = properties.isInitializeSchema();
 
-		return new PgVectorStore.Builder(jdbcTemplate, embeddingModel).withSchemaName(properties.getSchemaName())
```
ilayaperumalg, line 65: General thought: Given every vector store requires a client (jdbctemplate in this case) and an embedding model, can we make both of these as the default argument to the vector store builder?

vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/pg/vectorstore/PgVectorStore.java
```diff
@@ -123,77 +125,88 @@ public class PgVectorStore extends AbstractObservationVectorStore implements Ini
...
-	}
-
-	private PgVectorStore(String schemaName, String vectorTableName, boolean vectorTableValidationsEnabled,
-			JdbcTemplate jdbcTemplate, EmbeddingModel embeddingModel, int dimensions, PgDistanceType distanceType,
-			boolean removeExistingVectorStoreTable, PgIndexType createIndexMethod, boolean initializeSchema,
-			ObservationRegistry observationRegistry, VectorStoreObservationConvention customObservationConvention,
-			BatchingStrategy batchingStrategy, int maxDocumentBatchSize) {
-
-		super(observationRegistry, customObservationConvention);
+	// @Deprecated(forRemoval = true, since = "1.0.0-M5")
```
ilayaperumalg, line 128: I guess the plan was to keep the methods by deprecating instead of the removing?

vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/pg/vectorstore/PgVectorStore.java
```diff
@@ -628,11 +727,19 @@ public Builder withMaxDocumentBatchSize(int maxDocumentBatchSize) {
 		}
 
 		public PgVectorStore build() {
-			return new PgVectorStore(this.schemaName, this.vectorTableName, this.vectorTableValidationsEnabled,
-					this.jdbcTemplate, this.embeddingModel, this.dimensions, this.distanceType,
-					this.removeExistingVectorStoreTable, this.indexType, this.initializeSchema,
-					this.observationRegistry, this.searchObservationConvention, this.batchingStrategy,
-					this.maxDocumentBatchSize);
+			return PgVectorStore.builder(this.jdbcTemplate)
```
ilayaperumalg, line 730: Both `observationRegistry` and `searchObservationConvention` are missing in this builder setup?

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/pgvector/PgVectorStoreAutoConfiguration.java
markpollack: the design we are using is to have zero arg builder consistent throughout the project.  Reason is imagine a class that need 5 strings as required, if the builder takes 5 args, we haven't achieved any simplification by adding a fluent api.

vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/pg/vectorstore/PgVectorStore.java
markpollack: yes

markpollack:
The postgresql jdbc driver uses `org.postgresql` so instead of `org.springframework.ai.pg.vectorstore` using `org.springframework.ai.postgresql.vectorstore` would be a better naming.

markpollack:
updated reference docs and javadoc.  change to use `pgvector` in package name.

Merged in 677a18e3d4e54139181898e23ae8beb1cf2ad3ed

PR #1894, opened 2024-12-09
refactor: move Elasticsearch classes to dedicated package and use bui…

sobychacko:
…lder pattern


vector-stores/spring-ai-elasticsearch-store/src/main/java/org/springframework/ai/elasticsearch/vectorstore/ElasticsearchVectorStore.java
```diff
@@ -297,4 +305,94 @@ private String getSimilarityMetric() {
...
+		public ElasticsearchBuilder initializeSchema(boolean initializeSchema) {
+			this.initializeSchema = initializeSchema;
+			return this;
+		}
+
+		/**
+		 * Sets the batching strategy for vector operations.
+		 * @param batchingStrategy the batching strategy to use
+		 * @return the builder instance
+		 * @throws IllegalArgumentException if batchingStrategy is null
```
ilayaperumalg, line 365: Where do we throw IllegalArgumentException? Same question for other places we specify that the exception is thrown for invalid values.

sobychacko: In the method implementation, the `Assert.notNull(..)` call would throw the IAE.

markpollack:
updated ref docs and class javadoc for ElasticSearchVectorStore.

merged in fc1f92d11cacc79bdf8d41eb7f941a64fdf67482

PR #1898, opened 2024-12-10
refactor ChatOptions Builder

### LINKED ISSUE #1875: DefaultChatOptions should not be an inner class ChatOptionsBuilder, the build method should be inside the interface ChatOptions
This is similar to how we have ChatClient.

We can deprecate the top level ChatOptionsBuilder class and have the two sit side by side for one release cycle.

ilayaperumalg:
- Deprecate existing ChatOptionsBuilder and its inner class DefaultChatOptions
 - Create a new builder interface ChatOptions.Builder for building the Chat options
  - Create an explicit DefaultChatOptions
  - Create DefaultChatOptionBuilder which can create DefaultChatOptions

Resolves #1875

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -57,7 +57,6 @@ import org.springframework.ai.chat.observation.ChatModelObservationConvention;
 import org.springframework.ai.chat.observation.ChatModelObservationDocumentation;
 import org.springframework.ai.chat.observation.DefaultChatModelObservationConvention;
 import org.springframework.ai.chat.prompt.ChatOptions;
-import org.springframework.ai.chat.prompt.ChatOptionsBuilder;
 import org.springframework.ai.chat.prompt.Prompt;
 import org.springframework.ai.model.Media;
 import org.springframework.ai.model.ModelOptionsUtils;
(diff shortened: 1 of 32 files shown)


spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/ChatOptions.java
```diff
@@ -89,4 +89,82 @@ public interface ChatOptions extends ModelOptions {
 	 */
 	ChatOptions copy();
 
+	/**
+	 * Creates a new {@link ChatOptions.Builder} to create the default
+	 * {@link ChatOptions}.
+	 * @return Returns a new {@link ChatOptions.Builder}.
+	 */
+	static ChatOptions.Builder builder() {
```
ilayaperumalg, line 97: I think we need to have public access to this builder() for extensibility?

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -471,13 +470,13 @@ private List<AnthropicApi.Tool> getFunctionTools(Set<String> functionNames) {
 	}
 
 	private ChatOptions buildRequestOptions(AnthropicApi.ChatCompletionRequest request) {
-		return ChatOptionsBuilder.builder()
-			.withModel(request.model())
```
markpollack, line 475: need to keep the 'with' methods for once cycle of deprecation.

spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/ChatOptionsBuilder.java
```diff
@@ -18,6 +18,7 @@
 
 import java.util.List;
 
+@Deprecated(forRemoval = true, since = "1.0.0-M5")
```
markpollack, line 21: need a @deprecated javadoc command to point users to the new implementation.

markpollack:
merged in f252b2417a53db6be32480940a27159b1d45469c

PR #1902, opened 2024-12-10
refactor ChromaVectorStore builder API

sobychacko:
The commit restructures the ChromaVectorStore builder pattern to use a no-args constructor with fluent API for setting the ChromaApi.

This change:

 - Makes builder creation consistent with other vector stores
 - Moves ChromaApi validation to the doValidate method
 - Improves builder API ergonomics

The change requires updating all builder usages to use the new .chromaApi() method instead of passing it in the constructor.


markpollack:
great!  merged in e2825c98776e67605ac4e3d0722cff269e87a0a1

PR #1905, opened 2024-12-11
Fix Mistral AI Chat model function call usage calculation

ilayaperumalg:
- Fix the chat model's call() to calculate the cumulative usage
   - Use an explicit internalCall to pass the previous chat response so that accumulation can be done

 - Fix the chat model's stream() to calculate the cumulative usage
   - Fix MistralAi API to include usgae in ChatCompletionChunk
   - Use internalStream() to accumulate the usage

Add/update tests


tzolov:
Thanks @ilayaperumalg !

tzolov:
rebased and merged at 81dfd3b6b5af1d17b4b5bafcbd7c9e2dd928c922

PR #1907, opened 2024-12-11
Fix HTML document title on main page to "Introduction"  (again)

Lee-WonJun:
This is simply a restoration of #1615. The HTML document title was removed in commit 0b00e6f, and unless there is a specific reason for its removal, it would be better to re-add it.


markpollack:
thanks, not sure how this keeps popping up!

markpollack:
merged in 1d8e0b39de1070f0f5aba4743681ee7bfeeade81

Issue #1825, opened 2024-11-25
Provide a consistent experience for the usage of Builder pattern in Vector Store implementations


sobychacko:
Research into various builder pattern usages that is appropriate for the vector store implementations in spring-ai. 
Modify the current vector store impls to use the builder pattern consistently. 

- [x] Azure
- [x] Cassandra
- [x] Chroma
- [x] Coherence
- [x] Cosmos
- [x] ElasticSearch
- [x] Gemfire
- [x] Hanadb
- [x] Mariadb
- [x] Milvus
- [x] Mongodb
- [x] Neo4j
- [x] Opensearch
- [x] Oracle
- [x] PgVectorStore
- [x] Pinecone
- [x] Qdrant
- [x] Redis
- [x] Typesense
- [x] Weaviate
- [x] SimpleVectorStore

PR #1912, opened 2024-12-12
Cassandra vector store builder refactoring

sobychacko:
- Deprecate CassandraVectorStoreConfig
 - Move its contents to a top level Builder class within the CassandraVectorStore

Restructuring packages for Cassandra vector store and chat memory classes


sobychacko:
@michaelsembwever We are trying to make the vector store builder pattern consistent. See the recently merged chroma vector store for a template. Cassandra vector store seems to be more involved. Since you are the original author of the cassandra vector store, can you take a look at this PR and see what you think? Thank you!

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/memory/cassandra/CassandraChatMemoryAutoConfiguration.java
```diff
@@ -18,8 +18,8 @@
 
 import com.datastax.oss.driver.api.core.CqlSession;
 
-import org.springframework.ai.chat.memory.CassandraChatMemory;
-import org.springframework.ai.chat.memory.CassandraChatMemoryConfig;
+import org.springframework.ai.chat.memory.cassandra.CassandraChatMemory;
```
michaelsembwever, line 21: are all vectorstore modules moving their classes to subpackages ? (i don't see that yet with pgvector.  has there been some discussion/agreement to this somewhere we can link to ?)

vector-stores/spring-ai-cassandra-store/src/main/java/org/springframework/ai/vectorstore/cassandra/CassandraVectorStore.java
michaelsembwever, line 1: i can't read the diff here (because the change is clobbering the move).
would it be possible to do the move in a separate commit, to make the reviewing simpler…?

vector-stores/spring-ai-cassandra-store/src/main/java/org/springframework/ai/vectorstore/cassandra/CassandraVectorStore.java
```diff
@@ -0,0 +1,1038 @@
...
+	 *
+	 * To filter expression search against a metadata column configure it with
+	 * SchemaColumnTags.INDEXED
+	 *
+	 * The Cassandra Java Driver is configured via the application.conf resource found in
+	 * the classpath. See
+	 * https://github.com/apache/cassandra-java-driver/tree/4.x/manual/core/configuration
+	 *
+	 */
+	public static class CassandraBuilder extends AbstractVectorStoreBuilder<CassandraBuilder> {
```
michaelsembwever, line 737: does `CassandraBuilder` have to be an inner class ? 
no strong opinion, but i gut instinct is it's big enough to warrant being a separate java file.

vector-stores/spring-ai-cassandra-store/src/test/java/org/springframework/ai/vectorstore/cassandra/CassandraRichSchemaVectorStoreIT.java
```diff
@@ -92,66 +93,65 @@ class CassandraRichSchemaVectorStoreIT {
...
 				}
-				return java.lang.String.format("%s§¶%s", primaryKeys.get(2), primaryKeys.get(3));
+				return String.format("%s§¶%s", primaryKeys.get(2), primaryKeys.get(3));
 			})
-			.withDocumentIdTranslator(id -> {
+			.documentIdTranslator(id -> {
 				String[] parts = id.split("§¶");
 				String title = parts[0];
-				int chunk_no = 0 < parts.length ? Integer.parseInt(parts[1]) : 0;
+				int chunk_no = parts.length > 0 ? Integer.parseInt(parts[1]) : 0;
```
michaelsembwever, line 146: why this change ?

vector-stores/spring-ai-cassandra-store/src/test/java/org/springframework/ai/vectorstore/cassandra/WikiVectorStoreExample.java
```diff
@@ -90,36 +91,33 @@ public CassandraVectorStore store(CqlSession cqlSession, EmbeddingModel embeddin
...
-					return java.lang.String.format("%s§¶%s", primaryKeys.get(2), primaryKeys.get(3));
+					return String.format("%s§¶%s", primaryKeys.get(2), primaryKeys.get(3));
 				})
-
-				.withDocumentIdTranslator(id -> {
+				.documentIdTranslator(id -> {
 					String[] parts = id.split("§¶");
 					String title = parts[0];
-					int chunk_no = 0 < parts.length ? Integer.parseInt(parts[1]) : 0;
+					int chunk_no = parts.length > 0 ? Integer.parseInt(parts[1]) : 0;
```
michaelsembwever, line 116: why ?

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/cassandra/CassandraVectorStoreAutoConfiguration.java
```diff
@@ -63,25 +62,21 @@ public CassandraVectorStore vectorStore(EmbeddingModel embeddingModel, Cassandra
...
+		return CassandraVectorStore.builder()
+			.session(cqlSession)
+			.keyspace(properties.getKeyspace())
+			.table(properties.getTable())
+			.contentColumnName(properties.getContentColumnName())
+			.embeddingColumnName(properties.getEmbeddingColumnName())
+			.indexName(properties.getIndexName())
+			.fixedThreadPoolExecutorSize(properties.getFixedThreadPoolExecutorSize())
+			.disallowSchemaChanges(!properties.isInitializeSchema())
+			.returnEmbeddings(properties.getReturnEmbeddings())
```
michaelsembwever, line 74: there's a behaviour change here, in that we're no longer respecting the builder's default, imposing the autoConfiguration's default over it.  (the builder's default should be canonical)

i'm not sure how important this is… 🤷

michaelsembwever:
@sobychacko the overall direction of the refactoring looks good to me, nice work.

Threw in a few questions.  The reviewing need to separate config-build vs package renaming is the big one.
Otherwise, not having kept up with all project discussions and changes, I'm missing the project wide statement/agreement to move to this builder pattern and packaging naming.  My concern would be that we're only applying this to Chroma and Cassandra.  I assume that's not the case, but would still like an explicit reference link to this agreement :)

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/memory/cassandra/CassandraChatMemoryAutoConfiguration.java
sobychacko: Yes, we are in the process of moving the vector store classes under that structure - in this case - `org.springframework.ai.chat.memory.cassandra` and `springframework.ai.vectorstore.cassandra`.  The issue was that the vector store classes currently use the same package name in all the different modules - `springframework.ai.vectorstore`.

sobychacko:
@michaelsembwever Yes, we are moving to the builder pattern for all the vector store classes to make it consistent. Started with chroma,and there are a few PR's out there that's performing this migration on the other vector stores. Before the M5 release, we hope to complete this migration.

vector-stores/spring-ai-cassandra-store/src/main/java/org/springframework/ai/vectorstore/cassandra/CassandraVectorStore.java
sobychacko: Does this link work? https://github.com/spring-projects/spring-ai/blob/92b420c4a9c70ebf4b460c209706d2c9d7fbdcc5/vector-stores/spring-ai-cassandra-store/src/main/java/org/springframework/ai/vectorstore/cassandra/CassandraVectorStore.java

vector-stores/spring-ai-cassandra-store/src/main/java/org/springframework/ai/vectorstore/cassandra/CassandraVectorStore.java
sobychacko: We can consider that - but wanted to make the usage consistent across all the vector stores.

vector-stores/spring-ai-cassandra-store/src/test/java/org/springframework/ai/vectorstore/cassandra/CassandraRichSchemaVectorStoreIT.java
sobychacko: I will revert that.

vector-stores/spring-ai-cassandra-store/src/test/java/org/springframework/ai/vectorstore/cassandra/WikiVectorStoreExample.java
sobychacko: I will revert that.

vector-stores/spring-ai-cassandra-store/src/main/java/org/springframework/ai/vectorstore/cassandra/CassandraVectorStore.java
michaelsembwever: i would suggest that, given the lines of code and that the builder has a little more logic in it the a super simple builder.  (i had previously intentionally kept the classes separate for the sake of encapsulation between startup/configuration versus runtime behaviour).

it's not a hill i will die on.  if you take past PoVs into account, then i'm happy it's your call to make.

vector-stores/spring-ai-cassandra-store/src/main/java/org/springframework/ai/vectorstore/cassandra/CassandraVectorStore.java
michaelsembwever: i don't see any diff there^ 😔

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/memory/cassandra/CassandraChatMemoryAutoConfiguration.java
michaelsembwever: was hoping that there was some link you could provide to where the discussion had been had and landed on that conclusion and action.  but it's ok, i trust.

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/vectorstore/cassandra/CassandraVectorStoreAutoConfiguration.java
markpollack: Good catch, we should be consistent.  I'll create a new issue to capture this and discuss.

markpollack:
Thanks for your patience and understanding @michaelsembwever 

I've created https://github.com/spring-projects/spring-ai/issues/1949 to track the remaining comment in the discussion.

I've merged this after updating ref doc and class javadoc.

Merged in 25123a53640a2131347cc0573ac656bf5bf12186

sobychacko:
@michaelsembwever linking the original epic here: https://github.com/spring-projects/spring-ai/issues/1825

PR #1915, opened 2024-12-12
Bug Fix: Correct Parameter Type for KnnSearch$Builder.k Method in ElasticsearchVectorStore.java

yezhoujie:
## Bug Fix

Correct Parameter Type for KnnSearch$Builder.k Method in ElasticsearchVectorStore.java

## Description

This PR addresses a bug in ElasticsearchVectorStore.java where an incorrect parameter type was passed to the `KnnSearch$Builder.k` method of the ElasticSearchClient during the execution of similaritySearch. The issue caused the following error:

```
Handler dispatch failed: java.lang.NoSuchMethodError: 'co.elastic.clients.elasticsearch._types.KnnSearch$Builder co.elastic.clients.elasticsearch._types.KnnSearch$Builder.k(java.lang.Long)'
```

## Root Cause
The `KnnSearch$Builder.k` method in `elasticsearch-rest-client:8.15.4` requires an Integer parameter, but the code passed a long value, leading to the NoSuchMethodError.
```java
SearchResponse<Document> res = this.elasticsearchClient.search(
					sr -> sr.index(this.options.getIndexName())
						.knn(knn -> knn.queryVector(EmbeddingUtils.toList(vectors))
							.similarity(finalThreshold)
							.k((long) searchRequest.getTopK())
							.field("embedding")
							.numCandidates((long) (1.5 * searchRequest.getTopK()))
							.filter(fl -> fl.queryString(
									qs -> qs.query(getElasticsearchQueryString(searchRequest.getFilterExpression()))))),
					Document.class);
```

## Fix
The fix involves changing the parameter type from long to Integer when calling the `KnnSearch$Builder.k` method.

## Changes
Updated ElasticsearchVectorStore.java to pass an Integer parameter to the `KnnSearch$Builder.k` method.

### DIFF
--- a/vector-stores/spring-ai-elasticsearch-store/src/main/java/org/springframework/ai/vectorstore/ElasticsearchVectorStore.java
+++ b/vector-stores/spring-ai-elasticsearch-store/src/main/java/org/springframework/ai/vectorstore/ElasticsearchVectorStore.java
@@ -193,7 +193,7 @@ public class ElasticsearchVectorStore extends AbstractObservationVectorStore imp
 					sr -> sr.index(this.options.getIndexName())
 						.knn(knn -> knn.queryVector(EmbeddingUtils.toList(vectors))
 							.similarity(finalThreshold)
-							.k((long) searchRequest.getTopK())
+							.k(searchRequest.getTopK())
 							.field("embedding")
 							.numCandidates((long) (1.5 * searchRequest.getTopK()))
 							.filter(fl -> fl.queryString(


ilayaperumalg:
Hi @yezhoujie, 

Prior to [this update to use Spring Boot 3.4.2](https://github.com/spring-projects/spring-ai/commit/1c6132caf4179fbfda68d863cd3dbd2beadc0f70) change, Spring AI used elasticsearch-rest-client `8.13.3` which expected the  KnnSearch$Builder.k to have `long` type as the parameter. But, with Spring Boot 3.4.2, it is 8.15.5 which should work fine for you now. 

Also, the KnnSearch$Builder.numCandidates requires parameter type `Integer` and hence, the change was added [here](https://github.com/spring-projects/spring-ai/blob/a8e305d28a7e97f5d893bcbaeaa69b863c9b46e5/vector-stores/spring-ai-elasticsearch-store/src/main/java/org/springframework/ai/vectorstore/elasticsearch/ElasticsearchVectorStore.java#L267) to convert the double value to int. I would be curious to find out how this logic can be improved, perhaps, we could use Math.round() before casting to int. Please share your feedback. 

Closing this PR without merging as the underlying issue is already fixed with Spring Boot 3.4.2 update. Thank you for your contribution!

PR #1918, opened 2024-12-12
Fix Anthropic chat model functioncalling token usage

ilayaperumalg:
- Accumulate the token usage when functioncalling is used
   - Fix both call() as well as stream() operations

 - Add/update tests


tzolov:
rebased and merged at cadea8b7302c5cbc61525712006ad3b78ac84d8d

PR #1919, opened 2024-12-12
Update Anthropic3ChatOptions.java

bruno-oliveira:
The constructor needs to be public so that the options can be correctly copied over to the runtime call.


bruno-oliveira:
@tzolov @markpollack @sobychacko

markpollack:
Will review, thanks.  Note, a better anthropic experience can be found using the Bedrock Converse support that was recently added.

markpollack:
merged in 1453198c5b9b0e1996527747be584ebd7bcc5fd3

sorry it took so long.

PR #1921, opened 2024-12-12
Weaviate vector store builder refactoring

sobychacko:


markpollack:
Added tests for builder, updated ref docs and class level javadoc.

merged in d77c950ea625a51688747b57b83f1ebf2fda050b

PR #1922, opened 2024-12-12
Typesense vector store builder refactoring

sobychacko:


markpollack:
updated ref docs, added class level javadoc, and added builder test.

merged in 88776728464dbecac301ec20ed013b038717e8e5

PR #1923, opened 2024-12-12
Refactor FunctionCallingOptions Builder

### LINKED ISSUE #1874: Put builder class inside PortableFunctionCallingOptions
Currently the FunctionCallingOptionsBuilder is a separate class, it should be removed and instead a 'standard spring-ai' builder pattern be used inside PortableFunctionCallingOptions.

'standard spring-ai' builder at the moment is

* no 'with' prefix on methods
* check for passing in null values
* for a 'simple' case like options, where there aren't going to be multiple implementations, there isn't a need for a builder interface.

Also rename PortableFunctionCallingOptions to DefaultFunctionCallingOptions

ilayaperumalg:
- Deprecate existing FunctionCallingOptionsBuilder
  - Create FunctionCallingOptions.Builder which extends ChatOptions.Builder
  - Create DefaultFunctionCallingOptions which extends DefaultChatOptions and implements FunctionCallingOptions to serve the default FunctionCalling options
  - Create DefaultFunctionCallingOptionsBuilder to build the default functioncalling options
  - Update the usage of functioncalling options builder to use the newly added builder including the tests

Resolves #1874

### DIFF
--- a/models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/AnthropicChatModelIT.java
+++ b/models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/AnthropicChatModelIT.java
@@ -49,7 +49,7 @@ import org.springframework.ai.converter.ListOutputConverter;
 import org.springframework.ai.converter.MapOutputConverter;
 import org.springframework.ai.model.Media;
 import org.springframework.ai.model.function.FunctionCallback;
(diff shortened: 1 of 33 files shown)


markpollack:
merged b25d6e8d2d7ab1acdda2362aee1b4ff882892f13

as an aside, we should revisit what is in the portable option, some of them aren't that portable in retrospect.

PR #1926, opened 2024-12-13
feat: core abstraction for image classification models

d-lobo:
I'm currently working on another contribution, enabling the invocation of image classifying models deployed via huggingface.
As core abstractions for invoking classifying models is not yet implemented, I'd like to contribute it.
PR contains:

- core abstraction to invoke image classifying AI models 

If you see any room for improvements, kindly let me know.


markpollack:
without two implementations can't tell if the code is corect.

PR #1939, opened 2024-12-16
Refactor OpenAI chat options

ilayaperumalg:
- Deprecate existing OpenAI chat options
 - Add the options' methods by removing the prefix "with"
   - Update the following options
    - OpenAI chat options
    - OpenAI image options
    - OpenAI moderation options
    - OpenAI embedding options
    - OpenAI audiospeech options
    - OpenAI audio transcription options
 - Update referecens and docs

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -143,8 +143,7 @@ public class OpenAiChatModel extends AbstractToolCallSupport implements ChatMode
 	 * @throws IllegalArgumentException if openAiApi is null
 	 */
 	public OpenAiChatModel(OpenAiApi openAiApi) {
-		this(openAiApi,
-				OpenAiChatOptions.builder().withModel(OpenAiApi.DEFAULT_CHAT_MODEL).withTemperature(0.7).build());
+		this(openAiApi, OpenAiChatOptions.builder().model(OpenAiApi.DEFAULT_CHAT_MODEL).temperature(0.7).build());
 	}
 
 	/**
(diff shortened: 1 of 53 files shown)


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -143,8 +143,7 @@ public class OpenAiChatModel extends AbstractToolCallSupport implements ChatMode
 	 * @throws IllegalArgumentException if openAiApi is null
 	 */
 	public OpenAiChatModel(OpenAiApi openAiApi) {
-		this(openAiApi,
-				OpenAiChatOptions.builder().withModel(OpenAiApi.DEFAULT_CHAT_MODEL).withTemperature(0.7).build());
+		this(openAiApi, OpenAiChatOptions.builder().model(OpenAiApi.DEFAULT_CHAT_MODEL).temperature(0.7).build());
```
markpollack, line 146: we should check if these defaults correspond to what the autoconfig defaults are

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiEmbeddingOptions.java
```diff
@@ -96,21 +97,53 @@ public Builder() {
...
+
+		public Builder user(String user) {
+			this.options.setUser(user);
+			return this;
+		}
+
+		/**
+		 * @deprecated use {@link #model(String)} instead.
+		 */
 		public Builder withModel(String model) {
```
markpollack, line 123: need to use @Deprecated on the method, not just the javadoc

markpollack:
Added missing @Deprecated annotations, was a conflict in tooluse, it takes object not string.

Merged in 58efee6d909ccebbf693a4816de75a4e4c0ecf95

PR #1941, opened 2024-12-16
Refactor MoonshotChatOptions builder methods

apappascs:
- Refactor the builder methods to remove `with` as the prefix.
- Introduce new methods with updated naming conventions.
- Deprecate the existing `with*` methods to maintain backward compatibility.
- Update the `copy()` method to use the new builder methods.


ilayaperumalg:
Hi @apappascs , Thanks for the PR. We also need the documentation to be updated accordingly.

ilayaperumalg:
Hi @apappascs , Thanks for the PR. We also need the documentation to be updated accordingly.

ilayaperumalg:
LGTM, rebased and merged as 8c51f7a84

PR #1950, opened 2024-12-16
Refactor Azure model options builder methods

ilayaperumalg:
- Deprecate builder methods with the prefix `with` and add them without the prefix
 - Update references and docs


markpollack:
merged in da45244d99de04bb249b7c61e4e946accd768d03

PR #1952, opened 2024-12-16
Refactor Ollama options builder methods

ilayaperumalg:
- Refactor Ollama related options builder methods
   - Deprecate builder methods with the prefix `with`
   - Update docs and references


markpollack:
refactored builder class to be an inner class.

Breaking Change: builder() method now returns Builder instead of OllamaOptions

markpollack:
Disabled `OllamaChatModelIT.jsonSchemaFormatStructuredOutput` due to recent unrelated commit that is causing the test to fail.

markpollack:
merged in 7e4a187f0d310bb40ca6ef2fcce58789118f708f

PR #1953, opened 2024-12-16
Refactor OCI Gen AI builder methods

ilayaperumalg:
- Deprecate the builder methods with the prefix `with`
   - Fix OCI cohere chat model options and OCI Embedding model options
 - Update references and docs


markpollack:
merged in 95c32f05a34a183ffc44f570fdeca18069626949

PR #1958, opened 2024-12-17
Refactor QianFan model option builder methods

ilayaperumalg:
- Deprecate the builder methods with the prefix `with` and add the new methods
 - Update docs and references


markpollack:
merged in 18bc0cff25deeb9e727c0a26d75e5e310e8670e9

PR #1970, opened 2024-12-18
Add builder pattern to SimpleVectorStore

ilayaperumalg:
- Create SimpleVectorStoreBuilder which can be used to create the SimpleVectorStore instance
 - Remove the field embeddingModel from SimpleVectorStore as it can be retrieved from AbstractObservationVectorStore
 - Deprecate the existing SimpleVectorStore constructors and use the builder to instantiate it
 - Fix the tests


sobychacko:
Closed via https://github.com/spring-projects/spring-ai/commit/8ce837482584a42ba4921e825cc58fba1c076f61

PR #1975, opened 2024-12-18
Refactor SearchRequest builder methods

ilayaperumalg:
- Deprecate builder methods with the prefix `with`
 - Since there is a conflicting method named query(String) which returns the instance of SearchRequest, changed the existing builder method query(String) to queryString(String)
 - Update docs and references


markpollack:
There should be an inner Builder class and the the static method 

```
	public static SearchRequest query(String query) {
		Assert.notNull(query, "Query can not be null.");
		return new SearchRequest(query);
	}
```

should be deprecated.

PR #1976, opened 2024-12-18
Fix mariadb similarity search

ilayaperumalg:
- The similarity search doesn't need the embedding to be returned/stored back to Document


markpollack:
merged in 69cd3173f7e78469b94e81f5b9aea883f788c52d

PR #1978, opened 2024-12-19
Fix EmbeddingOptions

ilayaperumalg:
- Remove the explicit EMPTY check when merging the default embedding options
 - Use the builder to build the default embedding options


markpollack:
merged in adefca13f8ec86b7c1120c88c2d15cbfc6117412

PR #1981, opened 2024-12-19
Remove embedding from Document

### LINKED ISSUE #1781: Remove embedding from Document
This was a design decision way back taken from how langchain (python) was designed.  In practice, this doesn't seem useful.  It acts like a sort of cache, if you are writing the same document to the vector store multiple times, which doesn't seem realistic.   If you have a use case for keeping the embedding in the Document, please comment.

ilayaperumalg:
- This PR removes the reference of `embedding` from the `Document` object.
   - The Document is designed to contain the content and its metadata but not the embedding representation of its content. This provides clear separation of concern.
   - Prior to this change, all the vector stores are updated by not depending on the embedding from Document. Instead, use the embedding values returned by the Embedding Models directly: https://github.com/spring-projects/spring-ai/issues/1826

Resolves #1781

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/document/Document.java
@@ -123,12 +123,6 @@ public class Document {
 	private final Double score;
 
 	/**
-	 * Embedding of the document. Note: ephemeral field.
-	 */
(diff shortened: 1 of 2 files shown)


markpollack:
need to remove embedding in builder , use in tests and mention in javadoc.  will do on merge.

PR #1982, opened 2024-12-19
Fix DefaultFunctionCallingOptionsBuilder

ilayaperumalg:
- Reuse DefaultChatOptionsBuilder's options to refer both functioncalling and chat options - Set the parent class options with the FunctionCallingOptions when DefaultFunctionCallingOptionsBuilder is used


markpollack:
merged in bfc0b8285e7b60849940ece556721c66898940d9

PR #1983, opened 2024-12-19
Require essential dependencies in vector store builder constructors

sobychacko:
This commit refactors the builder pattern implementation across all VectorStore implementations to make the EmbeddingModel a required constructor parameter rather than an optional builder method. Key changes include:

- Move embeddingModel from being a builder method to a required constructor parameter
- Make embeddingModel final in AbstractVectorStoreBuilder
- Remove redundant validate() methods since EmbeddingModel validation now happens in constructor
- Update all VectorStore builder instantiations to pass EmbeddingModel in builder creation
- Add @Nullable annotations to appropriate methods in VectorStore interface

This change improves the API design by:
1. Enforcing that EmbeddingModel is provided at builder creation time
2. Removing the possibility of forgotten EmbeddingModel configuration
3. Simplifying the builder implementation by moving validation to construction
4. Making the dependency on EmbeddingModel more explicit in the API

Breaking Changes:
- VectorStore builders must now be created with an EmbeddingModel parameter
- The embeddingModel() builder method has been removed from all implementations

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/embedding/package-info.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/embedding/package-info.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2023-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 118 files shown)


markpollack:
merged in 1a6e79cbb986fc2b421140ff5154c0aa00f23770

PR #1986, opened 2024-12-20
feat(bedrock): deprecate InvokeModel API Chat modes in favor of Converse API

tzolov:
Following Amazon Bedrock's recommendations, deprecate individual model chat implementations (Anthropic, Cohere, Llama, Titan, etc.) in favor of the unified BedrockProxyChatModel. This change promotes the use of Bedrock's Converse API which offers benefits like model flexibility, extended functionality, and native tool support.

The Converse API does not support embedding operations, so these will remain in the current API and the embedding model functionality in the existing InvokeModel API will be maintained


markpollack:
merged in e941bc572d253128433857bd1a676a0594fcaf31

PR #1994, opened 2024-12-21
Improve inheritance hierarchy of ChatOptions/FunctionCallingOptions

ilayaperumalg:
- Convert ChatOptions.Builder to have `Builder<B extends Builder<B, C>, C extends ChatOptions>`
   - This will allow the subclasses which extend this builder to provide their implementation of ChatOptions i.e, in our case FunctionCallingOptions
   - Update the default builder implementations with this change
   - Update FunctionCallingOptions and its default builder impelementation to enforce this hiearchical change


markpollack:
I've reimplemented the options builders to get rid of the generics.  There were still hard to resolve warnings in the IDE and I kept going in circles.  It shouldn't be so complicated but that approach seems to lead to unwanted java generic warning/side effects.
 
 Now the approach is very simple and easy to understand.
 
 merged new impl in 0765b2ca88f110a9849ab869f1b3166e362af1cc

PR #1997, opened 2024-12-21
Modify SimpleLoggerAdvisor to use PrettyPrinter method for response logging

SasikiranJ:
* I have observed that default response logging in simple logger advisor printing the ChatResponse in raw json. This PR changes default print method used in the simpleloggeradvisor to pretty print the response.


ilayaperumalg:
@SasikiranJ Thanks for the PR!

PR #2000, opened 2024-12-23
Fix generic type in explicit type casting

linarkou:
Fix generic type of `responseConverter`. This typo was possible due to calling constructor of `FunctionInvokingFunctionCallback` with  raw type, so made it generic too.


ilayaperumalg:
@linarkou Thanks for the PR.

PR #2002, opened 2024-12-23
refactor(vectorstore): introduce Step Builder pattern for PineconeVec…

sobychacko:
…torStore

Replace the existing builder implementation with a type-safe Step Builder pattern to enforce required field initialization in a specific order. This change:

- Introduces a new builder() method that guides users through required fields
- Deprecates the old builder method with all parameters
- Makes field configuration more explicit and type-safe
- Updates all usage points to use the new builder pattern

The required fields must now be provided in sequence:
1. embeddingModel
2. apiKey
3. projectId
4. environment
5. indexName

This change improves API usability by preventing parameter confusion and ensuring all required fields are set before optional configuration.


markpollack:
rebased and merged in 69200e621dcf4240c612724427b61224d1bdc590

Issue #1370, opened 2024-09-17
Tools are called twice on streaming requests for Anthropic API


youssefwadie:
**Bug description**

When using the Anthropic API in Spring AI with streaming responses, tools (functions) are called twice. This behavior does not occur in blocking calls (i.e., ChatClient.call), where each function is invoked only once, as expected.

**Environment**
* Spring Boot version: 3.3.3
* Spring AI version: 1.0.0-M2
* Java version: 17
* OS: Windows 11

**Steps to reproduce**
1. Register a custom function using the Spring AI Anthropic API for a streaming request.
2. Trigger a request that involves streaming function calling.
3. Observe that the function is invoked twice during the stream, even though it should only be triggered once.

**Expected behavior**
The function should be called only once during the streaming response, just like the behavior observed in blocking calls.

**Minimal Complete Reproducible example**
I have replicated Dan Vega's example from his "Spring into AI" tutorial, which you can find here: [Dan Vega's Spring into AI example](https://github.com/danvega/spring-into-ai/tree/main/src/main/java/dev/danvega/functions).
But, I replaced the OpenAI client with Anthropic's API. You can find my version here: [Spring AI Anthropic Function Calling](https://github.com/youssefwadie/spring-ai-anthropic-function-calling).

The logs for both streaming and non-streaming requests demonstrate the issue. The tool is invoked twice in the streaming scenario but only once in the non-streaming scenario, as expected.

**Non-streaming request**
![Non streaming](https://github.com/user-attachments/assets/9b9fd476-1067-458b-b857-1dcfda7f2786)

---

**Streaming request**
![Streaming](https://github.com/user-attachments/assets/494a1b35-e2eb-4dc6-be5e-b0a43a94cade)

Claudio-code:
is that only the [switchMap](https://github.com/spring-projects/spring-ai/blob/d7fe07b0f1c2fcc24462c48ca826e0864da47a4b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java#L283) section is executed twice, not the entire method

mmazurkevich:
After some time for debugging, I actually found the root cause of the problem.

[...]

PR #2005, opened 2024-12-24
Fixed duplicate call bug in Anthropic stream apis

Claudio-code:

issue https://github.com/spring-projects/spring-ai/issues/1370

I fixed the duplicate call by passing the id of the previous response and using it to validate the next call, but the root cause was not identified and the loop continues to execute twice.

Adjusted test do validate correction [streamFunctionCallUsageTest](https://github.com/Claudio-code/spring-ai/blob/893a8dfc2cc2639d85b814f3d13bb8c651181385/models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/AnthropicChatModelIT.java#L339)


Claudio-code:
is that only the [switchMap](https://github.com/spring-projects/spring-ai/blob/d7fe07b0f1c2fcc24462c48ca826e0864da47a4b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java#L283) section is executed twice, not the entire method

PR #2018, opened 2024-12-29
Add o1 model to OpenAI ChatModel enum

dafriz:
The o1 model is out of preview and there are new model ids, see https://platform.openai.com/docs/models#o1


ilayaperumalg:
@dafriz Thanks for the PR! LGTM, merging.

Issue #1938, opened 2024-12-15
TopK parameter not working for values larger than the default (ElasticSearch), only working has a client-side limit


jcgouveia:
When creating a request for a query using Elasticsearch store
```
		SearchRequest request = SearchRequest.query(query);
		List<Document> docs = store.similaritySearch(request);
```

The default query size from ES is 10 results
SearchRequest.query(query) => 10 results

If I set a smaller value, eg 5
SearchRequest.query(query)withTopK(5);  => 5 results (OK)

If I set a larger value, eg 100
SearchRequest.query(query)withTopK(100);  => 10 results (Not OK) I was expecting 100

This happens because the **size** parameter of the ES SerachRequest (`co.elastic.clients.elasticsearch.core.SearchRequest`) is never set, the topK is applied only after the query results, which I think is wrong.
Doing the query directly on the _search ES rest api, I get the intended response -100 results (although not being a similarity query), eg
```
{
  "query": {
    "match_all": {}
  },
  "size": 100
}
```
Changing in debug the value of SearchRequest.size to 100, I can get also the intended result.

karol-antczak:
I have the same problem. Ended up making custom `VectorStore` implementation.

sobychacko:
This issue has been fixed via https://github.com/spring-projects/spring-ai/pull/2025.

PR #1423, opened 2024-09-26
Set redis query limit to match requested topK used in KNN search

dafriz:
The Jedis `Query` object sets limit to 10 by default. This fix aligns limit with topK requested. 

Fixes https://github.com/spring-projects/spring-ai/issues/1415

PR #2025, opened 2025-01-02
Set ElasticSearch size to match requested topK used in KNN search

dafriz:
Similar to https://github.com/spring-projects/spring-ai/pull/1423 but for ElasticSearch which defaults to a size of 10

Fixes https://github.com/spring-projects/spring-ai/issues/1938


vector-stores/spring-ai-elasticsearch-store/src/test/java/org/springframework/ai/vectorstore/elasticsearch/ElasticsearchVectorStoreIT.java
```diff
@@ -20,10 +20,7 @@
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.ZonedDateTime;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
ilayaperumalg, line 23: We avoid wildcard import. will fix when merging.

ilayaperumalg:
@dafriz Thanks for the PR fixing the issue. LGTM, tested and merged as 95675a85f

PR #2028, opened 2025-01-02
docs: fix grammatical issue in API provider description

chris-bailey:
Fix grammatical issue in API provider documentation

- Simplify redundant wording around API provider support
- Remove duplicate terms ('support')

Before:
Portable API support across AI providers for both synchronous and streaming API options are supported.

After:
Portable API support across AI providers for both synchronous and streaming options.


ilayaperumalg:
@chris-bailey Thanks for the PR!

PR #2029, opened 2025-01-02
Gemini response can contains text and function call.

GregoireW:
The function call feature can result in gemini respond multiple part like

 - part 1: message: " I understand what you want to do, I will check ..."
 - part 2:  function call  
 - part 3:  message "if it is not enough you will need to check yourself  .... "

Today, the response via spring-ai is the message will not launch the function because the code want every part to be a function call to do it.  There will be 3 outputs on the response  (1 message,  1 empty string, 1 message)

It seems it is ok to switch to 'anyMatch' instead of 'allMatch' on this subject, because after the function call, another call will be made to the model and a return like 

- part 1:  message "I understand what you want, I get the result and you can do .... "


markpollack:
@ddobrin can you please take a look?

ddobrin:
I'll add it to my @todo list @markpollack 

There was a new release of the GenAI SDK today and I'll see whether/how it improves this aspect

ericbottard:
Addressed differently via 8e8654eeeb17ea0053397938515227ebadb48ac7

PR #2051, opened 2025-01-07
Introduce LoggingMarkers marker for logging sensitive data

kpavlov:
Introduce a utility class `LoggingMarkers` providing an SLF4J marker for tagging log entries with different data security classification, such as:
- SENSITIVE_DATA_MARKER
- RESTRICTED_DATA_MARKER
- REGULATED_DATA_MARKER
- PUBLIC_DATA_MARKER

Update `BeanOutputConverter` to use the `SENSITIVE_DATA_MARKER` in error logs for invalid JSON conversions. Enhance tests to verify SENSITIVE_DATA_MARKER marker usage in logging.

Updated Maven configurations to dynamically reference the Java version using `${java.version}`. Added Kotlin compiler settings, including `jvmTarget` alignment with Java version and enabling `javaParameters` to ensure consistency and better compatibility across builds.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/util/LoggingMarkers.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/util/LoggingMarkers.java
@@ -1,0 +1,25 @@
+package org.springframework.ai.util;
+
+import org.slf4j.Marker;
+import org.slf4j.MarkerFactory;
+
+/**
+ * Utility class that provides predefined SLF4J {@link Marker} instances used in logging
+ * operations within the application. <br>
+ * This class is not intended to be instantiated.
+ */
+public class LoggingMarkers {
+
+	/**
+	 * Marker instance representing Personally Identifiable Information (PII) used in
+	 * logging operations to classify or tag log entries for sensitive data. This can be
+	 * utilized to allow selective filtering, handling, or analysis of log messages
+	 * containing PII.
+	 */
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
@kpavlov Could you check the test failure on this PR? Thanks

kpavlov:
@ilayaperumalg I have updated the PR and fixed incompatibility with JDK 17

markpollack:
The core of Spring AI does not depend on Spring Boot, which is usually a green flag for using slf4j.  Spring projects that do not depend on spring boot use commons logging, typically via the `LogAccessor` in Spring Framework.  We just merged a PR that changes us fom slf4j to `LogAccessor`.  Not sure what is the best practice in this case.  I'll inquire.

kpavlov:
LogAccessor wouldn't allow adding additional metadata to log messages. If so, my proposal for log message classification would not work. Please feel free to close this PR.

spring-ai-core/src/main/java/org/springframework/ai/util/LoggingMarkers.java
```diff
@@ -0,0 +1,69 @@
+package org.springframework.ai.util;
```
ilayaperumalg, line 1: Will need to add copywrite statement. will add it when merging.

ilayaperumalg:
@kpavlov Thanks for your contribution! Squashed, rebased and merged as b525309e8

PR #2053, opened 2025-01-07
Update getting-started.adoc

pawn-4-git:
Fix to 1.0.0-M5 because 1.0.0-SNAPSHOT is missing and build with Gradle fails. Maven side was not changed because we did not have a Maven environment and could not verify it immediately. If the change to M5 does not cause any problems, I will add the changes accordingly.

If a non-M5 version is better, please point it out to us.


pawn-4-git:
Closing due to a large possibility of pointing out errors.

PR #2063, opened 2025-01-11
Chat memory implementation for Neo4j

enricorampazzo:
I followed the Cassandra implementation as a guide. As you can see, this implementation supports all the `MessageType`s defined by Spring AI. At the moment, there's no TTL support for messages, unlike in Cassandra's implementation. 
The common `initialize-schema` parameter is also unsupported, as this implementation, at least at the moment, does not create any index or constraint.


enricorampazzo:
Sorry @ilayaperumalg , I will close and reopen this PR after signing all the commits correctly

PR #2065, opened 2025-01-12
Add OpenAI TTS Voice 

hgs-study:
OpenAI's list of TTS voices didn't seem to be up to date, so I added 3 new voices.("sage", "coral", "ash") 
Here is the reference link. (https://platform.openai.com/docs/guides/text-to-speech#voice-options)
Thank you.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiAudioApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiAudioApi.java
@@ -392,7 +392,13 @@ public class OpenAiAudioApi {
 			@JsonProperty("nova")
 			NOVA("nova"),
 			@JsonProperty("shimmer")
-			SHIMMER("shimmer");
+			SHIMMER("shimmer"),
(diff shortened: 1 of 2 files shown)


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiAudioApi.java
```diff
@@ -392,7 +392,13 @@ public enum Voice {
 			@JsonProperty("nova")
 			NOVA("nova"),
 			@JsonProperty("shimmer")
-			SHIMMER("shimmer");
+			SHIMMER("shimmer"),
```
ilayaperumalg, line 395: Need to update the javadoc for the same. Will update when merging.

ilayaperumalg:
@hgs-study Thanks for the PR! The changes LGTM, merging. 
As a side note, I suggest you to create separate branches on your work instead of working on your fork's main branch.

ilayaperumalg:
Rebased, squashed and merged as 7e303a06e

PR #2070, opened 2025-01-15
Neo4j module: remove unnecessary property removal.

### LINKED ISSUE #1932: Adding a Document to a neo4j vector store needlessly overwrites all its properties
https://github.com/spring-projects/spring-ai/blob/9e18652bb8251abbc3cb75a380d6eb2accc762ea/vector-stores/spring-ai-neo4j-store/src/main/java/org/springframework/ai/vectorstore/Neo4jVectorStore.java#L125

Because of this line, when an existing node is added to the vector store, its existing properties are deleted, and only those added as document metadata are persisted. 

This is needlessly destructive. I suggest to change line 125 to `SET u = properties(u)` so that existing properties are preserved

meistermeier:
Boils down the create or update logic even more.

Closes #1932

### DIFF
--- a/vector-stores/spring-ai-neo4j-store/src/main/java/org/springframework/ai/vectorstore/neo4j/Neo4jVectorStore.java
+++ b/vector-stores/spring-ai-neo4j-store/src/main/java/org/springframework/ai/vectorstore/neo4j/Neo4jVectorStore.java
@@ -207,12 +207,7 @@ public class Neo4jVectorStore extends AbstractObservationVectorStore implements 
 			var statement = """
 						UNWIND $rows AS row
 						MERGE (u:%s {%2$s: row.id})
-						ON CREATE
 							SET u += row.properties
-						ON MATCH
-							SET u = {}
-							SET u.%2$s = row.id,
-								u += row.properties
 						WITH row, u
 						CALL db.create.setNodeVectorProperty(u, $embeddingProperty, row[$embeddingProperty])
 					""".formatted(this.label, this.idProperty);


vector-stores/spring-ai-neo4j-store/src/main/java/org/springframework/ai/vectorstore/neo4j/Neo4jVectorStore.java
```diff
@@ -207,12 +207,7 @@ public void doAdd(List<Document> documents) {
 			var statement = """
 						UNWIND $rows AS row
 						MERGE (u:%s {%2$s: row.id})
-						ON CREATE
 							SET u += row.properties
```
enricorampazzo, line 210: I am nitpicking, but please consider modifying line 211 so that the `SET` statement is right under `MERGE `

meistermeier: Nitpicking back 😄  The set operation is a follow-up operation on the merge and therefore I would prefer to keep it indented for readability.

enricorampazzo: Good point :)

meistermeier:
Still interest in this or should I close the PR? Just saw that this is now more than a year open and nobody seems to be interested in.

PR #2077, opened 2025-01-16
small code improvements/clean-ups

ppjgit:
small code improvements/clean-ups, deleted unused method 'withEmbedding' from SimpleVectorStoreContent class


spring-ai-core/src/main/java/org/springframework/ai/vectorstore/SimpleVectorStoreContent.java
```diff
@@ -100,22 +99,10 @@ public SimpleVectorStoreContent(String id, String text, Map<String, Object> meta
...
 		this.embedding = Arrays.copyOf(embedding, embedding.length);
 	}
 
-	/**
-	 * Creates a new instance with an updated embedding vector.
-	 * @param embedding the new embedding vector, must not be null
-	 * @return a new instance with the updated embedding
-	 * @throws IllegalArgumentException if embedding is null or empty
-	 */
-	public SimpleVectorStoreContent withEmbedding(float[] embedding) {
```
ilayaperumalg, line 113: Public methods need to be deprecated first before getting deleted. I will re-add and deprecate this method instead.

ilayaperumalg:
Squashed, rebased and merged as 608b29cf9

PR #2084, opened 2025-01-19
Added index of tool call in the list of tool calls

ricken07:
Fix : `org.springframework.http.converter.HttpMessageNotReadableException: JSON parse error: Unrecognized field "index" (class org.springframework.ai.mistralai.api.MistralAiApi$ChatCompletionMessage$ToolCall), not marked as ignorable`

Mistral has added an extra field named `index` in `tool_calls` responses, which causes an exception when mapping the template response. 

This pull request corrects this problem. Currently, this `index` field is not used by Spring AI. 

However, in view of developments in assistant response models, wouldn't it be preferable to modify the `AssistantMessage` structure to make it generic and more adaptable ? @tzolov


tzolov:
Thank you @ricken07 !
Rebased, extended, squashed and merged at e92616b10e36f438bfc4def8c54945ba97f931f8
I've made some additional (non mistral related fixes as well)

Issue #2097, opened 2025-01-22
Double remote call on DefaultChatClient with content and chatResponse


Grogdunn:
**Bug description**
In DefaultChatClient#DefaultCallResponseSpec (also in streaming) call chatResponse or conent trigger the remote call twice.

**Environment**
Spring AI version M5 confirmed also on `main`

**Steps to reproduce**
simply call chatResponse and content and double remote service is invoked

**Expected behavior**
per "DefaultCallResponseSpec" access to same response (memoized)

**Minimal Complete Reproducible example**
```java
var response = chatClient.prompt()
		.advisors(new ReReadingAdvisor())
		.call();
response.content();
response.chatResponse();

```

ThomasVitale:
The API is designed to support only one terminal call operation via the `CallResponseSpec`: `content()`, `chatResponse()`, `entity()`, or `responseEntity()`.  I would say that it's by design that if the call action is called multiple times, then the model is called on each call.

That's the same behaviour present in other Client APIs in the Spring portfolio, such as `RestClient`.

If you need full access to the response (e.g. metadata), I suggest using the `chatResponse()` terminal operation and then extract the information you need from there (including the content).

Grogdunn:
Ok, but maybe like "java stream API" if a terminal operation is called twice the second call throws.

At the moment the feeling are the "content" call are a shortcut to invoke "chatResponse().getResult().getOutput().getText()" on SAME response.

Last 2 cents seems "DefaultResponseSpec" in DefaultRestClient save the `ClientHttpResponse clientResponse` and make the operation on that, but not invoke remote service each time that I invoke "getBody" (maybe throw if the body is a consumed)

tzolov:
Hi @Grogdunn,

I agree with @ThomasVitale - we should call the AI model endpoint for each call()/stream() invocation. AI endpoints aren't idempotent - identical inputs will return different responses, so caching previous responses isn't appropriate.

Could you explain how this impacts your use case? If you need caching, you could implement it as a custom advisor but be aware of the side effects of applying it.

Grogdunn:
Hi @tzolov, I agree with you for `call` or `stream` methods. Each `call` invocation should invoke LLM. The non-intuitive API is on `CallResponseSpec` or `DefaultCallResponseSpec` (the return of `call`).

As my first post:
```java
CallResponseSpec response = chatClient.prompt()
		.advisors(new ReReadingAdvisor())
		.call(); //Only one CALL method invocation
[...]

PR #2098, opened 2025-01-22
Double remote call on DefaultChatClient with content and chatResponse

Grogdunn:
This should fix the issue #2097 .


chemicL (commented):
Hey! Since I've been summoned here by @tzolov, here's my view: The approach using a `ThreadLocal` is not appropriate. I don't know what the solution should be (most probably an architectural decision or some usage guidelines), but I can say that `ThreadLocal` will cause issues, e.g. if the client is shared across Threads. Also, resubscribing to the `Flux` of `ChatResponse` is most probably not permitted and actually would probably end up exercising the LLM call twice as well. I do not see any tests here either validating the approach, I think that would be necessary to make progress, but best to tackle the question of the intended usage and design. Implementing caching in any way might as well be unexpected in most scenarios regardless if it relies on ThreadLocal or some other mechanism.

Grogdunn:
Ok fine, the ThreadLocal is not the way to acheve this, we will continue to discuss on Issue, maybe only a renaming is needed to avoid confusion.

markpollack:
thoughts @chemicL ?

chemicL:
@markpollack my view is that this PR should be closed. The actual issue or source of misunderstanding is addressed in https://github.com/spring-projects/spring-ai/issues/2097#issuecomment-2671339616

PR #2107, opened 2025-01-23
Refactor OpenAi Audio builders

apappascs:
- Refactor the builder methods to remove `with` as the prefix.
- Introduce new methods with updated naming conventions.
- Deprecate the existing `with*` methods to maintain backward compatibility.


ilayaperumalg:
@apappascs Thanks for the PR

PR #2108, opened 2025-01-24
Refactor Usage handling

ilayaperumalg:
- Remove model specific Usage implementations
 - Add `Object getNativeUsage()` to Usage interface
   - This will allow the model specific Usage data to be returned
   - At the client side, client needs to cast the return type of `getNativeUsage` into the corresponding Usage returned by the model API
 - Rename `generationTokens` to `completionTokens`
   - Since `completion` token name is more common among the models, renaming generation tokens into completion tokens
 - Change the prompt, completion and total token return types to Integer
 - Use DefaultUsage for most of the model specific usage handling
    - When initializing set the native usage to the model specific usage type


spring-ai-core/src/main/java/org/springframework/ai/chat/metadata/DefaultUsage.java
```diff
@@ -16,74 +16,120 @@
...
-	public DefaultUsage(@JsonProperty("promptTokens") Long promptTokens,
-			@JsonProperty("generationTokens") Long generationTokens, @JsonProperty("totalTokens") Long totalTokens) {
-		this.promptTokens = promptTokens != null ? promptTokens : 0L;
-		this.generationTokens = generationTokens != null ? generationTokens : 0L;
-		this.totalTokens = totalTokens != null ? totalTokens
-				: calculateTotalTokens(this.promptTokens, this.generationTokens);
+	public DefaultUsage(@JsonProperty("promptTokens") Integer promptTokens,
+			@JsonProperty("completionTokens") Integer completionTokens,
+			@JsonProperty("totalTokens") Integer totalTokens) {
+		this(promptTokens, completionTokens, totalTokens, null);
```
markpollack, line 92: The native usage argument should not be `null` but `Map.of`

spring-ai-core/src/main/java/org/springframework/ai/chat/metadata/DefaultUsage.java
```diff
@@ -96,18 +142,18 @@ public boolean equals(Object o) {
 		}
 		DefaultUsage that = (DefaultUsage) o;
 		return Objects.equals(this.promptTokens, that.promptTokens)
-				&& Objects.equals(this.generationTokens, that.generationTokens)
+				&& Objects.equals(this.completionTokens, that.completionTokens)
```
markpollack, line 145: equals should also include the `nativeUsage` field in the calculate, as well as for hascode/tostring

spring-ai-core/src/main/java/org/springframework/ai/chat/metadata/DefaultUsage.java
```diff
@@ -16,74 +16,120 @@
...
-	private final Long promptTokens;
+	private final Integer promptTokens;
 
+	private final Integer completionTokens;
+
+	@Deprecated(forRemoval = true, since = "1.0.0-M6")
 	private final Long generationTokens;
 
-	private final Long totalTokens;
+	private final int totalTokens;
```
markpollack, line 43: should be `Integer` no?

spring-ai-core/src/main/java/org/springframework/ai/chat/metadata/EmptyUsage.java
```diff
@@ -16,22 +16,30 @@
...
+import java.util.Map;
+
 /**
  * A EmpytUsage implementation that returns zero for all property getters
  *
  * @author John Blum
+ * @author Ilayaperumal Gopinathan
  * @since 0.7.0
  */
 public class EmptyUsage implements Usage {
```
markpollack, line 28: we should think about getting rid of this and use 'new DefaultUsage(0,0)

markpollack:
Added support to ser/deser the native usage object
Changed the json ser/deser approach to use a factory method instead of ctors
Updated equals/hashmap/toString to include native usage object
Added more tests including one for backwards deser from M5 format.
Added upgrade notes to docs

We should discuss getting rid of EmtpyUsage, can call DefaultUsage(0, 0)

merged in 4b64aa0ca638ce323db511749a8f593de8f70b7b

PR #2014, opened 2024-12-27
Remove UUID Enforcement for ID Column in PGVectorStore

jitokim:
resolve #1856

PR #2111, opened 2025-01-24
Fix Handling of UUID and Other ID Formats in PgVectorStore

jitokim:
Fix https://github.com/spring-projects/spring-ai/pull/2014 .

Refactor ID handling for different IdType formats
- Add handling for UUID, TEXT, INTEGER, SERIAL, BIGSERIAL formats in `convertIdToPgType` function.
- Implemented type conversion logic based on the IdType value (UUID, TEXT, INTEGER, SERIAL, BIGSERIAL).
- Add unit tests to validate correct conversion for UUID and non-UUID IdType formats.
  - `testToPgTypeWithUuidIdType`: Validates UUID handling.
  - `testToPgTypeWithNonUuidIdType`: Validates handling for non-UUID IdTypes.

### DIFF
--- a/vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/vectorstore/pgvector/PgVectorStore.java
+++ b/vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/vectorstore/pgvector/PgVectorStore.java
@@ -35,10 +35,8 @@ import org.slf4j.LoggerFactory;
 
 import org.springframework.ai.document.Document;
 import org.springframework.ai.document.DocumentMetadata;
-import org.springframework.ai.embedding.BatchingStrategy;
 import org.springframework.ai.embedding.EmbeddingModel;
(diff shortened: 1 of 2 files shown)


jitokim:
In the `afterPropertiesSet` function, when the `initializeSchema` property is `false`, we cannot guarantee that the table's ID type is UUID. Therefore, I added the `PgIdType` to determine how to cast the `document`'s ID. 

I would appreciate your feedback. @ilayaperumalg

jitokim:
How about PgVectorStore supporting only the UUID type?
The complexity might increase. 😂

ilayaperumalg:
@jitokim Thanks for the PR! Rebased, and merged as 4dbe73433afbf0a04a65585753a1f30af9af2ee3. Also, added a way to update the PgVectorAutoConfiguration to use the IdType configuration from PgVector properties via 54f49b55807ffcad591b14df12c6703b9ee548ab

jitokim:
@ilayaperumalg Thanks for review 😃

PR #2115, opened 2025-01-25
Improved ability to use OpenAiApi to support DeepSeek

ricken07:
This pull request contains an enhancement to OpenAiApi's completion request to support [DeepSeek](https://api-docs.deepseek.com/api/create-chat-completio), for which we've added two parameters useful for DeepSeek completion.


smhjamalii:
does it support Janus as well ?

ilayaperumalg:
@ricken07 Thanks for the PR! Rebased, squashed, added a couple of tests to verify the usage and merged as 45421b1c9

PR #2122, opened 2025-01-27
Refactor Amazon Bedrock builder methods

ilayaperumalg:
- Rename builder methods by removing `with` prefix
 - Deprecate the existing methods and mark them for removal post M6


markpollack:
@ilayaperumalg I guess this wasn't needed after all in that we are going to remove these deprecated bedrock chat models in M6?

markpollack:
closing without merging as we will remove the deprecated chat models.

PR #2127, opened 2025-01-27
Add filter-based deletion to Milvus vector store

sobychacko:
Add string-based filter deletion alongside the Filter.Expression-based deletion for Milvus vector store, providing consistent deletion capabilities with other vector store implementations.

Key changes:
- Add delete(Filter.Expression) implementation for Milvus store
- Leverage existing MilvusFilterExpressionConverter for filter translation
- Use Milvus client's native delete API with filter expressions
- Add comprehensive integration tests for filter deletion
- Support both simple and complex filter expressions

This maintains consistency with other vector store implementations while utilizing Milvus-specific APIs for efficient metadata-based deletion.


petros94:
Nice!

PR #2138, opened 2025-01-28
Support store and metadata OpenAI chat options

### LINKED ISSUE #2103: Feature Request: Support "store" and "metadata" options in OpenAiChatOptions.
**Current Behavior**

OpenAI supports the [store](https://platform.openai.com/docs/api-reference/chat/create#chat-create-store) option which can be used as part of the OpenAI [evals](https://platform.openai.com/docs/guides/evals) and model [distillation](https://platform.openai.com/docs/api-reference/chat/create#chat-create-store) processes.  In addition, the `metadata` field can be used to help filter messages for these processes (amongst other thing).  Currently the `OpenAiApi.ChatCompletionRequest` class supports the option fields, however there does not appear to be away to set either of these options using the ChatClient of ChatModel intefaces. 

**Expected Behavior**

It would be desirable for the OpenAiChatOptions class to support the `model` and `metadata` options.

Eg.

```
var promptOptions = OpenAiChatOptions.builder().store(true).metadata(Map.of("use", "training")).build();
chatClient.prompt().messages(userMessage).options(promptOptions).call().content();
```

**Context**
Currently attempting to be able to use stored prompts in OpenAI to evaluate the accuracy of prompts as well as create cheaper/streamlined models using the distillation and fine tuning process.  SpringAI does not currently appear to have a convenient method to set the `store` and `metadata` options using the ChatClient or ChatModel interfaces.

One alternative is using the low level  `OpenAiApi.ChatCompletionRequest` structure for building prompt JSON and use other libraries for HTTP calls to OpenAI, but this removes many of the value adds (e.g. Advisors) available in ChatClient.

ilayaperumalg:
- Add store and metadata as OpenAI Chat options
 - Add test

Resolves #2103

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatOptions.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatOptions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


markpollack:
added docs, merged in b08b9e8cf4cf8bd33994b7753b0c1eb4ee1bb8e0

Issue #2049, opened 2025-01-06
Advancing Tool Support in Spring AI


ThomasVitale:
This issue presents ideas and suggestions for consolidating and advancing tool support in Spring AI, before reaching the first GA version.

## Tool Registration

Tools are modelled via the `ToolCallback` API. By default, Spring AI supports registering tools from methods and from functional objects.

**Methods**

In the context of Java, a tool can naturally be modelled as a **method**. Spring AI support registering methods as tools via the `MethodToolCallback` API.

Declarative tool registration is supported via the `@Tool` annotation on methods.

```java
class MyTools {

    @Tool(description = "Get the list of books written by the given author available in the library")
    List<Book> booksByAuthor(String author) {
        return bookService.getBooksByAuthor(new Author(author));
    }

}
```

```java
@GetMapping("/chat")
String chat(String authorName) {
    return chatClient.prompt()
            .user("What books written by %s are available in the library?".formatted(authorName))
            .tools(new MyTools())
            .call()
            .content();
}
```

**Functions**

In special cases, there might be a wish for modelling tools as **functional objects** (`Function`/`BiFunction`/`Supplier`/`Consumer`). Spring AI support registering functional objects as tools via the `FunctionToolCallback` API.

```java
@GetMapping("/chat")
String chat(String authorName) {
    return chatClient.prompt()
            .user("What books written by %s are available in the library?".formatted(authorName))
            .tools(FunctionToolCallback.builder("booksByAuthor", author -> bookService().getBooksByAuthor(author))
                        .description("Get the list of books written by the given author available in the library")
                        .inputType(BookService.Author.class)
                        .build())
            .call()
            .content();
}
```

## Tool Execution

By default, tool execution is performed transparently by a `ChatModel` object using the `ToolCallingManager` API. In case of errors while executing tools, a `ToolCallExceptionConverter` is used to process the exception.

If you need full control over the tool execution, for example by asking the user permission before each tool execution, you can disable the internal tool execution feature and manage the tool calling workflow yourself using the `ToolCallingManager` API directly.

```java
ChatOptions chatOptions = ToolCallingChatOptions.builder()
			.toolCallbacks(ToolCallbacks.from(tools))
			.internalToolExecutionEnabled(false)
			.build();

Prompt prompt = new Prompt("What's the weather in Copenhagen?", chatOptions);

ChatResponse chatResponse = chatModel.call(prompt);

if (chatResponse.hasToolCalls()) {
[...]

PR #2148, opened 2025-01-29
Advancing Tool Support - Part 4

ThomasVitale:
* Adopted new tool calling logic in `OllamaChatModel`, while maintaining full API backward compatibility thanks to the `LegacyToolCallingManager`.
* Improved efficiency and robustness of merging options in prompts for Ollama.
* Update Ollama Autoconfiguration to use the new `ToolCallingManager`.
* Improved troubleshooting for new tool calling APIs and finalised changes for full backward compatibility.
* Updated Ollama Testcontainers dependency to 0.5.7.

Relates to gh-2049

### DIFF
--- a/models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatRequestTests.java
+++ b/models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatRequestTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,7 +29,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  * @author Christian Tzolov
  * @author Thomas Vitale
  */
-public class OllamaChatRequestTests {
+class OllamaChatRequestTests {
 
 	OllamaChatModel chatModel = OllamaChatModel.builder()
 		.ollamaApi(new OllamaApi())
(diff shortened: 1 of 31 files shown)


models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatRequestTests.java
```diff
@@ -29,17 +29,18 @@
...
+class OllamaChatRequestTests {
 
 	OllamaChatModel chatModel = OllamaChatModel.builder()
 		.ollamaApi(new OllamaApi())
 		.defaultOptions(OllamaOptions.builder().model("MODEL_NAME").topK(99).temperature(66.6).numGPU(1).build())
 		.build();
 
 	@Test
-	public void createRequestWithDefaultOptions() {
+	void createRequestWithDefaultOptions() {
```
ThomasVitale, line 40: These are the only Ollama tests that I had to change, and just because they are testing implementation details and not the public API. All the other tests I left untouched, supporting the backward compatibility validation.

spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/chat/model/ToolCallingAutoConfiguration.java
```diff
@@ -0,0 +1,78 @@
...
+
+	@Bean
+	@ConditionalOnMissingBean
+	ToolCallExceptionConverter toolCallExceptionConverter() {
+		return new DefaultToolCallExceptionConverter(false);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	ToolCallingManager toolCallingManager(ToolCallbackResolver toolCallbackResolver,
```
ThomasVitale, line 68: This can be re-used across all ChatModel instances.
In special cases, like Vertex, it can be overwritten to provide a variant using the non-standard schema config.

vector-stores/spring-ai-milvus-store/pom.xml
```diff
@@ -54,6 +54,15 @@
 			<version>${milvus.version}</version>
 		</dependency>
 
+		<!-- This is HORRIBLE! But IntelliJ doesn't want to work without it -->
```
ThomasVitale, line 57: There's something in the Milvus Java SDK causing a compilation error in IntelliJ if Lombok is not added as a dependency. And the error propagates to the spring-ai-spring-boot-auto-configure module, making it impossible to run auto tests from the IDE. I should probably remove this from the PR (I forgot to undo the change before committing), but since it's here I figured it'd be good to bring this up.

tzolov:
Thanks @ThomasVitale 
Rebased, cleaned and merged at b902ca2afb7fdd918f32eb40c0c3b0761aff4aca

PR #2154, opened 2025-02-01
Re-add slf4j logging

ilayaperumalg:
- This PR reverts the changes to update Apache Commons Logging and re-add the previously used slf4j logging

### DIFF
--- a/models/spring-ai-azure-openai/pom.xml
+++ b/models/spring-ai-azure-openai/pom.xml
@@ -59,6 +59,11 @@
 			<artifactId>spring-context-support</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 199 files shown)


models/spring-ai-azure-openai/pom.xml
```diff
@@ -59,6 +59,11 @@
 			<artifactId>spring-context-support</artifactId>
 		</dependency>
```
markpollack, line 61: I think all we need is

```
		<dependency>
			<groupId>org.slf4j</groupId>
			<artifactId>slf4j-api</artifactId>
			<version>${slf4j-api.version}</version>
		</dependency>
```

markpollack:
made the change to use slf4j directly.  Need to look, I think there are some version numbers in some of the slf4j deps, not managed via boot.

merged in 2932769883ea0b6bc6fecf23160119e4dd2be1c6

PR #2155, opened 2025-02-01
feat: Update OpenAI Chat Models to Latest Versions

apappascs:
This commit comprehensively updates the OpenAiApi `ChatModel` enum to include all the latest OpenAI chat completion models and enhances the documentation.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -326,119 +326,326 @@ public class OpenAiApi {
 	}
 
 	/**
-	 * OpenAI Chat Completion Models:
-	 *
+	 * OpenAI Chat Completion Models.
+	 * <p>
+	 * This enum provides a comprehensive list of chat completion models available through
+	 * the OpenAI API, along with their key features and links to the official OpenAI
+	 * documentation for further details.
+	 * <p>
+	 * The models are grouped by their capabilities and intended use cases. For each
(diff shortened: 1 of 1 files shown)


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -326,119 +326,326 @@ public <T> ResponseEntity<EmbeddingList<Embedding>> embeddings(EmbeddingRequest<
...
 		O1("o1"),
 		/**
-		 * Latest o1 model snapshot
+		 * Latest o1 model snapshot. Supports both text and image inputs, and produces
+		 * text outputs (including Structured Outputs).
+		 * <p>
+		 * Context window: 200,000 tokens. Max output tokens: 100,000 tokens. Knowledge
+		 * cutoff: October, 2023.
 		 */
 		O1_2024_12_17("o1-2024-12-17"),
```
ilayaperumalg, line 375: We need to remove the usage of snapshot model names in these Model enums for the following reasons:
- These enums require constant updates with the latest snapshots
- Post GA, we can no longer change these often and these references would become obsolete. 

I will go ahead and remove these snapshots and merge the PR.

ilayaperumalg:
@apappascs Thanks for the PR. Rebased and merged after removing the snapshot model names from the ChatModel enum: 74178b9f0

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
apappascs: Agree with you @ilayaperumalg .. good point.

PR #2160, opened 2025-02-02
Fix spell error and updated gemma model name

themanojshukla:
The `chat.model` property snippet was missing correct `.options` hierarchy, fixed it. Also fixed couple of spell-errors.


ilayaperumalg:
@themanojshukla Thanks for the PR!

Issue #2137, opened 2025-01-28
Vector store API to gain access to the native client


sobychacko:
Provide a getNativeClient method that returns the underlying vector store API class so users can always have an escape hatch to get client specific functionality.

PR #2161, opened 2025-02-02
GH-2137: Add getNativeClient API to VectorStore interface

sobychacko:
Fixes: #2137

Adds getNativeClient API to VectorStore interface allowing access to the underlying native client implementation.

This change:
- Adds getNativeClient() default method to VectorStore interface returning Optional<T>
- Implements getNativeClient() in all vector store implementations, exposing their respective native clients
- Adds integration tests verifying native client access for all implementations


markpollack:
Added docs.

Merged in 16a596f8b7f7fa024f6a0ee27d6581a34a57c16c

PR #2162, opened 2025-02-02
Advancing Tool Support - Part 5

ThomasVitale:
* Introduced new `@ToolParam` annotation for defining a description for tool parameters and marking them as (non)required.
* Improved the JSON Schema generation for tools, solving inconsistencies between methods and functions, and ensuring a predictable outcome.
* Added support for returning tool results directly to the user instead of passing them back to the model. Introduced new `ToolExecutionResult` API to propagate this information.
* Consolidated naming of tool-related options in `ToolCallingChatOptions`.
* Fixed varargs issue in `ChatClient` when passing `ToolCallback[]`.
* Introduced new documentation for the tool calling capabilities in Spring AI, and deprecated the old one.
* Bumped jsonschema dependency to 4.37.0.

Relates to gh-2049


models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
```diff
@@ -271,10 +272,19 @@ private ChatResponse internalCall(Prompt prompt, ChatResponse previousChatRespon
...
 		if (ToolCallingChatOptions.isInternalToolExecutionEnabled(prompt.getOptions()) && response != null
 				&& response.hasToolCalls()) {
-			var toolCallConversation = this.toolCallingManager.executeToolCalls(prompt, response);
-			// Recursively call the call method with the tool call message
-			// conversation that contains the call responses.
-			return this.internalCall(new Prompt(toolCallConversation, prompt.getOptions()), response);
+			var toolExecutionResult = this.toolCallingManager.executeToolCalls(prompt, response);
+			if (toolExecutionResult.returnDirect()) {
+				// Return tool execution result directly to the client.
+				return ChatResponse.builder()
```
ThomasVitale, line 278: When a given tool is marked with the option `returnDirect`, the tool call result is sent back to the user directly, wrapping each tool call result in a `Generation` object.

That is especially useful when building agents. For example, if I have a RAG tool, I might want to get the result back directly instead of having the main agent model post-process the RAG answer. Or I might want certain tools to end the reasoning loop of an agent.

spring-ai-core/src/main/java/org/springframework/ai/chat/client/ChatClient.java
```diff
@@ -216,9 +216,9 @@ interface ChatClientRequestSpec {
 
 		ChatClientRequestSpec tools(String... toolNames);
 
-		ChatClientRequestSpec tools(Object... toolObjects);
+		ChatClientRequestSpec tools(FunctionCallback... toolCallbacks);
```
ThomasVitale, line 219: The Java compiler complaints when trying to pass an array of FunctionCallback and requires casting. In this way, we keep the same API, but it's more explicit and no casting is required.

spring-ai-core/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
```diff
@@ -53,12 +53,12 @@ public interface ToolCallingChatOptions extends FunctionCallingOptions {
 	/**
 	 * Names of the tools to register with the ChatModel.
 	 */
-	Set<String> getTools();
+	Set<String> getToolNames();
```
ThomasVitale, line 56: Having the same API (`tools()`) in `ChatClient` and `ToolCallingChatOptions`, but with very different behaviours, didn't seem good.

Now, the `ChatClient` keeps having one single API with overloaded methods to support different use cases: `tools()`.
`ToolCallingChatOptions`, instead, reflects the two types of tool-related information it stores: `toolNames()` and `toolCallbacks()`.

@tzolov what do you think?

spring-ai-core/src/main/java/org/springframework/ai/model/tool/ToolCallingManager.java
```diff
@@ -39,7 +38,7 @@ public interface ToolCallingManager {
 	/**
 	 * Execute the tool calls requested by the model.
 	 */
-	List<Message> executeToolCalls(Prompt prompt, ChatResponse chatResponse);
+	ToolExecutionResult executeToolCalls(Prompt prompt, ChatResponse chatResponse);
```
ThomasVitale, line 41: I introduced this to propagate back whether the "returnDirect" option is enabled. Even without that use case, I think it's still a better return type than what I used before (`List<Message>`). It's more easily understandable.

spring-ai-core/src/main/java/org/springframework/ai/tool/annotation/ToolParam.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Marks a tool argument.
+ *
+ * @author Thomas Vitale
+ * @since 1.0.0
+ */
+@Target({ ElementType.PARAMETER, ElementType.FIELD, ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface ToolParam {
```
ThomasVitale, line 34: For the naming, I followed the most used convention in Spring Framework (`@RequestParam` `@CookieParam...`).

I made it tool-specific, but this could actually be used also when using Structured Outputs (as well as the entire JSON schema generation logic in `JsonSchemaGenerator`). But if we want that, we need to come up with a better name).

spring-ai-core/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Stream;
+
+/**
+ * Utilities to generate JSON Schemas from Java types and method signatures. It's designed
+ * to work well in the context of tool calling and structured outputs, aiming at ensuring
+ * consistency and robustness across different model providers.
+ * <p>
+ * Metadata such as descriptions and required properties can be specified using one of the
```
ThomasVitale, line 51: This was the main inconsistency I fixed in this PR. Methods and functions were treated differently, with functions having undocumented broader support for annotations. Now there's parity of behaviour between them when it comes to description and required properties.

tzolov:
Rebased and merged at 560baaae5ae96fb6a395b8c2dd0a386ea55d40f3

markpollack:
related to epic https://github.com/spring-projects/spring-ai/issues/2207

PR #2164, opened 2025-02-03
Add flexible API key management for OpenAI

markpollack:
Introduces a new API key management system that allows users to customize how API keys are provided and managed in their Spring AI applications. This change improves security and flexibility by:

- Adding core ApiKey interface and SimpleApiKey implementation
- Introducing @OpenAiApiKey qualifier for bean disambiguation
- Supporting custom API key providers for secure key management
- Adding auto-configuration support for API key injection
- Adding builder pattern for OpenAiApi configuration
- Deprecating public constructors in favor of builder API (since 1.0.0.M6)

The new system enables users to implement their own key management strategies while maintaining backward compatibility with property-based configuration.


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApiKey.java
```diff
@@ -0,0 +1,35 @@
...
+ * Qualifier annotation for OpenAI API key beans. Used to distinguish OpenAI API keys from
+ * other provider API keys.
+ *
+ * @author Mark Pollack
+ */
+@Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.TYPE, ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Qualifier
+public @interface OpenAiApiKey {
```
markpollack, line 33: change to `OpenAiChatApiKey` so it is specific to chat models, there will be `OpenAiEmbeddingApiKey`  different models can have different endpoints.

markpollack:
Don’t merge, need to remove the annotation to override the ApiKey implementation, the strategy isn’t working out.

markpollack:
merged in fc2690cdf8f9765862da9aad1845fec528320d21

Issue #2165, opened 2025-02-04
Simplify VectorStore.delete(List<String> idList) to return void


sobychacko:
Currently, the delete method returns Optional<Boolean>, which adds unnecessary complexity and is not meaningful for most implementations. The method should instead return void and rely on runtime exceptions to signal errors.

Changes needed:

- Update VectorStore interface to change delete() return type from Optional<Boolean> to void
- Modify all implementing classes to remove boolean return values
- Update related tests to remove Optional<Boolean> assertions

ilayaperumalg:
Merged as [6035516](https://github.com/spring-projects/spring-ai/commit/60355160445f09db85d4f3afa5848a5506833b53)
 ·

PR #2166, opened 2025-02-04
GH-2165: Simplify VectorStore delete method to return void

sobychacko:
- Change VectorStore.delete() and related implementations to return void instead of Optional<Boolean>
- Remove unnecessary boolean return values and success status checks across all vector store implementations
- Clean up tests by removing redundant assertions
- Implementations continue using runtime exceptions for error signaling


ilayaperumalg:
LGTM, merging

PR #2188, opened 2025-02-07
OpenAI - Adopt ToolCallingManager API

ThomasVitale:
- Update OpenAiChatModel to use the new ToolCallingManager API, while ensuring full API backward compatibility.
- Introduce Builder to instantiate a new OpenAiChatModel since the number of overloaded constructors is growing too big.
- Update documentation about tool calling and OpenAI support for that.
- Add extra validation to ensure the uniqueness of tool names when aggregated from different sources.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -195,29 +211,48 @@ public OpenAiChatModel(OpenAiApi openAiApi, OpenAiChatOptions options,
...
+		logger.warn("This constructor is deprecated and will be removed in the next milestone. "
+				+ "Please use the OpenAiChatModel.Builder or the new constructor accepting ToolCallingManager instead.");
+	}
 
+	public OpenAiChatModel(OpenAiApi openAiApi, OpenAiChatOptions defaultOptions, ToolCallingManager toolCallingManager,
+			RetryTemplate retryTemplate, ObservationRegistry observationRegistry) {
+		// We do not pass the 'defaultOptions' to the AbstractToolSupport,
+		// because it modifies them. We are using ToolCallingManager instead,
+		// so we just pass empty options here.
+		super(null, OpenAiChatOptions.builder().build(), List.of());
```
ilayaperumalg, line 237: Do we still need to extend `AbstractToolCallSupport` class?

ThomasVitale: We kept that to ensure API backward compatibility, but it will be removed right after M6 is released.

tzolov:
Rebased and merged at 52198ed9d774b0fc983d532fd7b6c582bb003ba9

Issue #2116, opened 2025-01-26
Support deepseek V3 & R1


impactCn:
The deepseek model cost reduces the ai model to a very low level. More and more people are using this open source model. I want to integrate it into spring ai so that more people can use this model directly.

ricken07:
DeepSpeek is compatible with openAi sdk, you can directly use deepSeek models with Spring AI Open AI integration as for Groq you just need to change the base URL and search for its api key in the configuration, I have submitted this  [Pull Request](https://github.com/spring-projects/spring-ai/pull/2115) to improve deepSeek support

robbyzhaox:
It is stated in DeepSeek's [api-docs](https://api-docs.deepseek.com/) that the API is compatible with OpanAI.

mbetrifork:
Spring-ai might not be perfectly compatible with reasoning models in general, in the sense that the application may want to deal with reasoning/thinking tokens independently of the final response.

markpollack:
I think we need to create a deepseek model module.

This commit https://github.com/spring-projects/spring-ai/commit/45421b1c93bbf19f5d7c5871f5064c9a4df8b067

added fields to the response of openai which are specific to deepseek.  This sort of leakage is happening in multiple places from models that are a super-set of openai api.  

There are a couple options on how to approach this.

1. We are having a RC1 next month, March 13.  I am not sure if this work can realistically be done by then.  It would be nice to have this model in spring-ai proper.
2. A more conservative option, that actually increases velocity, as you would be in full control of the repository, is to have the deepseek model development can take place as part of the new spring-ai-community organization.  You can read more about it here - https://github.com/spring-ai-community/community/wiki  Then when it is mature , it could be moved into spring-ai.

I think option 2 is more realistic, but let me know.  @mxsl-gr Thoughts?

mxsl-gr:
hi, @markpollack 

Option 2 does sound like the safer path. That said, I believe the timing might still work out. I should have some time over the next two weeks and will try updating the implementation based on the previous PR: https://github.com/spring-projects/spring-ai/pull/702.

If the updates bring it to a ready state before the RC1 deadline, we could merge with including it in Spring AI. Otherwise, I’m happy to migrate the work to the Spring AI Community organization and continue development there until it's ready to be promoted back into the main project.

apappascs:
Hi @mxsl-gr, I just saw the latest messages and the PR: https://github.com/spring-projects/spring-ai/pull/2192.

[...]

PR #2192, opened 2025-02-07
Add Reasoning Content support to OpenAiChatModel and related classes (For deepseek-reasoner)

apappascs:
This PR resolves: https://github.com/spring-projects/spring-ai/issues/2283 by adding reasoning content support to OpenAiChatModel and related classes, following the approach of this https://github.com/spring-projects/spring-ai/commit/45421b1c93bbf19f5d7c5871f5064c9a4df8b067 .

Reasoning Content Details
```
reasoning_content : string / nullable  
For DeepSeek-Reasoner model only. Represents the reasoning contents of the assistant's message before the final answer.  
```

Refer to the DeepSeek API documentation for more details: [API Docs](https://api-docs.deepseek.com/api/create-chat-completion).

Changes in This PR
	•	Added reasoningContent field to metadata in OpenAiChatModel.
	•	Updated ChatCompletionMessage to include reasoningContent.
	
With this implementation, we can build projects like [DeepClaude](https://github.com/getAsterisk/deepclaude), enabling the integration of DeepSeek R1’s logical reasoning capabilities with Anthropic Claude’s (or any other model’s) creative and coding prowess. This will provide a powerful, combined language model experience.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -258,7 +258,8 @@ public class OpenAiChatModel extends AbstractToolCallSupport implements ChatMode
 							"role", choice.message().role() != null ? choice.message().role().name() : "",
 							"index", choice.index(),
 							"finishReason", choice.finishReason() != null ? choice.finishReason().name() : "",
-							"refusal", StringUtils.hasText(choice.message().refusal()) ? choice.message().refusal() : "");
+							"refusal", StringUtils.hasText(choice.message().refusal()) ? choice.message().refusal() : "",
+							"reasoningContent", StringUtils.hasText(choice.message().reasoningContent()) ? choice.message().reasoningContent() : "");
 					// @formatter:on
 					return buildGeneration(choice, metadata, request);
 				}).toList();
@@ -346,7 +347,8 @@ public class OpenAiChatModel extends AbstractToolCallSupport implements ChatMode
 									"role", roleMap.getOrDefault(id, ""),
 									"index", choice.index(),
 									"finishReason", choice.finishReason() != null ? choice.finishReason().name() : "",
-									"refusal", StringUtils.hasText(choice.message().refusal()) ? choice.message().refusal() : "");
+									"refusal", StringUtils.hasText(choice.message().refusal()) ? choice.message().refusal() : "",
+									"reasoningContent", StringUtils.hasText(choice.message().reasoningContent()) ? choice.message().reasoningContent() : "");
 
 							return buildGeneration(choice, metadata, request);
 						}).toList();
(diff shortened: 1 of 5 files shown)


yangshaotuo (approved):
When can it be officially used

KonngExplorer (approved):

markpollack:
I think the number of changes that are occuring to make the openai model compatible with other models is starting to go too far, divergence starts to seem inevitable.  We should look to make it easy to extend the openai model support for different dedicated variations with minimal code, as compared to the current full blown cut and paste approach.  Moving this to the RC1 release so that we can better support deepseek as it's own first class model.

apappascs:
> I think the number of changes that are occuring to make the openai model compatible with other models is starting to go too far, divergence starts to seem inevitable. We should look to make it easy to extend the openai model support for different dedicated variations with minimal code, as compared to the current full blown cut and paste approach. Moving this to the RC1 release so that we can better support deepseek as it's own first class model.

Hi @markpollack 

Thanks for the feedback on the PR—makes total sense.

Just to make sure I fully understand your intention before moving forward—are you suggesting that instead of adapting the existing spring-ai-openai module for DeepSeek compatibility, we should treat DeepSeek as its own first-class model (like spring-ai-mistral-ai, spring-ai-moonshot, etc.) and place it under models/spring-ai-deepseek or something similar?

Let me know if that’s the direction you had in mind, or if you’d prefer another approach. Once confirmed, I’m happy to take the lead on the implementation.

Thanks again!

LinkCaau:
When using OpenAI compatibility mode, some models may extend the protocol, and the rapid development of large models makes it difficult for frameworks like Spring AI to adapt quickly and flexibly. Is there a better way to achieve a permanent solution?  
Examples:  
1. DeepSeek added a `reasoning_content` field.  
2. QianFan `finish_reason` field introduced a new type `normal`, but since the type is fixed as `ChatCompletionFinishReason`, it causes serialization failures.

markpollack:
Hi @apappascs 

yes, you understood me correctly.  

Please take a look at the comment here - https://github.com/spring-projects/spring-ai/issues/2116#issuecomment-2813935853

and let me know how you want to proceed.  As I'm going through the issues and doing triage, I see more and more of these deepseek issues that suggest it may take a bit of time to get the implementation stable in time for RC1 next May 13

I suggest first going into the spring-ai-community and then moving it over into spring-ai once it matures.

@mxsl-gr

apappascs:
Hi @markpollack,

Thanks for confirming.

I’ll get started on it. I believe we can have the implementation ready directly in spring-ai before the RC1 deadline on May 13. The spring-ai-community route does seem like a safer fallback if needed, but the goal is to have it solid enough to go straight into the main project.

markpollack:
@apappascs it seems like a community project to get active incubation and feedback would be the most effective route.  I suspect we will eventually need a `ReasoningClient` as well - will be interesting to see how it turns out.

https://github.com/spring-ai-community/community/issues

markpollack:
@mxsl-gr @apappascs There is now a deepseek model in spring ai to handle this without interfering with the OpenAiChatModel.  Closing this as any changes should be in that part of the code base.  

thanks!

PR #2194, opened 2025-02-07
Add vector store delete API ref docs with examples

sobychacko:
- Document delete APIs with ID lists and filter expressions
- Add versioning use case with metadata-based updates


markpollack:
merged in 86a73f9e87e97345aa58d5d128f3d6e051821841

PR #2209, opened 2025-02-10
Add missing integration tests for delete by ID API in vector store im…

sobychacko:
…plementations.

### DIFF
--- a/spring-ai-test/src/main/java/org/springframework/ai/test/vectorstore/BaseVectorStoreTests.java
+++ b/spring-ai-test/src/main/java/org/springframework/ai/test/vectorstore/BaseVectorStoreTests.java
@@ -42,6 +42,12 @@ import org.springframework.ai.vectorstore.filter.Filter;
  */
 public abstract class BaseVectorStoreTests {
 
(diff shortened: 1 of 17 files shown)


spring-ai-test/src/main/java/org/springframework/ai/test/vectorstore/BaseVectorStoreTests.java
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * Base test class for VectorStore implementations. Provides common test scenarios for
+ * delete operations.
+ *
+ * @author Soby Chacko
+ */
+public abstract class BaseVectorStoreTests {
+
+	protected abstract void executeTest(Consumer<VectorStore> testFunction);
```
ilayaperumalg, line 45: A javadoc here on what to expect for the implementations would be helpful

ilayaperumalg (commented):
Nice refactoring to the vector store ITs!

ilayaperumalg:
Rebased, squashed and merged as 4d692a542

Issue #2207, opened 2025-02-10
Switch to updated function calling APIs to use new tool calling APIs in  chat models


markpollack:
Updates needed for 

- [x] OpenAI
- [x] Azure OpenAI
- [x] Mistral
- [x] Bedrock Converse
- [ ] Anthropic
- [x] Google Vertex
- [x] Ollama

ThomasVitale:
OpenAI has been updated now

ThomasVitale:
MIstral has also been updated

ThomasVitale:
Bedrock, Anthropic and Gemini are also migrated. This task can be closed now.

ilayaperumalg:
Closing this as complete.

PR #2213, opened 2025-02-11
Mistral AI - Adopt ToolCallingManager API

ThomasVitale:
- Update MistralAiChatModel to use the new ToolCallingManager API, while ensuring full API backward compatibility.
- Introduce Builder to instantiate a new MistralAiChatModel since the number of overloaded constructors is growing too big.
- Update documentation about tool calling and Mistral AI support for that.
- Add extra validation to ensure the uniqueness of tool names when aggregated from different sources.

Relates to gh-2207


tzolov:
Thank you @ThomasVitale 
Rebased and merged at 6c3a2c39fcf6eebde8dac001e8d62b13dae1980d

PR #2223, opened 2025-02-12
refactor(bedrock): Migrate from function calling to tool calling

tzolov:
- Replace function calling with tool calling in BedrockProxyChatModel
- Deprecate function calling related code and APIs
- Add new tool calling manager and options
- Update builder pattern to remove "with" prefix from methods
- Update tests and documentation for tool calling

Part of the #2207 epic


ilayaperumalg:
Rebased and merged as a8e305d28

PR #2227, opened 2025-02-12
Deprecate redundant field in ModelObservationContext

ThomasVitale:
Each `ModelObservationContext` takes both a request object (e.g. `Prompt`) and an options object (e.g. `ChatOptions`). However, the options are already included in the request object. This PR deprecates the additional field, which will be removed in a subsequent release.

The reason why the extra field was there in the first place was due to the model implementations not handling request options correctly, requiring a dedicated setter. We started fixing the model implementations now, so we are deprecating the extra field, and we'll remove it in the next release, once we have completed the implementation of a fix for all model implementations.


tzolov:
rebased and merged at a307a2ba30e5d036a25ceab9858cdd6588a26607

tzolov:
rebased and merged at a307a2ba30e5d036a25ceab9858cdd6588a26607

PR #2228, opened 2025-02-12
Anthropic - Adopt ToolCallingManager API

ThomasVitale:
- Update AnthropicChatModel to use the new ToolCallingManager API, while ensuring full API backward compatibility.
- Introduce Builder to instantiate a new AnthropicChatModel since the number of overloaded constructors is growing too big.
- Update documentation about tool calling and Anthropic support for that.

Relates to gh-2207


models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -291,9 +347,18 @@ public Flux<ChatResponse> internalStream(Prompt prompt, ChatResponse previousCha
 				Usage accumulatedUsage = UsageUtils.getCumulativeUsage(currentChatResponseUsage, previousChatResponse);
 				ChatResponse chatResponse = toChatResponse(chatCompletionResponse, accumulatedUsage);
 
-				if (!isProxyToolCalls(prompt, this.defaultOptions) && this.isToolCall(chatResponse, Set.of("tool_use"))) {
-					var toolCallConversation = handleToolCalls(prompt, chatResponse);
-					return this.internalStream(new Prompt(toolCallConversation, prompt.getOptions()), chatResponse);
+				if (ToolCallingChatOptions.isInternalToolExecutionEnabled(prompt.getOptions()) && chatResponse.hasToolCalls() && chatResponse.hasFinishReasons(Set.of("tool_use"))) {
```
ThomasVitale, line 350: For Anthropic in streaming mode, checking for tool calls is not enough. We also need to check for the finish reason.

That was not necessary for OpenAI, Ollama and Mistral, for which it's enough to check for tool calls.

tzolov: Yeh, I remember that there were a cases requiring the finish reason. we have to review the bedrock-converse and the vertex gemini refactoring again.

tzolov:
Thanks @ThomasVitale 
Added minor doc improvements. Rebased and merged at b7dcfc1198aa724c48f492917efb7bf76ade185c

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
ThomasVitale: Yeah, I think we can add some auto-tests in those special cases, also as a way to document the behavior (ensuring tools are only called once).

I thought about adding some additional tests here for Anthropic, but we first need this defect fixed: https://github.com/spring-projects/spring-ai/issues/1370

Issue #2247, opened 2025-02-14
Stack overflow exception in trimAdjacentBlankLines() method due to inefficient regex in ExtractedTextFormatter


solenyk:
**Bug description**
When using **TikaDocumentReader**, I often encountered the following error:  
```
java
Caused by: java.lang.StackOverflowError: null  
    at java.base/java.util.regex.Pattern$Caret.match(Pattern.java:3896)  
    at java.base/java.util.regex.Pattern$Curly.match1(Pattern.java:4597)  
    at java.base/java.util.regex.Pattern$Curly.match(Pattern.java:4546)  
    at java.base/java.util.regex.Pattern$Dollar.match(Pattern.java:3996)  
    at java.base/java.util.regex.Pattern$Caret.match(Pattern.java:3906)  
    at java.base/java.util.regex.Pattern$GroupHead.match(Pattern.java:4969).  
```
After debugging, I found that the issue lies in the non-optimized regular expression from the trimAdjacentBlankLines() method in the ExtractedTextFormatter class. In my case, the problem occasionally occurred even with a files with small number of empty lines (~150) with default VM stack settings -Xmx8192m.

**Steps to reproduce**
For testing the occurrence of this error, I created an **XLSX file** with a large number of empty rows. I am attaching it to this issue.
[Stack_overflow_exception_test.xlsx](https://github.com/user-attachments/files/18801719/Stack_overflow_exception_test.xlsx)

destitutus:
We hit the same issue with PagePdfDocumentReader (not just Tika) and StackOverflowError is thrown from ExtractedTextFormatter.trimAdjacentBlankLines() during page text formatting in handleSinglePage().

Reproducer: A 2.2 MB PDF of a novel (dense prose, many paragraph breaks). PDFBox extracts the text fine, but the regex in trimAdjacentBlankLines() overflows the stack on the resulting page text.

```
Stack trace (bottom frames):
java.lang.StackOverflowError
    at java.base/java.lang.String.charAt(String.java:1626)
    at java.base/java.util.regex.Pattern$Caret.match(Pattern.java:3896)
    at java.base/java.util.regex.Pattern$GroupHead.match(Pattern.java:4969)
    at java.base/java.util.regex.Pattern$Loop.match(Pattern.java:5078)
    at java.base/java.util.regex.Pattern$GroupTail.match(Pattern.java:5000)
    ...
```

[...]

PR #2248, opened 2025-02-14
Optimize regex in trimAdjacentBlankLines() method of ExtractedTextFormatter to prevent stack overflow

solenyk:
These changes were made to close [issue 2247](https://github.com/spring-projects/spring-ai/issues/2247). The optimized regular expression works even with files containing 1500+ empty lines (I tested on files up to 2000 empty lines) on VM settings -Xss256k.

I also removed the `final` keyword from the class declaration because I see cases for its inheritance (for example, right now I was forced to copy the entire class in order to rewrite the regex for my code).


markpollack:
needs test

Issue #2243, opened 2025-02-14
`QueryTransformer` Throws "Unsupported parameter: 'temperature'" for Reasoning Models


dmmrch:
**Bug description**
When the application is configured with reasoning model, e.g.:
```
  ai:
    openai:
      chat:
        options:
          model: o3-mini
```
And we're applying any of available `QueryTransformer`'s – the application throw `Unsupported parameter: 'temperature'` during `Query transform(Query query);` phase. This happens because in the following section, `QueryTransform`er is using default available model, and  `temperature` param is always provided, regardless of the model type

```
public class CompressionQueryTransformer {
...
	@Override
	public Query transform(Query query) {

...
		var compressedQueryText = this.chatClient.prompt()
			.user(user -> user.text(this.promptTemplate.getTemplate())
				.param("history", formatConversationHistory(query.history()))
				.param("query", query.text()))
			**.options(ChatOptions.builder().temperature(0.0).build())**
			.call()
			.content();
...
	}

}
```
I thought of making PR with extending `ChatOptions` by new `reasoningEffort` param, but I'm not sure if this is a good idea, since it affects all `ChatOptions` heirs which has no introduced reasoning models yet, e.g. `MistralAiChatOptions`. 

If you suggest what would be the right way to approach this issue – I'd love to submit PR with fix

**Environment**
Spring AI: tried with `1.0.0-M5` and latest `main` branch state

**Steps to reproduce**
1. Configure Spring AI app with reasoning model (e.g. `o1` or `o3-mini`)
2. Apply either of available `QueryTransformer`'s (e.g. `CompressionQueryTransformer`)
3. Call `Query transform(Query query)` method of provided `QueryTransformer`

**Expected behavior**
`Query transform(Query query)` should work with reasoning models

PR #2258, opened 2025-02-17
fix: remove default temperature on all QueryTransformer implementations

apappascs:
resolving https://github.com/spring-projects/spring-ai/issues/2243

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/rag/preretrieval/query/transformation/CompressionQueryTransformer.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/rag/preretrieval/query/transformation/CompressionQueryTransformer.java
@@ -84,7 +84,7 @@ public class CompressionQueryTransformer implements QueryTransformer {
 			.user(user -> user.text(this.promptTemplate.getTemplate())
 				.param("history", formatConversationHistory(query.history()))
 				.param("query", query.text()))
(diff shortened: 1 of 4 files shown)


ilayaperumalg:
@apappascs Thanks for the PR!

spring-ai-core/src/main/java/org/springframework/ai/rag/preretrieval/query/transformation/CompressionQueryTransformer.java
```diff
@@ -84,7 +84,7 @@ public Query transform(Query query) {
 			.user(user -> user.text(this.promptTemplate.getTemplate())
 				.param("history", formatConversationHistory(query.history()))
 				.param("query", query.text()))
-			.options(ChatOptions.builder().temperature(0.0).build())
```
ilayaperumalg, line 87: @ThomasVitale Is there any specific reason why we had the temperature set here?

apappascs: Thank you for the feedback, @ThomasVitale. I have updated the documentation based on your comment.

v1r3n (approved):

markpollack:
ok to merge

ilayaperumalg:
@apappascs Thanks for the PR. Squashed, rebased and merged as 59fbd9e3e

PR #2278, opened 2025-02-19
fix sample code in bedrock-converse.adoc

dev-jonghoonpark:
fix sample code in bedrock-converse.adoc


ilayaperumalg:
@dev-jonghoonpark Thanks for the PR!

PR #2289, opened 2025-02-21
Modularise Spring AI Spring Boot autoconfigurations

ilayaperumalg:
- Split spring-ai-spring-boot-autoconfigure into modules

     - This PR addresses the restructuring of the following spring boot autoconfigurations:

       - spring-ai retry -> common
       - spring-ai chat client/model/memory -> chat - spring-ai chat/embedding/image observation -> observation - spring-ai chat/embedding models -> models

     - Update the Spring AI BOM and boot starters with the new autoconfigure modules


markpollack:
There is a really good reason to keep it separate, but I’m embarrassed to say I can’t remember the argument presented to me.  Let me sleep on it😁

tzolov:
@ThomasVitale, 
The main reason to keep auto-configurations separate from boot starters is that you can compose auto-configurations but not boot starters. If auto-configurations are nested inside starter modules, they become unusable for composition.
So collocating auto-configs inside the starters doesn't seem a good idea.

ThomasVitale:
Thanks for the explanation! That makes sense! I was more thinking about the "leaf" auto config (like "weaviate"), but I see that also in that case, if it ever needed to be composed into something else, having it inside the starter would be a limitation.

Issue #2298, opened 2025-02-23
Milvus Vector -  Support for Milvus Native Filter Expressions in doSimilaritySearch


waileong:
## Description
Currently, the implementation of `doSimilaritySearch` in `SearchRequest` does not provide direct support for passing native Milvus filter expressions. The existing approach converts `Filter.Expression` into a Milvus-compatible filter using `filterExpressionConverter`, which limits flexibility, especially for advanced filtering capabilities such as the `LIKE` operator, which is currently unsupported.

To address this limitation, we propose adding a new parameter, `String nativeExpression`, to allow direct passing of native Milvus filter expressions. This parameter should take priority over the existing `filterExpressionConverter.convertExpression(request.getFilterExpression())` logic when specified.

## Proposed Changes
1. Add a new optional parameter `String nativeExpression` to `SearchRequest`.
2. Modify `doSimilaritySearch` to prioritize `nativeExpression` if provided, bypassing `filterExpressionConverter.convertExpression`.
3. Ensure that `nativeExpression` integrates seamlessly with the existing search request pipeline.

## Rationale
- Enables support for Milvus-specific filter expressions, including operators like `LIKE`, `IN`, and other advanced filtering capabilities.
- Reduces dependency on `filterExpressionConverter`, providing more flexibility in query execution.
- Improves compatibility with Milvus query syntax for users needing finer control over search filtering.

## Example Usage
```java
SearchRequest request = SearchRequest.builder()
    .query("example query")
    .topK(10)
    .similarityThreshold(0.8)
    .nativeExpression("city LIKE 'New%' AND price > 1000") // Direct Milvus expression
    .build();
```

## Expected Behavior
If `nativeExpression` is provided, it should be used in the search query instead of the converted filter expression. If `nativeExpression` is `null`, fallback to the existing behavior.

## Additional Context
Milvus documentation on boolean expressions: [https://milvus.io/docs/boolean.md](https://milvus.io/docs/boolean.md)

This feature would greatly enhance the flexibility of `doSimilaritySearch`, allowing for more complex and native-compatible queries in Milvus.

waileong:
Related to #2294

waileong:
@ilayaperumalg, I would appreciate it if you could look into this issue.

ilayaperumalg:
@waileong Thank you for creating descriptive and report and the follow up PR.

ilayaperumalg:
@waileong Thank you for your contribution! The changes are merged as e128a39ec

PR #2300, opened 2025-02-24
Updates the `SearchRequest` class to be non-final and adds a `MilvusSearchRequest` subclass that includes Milvus-specific fields for native expressions and search parameters.

### LINKED ISSUE #2294: Milvus Vector - Unable to Set `nprobe` in `doSimilaritySearch` for Default `IVF_FLAT` When `initializeSchema = true`
# Milvus Vector - Unable to Set `nprobe` in `doSimilaritySearch` for Default `IVF_FLAT` When `initializeSchema = true`

## Description
We are encountering an issue in `MilvusVectorStore` where it is **not possible to set `nprobe`** when performing similarity searches (`doSimilaritySearch`) if the index type is **IVF_FLAT**. 

By default, `IVF_FLAT` is selected when `initializeSchema = true`, but `nprobe` is not explicitly set, leading to **poor recall** or **zero results** in some cases.

## Expected Behavior
- Users should be able to **set `nprobe`** when performing similarity searches with `IVF_FLAT`.
- A way to **override `nprobe`** in `doSimilaritySearch()` should be provided via either:
  - A configurable parameter in `SearchRequest`
  - A fallback to a reasonable default value (e.g., `nprobe = 256`)

## Current Behavior
- **IVF_FLAT is used as the default index**, but `nprobe` is **not explicitly set** in `doSimilaritySearch()`.
- This leads to **low recall** or **0 search results** if `nprobe` defaults to a very small value (`1`).
- There is **no exposed method** to override `nprobe` in `SearchRequest`.

## Steps to Reproduce
1. Initialize a `MilvusVectorStore` instance with `initializeSchema = true`.  
2. Add some vector data to the store.  
3. Perform a similarity search using `doSimilaritySearch()`.  
4. Observe that `nprobe` is **not explicitly set**, leading to **poor recall** or **zero results**.

### Code Example:
```java
MilvusVectorStore vectorStore = MilvusVectorStore.builder(milvusClient, embeddingModel)
    .initializeSchema(true)  // Uses default IVF_FLAT
    .build();

SearchRequest searchRequest = SearchRequest.query("example query")
    .withTopK(5)
    .withSimilarityThreshold(0.7);  // No way to set nprobe

List<Document> results = vectorStore.similaritySearch(searchRequest);

// Expecting search results, but sometimes returns 0 records due to missing nprobe

waileong:
- Updated `SearchRequest.java` to make the class non-final.
- Added `MilvusSearchRequest` with specific Milvus parameters such as `nativeExpression` and `searchParamsJson`.
- Modified `doSimilaritySearch` method in `MilvusVectorStore` to handle these new fields from `MilvusSearchRequest`.

Closes #2294, #2298

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/SearchRequest.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/vectorstore/SearchRequest.java
@@ -33,7 +33,7 @@ import org.springframework.util.Assert;
  * @author Thomas Vitale
  * @author Ilayaperumal Gopinathan
  */
-public final class SearchRequest {
+public class SearchRequest {
 
 	/**
 	 * Similarity threshold that accepts all search scores. A threshold value of 0.0 means
@@ -69,6 +69,16 @@ public final class SearchRequest {
 			.topK(originalSearchRequest.getTopK())
 			.similarityThreshold(originalSearchRequest.getSimilarityThreshold())
 			.filterExpression(originalSearchRequest.getFilterExpression());
+	}
+
+	public SearchRequest() {
+	}
+
+	protected SearchRequest(SearchRequest original) {
+		this.query = original.query;
+		this.topK = original.topK;
+		this.similarityThreshold = original.similarityThreshold;
+		this.filterExpression = original.filterExpression;
 	}
 
 	public String getQuery() {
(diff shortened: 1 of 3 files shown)


waileong:
@ilayaperumalg I've added the test case and documentation. Please review and let me know if any further changes are needed. Thanks!

ilayaperumalg:
@waileong Thank you for updating the doc and tests. Rebased and merged as e128a39ec

Issue #2301, opened 2025-02-24
There is a bug in MCP client auto configuration, will cause: IllegalStateException: Multiple tools with the same name


banalite:
Please do a quick search on GitHub issues first, there might be already a duplicate issue for the one you are about to create.
If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

**Bug description**
I am trying to using mcpSyncClient to list tools, but there is a exeception: java.lang.IllegalStateException: Multiple tools with the same name.

**Environment**
Please provide as many details as possible:  Version: 1.0.0-M6. And main baranch: 1.0.0-SNAPSHOT

**Steps to reproduce**
Just using McpSyncClient to list tools

**Expected behavior**

**Minimal Complete Reproducible example**
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.ai.tool.resolution.ToolCallbackResolver]: Factory method 'toolCallbackResolver' threw exception with message: Multiple tools with the same name (getAlerts, getWeatherForecastByLocation)
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.lambda$instantiate$0(SimpleInstantiationStrategy.java:199) ~[spring-beans-6.2.3.jar:6.2.3]
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiateWithFactoryMethod(SimpleInstantiationStrategy.java:88) ~[spring-beans-6.2.3.jar:6.2.3]
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:168) ~[spring-beans-6.2.3.jar:6.2.3]
	at org.springframework.beans.factory.support.ConstructorResolver.instantiate(ConstructorResolver.java:653) ~[spring-beans-6.2.3.jar:6.2.3]
	... 21 common frames omitted
Caused by: java.lang.IllegalStateException: Multiple tools with the same name (getAlerts, getWeatherForecastByLocation)
	at org.springframework.ai.mcp.SyncMcpToolCallbackProvider.validateToolCallbacks(SyncMcpToolCallbackProvider.java:126) ~[spring-ai-mcp-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
[...]

PR #2305, opened 2025-02-24
Fix bug #2301. The SseHttpClientTransportAutoConfiguration will always run for the class name in @ConditionalOnMissingClass is wrong.

banalite:
For the reason in the title, using McpSyncClient to list tools will always failed. As there will be two NamedClientMcpTransports connect to one same server, then will be duplicate tool names from McpSyncClient.listTools(), and this will not pass the check in SyncMcpToolCallbackProvider.validateToolCallbacks(ToolCallback[] toolCallbacks).

### DIFF
--- a/auto-configurations/spring-ai-mcp-client/src/test/java/org/springframework/ai/autoconfigure/mcp/client/SseHttpClientTransportAutoConfigurationTest.java
+++ b/auto-configurations/spring-ai-mcp-client/src/test/java/org/springframework/ai/autoconfigure/mcp/client/SseHttpClientTransportAutoConfigurationTest.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2025-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


tzolov:
@banalite thank you for the contribution. 
The issue was fixed in https://github.com/spring-projects/spring-ai/commit/dcdfd7a72ab3b4fbc5391e28b9bb381b24bb4ba7 
I like the verification tests you've added. Would you please rebase your PR on top of the upstream main so that we can merge your tests?

banalite:
> @banalite thank you for the contribution. The issue was fixed in [dcdfd7a](https://github.com/spring-projects/spring-ai/commit/dcdfd7a72ab3b4fbc5391e28b9bb381b24bb4ba7) I like the verification tests you've added. Would you please rebase your PR on top of the upstream main so that we can merge your tests?

Hi, I have rebased my PR on top of the upstream main.

tzolov:
Thank you @banalite. I'm afraid this test isn't doing much as the sseHttpClientTransportAutoConfiguration is not a bean so it will never be present?

tzolov:
Replaced by https://github.com/spring-projects/spring-ai/pull/2451

PR #2313, opened 2025-02-25
Upgrade Milvus java sdk to 2.5.4

ilayaperumalg:
- Upgrade Milvus java sdk to 2.5.4
  - Remove use of fastjson
    - Use com.google.gson.JsonObject as Milvus SDK replaces fastjson with gson
  - Change SearchResult RowRecord similarity metric name to score
     - Milvus SDK 2.5.4 uses "score" instead of "distance"

### DIFF
--- a/spring-ai-test/src/main/java/org/springframework/ai/test/vectorstore/BaseVectorStoreTests.java
+++ b/spring-ai-test/src/main/java/org/springframework/ai/test/vectorstore/BaseVectorStoreTests.java
@@ -110,7 +110,8 @@ public abstract class BaseVectorStoreTests {
 			assertThat(results.get(0).getId()).isEqualTo(documents.get(2).getId());
 			Map<String, Object> metadata = results.get(0).getMetadata();
 			assertThat(normalizeValue(metadata.get("country"))).isEqualTo("BG");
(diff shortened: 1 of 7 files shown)


spring-ai-test/src/main/java/org/springframework/ai/test/vectorstore/BaseVectorStoreTests.java
```diff
@@ -110,7 +110,8 @@ protected void deleteById() {
 			assertThat(results.get(0).getId()).isEqualTo(documents.get(2).getId());
 			Map<String, Object> metadata = results.get(0).getMetadata();
 			assertThat(normalizeValue(metadata.get("country"))).isEqualTo("BG");
-			assertThat(normalizeValue(metadata.get("year"))).isEqualTo("2023");
+			// the values are converted into Double
+			assertThat(normalizeValue(metadata.get("year"))).isEqualTo("2023.0");
```
ilayaperumalg, line 114: The metadata for number values get converted into double to avoid data loss. But, there can be a better way to handle.

sobychacko:
Merged upstream via https://github.com/spring-projects/spring-ai/commit/0065949d5f771717d31584c5b582d0eb30f62c68.

PR #2326, opened 2025-02-26
correct wrong property name

vinky-w:
correct wrong property name


ilayaperumalg:
@vinky-w Thanks for the PR!

PR #2333, opened 2025-02-26
Fix bug SseHttpClientTransportAutoConfiguration

Yufeng0918:
- fix bug @ConditionalOnMissingClass class declaration
- change from "public class WebFluxSseClientTransport" to "WebFluxSseClientTransport"


tzolov (approved):
Nice catch. Thanks @Yufeng0918

PR #2338, opened 2025-02-27
fix: reverse ToolContext validation logic in MethodToolCallback

### LINKED ISSUE #2337: Reverse the MethodToolCallback's ToolContext validation logic
Change the validation logic in `MethodToolCallback` to check if a `ToolContext` argument is defined by the method but not provided in the request.
This would ensure that methods that expect a ToolContext parameter receive one.

Current logic is checking if a ToolContext is provided but not supported by the method (e.g. the reverse)

tzolov:
Changes the validation logic in MethodToolCallback to check if a ToolContext is required by the method but not provided, rather than checking if a ToolContext is provided but not supported by the method. This ensures methods that expect a ToolContext parameter receive one.

Updates tests cases to reflect the new validation logic

Resolves #2337


ThomasVitale (approved):
Nice!

PR #2353, opened 2025-03-01
feat: Add equals, hashCode, deep copy, and tests to AnthropicChatOptions

apappascs:
This commit enhances `AnthropicChatOptions` by:

- Adding `equals` and `hashCode` methods for proper object comparison.
- Implementing a deep `copy()` method, creating new instances of mutable collections (List, Set, Map, Metadata) to prevent shared state.
- Adding `AnthropicChatOptionsTests` to verify `copy()`, builders, setters, and default values.


tzolov:
Thanks @apappascs for fixing the copy idempotency issues and adding the test

apappascs:
Thank you @tzolov for the review. I will create one pr for each model respectively

PR #2358, opened 2025-03-02
feat: Add equals, hashCode, deep copy, and tests to MiniMaxChatOptions

apappascs:
This commit enhances MiniMaxChatOptions by:
- Updating `equals` and `hashCode` methods for proper object comparison.
- Updating `copy()` method, creating new instances of mutable collections (List, Set, Map, Metadata) to prevent shared state.
- Adding `MiniMaxChatOptionsTests` to verify `copy()`, builders, setters, and default values.


markpollack:
rebased and added a few more tests and defensive copies of some getters.  merged into main d9e30379843de8ceb987e8257d3be69e526ca16d and backported to 1.0.x

thanks!

Issue #2373, opened 2025-03-03
Cannot disable OpenAiModerationModel when using OpenAiAutoConfiguration


gm2552:
**Bug description**
All of the Beans in the `OpenAiAutoConfiguration` class that require an API key have an option to disable the particular bean with the `@ConditionalOnProperty` with the exception of the OpenAiModerationModel.  This has two potential issues:

- Inconsistent AutoConfiguration pattern as all of the other Client or Model beans support an `enabled` property.
- If I want to use constructs/classes in the OpenAI library but do not necessarily need a client connection (i.e. I don't need to set the apiKey), I cannot load my application without setting a bogus API key.  The alternative is to disable AutoConfiguration for this class, but are scenarios where I still may want to auto configure a particular client.

**Environment**
Spring Boot 3.4.x and spring-ai M6.

**Steps to reproduce**
Set the `enabled` flag for all models and client to false and do not set an apiKey (do not disable auto configuration).  The application will crash with the following error:

```
OpenAI API key must be set. Use the connection property: spring.ai.openai.api-key or spring.ai.openai.moderation.api-key property
```

**Expected behavior**
I would expect there to be a consistent pattern for disabling the moderation client.  Something like:
```
spring:
  ai:
    openai:
      moderation:
        enabled: false
```

**Minimal Complete Reproducible example**
Please provide a failing test or a minimal complete verifiable example that reproduces the issue.
Bug reports that are reproducible will take priority in resolution over reports that are not reproducible.

Just create a trivial SpringBootApplication with the OpenAI starter.  Use the following application.yaml.

```
spring:
  ai:
    openai:
      chat:
        enabled: false 
      embedding:
        enabled: false
      image:
        enabled: false
      audio:
        transcription:
          enabled: false
        speech:
          enabled: false
```

sobychacko:
This issue has been fixed since `1.0.0-M7` by the `spring.ai.model.moderation` property. Use `spring.ai.model.moderation=none` to disable.

PR #2374, opened 2025-03-03
Adding @ConditionalOnProperty configuration for OpenAiModerationModel.

gm2552:
Added the ability to disable auto configuration of the OpenAiModerationModel bean.  Addresses reported bug [2373](https://github.com/spring-projects/spring-ai/issues/2373)

### DIFF
--- a/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/openai/OpenAiAutoConfiguration.java
+++ b/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/openai/OpenAiAutoConfiguration.java
@@ -253,6 +253,8 @@ public class OpenAiAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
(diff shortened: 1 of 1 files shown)


spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/openai/OpenAiAutoConfiguration.java
```diff
@@ -253,6 +253,8 @@ public OpenAiAudioTranscriptionModel openAiAudioTranscriptionModel(OpenAiConnect
 
 	@Bean
 	@ConditionalOnMissingBean
+	@ConditionalOnProperty(prefix = OpenAiModerationProperties.CONFIG_PREFIX, name = "enabled", havingValue = "true",
```
ilayaperumalg, line 256: I don't seem to find "enabled" property in OpenAiModerationProperties.

ThomasVitale:
@gm2552 thanks for your contribution! I can see that 1.0.0-M7 changed the way to enable/disable model auto-configurations. For example, for moderation model, the `spring.ai.model.moderation=openai` property is set automatically when OpenAI is in the class path. And it can be disabled by setting it to `spring.ai.model.moderation=none`. Would that address your use case?

PR #2386, opened 2025-03-04
Migrate vector store auto-configurations to dedicated modules

sobychacko:
Move vector store auto-configuration classes to dedicated modules under auto-configurations/vector-stores/:

- Creates separate modules for Milvus, Pinecone, Qdrant, and Typesense vector stores
- Moves CommonVectorStoreProperties to spring-ai-core for better reusability
- Updates pom.xml dependencies to maintain proper relationships between modules
- Name the artifacts based on the pattern spring-ai-autoconfigure-vector-store-<implementation>. For example - spring-ai-autoconfigure-vectore-store-milvus
- Package names follow the pattern org.springframework.ai.vectorstore.<implementation>.autoconfigure
- Naming the correspondinbg starter modules accordingly (spring-ai-starter-vector-store-milvus for example).

This change improves modularity by allowing each vector store implementation to be independently versioned and maintained, continuing the migration pattern established with previous vector stores.


spring-ai-spring-boot-starters/spring-ai-starter-vector-store-milvus/pom.xml
```diff
@@ -48,6 +48,12 @@
             <version>${project.parent.version}</version>
         </dependency>
```
ilayaperumalg, line 50: The dependency `spring-ai-spring-boot-autoconfigure` needs to be removed. I will fix when merging.

ilayaperumalg:
The changes look good. Merged as 0f02e319c

PR #2406, opened 2025-03-07
Migrate vector store observation autoconfiguration to dedicated module

sobychacko:
Update the vector store starters with the new observation autoconfig dependency

Other maven configuraiton changes


ilayaperumalg:
Rebased and merged as d077474b2

Issue #1776, opened 2024-11-20
Support for FILTER_EXPRESSION with RetrievalAugmentationAdvisor


habuma:
As `RetrievalAugmentationAdvisor` is expected to replace `QuestionAnswerAdvisor`, it should also support `FILTER_EXPRESSION` the way that `QuestionAnswerAdvisor` does.

This makes it possible to configure the advisor as a default advisor when creating a `ChatClient` and then plug in a filter expression for searching the vector store when submitting a prompt.

I suspect, based on how this works with `RetrievalAugmentationAdvisor` that the change required would be made to `VectorStoreDocumentRetriever`, as that's the place and implementation of `DocumentRetriever` that would be able to make use of the expression.

habuma:
Just to add some context, here's some code showing what I'm asking for...

```
@RestController
public class AskController {

    private final ChatClient chatClient;
    private final RetrievalAugmentationAdvisor ragAdvisor;

    // If I define the ChatClient with the advisor here...
    AskController(ChatClient.Builder chatClientBuilder, VectorStore vectorStore) {
        var documentRetriever = VectorStoreDocumentRetriever.builder()
                .vectorStore(vectorStore)
                .build();
        this.ragAdvisor = RetrievalAugmentationAdvisor.builder()
                .documentRetriever(documentRetriever)
                .build();
        this.chatClient = chatClientBuilder
                .defaultAdvisors(ragAdvisor)
                .build();
    }

    @PostMapping("/ask")
    public Answer ask(@RequestBody Question question) {
        String someRequestTimeMetadataValue = ...;
        return chatClient.prompt()
                .user(question.question())
                // ...then how can I customize the query sent to the vector store
                //    to filter on someRequestTimeMetadataValue?
                .call()
                .entity(Answer.class);
    }

}
```

It doesn't have to be exactly the same as with `QuestionAnswerAdvisor`, but I need to have some way of refining the vector store query based on request-time values.

ThomasVitale:
Using a similar approach as the one in QuestionAnswerAdvisor, I added the capability to pass a request-time filter expression in this PR. @habuma @marcushellberg would that work for your use case?

https://github.com/spring-projects/spring-ai/pull/2418

marcushellberg:
Yes, that seems to be exactly what I was looking for yesterday when trying it out!

PR #2418, opened 2025-03-09
Request-time filter expressions for RAG

ThomasVitale:
When using the RetrievalAugmentationAdvisor with the VectorStoreDocumentRetriever, it’s now possible to provide a filter expression at request-time as an advisor context variable with key VectorStoreDocumentRetriever.FILTER_EXPRESSION.

Fixes gh-1776


ilayaperumalg:
@ThomasVitale Thanks for the PR! Rebased and merged as 5a4e9f510

PR #2432, opened 2025-03-10
feat(mcp): handle ToolContext in MCP tool callbacks

### LINKED ISSUE #2378: SyncMcpToolCallback call fails with UnsupportedOperationException if ToolContext is set
**Bug description**
Integrating and calling an MCP Tool with an existing Spring AI application that uses a ToolContext will result in an UnsupportedOperationException.

**Environment**
Spring AI 1.0.0-M6

**Steps to reproduce**

1. Integrate https://github.com/modelcontextprotocol/servers/tree/main/src/everything into a Spring AI application using https://docs.spring.io/spring-ai/reference/api/mcp/mcp-client-boot-starter-docs.html
2. Set the ToolContext in your Spring AI application
3. Call one of the tools provided by the everything MCP server, e.g. printEnv
4. Observe "UnsupportedOperationException: Tool context is not supported!" for SyncMcpToolCallback

Exception is raised here: https://github.com/spring-projects/spring-ai/blob/2394ac82ad1229ffefa70e6a338d249890df24b7/spring-ai-core/src/main/java/org/springframework/ai/tool/ToolCallback.java#L57

**Expected behavior**
The MCP printEnv tool is called, ignoring the ToolContext (assuming this concept is not supported in MCP).

tzolov:
- Implement the new ToolCallback.call(String, ToolContext) method in both Sync and Async MCP tool callbacks.
- Since MCP tools don't support tool context, the implementation ignores the context parameter and delegates to the existing call(String) method. Added test to verify the behavior.

Resolves #2378

### DIFF
--- a/mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallback.java
+++ b/mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallback.java
@@ -23,6 +23,7 @@ import io.modelcontextprotocol.spec.McpSchema.CallToolRequest;
 import io.modelcontextprotocol.spec.McpSchema.CallToolResult;
 import io.modelcontextprotocol.spec.McpSchema.Tool;
 
+import org.springframework.ai.chat.model.ToolContext;
 import org.springframework.ai.model.ModelOptionsUtils;
 import org.springframework.ai.tool.ToolCallback;
 import org.springframework.ai.tool.definition.ToolDefinition;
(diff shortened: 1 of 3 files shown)


mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallback.java
```diff
@@ -111,4 +112,10 @@ public String call(String functionInput) {
 		return ModelOptionsUtils.toJsonString(response.content());
 	}
 
+	@Override
+	public String call(String toolArguments, ToolContext toolContext) {
```
ilayaperumalg, line 116: I would like to understand why the default implementation of call(String, ToolContext) at ToolCallback is not enough as it handles null check and call(String) eventually.

PR #2434, opened 2025-03-10
Fix PgVectorStore doDelete function as batch update

CChuYong:
This pull request includes fixing minor issues on `PgVectorStore.doDelete()`.

- handling different types of id
  - `PgVectorStore` can handle different kind of IdTypes from #2111.
  - but `doDelete()` function only accepts UUID, so method will throw `IllegalArgumentException` when id type isn't UUID.
  - so used `convertIdToPgType()` function to correctly map id types.
- used `JdbcTemplate.batchUpdate()` to delete bulk of ids.

### DIFF
--- a/vector-stores/spring-ai-pgvector-store/src/test/java/org/springframework/ai/vectorstore/pgvector/PgVectorStoreIT.java
+++ b/vector-stores/spring-ai-pgvector-store/src/test/java/org/springframework/ai/vectorstore/pgvector/PgVectorStoreIT.java
@@ -75,6 +75,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  * @author Christian Tzolov
  * @author Thomas Vitale
  * @author Jihoon Kim
+ * @author CChuYong
  */
 @Testcontainers
 @EnabledIfEnvironmentVariable(named = "OPENAI_API_KEY", matches = ".+")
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
@CChuYong Thanks for the PR fixing the issue. Could you add a test to validate the fix?

CChuYong:
@ilayaperumalg Sure! i added a test for testing bulk operations of PgVectorStore with different types of id. [ea4632b](https://github.com/spring-projects/spring-ai/pull/2434/commits/ea4632b1df51557ee9ec68d50ab3a5046f5164e5)

vector-stores/spring-ai-pgvector-store/src/test/java/org/springframework/ai/vectorstore/pgvector/PgVectorStoreIT.java
```diff
@@ -232,6 +233,48 @@ public void testToPgTypeWithNonUuidIdType() {
 			});
 	}
 
+	@Test
+	public void testBulkOperationWithUuidIdType() {
```
ilayaperumalg, line 237: Thanks for adding the test. This test runs fine when running as a standalone but fails when all the tests within this class run. Looks like some race condition there. Could you check?

CChuYong: Thanks for your review!
i figured out that `getNativeClientTest` also initializes vector table, so i also added code for dropping table in `getNativeClientTest`. [fb829af](https://github.com/spring-projects/spring-ai/pull/2434/commits/fb829afbaf003570828e50646715c23fd81f31fc)

ilayaperumalg:
@CChuYong Thanks for your contribution! Rebased and merged as 268248ba3

PR #2436, opened 2025-03-10
feat(mcp): Prefix tool names with client name to avoid naming conflicts

### LINKED ISSUE #2393: MCP Client cannot connect to two or more MCP Servers with the same tool names
If an MCP Client is configured with two or more MCP Servers that expose tools with the same name, the client fails to start with and `IllegalStateException: Multiple tools with the same name (doSomethingCool)`.

I understand why Spring AI checks for multiple tools with the same name, but I can see how as MCP becomes more prominent and applications are configured to attach to multiple MCP Servers, the likelihood of this happening are going to increase. And, ultimately it's not the client's fault that there are two servers that have the same name.

Perhaps it would be possible (in the case of MCP) to prepend the tool name with a namespace, either based on the MCP Server's name or even just some UUID or random value. 

FWIW, Claude Desktop doesn't seem to have this issue.  It's unclear to me how they deal with it. Hopefully they do something like namespacing, but I suppose that it's also possible that when a duplicate tool is found at startup, it overwrites the original with the same name (which, if so, would be unfortunate).

tzolov:
Enhances MCP tool naming by prefixing tool names with their client + transport name to prevent conflicts when multiple MCP clients expose tools with the same name.

- Modifies client info in McpClientAutoConfiguration to include transport name
- Updates SyncMcpToolCallback and AsyncMcpToolCallback to prefix tool names with client + transport name
- Adds test coverage for tools with identical names but different client info

Resolves #2393

### DIFF
--- a/auto-configurations/spring-ai-mcp-client/src/main/java/org/springframework/ai/autoconfigure/mcp/client/McpClientAutoConfiguration.java
+++ b/auto-configurations/spring-ai-mcp-client/src/main/java/org/springframework/ai/autoconfigure/mcp/client/McpClientAutoConfiguration.java
@@ -111,6 +111,18 @@ import org.springframework.util.CollectionUtils;
 public class McpClientAutoConfiguration {
 
 	/**
+	 * Create a dynamic client name based on the client name and the name of the server
+	 * connection.
+	 * @param clientName the client name as defined by the configuration
+	 * @param serverConnectionName the name of the server connection being used by the
+	 * client
+	 * @return the connected client name
+	 */
+	private String connectedClientName(String clientName, String serverConnectionName) {
+		return clientName + " - " + serverConnectionName;
+	}
+
+	/**
 	 * Creates a list of {@link McpSyncClient} instances based on the available
 	 * transports.
 	 *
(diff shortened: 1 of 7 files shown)


auto-configurations/spring-ai-mcp-client/src/main/java/org/springframework/ai/autoconfigure/mcp/client/McpClientAutoConfiguration.java
```diff
@@ -144,8 +144,8 @@ public List<McpSyncClient> mcpSyncClients(McpSyncClientConfigurer mcpSyncClientC
 		if (!CollectionUtils.isEmpty(namedTransports)) {
 			for (NamedClientMcpTransport namedTransport : namedTransports) {
 
-				McpSchema.Implementation clientInfo = new McpSchema.Implementation(commonProperties.getName(),
-						commonProperties.getVersion());
+				McpSchema.Implementation clientInfo = new McpSchema.Implementation(
+						commonProperties.getName() + " - " + namedTransport.name(), commonProperties.getVersion());
```
ilayaperumalg, line 148: It would be good to move this logic to a method for reusability and consistency across.

tzolov: sure. It is addressed

ilayaperumalg:
Rebased and merged as 957150128

Issue #2447, opened 2025-03-12
SyncMcpToolCallback call method NPE


Fj-ivy:
**Bug description**
A clear and concise description of what the bug is about.
I create mcp client to invoke @modelcontextprotocol/server-filesystem by stdio.  It can invoke server-filesystem success, But when 
SyncMcpToolCallback call method response field error is null,  it will NPE.

![Image](https://github.com/user-attachments/assets/1a62e63f-2bce-4260-8567-77ffeb2943de)

**Environment**
Spring AI version: 1.0.0-SNAPSHOT
Java Version: 21

**Steps to reproduce**

- application.properties

 ```
 spring.application.name=mcp-examples
  spring.ai.ollama.base-url=http://localhost:11434
  spring.ai.ollama.chat.model=qwen2.5:latest
  spring.ai.mcp.client.enabled=true
  spring.ai.mcp.client.name=mcp-client
  spring.ai.mcp.client.version=1.0.0
  spring.ai.mcp.client.type=SYNC
  spring.ai.mcp.client.request-timeout=30s
  spring.ai.mcp.client.stdio.servers-configuration=classpath:/mcp-servers-config.json
```

- mcp-servers-config.json
```
{
      "mcpServers": {
        "filesystem": {
          "command": "npx",
          "args": [
            "-y",
            "@modelcontextprotocol/server-filesystem",
            "."
          ]
        }
      }
    }
```

- controller
```
@RestController
public class FileSystemController {

    @Resource
    private OllamaChatModel ollamaChatModel;

    @Resource
    private SyncMcpToolCallbackProvider toolCallbackProvider;

    @GetMapping("/chat")
    public String call(@RequestParam String input) {
        ChatClient chatClient = ChatClient.builder(ollamaChatModel)
                .defaultTools(toolCallbackProvider.getToolCallbacks())
                .build();
        return chatClient.prompt(input).call().content();
    }
}
```

- visit http://localhost:8080/chat?input=create folder mcp

Fj-ivy:
https://github.com/spring-projects/spring-ai/pull/2452

PR #2452, opened 2025-03-12
SyncMcpToolCallback call method NPE

Fj-ivy:
fix SyncMcpToolCallback call method NPE when response error is null


tzolov:
Nice catch @Fj-ivy Can you please apply the same fix to the AsyncMcpToolCallback?

tzolov:
@Fj-ivy, can you elaborate on when do you see this NPE? Is it a particular server missing to set it?

Fj-ivy:
> @Fj-ivy, can you elaborate on when do you see this NPE? Is it a particular server missing to set it?

 Steps to reproduce refer this issue  https://github.com/spring-projects/spring-ai/issues/2447.

tzolov:
> > @Fj-ivy, can you elaborate on when do you see this NPE? Is it a particular server missing to set it?
> 
> Steps to reproduce refer this issue #2447.

I see. We can not assume that all MCP servers implementations will be spec compliant.

tzolov:
Ok, I will fix the AsyncMcpToolCallback will merging this PR

PR #2453, opened 2025-03-12
minor fix azure openai chat model doc

magware-dev:
This fixes a minor issue in the azure openai chat model doc `OpenAIClientBuilderCustomizer` usage example

When I created https://github.com/spring-projects/spring-ai/pull/2224 I used an old code version that I had locally where the class name was outdated.

This fixes the documentation.

ping @sobychacko


ilayaperumalg:
@magware-dev Thanks for the PR!

Issue #2456, opened 2025-03-13
PromptTemplate validation prevents the use of StringTemplate built-in functions


asw12:
**Bug description**
PromptTemplate has a validate method that compares input variables used in the template (pulled straight from the underlying Antlr TokenStream) to the parameters provided in an attempt to verify that there are no unbound variables. However, [StringTemplate has some built-in functions](https://github.com/antlr/stringtemplate4/blob/master/doc/cheatsheet.md#functions), and this validation is preventing their use.

**Environment**
Spring AI 1.0.0-M6
Java 17

**Steps to reproduce**
I was attempting to make it so that the prompt added by `VectorStoreChatMemoryAdvisor` would be added conditionally on whether there was actually data pulled from the chat history. Something like:
```
   String chatMemoryPrompt = """
                {if(strlen(long_term_memory))}

                Use the long term conversation memory from the LONG_TERM_MEMORY section to provide accurate answers.

                ---------------------
                LONG_TERM_MEMORY:
                {long_term_memory}
                ---------------------
                {endif}
                """;

    VectorStoreChatMemoryAdvisor.builder(chatHistoryStore)
                    .systemTextAdvise(chatMemoryPrompt)
                    .build();
```
But the validation considers `strlen` to be a variable that is unbound, and fails.

**Expected behavior**
I think the same conclusion from this can be reached as https://github.com/spring-projects/spring-ai/issues/1428 - make the validation optional.
But if we wanted to continue the validation "whack-a-mole", there exists a `org.stringtemplate.v4.compiler.Compiler#funcs` static map whose keys could be used to find the names of these built-in functions, and omit these in validation.
https://github.com/antlr/stringtemplate4/blob/master/src/org/stringtemplate/v4/compiler/Compiler.java#L81

sunyuhan1998:
I submitted a PR to fix this issue: https://github.com/spring-projects/spring-ai/pull/2458

markpollack:
Hi.  @sunyuhan1998 and thoughts on having the validation simply disabled by default - or perhaps just removed all together?

sunyuhan1998:
> Hi. [@sunyuhan1998](https://github.com/sunyuhan1998) thanks so much. I'm thinking to disable validation by default, I really haven't experienced cases where someone didn't replace the template values and was happy to have eager failure. Thoughts?

[...]

PR #2458, opened 2025-03-13
Fixed the issue where using built-in functions in PromptTemplate caused validation failures, and added a method:skipValidate() to skip validation for PromptTemplate.

sunyuhan1998:
As mentioned in https://github.com/spring-projects/spring-ai/issues/2456, when using `PromptTemplate`, if the template contains certain built-in functions, it fails validation during `render()`. I made the following two fixes and optimizations:

1. By comparing the function names in `org.stringtemplate.v4.compiler.Compiler#funcs`, I skip the validation of these built-in functions.
2. Added a `skipValidate()` method to allow users to actively skip this validation.

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java
@@ -29,6 +29,7 @@ import java.util.Set;
 import org.antlr.runtime.Token;
 import org.antlr.runtime.TokenStream;
 import org.stringtemplate.v4.ST;
+import org.stringtemplate.v4.compiler.Compiler;
 import org.stringtemplate.v4.compiler.STLexer;
 
 import org.springframework.ai.chat.messages.Message;
@@ -46,6 +47,8 @@ public class PromptTemplate implements PromptTemplateActions, PromptTemplateMess
 	private ST st;
 
 	private Map<String, Object> dynamicModel = new HashMap<>();
+
+	private boolean skipRenderValidate = false;
 
 	public PromptTemplate(Resource resource) {
 		try (InputStream inputStream = resource.getInputStream()) {
@@ -109,6 +112,10 @@ public class PromptTemplate implements PromptTemplateActions, PromptTemplateMess
 	public void add(String name, Object value) {
 		this.st.add(name, value);
 		this.dynamicModel.put(name, value);
+	}
+
+	public void skipValidate() {
+		this.skipRenderValidate = true;
 	}
 
 	public String getTemplate() {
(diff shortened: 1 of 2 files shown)


spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java
```diff
@@ -111,6 +114,10 @@ public void add(String name, Object value) {
 		this.dynamicModel.put(name, value);
 	}
 
+	public void skipValidate() {
```
markpollack, line 117: should prob be in the constructor.

sunyuhan1998: You're right, it makes more sense to specify it in the constructor. I've already submitted a new commit with the necessary changes. Please review it when you have a moment.

sunyuhan1998: I've added another commit, which disables validation during render() by default.

ThomasVitale:
@sunyuhan1998  Thanks for your contribution! https://github.com/spring-projects/spring-ai/pull/2780 introduced a new `TemplateRenderer` API to decouple `PromptTemplate` from the specific template engine used. The ST logic has been moved to `StTemplateRenderer` which now allows configuring the validation behaviour (`throw`, `warn`, `none`). This new structure also allows providing a fully custom `TemplateRenderer` to use ST in a different way than the default implementation and get full flexibility.

Would that address the case where built-in functions are used and the validation needs to be skipped?

sunyuhan1998:
> @sunyuhan1998 Thanks for your contribution! #2780 introduced a new `TemplateRenderer` API to decouple `PromptTemplate` from the specific template engine used. The ST logic has been moved to `StTemplateRenderer` which now allows configuring the validation behaviour (`throw`, `warn`, `none`). This new structure also allows providing a fully custom `TemplateRenderer` to use ST in a different way than the default implementation and get full flexibility.
> 
> Would that address the case where built-in functions are used and the validation needs to be skipped?

Hi, @ThomasVitale 
    I've reviewed this PR, and I think decoupling the specific template engine implementation from PromptTemplate is a wise choice. The StTemplateRenderer is also more user-friendly now that it supports different validation behaviors. 

    However, regarding the built-in functions, do you think we should still skip their validation? Consider this scenario: I'm using the new StTemplateRenderer, and I want an exception to be thrown during the render() call if I forget to replace certain variables in the prompt. So, I set validationMode to THROW. But since I'm using built-in functions, even though I've replaced all the variables, it seems I would still receive an IllegalStateException?

sunyuhan1998:
> @sunyuhan1998 Thanks for your contribution! #2780 introduced a new `TemplateRenderer` API to decouple `PromptTemplate` from the specific template engine used. The ST logic has been moved to `StTemplateRenderer` which now allows configuring the validation behaviour (`throw`, `warn`, `none`). This new structure also allows providing a fully custom `TemplateRenderer` to use ST in a different way than the default implementation and get full flexibility.
> 
> Would that address the case where built-in functions are used and the validation needs to be skipped?

Hi, @ThomasVitale! 
I've reviewed this PR, and I think decoupling the specific template engine implementation from PromptTemplate is a wise choice. The StTemplateRenderer is also more user-friendly now that it supports different validation behaviors. However, regarding the built-in functions, do you think we should still skip their validation? Consider this scenario: I'm using the new StTemplateRenderer, and I want an exception to be thrown during the render() call if I forget to replace certain variables in the prompt. So, I set validationMode to THROW. But since I'm using built-in functions, even though I've replaced all the variables, it seems I would still receive an IllegalStateException?

markpollack:
This change is still good, but needs to be updated to reflect the changes to switching to a TemplateRenderer interface.  The StringTemplate implementation is in `StTemplateRenderer` which does allow to disable validation completely, but the work you did here @sunyuhan1998 is still a nice addition for when validation is enabled and it skips built in functions.  Would you be able to submit a new PR with the same functionality but on the latest code base please?

markpollack:
forgot to close.  Thanks @sunyuhan1998 additional changes in https://github.com/spring-projects/spring-ai/commit/0e15197298c0848b78a746f3d740191e6a6aee7a

PR #2463, opened 2025-03-13
Add PgIdType based schema generation for PgVectorStore

CChuYong:
This pull request includes id-type based schema generation support in PgVectorStore.

- Added id type based schema generation
  - currently schema initialization only supports UUID Type.
  - even if I specify a type other than uuid, the type of the table's id column becomes uuid.
  - so i added type based ddl execution when `initializeSchema` properties set to `true`
  - also only create uuid extension when id type set to UUID.
- Added test for each types.


ilayaperumalg:
@CChuYong Thanks for your PR adding this feature. Rebased and merged as c02430ea6

Issue #2485, opened 2025-03-16
Support Gemini OpenAI compatibility Embeddings


Lunatix01:
Gemini Supports OpenAI compatibility for both Chat models and Embeddings. 

For the Chat model it works perfectly fine with `SpringAI`, but not for the embeddings.

So, using Gemini embeddings and [OpenAI embedding](https://docs.spring.io/spring-ai/reference/api/embeddings/openai-embeddings.html) (Spring Ai) will cause a `NullPointerException`.

Here is the reason why it causes a `NullPointerException`:

So in `OpenAiEmbeddingModel.class` we have this piece of code:
```java
private DefaultUsage getDefaultUsage(OpenAiApi.Usage usage) {
	return new DefaultUsage(usage.promptTokens(), usage.completionTokens(), usage.totalTokens(), usage);
}
```
As far as I have checked, Gemini embedding responses doesn't provide usage stats, so calling 
`usage.promptTokens(), usage.completionTokens(), usage.totalTokens()` each will create an `NPE` exception. 

So it would be nice if we could have such an implementation to fix this issue, to be honest, it's a feature, not an issue.

Currently I mock the values, since I don't use the stats, here is an example code:

https://github.com/Lunatix01/ragscan/pull/5/files#diff-2e6f569f1db170073e027f80b1ee4c8623be4b953cc54553ae2fe53dfc1e5ec2R117-R124

dev-jonghoonpark:
related doc : https://ai.google.dev/gemini-api/docs/openai

ilayaperumalg:
@dev-jonghoonpark Thanks for the fix. Merged as [77ee57c](https://github.com/spring-projects/spring-ai/commit/77ee57cec0aabdb4a1e26542ea346b7236ee042d)

PR #2488, opened 2025-03-17
allow null usage for OpenAI embedding model

dev-jonghoonpark:
The issue described in https://github.com/spring-projects/spring-ai/issues/2485, where a NullPointerException (NPE) occurs when the `usage` field is null in OpenAI-compatible models, has been fixed.


ilayaperumalg:
@dev-jonghoonpark Thanks for fixing this.

PR #2489, opened 2025-03-17
fix: Fix validate prefix and toolName logic in prefixedToolName method

GOODBOY008:
fix: Fix validate prefix and toolName logic in prefixedToolName method


ilayaperumalg:
@GOODBOY008 Thanks for fixing the issue.

PR #2500, opened 2025-03-17
extract onFinishReason() method

ghdcksgml1:
Hello,

I would like to propose a refactoring. 
While working on `CustomAdvisor`, I found that the `onFinishReason()` method could be extracted into a utility class.

This method is particularly useful when implementing `StreamAroundAdvisor`, as it helps determine whether the stream has finished. 
Copying and using this method every time we implement `CustomAdvisor` seems quite cumbersome. 
How about extracting it into a utility class?

Thank you.


ghdcksgml1:
sorry. I applied spring javaformat.

ilayaperumalg:
@ghdcksgml1 No problem, thanks for the PR! Rebased, squashed and merged as e4357ba2c

PR #2507, opened 2025-03-18
fix: Update import and remove qodana configuration

GOODBOY008:
fix: Update import and remove qodana configuration
- Update ToolCallback import to use relative package path
- Remove qodana.yaml file from the project


ilayaperumalg:
@GOODBOY008 Thanks for the code cleanup. Rebased and merged as 6b25b6277

PR #2550, opened 2025-03-21
Update vectordbs doc,batching-strategy is remove

xwh1108:
Update vectordbs doc,batching-strategy is remove


ilayaperumalg:
@xwh1108 Thanks for the PR fixing the documentation for the vector stores as the batch strategy is no longer property driven.

PR #2551, opened 2025-03-21
Add Missing Audio Formats (WAV, PCM)

gad2103:
Current code does not support all the audio formats that OAI returns. My company needs WAV and PCM formats for our use case (anyone using telephony will want this).


ilayaperumalg:
@gad2103 Thanks for the PR adding the missing audio formats into OpenAI Audio API. Fixed the formatting issues, rebased and merged as 64a4f7978

PR #2557, opened 2025-03-23
Fix: Added missing return statement in Couchbase Cluster bean example

PSriVarshan:
This PR fixes an issue in the documentation where the Couchbase Cluster bean method was missing a return statement, leading to a compilation error. The fix ensures proper Java syntax and prevents confusion.


ilayaperumalg:
@PSriVarshan Thanks for the PR!

PR #2570, opened 2025-03-25
Update/remove autoconfiguration enable/disable configurations

ilayaperumalg:
- Since the model autoconfiguration enable/disable flags are no longer used, remove them

   - Currently, the model autoconfigurations can be enabled/disabled via top level Spring AI properties such as spring.ai.model.chat/embedding/image/moderation=<model-provider-name>

   - Update documentation to add note section about this change

- Update note on the autoconfiguration section to point to the configuration changes

- Align the vertex ai text/multimodal keys in line with the other properties


tzolov:
Thank you @ilayaperumalg 
Rebased and merged at 9eacd40aa97b56a3a958f69656f2fd0aae96a4ac

Issue #2366, opened 2025-03-03
New JSON schema generation includes ToolContext, while old implementation skips it.


DamonBao:
The old implementation (MethodInvokingFunctionCallback) skips ToolContext when generating JSON schemas for method inputs, but the new implementation (JsonSchemaGenerator) does not.

Old implementation:
```
protected String generateJsonSchema(Map<String, Class<?>> namedClasses) {
    try {
        JsonSchemaGenerator schemaGen = new JsonSchemaGenerator(this.mapper);
        ObjectNode rootNode = this.mapper.createObjectNode();
        rootNode.put("$schema", "https://json-schema.org/draft/2020-12/schema");
        rootNode.put("type", "object");
        ObjectNode propertiesNode = rootNode.putObject("properties");
        for (Map.Entry<String, Class<?>> entry : namedClasses.entrySet()) {
            String className = entry.getKey();
            Class<?> clazz = entry.getValue();
            if (ClassUtils.isAssignable(clazz, ToolContext.class)) {
                this.isToolContextMethod = true;
                continue;
            }
            JsonSchema schema = schemaGen.generateSchema(/clazz);
            JsonNode schemaNode = this.mapper.valueToTree(schema);
            propertiesNode.set(className, schemaNode);
        }
        return this.mapper.writerWithDefaultPrettyPrinter().writeValueAsString(rootNode);
    } catch (Exception e) {
        throw new RuntimeException(e);
    }
}
```
New implementation:
```
public static String generateForMethodInput(Method method, SchemaOption... schemaOptions) {
    ObjectNode schema = JsonParser.getObjectMapper().createObjectNode();
    schema.put("$schema", SchemaVersion.DRAFT_2020_12.getIdentifier());
    schema.put("type", "object");
    ObjectNode properties = schema.putObject("properties");
    List<String> required = new ArrayList<>();
    for (int i = 0; i < method.getParameterCount(); i++) {
        String parameterName = method.getParameters()[i].getName();
        Type parameterType = method.getGenericParameterTypes()[i];
        if (isMethodParameterRequired(method, i)) {
            required.add(parameterName);
        }
        ObjectNode parameterNode = SUBTYPE_SCHEMA_GENERATOR.generateSchema(parameterType);
        String parameterDescription = getMethodParameterDescription(method, i);
        if (StringUtils.hasText(parameterDescription)) {
            parameterNode.put("description", parameterDescription);
        }
        properties.set(parameterName, parameterNode);
    }
[...]

PR #2580, opened 2025-03-27
ToolContext is now excluded from JSON Schema Generation

ThomasVitale:
When generating the JSON Schema for a tool input from a method, ToolContext is now excluded since it's not something we want the model to provide. The framework takes care of passing a value for it when actually executing the tool call.

Fixes gh-2366


tzolov (approved):
LGTM

PR #2585, opened 2025-03-28
feat(tool): Add ToolExecutionEligibilityChecker interface

tzolov:
Introduce a new ToolExecutionEligibilityChecker interface to provide a more flexible way to determine when tool execution should be performed based on model responses. This abstraction replaces the hardcoded logic previously scattered across the codebase.

- Adds a new ToolExecutionEligibilityChecker interface in spring-ai-core
- Integrates the checker into OpenAiChatModel with appropriate defaults
- Updates OpenAiChatAutoConfiguration to support the new interface
- Provides a default implementation that maintains backward compatibility

### DIFF
--- a/spring-ai-core/src/main/java/org/springframework/ai/model/tool/ToolExecutionEligibilityChecker.java
+++ b/spring-ai-core/src/main/java/org/springframework/ai/model/tool/ToolExecutionEligibilityChecker.java
@@ -1,0 +1,80 @@
+/*
+* Copyright 2025 - 2025 the original author or authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* https://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package org.springframework.ai.model.tool;
+
+import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


spring-ai-core/src/main/java/org/springframework/ai/model/tool/ToolExecutionEligibilityChecker.java
```diff
@@ -0,0 +1,80 @@
...
+public interface ToolExecutionEligibilityChecker extends Function<ChatResponse, Boolean> {
+
+	/**
+	 * Determines if tool execution should be performed based on the prompt options and
+	 * chat response.
+	 * @param promptOptions The options from the prompt
+	 * @param chatResponse The response from the chat model
+	 * @return true if tool execution should be performed, false otherwise
+	 */
+	default boolean isToolExecutionRequired(ChatOptions promptOptions, ChatResponse chatResponse) {
```
ThomasVitale, line 40: I would suggest keeping only this method in the interface and make it a `Predicate`, returning just true/false.

The other two methods are already implemented elsewhere and the implementation should never change. And if customisations are needed, this method here would allow that, without having to change the other two.

@tzolov what do you think?

ThomasVitale: Following that train of thought, I would name this interface `ToolExecutionEligibilityPredicate`. The default implementation can simply consist of the following:

```java
return ToolCallingChatOptions.isInternalToolExecutionEnabled(prompt.getOptions()) && response != null
				&& response.hasToolCalls()
```

tzolov:
Thanks for the suggestions @ThomasVitale . I've adjusted the PR. 
Will add support for the other (supported) models as well

ilayaperumalg:
Rebased and merged as 0d748e2b8

Issue #2423, opened 2025-03-10
`toolName cannot be null or empty` in  Flux


fxz26284407:
**Bug description**
I found an issue when making MCP calls, where an error occurs when the requesting controller returns Flux<> in reactive mode
```java
java.lang.IllegalArgumentException: toolName cannot be null or empty
	at org.springframework.util.Assert.hasText(Assert.java:253) ~[spring-core-6.2.3.jar:6.2.3]
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
```

**Environment**
java: oracle openJDK17
Spring AI: 1.0.0-M6
LM Studio: 0.3.12
Model: Qwen2.5

**Steps to reproduce**
```shell
cd spring-mcp-test
git checkout EmptyToolName
mvn clean package
java -jar mcp-server/target/mcp-server.jar
java -jar mcp-client/target/mcp-client.jar
```
```shell
curl http://localhost:8080/ai/mcp?message=Get%scurrent%sDate
# or message=List the files in the UV directory of drive D
```

**Expected behavior**
Some information about files.

**Minimal Complete Reproducible example**
Check out the project [https://github.com/fxz26284407/spring-mcp-test](https://github.com/fxz26284407/spring-mcp-test)

dev-jonghoonpark:
I tried running the example file you posted and it returns a message like this

```
What specific help do you need regarding listing files? Please provide more details.%
```

No errors occurred

I tested it with the `gpt-4o-mini` model of Open AI

fxz26284407:
> I tried running the example file you posted and it returns a message like this
> 
> ```
> What specific help do you need regarding listing files? Please provide more details.%
> ```
> 
> No errors occurred
> 
> I tested it with the `gpt-4o-mini` model of Open AI

Sorry, the message parameter of the URL was written incorrectly. It should be like this
```shell
curl http://localhost:8080/ai/mcp?message=Get%scurrent%sDate
```

tzolov:
@fxz26284407 I've checked out your project, updated it to latest spring-ai 1.0.0-SNAPSHOT
within the EmptyToolName branch I started the server and the client apps (using gpt-4o)

Then thit the http://localhost:8080/ai/mcp?message=Get%scurrent%sDate and got
`What specific directory would you like to list the files from?`

No errors

fxz26284407:
[...]

Issue #2417, opened 2025-03-09
stream toolInput cannot be null or empty


WilliamTan778:
**Bug description**
I have a tool function that retrieves dates. The interface runs in call mode and returns all correct results, but when I execute it in stream mode, it always reports an error

java.lang.IllegalArgumentException: toolInput cannot be null or empty
	at org.springframework.util.Assert.hasText(Assert.java:240)
	Suppressed: The stacktrace has been enhanced by Reactor, refer to additional information below: 
Assembly trace from producer [reactor.core.publisher.FluxFlatMap] :
	reactor.core.publisher.Flux.flatMap(Flux.java:5471)
	org.springframework.ai.openai.OpenAiChatModel.lambda$internalStream$13(OpenAiChatModel.java:439)
Error has been observed at the following site(s):
	*__________Flux.flatMap ⇢ at org.springframework.ai.openai.OpenAiChatModel.lambda$internalStream$13(OpenAiChatModel.java:439)
	|_       Flux.doOnError ⇢ at org.springframework.ai.openai.OpenAiChatModel.lambda$internalStream$13(OpenAiChatModel.java:458)
	|_       Flux.doFinally ⇢ at org.springframework.ai.openai.OpenAiChatModel.lambda$internalStream$13(OpenAiChatModel.java:459)
	|_    Flux.contextWrite ⇢ at org.springframework.ai.openai.OpenAiChatModel.lambda$internalStream$13(OpenAiChatModel.java:460)
	|_   Flux.doOnSubscribe ⇢ at org.springframework.ai.chat.model.MessageAggregator.aggregate(MessageAggregator.java:94)
	|_        Flux.doOnNext ⇢ at org.springframework.ai.chat.model.MessageAggregator.aggregate(MessageAggregator.java:105)
	|_    Flux.doOnComplete ⇢ at org.springframework.ai.chat.model.MessageAggregator.aggregate(MessageAggregator.java:144)
	|_       Flux.doOnError ⇢ at org.springframework.ai.chat.model.MessageAggregator.aggregate(MessageAggregator.java:171)
	*__Flux.deferContextual ⇢ at org.springframework.ai.openai.OpenAiChatModel.internalStream(OpenAiChatModel.java:345)
	|_             Flux.map ⇢ at org.springframework.ai.chat.client.DefaultChatClient$DefaultChatClientRequestSpec$2.aroundStream(DefaultChatClient.java:699)
[...]

PR #2589, opened 2025-03-29
fix(openai): prevent incorrect ToolCall merging caused by empty id and name strings

yangtuooc:
Empty tool call id and name strings causing improper ToolCall merging during OpenAI stream
processing, resulting in incorrect function calling responses. This fix ensures both null
and empty strings are properly validated before merging operations.

Fixes https://github.com/spring-projects/spring-ai/issues/2417 (https://github.com/spring-projects/spring-ai/issues/2417)
Fixes https://github.com/spring-projects/spring-ai/issues/2423 (https://github.com/spring-projects/spring-ai/issues/2423)


DannySortino:
I shaded this change in to test since I have similar issue. We should also look to fix for toolname being blank - line 147 specifically. When i did same has text check, it made the merge work even when name was not present. Possibly worth adding that into this fix as well.

yangtuooc:
> I shaded this change in to test since I have similar issue. We should also look to fix for toolname being blank - line 147 specifically. When i did same has text check, it made the merge work even when name was not present. Possibly worth adding that into this fix as well.

Thanks for your feedback! I'll update the PR. May I ask in what scenario you encountered this issue?

PR #2619, opened 2025-04-02
docs: send traces to OpenTelemetry backends

jannikmaierhoefer:
This PR adds a section that shows how to send Spring AI traces to an OpenTelemetry backend such as Langfuse.

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/observability/index.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/observability/index.adoc
@@ -315,3 +315,138 @@ whereas data is exported as span attributes if you use an OpenZipkin tracing bac
 |===
 
 WARNING: If you enable the inclusion of the vector search response data in the observations, there's a risk of exposing sensitive or private information. Please, be careful!
+
+== Example: Sending traces to an OpenTelemetry Backend
+
+This guide shows how to send traces to https://langfuse.com/[Langfuse]'s OpenTelemetry endpoint.
+
+[NOTE]
+====
+Visit the https://github.com/langfuse/langfuse-examples/tree/main/applications/spring-ai-demo[Langfuse Example Repo] for a fully instrumented example application.
+====
+
+=== Step 1: Enable OpenTelemetry in Spring AI
+
+*Add OpenTelemetry and Spring Observability Dependencies* (Maven):
+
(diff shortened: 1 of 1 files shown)


dev-jonghoonpark:
🙌 Please fix the DCO check by signing your commit. `git commit --amend --signoff`

jannikmaierhoefer:
Thanks @dev-jonghoonpark :) Just signed off the commit!

jannikmaierhoefer:
Hi @dev-jonghoonpark what do you think about this addition? Let me know if there is anything I can add/edit :)

dev-jonghoonpark:
Sorry for the late reply. @jannikmaierhoefer 

I think it's a good guide.

It inspired me with a couple of ideas that might help in the future:
- Creating a separate document as a sub-item for better organization.
- Developing a module for easier setup.

But unfortunately, I don't have the authority to approve PRs.
I am just one of the contributors and not a member.😭

jannikmaierhoefer:
Thanks @dev-jonghoonpark! @tzolov what do you think about this addition?

ThomasVitale:
Thanks so much for your contribution! I think it's great to show the integration with Langfuse! I would suggest perhaps to create a dedicated page under the Observability section to showcase the different integrations (starting with a Langfuse page). What do you think?

About the example, I'm afraid the combination of dependencies in the docs is not supported in Spring Boot. It's combining the usage of the OpenTelemetry support in Spring Boot/Micrometer with the OpenTelemetry Spring Boot Starter from the OpenTelemetry Java Instrumentation. They are two mutually exclusive approaches, and I'm not sure about having them both in the same application, since they work in a very different way. Since the Spring AI instrumentation is provided via Micrometer, would it be possible to describe the scenario with that approach (without the OpenTelemetry Spring Boot Starter)?

Besides the project specific instrumentation, Spring AI fully relies on the observability backbone in Spring Boot. Maybe we should point to the Spring Boot documentation about exporting OpenTelemetry traces from a Spring Boot app? There might be differences between Spring Boot versions, so having the docs there might help identifying the right approach based on the used Spring Boot version. Perhaps @jonatan-ivanov has some suggestions.

jonatan-ivanov:
My two cents: I don't think we should do this for every single backend that the Observation API supports or can support. The PR feels like advertising to me. As @ThomasVitale mentioned above, this is not the recommended approach (I have doubts that it even works).

I think the Spring AI docs could be enhanced by linking to the Boot and Micrometer docs but I don't think it should duplicate them and list every single backend that Boot/Micrometer supports.

PR #2634, opened 2025-04-03
apply builder pattern to OllamaApi

dev-jonghoonpark:
apply Builder pattern to `OllamaApi` to provide a consistent development design

related: https://github.com/spring-projects/spring-ai/pull/2164, https://github.com/spring-projects/spring-ai/pull/2817

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
```diff
@@ -51,58 +50,74 @@
...
 
-		this.responseErrorHandler = new OllamaResponseErrorHandler();
+	/**
+	 * Create a new OllamaApi instance
+	 * @param baseUrl The base url of the Ollama server.
+	 * @param restClientBuilder The {@link RestClient.Builder} to use.
+     * @param webClientBuilder The {@link WebClient.Builder} to use.
+	 * @param responseErrorHandler Response error handler.
+	 */
+	public OllamaApi(String baseUrl, RestClient.Builder restClientBuilder, WebClient.Builder webClientBuilder, ResponseErrorHandler responseErrorHandler) {
```
ThomasVitale, line 105: Consider making this constructor private to ensure the Builder is the only option for constructing an OllamaApi instance.

ThomasVitale:
@dev-jonghoonpark thanks for your contribution!

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
dev-jonghoonpark: @ThomasVitale Hello, thank you for your feedback.

Considering backward compatibility, would it be okay to directly change it to private?
If there's no issue, I think making it private would be fine.

As mentioned in the description of this PR, it references markpollack's PR, #2164.

It seems that `OpenAiApi.java` still has a public constructor.
https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java#L99

ThomasVitale: I might have misread the change. I thought this constructor was newly added in this PR. If it was already there, then I agree we should consider backward compatibility. In that case, I would deprecate it with a note to use the Builder (and in the next release we can make it private).

dev-jonghoonpark: Oh, you are right.
I mistakenly thought it was an existing constructor since I hadn't looked at it in a while.

I will update it to private.

dev-jonghoonpark:
conflict resolved.

ilayaperumalg:
@dev-jonghoonpark Thanks for adding the builder pattern to OllamaApi. Rebased and merged as d48f4f63d

Issue #2618, opened 2025-04-02
Parameters not saved in ChatMemory when using PromptUserSpec


eugeniace:
When trying to implement ChatMemory with PromptUserSpec messages, only the template is saved in history, without the actual parameters being replaced.

Example:

```
    private ChatClient.ChatClientRequestSpec buildChatClientRequestSpecRAGAdvisor(Question question, String conversationId) {
        return chatClient.prompt()
                .user(userSpec -> userSpec
                        .text("{question} (error: '{error}')?")
                        .param("question", question.question())
                        .param("error", question.error()!=null?question.error():""))
                .advisors( advisorSpec -> advisorSpec.param(AbstractChatMemoryAdvisor.CHAT_MEMORY_CONVERSATION_ID_KEY, conversationId))            ;
    }
```
In the history a user message like this is saved: "{question} (error: {error})", without the actual values being used.
I have tested with all types of ChatMemory. The problems comes from the advisor code (eg. PromptChatMemoryAdvisor), method before:

```
this.getChatMemoryStore().add(this.doGetConversationId(request.adviseContext()), userMessage);
```

ThomasVitale:
Thanks for raising this issue. The problem should be solved now for all Advisors since ChatClient renders the input prompt before executing the advisor chain. The change was included in the 1.0.0-M8 release. More information in the upgrade notes: https://docs.spring.io/spring-ai/reference/upgrade-notes.html#_chat_client Does that solve your issue?

PR #2636, opened 2025-04-03
Use actual parameter values in userText when saving to chat memory (#…

eugeniace:
Use actual parameter values in userText when saving to chat memory (#2618)


ThomasVitale:
@eugeniace thanks for your contribution. This problem should be solved now for all Advisors since ChatClient renders the input prompt before executing the advisor chain. The change is currently in the snapshot release and will be included in the upcoming 1.0.0-M8. More information in the upgrade notes: https://docs.spring.io/spring-ai/reference/upgrade-notes.html#_chat_client Does that solve your issue?

PR #2638, opened 2025-04-04
Add utility class for dynamic MCP stdio connections

baekgol:
## Overview

This PR adds a new utility class `McpClientStdioUtils` that enables dynamic creation and management of MCP client connections at runtime using stdio transport. This functionality allows applications to add new MCP connections without relying solely on configuration properties.

## Changes

- Added `McpClientStdioUtils` class with:
  - Support for both synchronous and asynchronous MCP client creation
  - Consistent configuration using common properties
  - Runtime connection management methods
  - Comprehensive Javadoc documentation

## Use Cases

- Dynamic management of MCP connections without application restart
- Programmatic addition of connections from user input or configuration
- Runtime integration with newly discovered MCP servers


markpollack:
@baekgol There is just one class, with all private static methods.  It does not seem to address the functionality mentioned in the description.  Can you please elaborate?

tzolov:
Thank you @baekgol , 
It is trivial to create [transport for any stdio, sse and/or streamable-http](https://modelcontextprotocol.io/sdk/java/mcp-client#client-transport) and they use it to create McpClient (e.g. `McpClient.sync(transport)` or `McpClient.async(transport)`. I don't see the reason of adding additional utils for it.

PR #2639, opened 2025-04-04
feat: Add three request body parameters to Mistral AI Chat Completion 

jee14:
Add `presence_penalty`, `frequency_penalty`, and `n` parameters
Following Mistral AI API specifications as referenced in https://docs.mistral.ai/api/#tag/chat


xchopin:
Exactly what I was looking for. I was about to code it and make a PR, thank you

PR #2643, opened 2025-04-05
Update deepseek-chat.adoc

WRCoding:
With the release of the latest version of DeepSeek, DeepSee-V3 -0324, the problem of previous Function Calling has been solved. DeepSeek official website has also deleted the previous Function Calling precautions. Therefore, this document deletes the Note to avoid misleading users.

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/deepseek-chat.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/deepseek-chat.adoc
@@ -6,8 +6,6 @@ Spring AI integrates with DeepSeek AI by reusing the existing xref::api/chat/ope
 
 image::spring-ai-deepseek-integration.jpg[w=800,align="center"]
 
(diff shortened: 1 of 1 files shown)


apappascs:
Thanks for the PR, @WRCoding!
I just ran the tests, and it looks like the `deepseek-reasoner` model still doesn’t support function calling—only `deepseek-chat` does.

```
org.springframework.ai.retry.NonTransientAiException: 400 - {"error":{"message":"deepseek-reasoner does not support Function Calling","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
```

apappascs:
The same applies to non-blocking calls.

```
org.springframework.web.reactive.function.client.WebClientResponseException$BadRequest: 400 Bad Request from POST https://api.deepseek.com/v1/chat/completions

	at org.springframework.web.reactive.function.client.WebClientResponseException.create(WebClientResponseException.java:321)
	Suppressed: The stacktrace has been enhanced by Reactor, refer to additional information below: 
Error has been observed at the following site(s):
	*__checkpoint ⇢ 400 BAD_REQUEST from POST https://api.deepseek.com/v1/chat/completions [DefaultWebClient]

```

PR #2670, opened 2025-04-08
[anthropic] fix issue #1370 with tool call duplication

mmazurkevich:
Without this fix during the stream event handling when `EventType.MESSAGE_STOP` occurs, the latest content block was resent again and it caused to the additional tool call(if it was the latest event)
Also I've described the root cause in [the comment to the issue](https://github.com/spring-projects/spring-ai/issues/1370#issuecomment-2783492564)


ilayaperumalg:
@mmazurkevich Thanks for debugging and submitting the PR to fix the issue.

mmazurkevich:
@ilayaperumalg I've updated the formatting and force-pushed the changes because of the failed [GH Action](https://github.com/spring-projects/spring-ai/actions/runs/14333097975)

ilayaperumalg:
@mmazurkevich With this change, some of the stream+toolcalling tests are failing: 

AnthropicChatClientIT#streamFunctionCallTest
AnthropicChatModelIT#streamFunctionCallTest
AnthropicChatModelIT#streamFunctionCallUsageTest

Could you check?

mmazurkevich:
@ilayaperumalg thank you for the ping. The failed test highlighted the problem that was in the code even before the change. Previously, `internalStream` used switchMap to process `ChatCompletionResponses`, which caused the active stream (including potential recursive calls) to be canceled whenever a new response arrived. This led to incomplete processing of streaming tool calls and unexpected behavior when handling `tool_use` events.

The fix has been pushed. Also, I've run all 50 tests locally with my own ANTHROPIC_API_KEY looks like all of them passed

ilayaperumalg:
@mmazurkevich Thank you so much for the quick response,  fixing the issue and verifying all the tests! I will get this merged soon.

ilayaperumalg:
@mmazurkevich Thanks for the PR and following up with fixing the tests. Rebased, updated the AnthropicApiIT to use the latest builder methods and merged as e6fbc28a5

PR #2676, opened 2025-04-09
Use inline literal in documents

quaff:
`https://example.com` is rendered to a link which the text is `example.com` without https scheme, we should use literal here.

See https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#literals-and-source-code


ericbottard:
Applied on main via 54f51271dcd0c766e103efaadf8e78a458d542cb

PR #2706, opened 2025-04-11
Add request body parameters to Mistral AI Chat Completion

jee14:
Relates to: #2639

Add presence_penalty, frequency_penalty, and n parameters Following Mistral AI API specifications as referenced in https://docs.mistral.ai/api/#tag/chat

### DIFF
--- a/models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
+++ b/models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
@@ -103,9 +103,9 @@ public class MistralAiChatOptions implements ToolCallingChatOptions {
 
 	/**
 	 * Number between -2.0 and 2.0. frequency_penalty penalizes the repetition of words
-	 * based on their frequency in the generated text. A higher frequency penalty discourages
-	 * the model from repeating words that have already appeared frequently in the
-	 * output, promoting diversity and reducing repetition.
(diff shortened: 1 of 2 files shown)


models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
```diff
@@ -438,6 +479,21 @@ public Builder stop(List<String> stop) {
...
+			this.options.frequencyPenalty = frequencyPenalty;
+			return this;
+		}
+
+		public Builder presencePenalty(Double presencePenalty) {
+			this.options.presencePenalty = presencePenalty;
+			return this;
+		}
+
+		public Builder N(Integer n) {
```
ThomasVitale, line 492: This method name should probably be lowercases as `n(Integer n)`. I know that in OpenAI we're using the uppercase letter, but that will need to be fixed as well. Thanks!

ThomasVitale:
@jee14 this is great, thanks so much for your contribution!

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
jee14: Fixed the method name to lowercase as suggested. Thanks for the review!

sobychacko:
Merged via https://github.com/spring-projects/spring-ai/commit/b74e30816c30f73df938dacb344052261319618a.

Issue #2714, opened 2025-04-12
Tools MethodToolCallback throwing an IllegalArgumentException


yuchaozhou:
**Bug description**
In the latest version of the Spring AI framework (m7), the `validateToolContextSupport` method has a bug introduced during refactoring. Specifically, the variable handling in the `if` condition is incorrect, causing an `IllegalArgumentException` to be thrown erroneously when `ToolContext` is non-null and its context is non-empty. This behavior worked correctly in version m6, but the refactoring in m7 introduced this issue.

Below is a code comparison:

**m6 version**:
```java
private void validateToolContextSupport(@Nullable ToolContext toolContext) {
    var isToolContextRequired = toolContext != null && !CollectionUtils.isEmpty(toolContext.getContext());
    var isToolContextAcceptedByMethod = Stream.of(toolMethod.getParameterTypes())
        .anyMatch(type -> ClassUtils.isAssignable(type, ToolContext.class));
    if (isToolContextRequired && !isToolContextAcceptedByMethod) {
        throw new IllegalArgumentException("ToolContext is not supported by the method as an argument");
    }
}
```

**m7 version**:
```java
	private void validateToolContextSupport(@Nullable ToolContext toolContext) {
		var isNonEmptyToolContextProvided = toolContext != null && !CollectionUtils.isEmpty(toolContext.getContext());
		var isToolContextAcceptedByMethod = Stream.of(toolMethod.getParameterTypes())
			.anyMatch(type -> ClassUtils.isAssignable(type, ToolContext.class));
		if (isToolContextAcceptedByMethod && !isNonEmptyToolContextProvided) {
			throw new IllegalArgumentException("ToolContext is required by the method as an argument");
		}
	}
```

Although the code appears identical, in m7, the evaluation of `isToolContextAcceptedByMethod` seems to misbehave, likely due to incorrect variable scoping or context handling during refactoring, causing the exception to be thrown unexpectedly.

**Environment**
- **Spring AI version**: spring-ai-M7
- **Java version**: [Java 17]

**Steps to reproduce**
Using the Tools tool will cause this issue

**Expected behavior**
In version m7, the `validateToolContextSupport` method should behave as it did in m6, throwing an `IllegalArgumentException` only when `ToolContext` is non-null, its context is non-empty, and the method’s parameters do not support the `ToolContext` type. Currently, m7 incorrectly throws the exception, leading to functional issues.

**Minimal Complete Reproducible example**
Using Tools

PR #2718, opened 2025-04-12
Fixed incorrect variable usage in MethodToolCallback.java

yuchaozhou:
In version m7, the validateToolContextSupport method should behave as it did in m6, throwing an IllegalArgumentException only when ToolContext is non-null, its context is non-empty, and the method’s parameters do not support the ToolContext type. Currently, m7 incorrectly throws the exception, leading to functional issues.

Relates to: [#2714](https://github.com/spring-projects/spring-ai/issues/2714)


markpollack:
I suspect his is outdated.  @ilayaperumalg ?

sobychacko:
Thanks @yuchaozhou, and for spotting this early. It overlaps with #6557, which fixes the same `validateToolContextSupport` issue in `MethodToolCallback` but goes a bit further: it removes the faulty validation entirely, converts a null `ToolContext` to an empty one in `buildMethodArguments`, and adds test coverage. This one is also fairly old now and no longer applies cleanly to the refactored code, so we are going to go with #6557 and close this.

PR #2723, opened 2025-04-13
Update anthropic-chat.adoc

WRCoding:
In the current version 1.0.0-SNAPSHOT, AnthropicChatModel does not provide a constructor with two parameters. If there are only two parameters, you should use Builder to construct AnthropicChatModel. The current document misleads users and needs to be updated.


WRCoding:
image
 That's the way it's written in the current document.

WRCoding:
image
 Constructor methods provided in AnthropicChatModel

PR #2729, opened 2025-04-14
fix: predicate should be wired in even if toolCallingManager is not null

garethjevans:


tzolov:
Thanks for catching this @garethjevans

PR #2730, opened 2025-04-14
Fix ToolCallingChatOptions isInternalToolExecutionEnabled

ilayaperumalg:
- The ModelOptionsUtils merges the ToolCallingChatOptions with the provider chat options
   - The underlying beanwrapper implementation expects the boolean option to have the getter method starting with "get" instad of "is" which makes the current `isInternalToolExecutionEnabled` not used by the merge operation at all. To fix this, rename the `isInternalToolExecutionEnabled` method to `getInternalToolExecutionEnabled` and deprecate the existing `isInternalToolExecutionEnabled` method.


tzolov:
Thanks @ilayaperumalg !
rebased and merged at 76a1dbe747f9cdc13df13ef17cfb8967bfba8ea8

Issue #2655, opened 2025-04-06
Make ChatClient and Advisor APIs more robust, consistent, and flexible


ThomasVitale:
### Advisor API

The Advisor API went through some changes over time. Initially, there was a concept of "advisor type", with separate interfaces for "before", "after", and "around" advisors. Currently, there is only one type of advisor (around). Still, the naming of many APIs and the observation context include the legacy "around" concept. We should probably get rid of that to avoid confusion.

### AdvisedRequest

The `AdvisedRequest` API contains the same information of a `Prompt`, but in a de-structured format. Furthermore, for the same information, multiple ways exist to store it. This causes lots of troubles in all consumers of an `AdvisedRequest` object (e.g. advisors and observations), in particular:

- A consumer doesn't know if some of the "de-structured" information has already been used to reach a more structured format. For example, there is no way to tell if the `userParams` have already been used to render a version of the `userText`. In that case, the workaround is for each consumer to render the `userText`, even if not needed. Some issues registered about this: 
- A consumer must be aware that prompt messages can be found in multiple places and must be aware of the rules for combining the information into a unified view. For example, the user prompt or system message might be found in one of the dedicated fields (`userText`, `systemText`) or in the more generic `messages`. In that case, the consumer must come up with a strategy to infer the actual user message and the actual system message, since there can be multiple ones. Some issues registered about this:
- A consumer must be aware that tools can be found in multiple places and must be aware of the rules for combining the information into a unified view. For example, tool names might be found in the dedicated field (`toolNames`) to in the more generic `chatOptions`. When both places contain toolNames, how should possible conflicts be handled? Some issues registered about this:
- When both a `systemText` and a SystemMessage in `messages` are provided, it's not clear which one is used in the call to the chat model. Some issues registered about this: 

Furthermore, an `AdvisedRequest` contains a mutable `ChatModel` instance which is very confusing. Consumers of an `AdvisedRequest` have the power to replace the `ChatModel` instance, but that operation would have no effect. Actually, it could lead to wrong information provided downstream. If a subsequent consumer relies on that field to know which model is used by the chain of advisors, it could receive wrong information. It should probably be removed.

[...]

Issue #1687, opened 2024-11-06
StringTemplate should not be the only prompt template mechanism


johnsonr:
StringTemplate should not be the only prompt template mechanism used in Spring AI. Nor should it be the default.

StringTemplate is presently deeply baked into Spring AI. Consider `AdvisedRequest.toPrompt` and `PromptTemplate`. `ChatClient.defaultSystem` will automatically use ST.

The template engine should be pluggable, and a more familiar default such as Mustache should be used, ideally aligning as far as possible with existing Spring support. Jinjava provides solid JVM support for Jinja and it should also be an option because of its pervasive use in the Gen AI space.

Reasons:

- StringTemplate is not particularly popular and is not a common choice in the broader Spring ecosystem. Forcing its use creates unnecessary friction for new adopters, and frustration for existing advanced users.
- Most existing IP around Gen AI comes out of the Python ecosystem. StringTemplate does not have a Python implementation. Thus supporting only StringTemplate makes it impossible to benefit from the many valuable prompts produced by this ecosystem. Not only is reuse impossible; porting even simple templates from Jinja to StringTemplate is harder than to, say, Mustache.
- Pluggability has always been a core Spring value. Spring MVC supported different template engines from its inception. Hardcoding one template engine is not Spring-like.
- StringTemplate syntax is quirky and likely to be off-putting to developers, even if they're familiar with other templating technologies. 

While variable interpolation is fine, more complex things are quirky with StringTemplate. For example, looping is a common requirement in prompt templates. Consider how it looks in various popular template languages:

Mustache (familiar to many developers on all stacks):

```mustache
{{#items}}
    {{.}}
{{/items}}
```

Jinja (the norm in Python and probably the most widely used in Gen AI):
```jinja
{% for item in items %}
    {{ item }}
{% endfor %}
```

Both of these are easy for any developer to grasp, even if they haven't seen that particular syntax.

Now take StringTemplate:
```
list(items) ::= 
<items:{item | <item>}; separator="\n">
>>
```
This looks more like a BNF grammar than a typical template.

ilayaperumalg:
There is some related discussion around this topic [here](https://github.com/spring-projects/spring-ai/issues/355)

quaff:
GH-2616 is related.

markpollack:
This has now been fixed, the PR is https://github.com/spring-projects/spring-ai/pull/2920 and searching for TemplateRenderer in the docs will show you how to use it.

PR #2780, opened 2025-04-17
Introduce TemplateRenderer for prompt templating

ThomasVitale:
- Introduce new TemplateRenderer API providing the logic for rendering an input template.
- Update the PromptTemplate API to accept a TemplateRenderer object at construction time.
- Move ST logic to StTemplateRenderer implementation, used by default in PromptTemplate. Additionally, make start and end delimiter character configurable.

Relates to gh-2655, gh-1687.

Second PR will follow to use TemplateRenderer in ChatClient. See usage examples here: https://github.com/ThomasVitale/spring-ai/pull/1/files#diff-38f2fe47de01515938c5201ff9af94a7fa869b2760c515e2472f83cb42271518R103

Untitled-2025-04-16-0204

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/template/ValidationMode.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/template/ValidationMode.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.template;
+
+/**
+ * Validation modes for template renderers.
+ *
+ * @author Thomas Vitale
+ * @since 1.0.0
+ */
+public enum ValidationMode {
+
+	/**
+	 * If the validation fails, an exception is thrown. This is the default mode.
+	 */
+	THROW,
+
+	/**
+	 * If the validation fails, a warning is logged. The template is rendered with the
+	 * missing placeholders/variables. This mode is not recommended for production use.
+	 */
+	WARN,
+
+	/**
+	 * No validation is performed.
+	 */
+	NONE;
+
+}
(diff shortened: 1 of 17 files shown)


spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java
```diff
@@ -179,59 +190,89 @@ public Prompt create(ChatOptions modelOptions) {
...
-				}
-			}
-			else if (token.getType() == STLexer.RDELIM) {
-				isInsideList = false;
-			}
-			else if (!isInsideList && token.getType() == STLexer.ID) {
-				inputVariables.add(token.getText());
-			}
-		}
+		throw new UnsupportedOperationException(
```
ThomasVitale, line 209: This is the only behavioural breaking change in this PR. What do you think?

This method was used internally for validation and testing (and it's still used that way, but inside StTemplateRenderer). This is very specific not only based on the template engine, but also on how it's configured, so I wouldn't expose this information here. If we need some additional introspection into the internals of the template engine, it should be the TemplateRenderer giving access to that.

spring-ai-model/src/main/java/org/springframework/ai/template/st/StTemplateRenderer.java
```diff
@@ -0,0 +1,189 @@
...
+
+	private static final String VALIDATION_MESSAGE = "Not all variables were replaced in the template. Missing variable names are: %s.";
+
+	private static final char DEFAULT_START_DELIMITER_TOKEN = '{';
+
+	private static final char DEFAULT_END_DELIMITER_TOKEN = '}';
+
+	private static final ValidationMode DEFAULT_VALIDATION_MODE = ValidationMode.THROW;
+
+	private final char startDelimiterToken;
```
ThomasVitale, line 50: Solves gh-355

spring-ai-model/src/main/java/org/springframework/ai/template/TemplateRenderer.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Map;
+import java.util.function.BiFunction;
+
+/**
+ * Renders a template using a given strategy.
+ *
+ * @author Thomas Vitale
+ * @since 1.0.0
+ */
+public interface TemplateRenderer extends BiFunction<String, Map<String, Object>, String> {
```
markpollack, line 28: hard to know at this point, but wondering if this and the impls should go into spring-ai-commons as if there is a spring-ai-client-image in the future, it would also need templating support.

spring-ai-commons/src/main/java/org/springframework/ai/template/ValidationMode.java
```diff
@@ -0,0 +1,43 @@
...
+
+package org.springframework.ai.template;
+
+/**
+ * Validation modes for template renderers.
+ *
+ * @author Thomas Vitale
+ * @since 1.0.0
+ */
+public enum ValidationMode {
```
ThomasVitale, line 25: @markpollack  I would keep this internal to StTemplateRenderer since I'm not sure it's generic enough to be useful across template engines, for now. We can re-evaluate once we have concrete examples of different template engines. What do you think?

spring-ai-model/src/main/java/org/springframework/ai/template/TemplateRenderer.java
markpollack: ok, changes made. tx!

PR #2788, opened 2025-04-17
Make BeanOutputConverter more extensible

### LINKED ISSUE #1985: Support for BeanOutputConverter with non-standard enum mapping
**Expected Behavior**

I'd like to see ``@JsonProperty`` style of enums with ``BeanOutputConverter`` - it'd be great to have an ability to give AI assistant a range of constant values, which aren't just Java String names. That is, I'd love to see something like:

```java
	@PostConstruct
	public void doSomething() {
		var myWrapper = ChatClient.create(chatModel)
				.prompt("Select randomly either \"value.a\" or \"value.b\"")
				.call()
				.entity(MyWrapper.class);
		System.out.println(myWrapper.values);
	}


	public record MyWrapper(EnumValues values) {

	}

	public enum EnumValues {
		@JsonProperty("value.a")
		A,
		@JsonProperty("value.b")
		B
	}
```

being a correct way to use structured outputs

**Current Behavior**

As far as I can see, ``BeanOutputConverter`` is not really that much customizable. All the parametrization is done inside of a private method (``generateSchema``). In theory, one could even use ``@JsonProperty`` - it is recognized by the underling Jackson, but not by the schema generator, leading to interesting problems.

Generated schema for the code above:
```json
{
  "$schema" : "https://json-schema.org/draft/2020-12/schema",
  "type" : "object",
  "properties" : {
    "values" : {
      "type" : "string",
      "enum" : [ "A", "B" ]
    }
  },
  "additionalProperties" : false
}
```

Stack trace:
```
Caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `xyz.SpringAiTest$EnumValues` from String "A": not one of the values accepted for Enum class: [value.a, value.b]
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 2, column: 13] (through reference chain: xyz.SpringAiTest$MyWrapper["values"])
	at com.fasterxml.jackson.databind.exc.InvalidFormatException.from(InvalidFormatException.java:67) ~[jackson-databind-2.18.2.jar:2.18.2]
```

**Context**

I wanted to integrate with a service that uses non-standard constants directly, and to select values using AI assistant. 

There are other ways this can be done, but as far as I understand, they either include:
1. Copying ``BeanOutputConverter`` and modifying its code
2. Implementing your own ``StructuredOutputConverter``. For a similar case, I've tried to modify the generated schema myself to account for ``@JsonProperties`` and it works, but it's very manual and brittle
3. Mapping this enum to a String via some different code. Not the worst thing in the world, but it means that I need to refer in the conversations with the AI to the values using Java enum values, as well, as also having some code to map them.

markpollack:
## Summary
- Makes BeanOutputConverter class more subclass-friendly by changing access modifiers for key fields and methods
- Extracts template string in getFormat() to allow customization by subclasses
- Provides better extension points for customizing behavior

Fixes #1985


ThomasVitale:
@markpollack since `BeanOutputConverter` implements `StructuredOutputConverter`, would it be enough to create a custom `StructuredOutputConverter` implementation if a different logic is desired? 

I can already define a `MyBeanOutputConverter` which implements `StructuredOutputConverter` and define my custom logic for the schema generation.

I'm afraid that by opening up methods and fields in `BeanOutputConverter`, it will be more difficult to evolve the structured output support due to backward compatibility. We had a similar issue in the old `FunctionCallback` implementations, which exposed methods to customise schemas and ObjectMapper, and that made it difficult to change the internal implementation since all the internal methods were protected. In that case, we solved it by introducing a different API (ToolCallback), but here it would be more challenging.

I would tend to suggest evolving `BeanOutputConverter` to be a final class. Customizations would still be possible either via a delegate pattern or by introducing a custom implementation.

What do you think?

xak2000:
It would be great to actually make `JsonSchemaGenerator` customizable.

E.g. currently there is no way to modify the options list of `JacksonModule` that I needed to enable `FLATTENED_ENUMS_FROM_JSONVALUE` and `FLATTENED_ENUMS_FROM_JSONPROPERTY` options.

For these specific 2 options I think they should be enabled by default, but it would be still very good to have ability to customize.

Of course it is always possible to just write and use custom schema generator, but it is quite a lot of error-prone code, so an out-of-the-box configurability would be much appreciated.

PR #2789, opened 2025-04-17
make responseMessage in AzureOpenAiChatModel.buildGeneration null-safe

### LINKED ISSUE #2691: Nullpointer when using Azure Open AI and streaming
**Bug description**
While using the AzureOpenAI chat client streaming won't work. It gets into this nullpointer. When I subscribe and print out the .content() Flux; I noticed that the last received token is `null`.

2025-04-10 17:38:22.611 [http-nio-8080-exec-7] ERROR o.a.c.c.C.[.[.[.[dispatcherServlet].log - Servlet.service() for servlet [dispatcherServlet] threw exception
java.lang.NullPointerException: Cannot invoke "com.azure.ai.openai.models.ChatResponseMessage.getToolCalls()" because "responseMessage" is null
	at org.springframework.ai.azure.openai.AzureOpenAiChatModel.buildGeneration(AzureOpenAiChatModel.java:498)
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Assembly trace from producer [reactor.core.publisher.FluxMapFuseable] :
	reactor.core.publisher.Flux.map(Flux.java:6588)
	org.springframework.ai.azure.openai.AzureOpenAiChatModel.lambda$internalStream$13(AzureOpenAiChatModel.java:381)

**Environment**
Spring-AI 1.0.0-M6
Chat Model: Azure OpenAI

**Steps to reproduce**
Set a custom content-filer on your model: Azure AI Foundry->Safety+Security->Create content filter->Output Filter->Streaming mode (Preview)->Asynchronous Filter
`chatClient.prompt().user("How are you?").stream().content().doOnEach(data -> System.out.println(data.get()));
`

//output
How
 can
 I
 assist
 you
 today
?
null

**Expected behavior**
AzureOpenAiChatModel should be null safe or null values should be filtered out.

See above. When I switch to another vendor like Anthrophic the result is as expected (without a null at the end of the stream)

berjanjonker:
Fixes #2691

### DIFF
--- a/models/spring-ai-azure-openai/src/test/java/org/springframework/ai/azure/openai/AzureOpenAiChatModelIT.java
+++ b/models/spring-ai-azure-openai/src/test/java/org/springframework/ai/azure/openai/AzureOpenAiChatModelIT.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 import com.azure.ai.openai.OpenAIClientBuilder;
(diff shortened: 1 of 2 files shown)


models/spring-ai-azure-openai/src/test/java/org/springframework/ai/azure/openai/AzureOpenAiChatModelIT.java
```diff
@@ -106,6 +107,26 @@ void testMessageHistory() {
 		assertThat(response.getResult().getOutput().getText()).containsAnyOf("Blackbeard");
 	}
 
+	@Test
```
markpollack, line 110: This test passes without the fix to the ChatModel, so I'm not sure what to make of it.  Perhaps this is an intermittent error in what the model returns?

markpollack: See comments in https://github.com/spring-projects/spring-ai/issues/2691#issuecomment-2819279453 as well.

berjanjonker: As explained  in #2691 this PR fixes a NPE when [Asynchronous Filter](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cuser-prompt%2Cpython-new#asynchronous-filter) is enabled for your model. As it fixes the streaming output, I don't know for sure if the content filtering works as expected with this option enabled.

PR #2794, opened 2025-04-18
fix(MessageChatMemory) When adding data to Message Chat Memory Data, the actual data was not replaced

zhangshenghang:
When I use something like `ReReadingAdvisor`, I will use `{re2_input_query}` for parameter substitution.

image

However, when `MessageChatMemoryAdvisor` executes `this.getChatMemoryStore().add`, it does not perform replacement and will directly save `{re2_input_query}` to `ChatMemory`. This is incorrect.

```java
// 4. Add the new user input to the conversation memory.
		UserMessage userMessage = new UserMessage(request.userText(), request.media());
		this.getChatMemoryStore().add(this.doGetConversationId(request.adviseContext()), userMessage);
```

image

We need to replace `{re2_input_query}` with real content. Modified code:
```java
String processedUserText = request.userText();
		if (!CollectionUtils.isEmpty(request.userParams())) {
			processedUserText = new PromptTemplate(processedUserText, request.userParams()).render();
		}
		UserMessage userMessage = new UserMessage(processedUserText, request.media());

		this.getChatMemoryStore().add(this.doGetConversationId(request.adviseContext()), userMessage);
```

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
@@ -19,6 +19,8 @@ package org.springframework.ai.chat.client.advisor;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.springframework.ai.chat.prompt.PromptTemplate;
+import org.springframework.util.CollectionUtils;
 import reactor.core.publisher.Flux;
 
 import org.springframework.ai.chat.client.advisor.api.AdvisedRequest;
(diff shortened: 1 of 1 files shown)


zhangshenghang:
@tzolov @markpollack @ilayaperumalg PTAL.

ThomasVitale:
@zhangshenghang thanks for your contribution! This problem should be solved now for all Advisors since ChatClient renders the input prompt before executing the advisor chain. The change is currently in the snapshot release and will be included in the upcoming 1.0.0-M8. More information in the upgrade notes: https://docs.spring.io/spring-ai/reference/upgrade-notes.html#_chat_client Does that solve your issue?

PR #2802, opened 2025-04-18
chore: Use stream flatmap when build toolcallbacks.

He-Pin:
Motivation:
Use stream flatmap to make code a little clean.


ilayaperumalg:
@He-Pin Thanks for the PR simplifying the code.

PR #2822, opened 2025-04-21
Fix array-to-List bug and clarify synchronous usage in “Async” classes

kimsunghyun1995:
Array to List conversion

Updated methods like getToolCallbacksFromSyncClients to flatten ToolCallback[] properly using Arrays.asList(...) or streams.

Previously, using List.of(array) created a single-element list wrapping the entire array, causing incorrect behavior.

### DIFF
--- a/mcp/common/src/main/java/org/springframework/ai/mcp/McpToolUtils.java
+++ b/mcp/common/src/main/java/org/springframework/ai/mcp/McpToolUtils.java
@@ -16,6 +16,7 @@
 
 package org.springframework.ai.mcp;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


tzolov:
@kimsunghyun1995 I've added [few tests](https://github.com/spring-projects/spring-ai/pull/2946), like: https://github.com/spring-projects/spring-ai/blob/5634d892fa16e7486ac92511bec4fed36c93e83d/mcp/common/src/test/java/org/springframework/ai/mcp/ToolUtilsTests.java#L257

But they seems to show correct behavior from the existing code. Can you please add a test to show the https://github.com/spring-projects/spring-ai/pull/2822#issue-3007678099 issue?

tzolov:
I'm closing this as could not reproduce it and we have not received additional feedback.

Issue #2804, opened 2025-04-18
Add image URL support for Anthropic model


mmazurkevich:
**Description**
`org.springframework.ai.anthropic.api.AnthropicApi.ContentBlock` doesn't support `Source` with image URL. Currently, it's possible to attach images only in base64 format, while API allows you to do so. 
What is currently implemented:
<img height="420"  alt="Image" src="https://github.com/user-attachments/assets/1843a445-5dc9-45a6-be9f-d002a9f1d846" />

Requesting API support:
<img height="420"  alt="Image" src="https://github.com/user-attachments/assets/a83b653f-c1ae-49af-914d-2b52e6dc46e6" />

dev-jonghoonpark:
Can you share the documentation URL you captured?

mmazurkevich:
@dev-jonghoonpark  Sure, https://docs.anthropic.com/en/api/messages#body-messages-content please check the `messages.content.source` field

sobychacko:
The PR has been merged a while back. Closing the issue.

PR #2825, opened 2025-04-21
Make allow image url for Anthropic Message API

dev-jonghoonpark:
related issue: https://github.com/spring-projects/spring-ai/issues/2804

I have improved the Anthropic API to support image URLs.

---

The following areas still need improvement:
- In Spring AI, the `MimeType` is required in the `Media` object, but the Anthropic API does not require the MimeType. Therefore, we need to explicitly provide a MimeType (any value can be used).

example:
```
String response = ChatClient.create(this.chatModel).prompt()
        .options(AnthropicChatOptions.builder().model(modelName).build())
        .user(u -> u.text("Explain what do you see on this picture?").media(MimeTypeUtils.IMAGE_PNG, url))
        .call()
        .content();
```

`MimeTypeUtils.IMAGE_PNG` is not necessary for the Anthropic API, but it must be included to meet the requirements of Spring AI.

I thought that creating a `media(URL url)` method that does not require a `MimeType` could cause confusion with other models that mandate the use of a `MimeType.`

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -88,6 +88,7 @@ import org.springframework.util.StringUtils;
  * @author Thomas Vitale
  * @author Claudio Silva Junior
  * @author Alexandros Pappas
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
@dev-jonghoonpark Thanks for the PR. Rebased and merged as c6ccc1a4a

PR #2826, opened 2025-04-21
Delete non-existing methods

wayhelper:
The boxed() method is not a method in List, but a method in Java's IntStream class.


spring-ai-commons/src/main/java/org/springframework/ai/transformer/splitter/TokenTextSplitter.java
```diff
@@ -142,7 +142,7 @@ protected List<String> doSplit(String text, int chunkSize) {
 
 	private List<Integer> getEncodedTokens(String text) {
 		Assert.notNull(text, "Text must not be null");
-		return this.encoding.encode(text).boxed();
```
sobychacko, line 145: `boxed()` is a method from the library that the `TokenTextSplitter` uses that provides certain guarantees and I think we need to use it. See here for more details: https://github.com/knuddelsgmbh/jtokkit/blob/main/lib/src/main/java/com/knuddels/jtokkit/api/IntArrayList.java#L146

sobychacko:
Please confirm the comments above, and if you agree, we can close this PR.

PR #2827, opened 2025-04-21
chore: fix typo

He-Pin:
Fix typo, it should be `asyncSpec`


ilayaperumalg:
@He-Pin Thanks for fixing the code!

PR #2853, opened 2025-04-23
chore: Add ToolContextCreator.

He-Pin:
Motivation:
Add an extension point for users to extend and create a subclass of  `ToolContext`

### DIFF
--- a/auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
+++ b/auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
@@ -22,6 +22,9 @@ import java.util.List;
 import io.micrometer.observation.ObservationRegistry;
 
 import org.springframework.ai.chat.model.ChatModel;
+import org.springframework.ai.chat.model.ToolContext;
+import org.springframework.ai.chat.model.ToolContextCreator;
+import org.springframework.ai.model.tool.DefaultToolContextCreator;
 import org.springframework.ai.model.tool.ToolCallingManager;
 import org.springframework.ai.tool.ToolCallback;
 import org.springframework.ai.tool.ToolCallbackProvider;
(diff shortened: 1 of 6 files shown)


ThomasVitale:
@He-Pin thanks for your contribution. Could you share some more details about your use case for such a change? Thanks!

markpollack:
interested as well - related to the other PR, this enables an execution strategy for tools.  Also, as this would require some tests cases before merging.

He-Pin:
We are building a mcp proxy server and need some extension point.

I will share more detail once I leave the airport

He-Pin:
We are building an mcp proxy server and need some extension point.

I will share more detail once I leave the airport

ThomasVitale:
@He-Pin thanks, looking forward to that! It sounds a very interesting use case! The `ToolCallingManager` is an interface, so it should be easy to customise to support more use cases. The `DefaultToolCallingManager` is just an example of a default implementation. Before changing the default implementation by introducing new APIs, I would suggest considering whether a separate implementation would work, which should guarantee full flexibility and customisations. That was the intent behind the `ToolCallingManager` API.

If you define your custom `ToolCallingManager` as a bean, Spring AI will pick it up and use it instead of `DefaultToolCallingManager` automatically.

```java
@Bean
ToolCallingManager myToolCallingManager() {
    return new MyMcpToolCallingManager();
}
```

Inside `MyMcpToolCallingManager`, if there's a functionality you don't need to customise, you can also delegate to `DefaultToolCallingManager`.

I'm mentioning this also considering your other proposal: https://github.com/spring-projects/spring-ai/pull/2855.

He-Pin:
yes, thanks for that , but another thing is the current ToolContext is final, and then we can only put our CTX as an special field in it, if we can subclass it , we can separate our ctx from the origin CTx

It would be better that there is an addition generic CtxData ctx parameter in the call method,  but that will break the current binary compatibility. 

So that's what I was thinking,  we can even pass our ctx with thread local, but that's not ideal

markpollack:
@ThomasVitale what was the motivation for the `ToolContext` to be final?

https://github.com/spring-projects/spring-ai/commit/83294023cd4946c46b42df01b686bcc70418ea70

tzolov:
> yes, thanks for that , but another thing is the current ToolContext is final, and then we can only put our CTX as an special field in it, if we can subclass it , we can separate our ctx from the origin CTx

@He-Pin what is the problem to put your CTX as a special field in the existing ToolContext?

severeserpent:
> > yes, thanks for that , but another thing is the current ToolContext is final, and then we can only put our CTX as an special field in it, if we can subclass it , we can separate our ctx from the origin CTx
> 
> @He-Pin what is the problem to put your CTX as a special field in the existing ToolContext?



This is also not possible as it gets overwritten. https://github.com/spring-projects/spring-ai/issues/4344

No? @He-Pin @tzolov ?

PR #2855, opened 2025-04-23
chore: Add around tool calling execution

He-Pin:
Add an extension point for tool calling

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
@@ -203,7 +203,7 @@ public class DefaultToolCallingManager implements ToolCallingManager {
 
 			String toolResult;
 			try {
-				toolResult = toolCallback.call(toolInputArguments, toolContext);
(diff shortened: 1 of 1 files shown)


ThomasVitale:
@He-Pin thanks for your contribution. Could you share some more details about your use case for such a change? Thanks!

markpollack:
is this to implement a client side parallel tool execution feature?

He-Pin:
We need the enhance the function input and result,  and which can be rescheduled to a virtual thread pool tool. Otherwise,  I have to using an AOP

ThomasVitale:
I'd like to share the same input as the other task: https://github.com/spring-projects/spring-ai/pull/2853#issuecomment-2839272262

In addition to that, I should add that `DefaultToolCallingManager` should have been `final`, but it kind of slipped away. It's not designed to be extended, but only as a reference default implementation. So I wouldn't introduce abstract methods. The idea is to provide a custom implementation of the `ToolCallingManager` for customisations and extensions (possibly couple with the delegate pattern).

What do you think?

He-Pin:
I need two things 
1. Enhance the input and output of the call, an aop
2. Passing our ctx data without interfering the current toolcontext

Issue #2863, opened 2025-04-23
OpenAI o1 models support 'developer' role instead of 'system' role.


dolukhanov:
**Bug description**
According to the [OpenAI Guide](https://platform.openai.com/docs/guides/text#message-roles-and-instruction-following) and the [Model Spec](https://model-spec.openai.com/2025-02-12.html#chain_of_command) the former 'system' role' has now been replaced by the 'developer' role.

The o1 range of models reject the system role within a prompt

**Environment**
Java: 21
Spring AI: M7
Spring Boot: 3.4.4

**Steps to reproduce**

Create a Prompt containing a SystemMessage and a UserMessage using an o1 model e.g. o1-mini-2024-09-12

The call fails with:

```
Caused by: org.springframework.ai.retry.NonTransientAiException: 400 - {
  "error": {
    "message": "Unsupported value: 'messages[0].role' does not support 'system' with this model.",
    "type": "invalid_request_error",
    "param": "messages[0].role",
    "code": "unsupported_value"
  }

```

**Expected behavior**
A successful call to Open AI

**Minimal Complete Reproducible example**

```
@Component
public class TestOpenAI
{
    private final static Logger log = LoggerFactory.getLogger( TestOpenAI.class );

    private OpenAiChatModel openAiChatModel;

    public TestOpenAI( OpenAiChatModel openAiChatModel )
    {
        this.openAiChatModel = openAiChatModel;
    }

    @PostConstruct
    public void run()
    {
        log.info( "Running OpenAI" );

        OpenAiChatOptions options = OpenAiChatOptions.builder().model( "o1-mini-2024-09-12" ).build();

        SystemMessage systemMessage = new SystemMessage(
                "You are a helpful assistant knowledgable about AI Platforms" );

        UserMessage userMessage = new UserMessage( "Tell me about OpenAI" );

        List<Message> openAiMessages = List.of( systemMessage, userMessage );
        Prompt prompt = new Prompt( openAiMessages, options );

        ChatResponse call = openAiChatModel.call( prompt );
        Generation generation = call.getResults().get( 0 );

        System.out.println( generation.getOutput().getText() );

    }
}
```

dolukhanov:
[Pull request](https://github.com/spring-projects/spring-ai/pull/2864) added.

markpollack:
The PR adds a new message type, but doesn't solve the ease of use problem in sending this new message type for openai when using chat client.  needs some more thought.

andresssantos:
[...]

PR #2864, opened 2025-04-23
Added a Developer role, as used by OpenAI replacing the system role

dolukhanov:
As per this [issue](https://github.com/spring-projects/spring-ai/issues/2863) - OpenAI has replaced the system role with the developer role.

o1 models now mandate the developer role to be used instead of the system role.


markpollack:
I believe to make this work end to end it will need to be supported in ChatClient no?

markpollack:
how to use this in practice with ChatClient?

markpollack:
Closing to focus discussion on the other PR and that this also will not be in the GA release.  See https://github.com/spring-projects/spring-ai/issues/3103

Issue #1694, opened 2024-11-07
Anthropic module to support setting the base_url and the uri, not only base_url


ddobrin:
Anthropic module does not allow the user to set both:

1. spring.ai.anthropic.base-url (can be set)
2. spring.ai.anthropic.uri (can not be set, name can be changed)
It hard-codes the URI to **/v1/messages** and should be a user-settable value, including an empty String

Similar example: OpenAI module
https://docs.spring.io/spring-ai/reference/api/chat/openai-chat.html#_configuration_properties

Rationale:
If the Anthropic model is deployed in a managed environment, the base URL and URI should be user-settable, allowing also for an empty value for the uri, as the model could be hosted in different managed or self-hosted environments

ilayaperumalg:
Closing this as complete via #2869

PR #2869, opened 2025-04-24
Added the possibility to configure the path in Anthropic client like OpenAI client implementation

Claudio-code:
Issue https://github.com/spring-projects/spring-ai/issues/1694

Added the possibility to configure the path in the anthropic client with its property:

spring.ai.anthropic.completions-path=message-path


ilayaperumalg:
@Claudio-code Thanks for the PR! Rebased and updated the changes to include the recent builder changes and merged as d7eb9bb32

PR #2885, opened 2025-04-25
fix:toolcall return string

Cui-xf:
Signed-off-by: Cui-xf <cui_xf293@qq.com>

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
@@ -58,6 +58,10 @@ public final class DefaultToolCallResultConverter implements ToolCallResultConve
 			final var imgB64 = Base64.getEncoder().encodeToString(buf.toByteArray());
 			return JsonParser.toJson(Map.of("mimeType", "image/png", "data", imgB64));
 		}
+		else if (result instanceof String stringResult) {
+			logger.debug("Returning string result.");
+			return stringResult;
+		}
 		else {
 			logger.debug("Converting tool result to JSON.");
 			return JsonParser.toJson(result);


spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
```diff
@@ -58,6 +58,10 @@ public String convert(@Nullable Object result, @Nullable Type returnType) {
 			final var imgB64 = Base64.getEncoder().encodeToString(buf.toByteArray());
 			return JsonParser.toJson(Map.of("mimeType", "image/png", "data", imgB64));
 		}
+		else if (result instanceof String stringResult) {
```
markpollack, line 61: the current assumption is that if is not a rendered image, we assume it is json (also a string)  in this current PR, the json conversion in the else statement will never be accessed.  Can you explain more of the surrounding use case that motivated you to make this change?

markpollack: Also wondering why `JsonParser.toJson(result);` doesn't use `Type returnType`

markpollack: any thoughts @tzolov ?

Cui-xf: Hey, thanks for the review!
I noticed that when I use the org.springframework.ai.tool.annotation.Tool annotation and the method returns a String, the result gets escaped (see screenshot below). I’m not really sure if this is the expected behavior, or if there’s any doc about it.
If this is actually, I will to close the PR.
Thanks!

image
image

tzolov:
@Cui-xf this is an interesting case:

Currently the `DefaultToolCallResultConverter` assumes that all string values are of type JSON (including) the string literals.
Note that `"Foo Bar"` is not JSON but `toJson("Foo Bar")` converts it into a valid json: `""Foo Bar""`.

Here is an example how the DefaultToolCallResultConverter behvaves:

```java
public class Main {

    private static final ObjectMapper OBJECT_MAPPER = JsonMapper.builder().build();

        var value1 = "Foo Bar";
        System.out.println("Value 1: " + value1);
        System.out.println(String.format("- raw value: %s, is JSON: %s", value1, isValidJSONJackson(value1)));
        System.out.println(String.format("- toJson(value): %s, is JSON: %s", toJson(value1),
                isValidJSONJackson(toJson(value1))));

        var value2 = "{\"hello\": \"world\"}";
        System.out.println("Value 2: " + value2);
        System.out.println(String.format("- raw value: %s, is JSON? %s", value1, isValidJSONJackson(value2)));
        System.out.println(String.format("- toJson(value): %s, is JSON? %s", toJson(value2),
                isValidJSONJackson(toJson(value2))));
    }

    public static boolean isValidJSONJackson(String jsonString) {
        try {
            OBJECT_MAPPER.readTree(jsonString);
            return true;
        } catch (Exception e) {
            return false;
        }
    }

    	public static String toJson(@Nullable Object object) {
		try {			
			return OBJECT_MAPPER.writeValueAsString(object);
		}
		catch (JsonProcessingException ex) {
			throw new IllegalStateException("Conversion from Object to JSON failed", ex);
		}
	}
}
```

Would generate:
```
Value 1: Foo Bar
- raw value: Foo Bar, is JSON: false
- toJson(value): "Foo Bar", is JSON: true

Value 2: {"hello": "world"}
- raw value: Foo Bar, is JSON? true
- toJson(value): "{\"hello\": \"world\"}", is JSON? true
```

Before we try to change the default behavior perhaps you can try to  implement a custom  ToolCallResultConverter implementation? 
https://docs.spring.io/spring-ai/reference/api/tools.html#_result_conversion

ilayaperumalg:
@Cui-xf Closing this PR as the custom ToolResultConverter is the potential way to fix this issue. Thanks for the contribution!

PR #2781, opened 2025-04-17
Fixed message order for JDBC Chat Memory

linarkou:
This PR solves some problems with message ordering:
1) JdbcChatMemory fetches rows in DESC order, and MessageChatMemoryAdvisor get and add all messages in that order - from last to first. So messages list needs to be reversed.
2) After batch insert without specifying timestamp manually  all rows have the same timestamp (because database sets current_timestamp by default). Sooo after fetching rows from database the message order is unpredictable - they have same timestamp.

PR #2890, opened 2025-04-25
Chat Memory Enhancements

ThomasVitale:
* ChatMemory will become a generic interface to implement different memory management strategies. It’s been moved from the “”spring-ai-client-chat” package to “spring-ai-model” package while retaining the same package, so it’s transparent to users.
* A MessageWindowChatMemory has been introduced to provide support for a chat memory that keeps at most N messages in the memory.
* A ChatMemoryRepository interface has been introduced to support different storage strategies for the chat memory. It’s meant to be used as part of a ChatMemory implementation. This is different than before, where the storage-specific implementation was directly tied to the ChatMemory. This design is familiar to Spring users since it’s used already in the ecosystem. The goal was to use a programming model similar to Spring Session and Spring Data.
* The JdbcChatMemory has been supersed by JdbcChatMemoryRepository.
* A ChatMemory bean is auto-configured for you whenever using one of the Spring AI Model starters. By default, it uses the MessageWindowChatMemory implementation and stores the conversation history in memory. If a different repository is already configured (e.g., Cassandra, JDBC, or Neo4j), Spring AI will use that instead.
* First-class documentation has been introduced to describe the ChatMemory API and related features.
* All the changes introduced in this PR are backward-compatible.


ThomasVitale:
@leijendary we are introducing a new `ChatMemoryRepository` API to decouple the storage mechanism (JDBC, Cassandra...) from the memory management strategy (last N messages, last N tokens...). In this pull request, as the first mover, I have extended the JDBC chat memory support to use the new APIs. The existing ones are still there, but deprecated. Since you implemented the initial JDBC chat memory support, it would be great if you could have a look at the JDBC changes in this PR and share your feedback. Thank you!

The [new documentation](https://github.com/spring-projects/spring-ai/blob/772be8b5ba4551d6261e86e1997e2525d00d6eef/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat-memory.adoc) explains how these new APIs will work.

The [upgrade notes](https://github.com/spring-projects/spring-ai/blob/772be8b5ba4551d6261e86e1997e2525d00d6eef/spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc#chat-memory) explains how to migrate to the new APIs.

spring-ai-model/src/main/java/org/springframework/ai/chat/memory/ChatMemoryRepository.java
```diff
@@ -16,30 +16,24 @@
...
-	// TODO: consider a non-blocking interface for streaming usages
+public interface ChatMemoryRepository {
 
-	default void add(String conversationId, Message message) {
-		this.add(conversationId, List.of(message));
-	}
+	List<String> findConversationIds();
 
-	void add(String conversationId, List<Message> messages);
+	List<Message> findMessages(String conversationId);
```
markpollack, line 33: findByConversationId instead?  just thinking of spring data repo conventions for finders.

spring-ai-model/src/main/java/org/springframework/ai/chat/memory/ChatMemoryRepository.java
```diff
@@ -16,30 +16,24 @@
...
 
-	List<Message> get(String conversationId, int lastN);
+	void saveMessages(String conversationId, List<Message> messages);
 
-	void clear(String conversationId);
+	void deleteMessages(String conversationId);
```
markpollack, line 37: deleteByConversationId  ?

spring-ai-model/src/main/java/org/springframework/ai/chat/memory/ChatMemoryRepository.java
```diff
@@ -16,30 +16,24 @@
...
-	default void add(String conversationId, Message message) {
-		this.add(conversationId, List.of(message));
-	}
+	List<String> findConversationIds();
 
-	void add(String conversationId, List<Message> messages);
+	List<Message> findMessages(String conversationId);
 
-	List<Message> get(String conversationId, int lastN);
+	void saveMessages(String conversationId, List<Message> messages);
```
markpollack, line 35: saveAll ?

memory/spring-ai-model-chat-memory-jdbc/src/main/java/org/springframework/ai/chat/memory/jdbc/JdbcChatMemoryRepository.java
```diff
@@ -0,0 +1,152 @@
...
+		@Nullable
+		public Message mapRow(ResultSet rs, int i) throws SQLException {
+			var content = rs.getString(1);
+			var type = MessageType.valueOf(rs.getString(2));
+
+			return switch (type) {
+				case USER -> new UserMessage(content);
+				case ASSISTANT -> new AssistantMessage(content);
+				case SYSTEM -> new SystemMessage(content);
+				case TOOL -> null;
```
markpollack, line 124: is returning null ok here?

memory/spring-ai-model-chat-memory-jdbc/src/main/java/org/springframework/ai/chat/memory/jdbc/JdbcChatMemoryRepository.java
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * An implementation of {@link ChatMemoryRepository} for JDBC.
+ *
+ * @author Jonathan Leijendekker
+ * @author Thomas Vitale
+ * @since 1.0.0
+ */
+public class JdbcChatMemoryRepository implements ChatMemoryRepository {
+
+	private static final String QUERY_GET_IDS = """
```
markpollack, line 42: as with vector store schemas, in the future we need to let users customize their schema, e.g. table and row names.

auto-configurations/models/chat/memory/spring-ai-autoconfigure-model-chat-memory-jdbc/src/main/java/org/springframework/ai/model/chat/memory/jdbc/autoconfigure/JdbcChatMemoryDataSourceScriptDatabaseInitializer.java
```diff
@@ -25,6 +25,11 @@
...
 import org.springframework.boot.sql.init.DatabaseInitializationSettings;
 
+/**
+ * Performs database initialization for the JDBC Chat Memory Repository.
+ *
+ * @since 1.0.0
+ */
 class JdbcChatMemoryDataSourceScriptDatabaseInitializer extends DataSourceScriptDatabaseInitializer {
 
 	private static final String SCHEMA_LOCATION = "classpath:org/springframework/ai/chat/memory/jdbc/schema-@@platform@@.sql";
```
markpollack, line 35: I don't see where the classpath resources are located...

memory/spring-ai-model-chat-memory-jdbc/src/main/java/org/springframework/ai/chat/memory/jdbc/JdbcChatMemoryRepository.java
ThomasVitale: Makes sense! I kept the original logic from JdbcChatMemory, but it does make sense to handle tool messages nicely. I'll do that.

memory/spring-ai-model-chat-memory-jdbc/src/main/java/org/springframework/ai/chat/memory/jdbc/JdbcChatMemoryRepository.java
```diff
@@ -0,0 +1,152 @@
...
+			SELECT conversation_id FROM ai_chat_memory
+			""";
```
markpollack, line 44: select distinct instead otherwise get duplicates?

spring-ai-model/src/main/java/org/springframework/ai/chat/memory/ChatMemoryRepository.java
ThomasVitale: Done

spring-ai-model/src/main/java/org/springframework/ai/chat/memory/ChatMemoryRepository.java
ThomasVitale: Done

spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
```diff
@@ -0,0 +1,147 @@
...
+		Assert.hasText(conversationId, "conversationId cannot be null or empty");
+		this.chatMemoryRepository.deleteMessages(conversationId);
+	}
+
+	private List<Message> process(List<Message> memoryMessages, List<Message> newMessages) {
+		List<Message> processedMessages = new ArrayList<>();
+
+		boolean hasNewSystemMessage = newMessages.stream()
+			.filter(message -> message instanceof SystemMessage)
+			.anyMatch(message -> !memoryMessages.contains(message));
```
markpollack, line 90: this is a loop within a loop.

```
		Set<Message> memoryMessagesSet = new HashSet<>(memoryMessages);

		boolean hasNewSystemMessage = newMessages.stream()
				.filter(SystemMessage.class::isInstance) 
				.anyMatch(message -> !memoryMessagesSet.contains(message));
```				
makes it more efficient

memory/spring-ai-model-chat-memory-jdbc/src/main/java/org/springframework/ai/chat/memory/jdbc/JdbcChatMemoryRepository.java
ThomasVitale: I changed it to return an empty ToolResponseMessage, because the storage saves the content of "getText()" and that method returns always an empty string for ToolResponseMessage.

We need a separate enhancement if we want to store the list of tool call results as well.

spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
markpollack: Is it correct that the UserMessage object equals/hashcode doesn't take into account media?  we aren't storing it, so yes, but prob it needs to be updated?

memory/spring-ai-model-chat-memory-jdbc/src/main/java/org/springframework/ai/chat/memory/jdbc/JdbcChatMemoryRepository.java
ThomasVitale: Done. This comment made me realise the current SQL schema definition do not define a primary key. I fixed that as well.

spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
ThomasVitale: Done

memory/spring-ai-model-chat-memory-jdbc/src/main/java/org/springframework/ai/chat/memory/jdbc/JdbcChatMemoryRepository.java
ThomasVitale: Postponed the primary key discussion for a separate task.

spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
ThomasVitale: About the UserMessage, I guess it's not correct that we are not considering the media. Agreed that it should be updated.

For SystemMessages, that works fine since it only has text. But for other message types, we should probably override the equals/hashcode inherited from the AbstractMessage.

ThomasVitale:
Committed in https://github.com/spring-projects/spring-ai/commit/0024e4dd497be3222367508e5c99a37068afd966

linarkou:
@ThomasVitale thank you for the new api, really fine!
But I think fetching all rows from the database, loading it to memory and then keeping only last N seems to be not very efficient - maybe better to keep an ability to fetch only N rows in `ChatMemoryRepository`?

memory/spring-ai-model-chat-memory-jdbc/src/main/java/org/springframework/ai/chat/memory/jdbc/JdbcChatMemory.java
```diff
@@ -38,14 +39,17 @@
...
+ * {@link ChatMemory} implementations) with a {@link JdbcChatMemoryRepository} instance.
  */
+@Deprecated
 public class JdbcChatMemory implements ChatMemory {
 
 	private static final String QUERY_ADD = """
 			INSERT INTO ai_chat_memory (conversation_id, content, type) VALUES (?, ?, ?)""";
 
 	private static final String QUERY_GET = """
-			SELECT content, type FROM ai_chat_memory WHERE conversation_id = ? ORDER BY "timestamp" DESC LIMIT ?""";
```
linarkou, line 48: It wasn't correct to remove descending order here - now it will always fetch only first N rows instead of last N.
I'll fix message order in #2781.

ThomasVitale: Thanks for the feedback! The `ChatMemoryRepository` as designed now keeps in storage only the messages allowed by the specific `ChatMemory` strategy. Right now, the only built-in strategy is `MessageWindowChatMemory`, which keeps at most M messages. The logic for sorting out which messages to keep/remove is handled within `ChatMemory` before storing them, whereas `ChatMemoryRepository` stores exactly what the `ChatMemory` defines. 

When you call `JdbcChatMemoryRepository.findByConversationId()`, you want to return all the messages because those are the ones determined as relevant to keep in memory. That's why I removed the "DESC". And that's why the `lastN()` method is deprecated and will be removed (besides leading to wrong results). When you call `JdbcChatMemoryRepository.saveAll()`, you practically overwrite the current list of messages with the new list (which has been processed already based on a given memory management strategy, such as # of messages or # of tokens).

If the message window is set to 15 messages, I will always have at most 15 messages stored in the database for a given `conversationId`. Does it make sense? I'm sure we can optimize the operations, especially the `saveAll()` one, since it's not the best. But I guess we need to introduce identifiers for each row (right now they don't have a primary key). Possibly related to https://github.com/spring-projects/spring-ai/issues/2902. If we make the schema definition customizable, it would be possible to enable different types of implementations, such as having one row per `conversationId`, stored as a JSON BLOB since they are handled as a single unit anyway.

Considering also your other comment, I wonder if we need some kind of specialization of the API or perhaps new APIs. `ChatMemory` as designed now is tailored towards standard short-term memory, keeping only the messages identified by the specific strategy.

If we want to keep the entire history and then make decisions at query time (with standard capabilities for filtering, selecting, and so on), I think we might need a different strategy. That would probably require a separate API since it wouldn't be `chat memory`, but it would be `chat history`. And for that we would benefir from the existing Spring Data APIs. I'll create a separate issue to talk about that since it's strictly related to our wish to surface memory as first-class citizen in ChatClient, and when doing that we need to support actual memory (short-term) but also chat history (long-term).

ThomasVitale: About what kind of messages end up in the memory, right now the built-in advisors in `ChatClient `never store system messages and tool response messages. But since `ChatMemory `is a core API that can be used outside `ChatClient `(for example, directly with `ChatModel `as shown [here](https://docs.spring.io/spring-ai/reference/api/chat-memory.html#_memory_in_chat_model)), we need to support all message types.

ThomasVitale: Oh yeah, good catch! I hadn't noticed that. Thanks for reporting it! That should still be there for the old implementation. Are you fixing it in your existing PR or should I create a separate task?

Thanks so much for all your contributions!

linarkou: Already fixed it in my PR

Issue #2518, opened 2025-03-19
Remove deprecation in ObservationContext for models


ThomasVitale:
In 1.0.0-M6, the ".requestOptions()" method for ModelObservationContext instances has been deprecated and marked for removal. This issue is to track the changes needed to use the new strategy and remove the deprecated logic.

- [x] Anthropic
- [ ] Azure OpenAI
- [ ] Bedrock Converse
- [x] Mistral AI => https://github.com/spring-projects/spring-ai/pull/2519
- [ ] OCI GenAi
- [x] Ollama => https://github.com/spring-projects/spring-ai/pull/2520
- [x] OpenAI => https://github.com/spring-projects/spring-ai/pull/2521
- [x] Transformers
- [ ] VertexAI Embedding
- [ ] VertexAI Gemini
- [ ] Minimax
- [ ] ZhiPu

PR #2896, opened 2025-04-26
GH-2518: Remove requestOptions from observation context objects

sobychacko:
Fixes: #2518

Issue: https://github.com/spring-projects/spring-ai/issues/2518

This commit removes the deprecated requestOptions field from ChatModelObservationContext and EmbeddingModelObservationContext classes. Instead of passing options separately, the code now retrieves them directly from the request objects (prompt.getOptions() or embeddingRequest.getOptions()).

Key changes:
- Removed requestOptions parameter from observation context builders
- Updated all model implementations to stop passing options separately
- Fixed EmbeddingRequest handling in several model implementations
- Added buildEmbeddingRequest method in models to properly merge options

This change simplifies the API and removes duplication, as options are already available in the request objects themselves.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/observation/DefaultChatModelObservationConvention.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/observation/DefaultChatModelObservationConvention.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@ import java.util.StringJoiner;
 import io.micrometer.common.KeyValue;
 import io.micrometer.common.KeyValues;
 
+import org.springframework.ai.chat.prompt.ChatOptions;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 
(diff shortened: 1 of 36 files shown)


spring-ai-model/src/main/java/org/springframework/ai/chat/observation/DefaultChatModelObservationConvention.java
```diff
@@ -48,9 +50,9 @@ public String getName() {
 
 	@Override
 	public String getContextualName(ChatModelObservationContext context) {
-		if (StringUtils.hasText(context.getRequestOptions().getModel())) {
-			return "%s %s".formatted(context.getOperationMetadata().operationType(),
-					context.getRequestOptions().getModel());
+		ChatOptions options = context.getRequest().getOptions();
+		if (StringUtils.hasText(options.getModel())) {
```
ThomasVitale, line 54: Unfortunately, a `Prompt` object can return `null` `ChatOptions`. Before, when handling options separately, we were enforcing the non-nullability. Now, we need to take into account that whenever we call the `getOptions()` method from a `Prompt` object, we might get back a `null` object. This line might throw a `NullPointerException`.

It would be nice if, in the future, we could enhance the `Prompt` class to always return a valid `ChatOptions` object.

ilayaperumalg: This is addressed via 4eeeb83d3f70e968b72b3166c6c1324c811246e2

PR #2926, opened 2025-04-28
Add default ChatOptions to Prompt

ilayaperumalg:
- Remove @Nullable on ChatOptions
 - By default, pass the default ChatOptions from the DefaultChatOptionsBuilder


markpollack:
updated `copy` and `merge` to reflect the chat options not being null.

merged in 4eeeb83d3f70e968b72b3166c6c1324c811246e2

PR #2956, opened 2025-04-30
Added a "skipBuiltInFunctionsValidation" option to StTemplateRenderer

sunyuhan1998:
Added a "skipBuiltInFunctionsValidation" option to StTemplateRenderer, which defaults to false. When enabled, it skips validation of built-in functions in the template, preventing these built-in functions from being validated as variables to be replaced.

### DIFF
--- a/spring-ai-template-st/src/main/java/org/springframework/ai/template/st/StTemplateRenderer.java
+++ b/spring-ai-template-st/src/main/java/org/springframework/ai/template/st/StTemplateRenderer.java
@@ -24,6 +24,7 @@ import org.springframework.ai.template.TemplateRenderer;
 import org.springframework.ai.template.ValidationMode;
 import org.springframework.util.Assert;
 import org.stringtemplate.v4.ST;
+import org.stringtemplate.v4.compiler.Compiler;
 import org.stringtemplate.v4.compiler.STLexer;
 
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


ThomasVitale:
Thanks so much for this, very useful! I was wondering if we should reframe this functionality from "skip built-in functions validation" to "enable built-in functions support". What do you think?

Basically, I would consider whether it makes sense to rename "skipBuiltInFunctionsValidation" to "supportBuiltInFunctions" or "supportStFunctions"

sunyuhan1998:
> Thanks so much for this, very useful! I was wondering if we should reframe this functionality from "skip built-in functions validation" to "enable built-in functions support". What do you think?
> 
> Basically, I would consider whether it makes sense to rename "skipBuiltInFunctionsValidation" to "supportBuiltInFunctions" or "supportStFunctions".
> 
> @sunyuhan1998 what do you think?

Hi @ThomasVitale 
I agree with your suggestion. I did struggle a bit with the name of this option, and I think supportStFunctions is a good name. I've already submitted a PR, please feel free to review it when you have time. Thank you!

ThomasVitale (approved):

ThomasVitale:
Looks good, thank you!

markpollack:
I'll merge this, but I think we can revisit and remove the option, as setting it to true doesn't seem like a realistic use case.  why do we care to enforce what features of the template engine are used, we just need to avoid these built in functions when doing validation always.

markpollack:
merged in 788fc3afe5c7ea8793a47a32d9e1f511c95477e3

PR #2963, opened 2025-05-01
Inner advisor should handle output format instructions

ThomasVitale:
Output format instructions should not be included until the very last advisor runs, otherwise there's a risk of templating failure if more than one advisor tries to render the prompt template. This change guarantees the output format instructions are always included right before calling the chat model, without the risk of previous advisors interfering with it.

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientAttributes.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientAttributes.java
@@ -21,10 +21,7 @@ package org.springframework.ai.chat.client;
  *
  * @author Thomas Vitale
  * @since 1.0.0
- * @deprecated only introduced to smooth the transition to the new APIs and ensure
- * backward compatibility
  */
-@Deprecated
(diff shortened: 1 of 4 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientAttributes.java
```diff
@@ -33,7 +30,6 @@ public enum ChatClientAttributes {
 	ADVISORS("spring.ai.chat.client.advisors"),
 	@Deprecated // Only for backward compatibility until the next release.
 	CHAT_MODEL("spring.ai.chat.client.model"),
-	@Deprecated // Only for backward compatibility until the next release.
 	OUTPUT_FORMAT("spring.ai.chat.client.output.format"),
```
ThomasVitale, line 33: Not a great approach using the advisor context for this, but we might re-evaluate later when considering the new structured output support.

markpollack:
added a simple test with streaming - though we may have it elsewhere...

merged in 90cab219d28e56e8ac0d76ef456d6351dab1ed04

Thanks!

PR #2964, opened 2025-05-01
feat: Improve validation in MethodToolCallbackProvider

tzolov:
This ensures that validation errors are caught early during object construction rather than later when methods are called, providing better error feedback.

- Add validation for tool-annotated methods during construction
- Validate duplicate tool names in constructor instead of only at getToolCallbacks() time
- Add comprehensive test suite for MethodToolCallbackProvider

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallbackProvider.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallbackProvider.java
@@ -19,6 +19,7 @@ package org.springframework.ai.tool.method;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -44,6 +45,7 @@ import org.springframework.util.ReflectionUtils;
  * {@link Tool}-annotated methods.
  *
  * @author Thomas Vitale
+ * @author Christian Tzolov
  * @since 1.0.0
  */
 public final class MethodToolCallbackProvider implements ToolCallbackProvider {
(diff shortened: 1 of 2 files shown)


spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallbackProvider.java
```diff
@@ -55,7 +57,25 @@ public final class MethodToolCallbackProvider implements ToolCallbackProvider {
...
+		for (Object toolObject : toolObjects) {
+			List<Method> toolMethods = Stream
+				.of(ReflectionUtils.getDeclaredMethods(
+						AopUtils.isAopProxy(toolObject) ? AopUtils.getTargetClass(toolObject) : toolObject.getClass()))
+				.filter(toolMethod -> toolMethod.isAnnotationPresent(Tool.class))
+				.filter(toolMethod -> !isFunctionalType(toolMethod))
+				.toList();
+
+			if (toolMethods.isEmpty()) {
+				throw new IllegalStateException("No @Tool annotated methods found in " + toolObject);
```
ThomasVitale, line 76: Consider extending this message to include "Did you mean to pass a ToolCallback or ToolCallbackProvider? If so, you have to use .toolCallbacks() instead of .tool()".

It could also be a warn log message.

spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallbackProvider.java
```diff
@@ -55,7 +57,25 @@ public final class MethodToolCallbackProvider implements ToolCallbackProvider {
 	private MethodToolCallbackProvider(List<Object> toolObjects) {
 		Assert.notNull(toolObjects, "toolObjects cannot be null");
 		Assert.noNullElements(toolObjects, "toolObjects cannot contain null elements");
+		assertToolAnnotatedMethodsPresent(toolObjects);
 		this.toolObjects = toolObjects;
+		validateToolCallbacks(getToolCallbacks());
+	}
+
+	private void assertToolAnnotatedMethodsPresent(List<Object> toolObjects) {
```
ThomasVitale, line 65: very nice!

spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallbackProvider.java
markpollack: I think we are seeing that this is a pretty critical fix, the app won't work as expected so I think an exception rather than a warn message is appropriate.  agree wrt to the log message.

ryanjbaxter: Agree.  Also they could mean to call `defaultToolCallbacks` or `defaultTool`.

markpollack:
merged in 6c52c99291d8a3d1d8f04d8b9845e3e071245ba4 

kept the exception, can change if needed.  let's see how this plays.

PR #2965, opened 2025-05-01
Mark the builder methods in `OpenAiImageOptions.Builder` that start with `with` as `@Deprecated`.

dev-jonghoonpark:
- Mark the builder methods in `OpenAiImageOptions.Builder` that start with `with` as `@Deprecated`. 
  - Note: The removal of the `with` prefix from builder methods, as described in PR #1939, was not fully completed for the `OpenAiImageOptions` class.


markpollack:
Since we are going to RC, we will remove these as there isn't a deprecation cycle left.

markpollack:
Updated the class and test to match with OpenAiChatOptions

merged in 9ca3a7d15ff3314265448622ffd3afa824efbb9b

tx!

PR #3002, opened 2025-05-05
Remove deprecations in JdbcChatMemory

ThomasVitale:
* Remove deprecated APIs for JdbcChatMemory.
* Improve documentation about chat memory vs. chat history.
* Fix mismatch between docs vs code for max messages in MessageWindowChatMemory.


spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
```diff
@@ -40,7 +40,7 @@
  */
 public final class MessageWindowChatMemory implements ChatMemory {
 
-	private static final int DEFAULT_MAX_MESSAGES = 200;
+	private static final int DEFAULT_MAX_MESSAGES = 20;
```
markpollack, line 43: how did you come up with 20 as a heuristic BTW?

markpollack:
merged

PR #2662, opened 2025-04-07
Improve JDBC Chat Memory

eddumelendez:
* Apply database initialization checks
* Switch initializeSchema to `DatabaseInitializationMode`

Signed-off-by: Eddú Meléndez <eddu.melendez@gmail.com>

PR #3010, opened 2025-05-06
Align JdbcChatMemoryProperties with other modules in Spring Boot

quaff:
Such as `BatchProperties.Jdbc` and `IntegrationProperties.Jdbc`

1. Change type of `initialize-schema` from `boolean` to `DatabaseInitializationMode`
2. Allow to customize schema location


linarkou:
Can you please add username/password properties like in`spring.sql.init`?
Usually not all users have DDL permissions, so it would be great to customize user here.
See [DataSourceInitializationConfiguration](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/sql/init/DataSourceInitializationConfiguration.java) for reference

quaff:
> Can you please add username/password properties like in`spring.sql.init`? Usually not all users have DDL permissions, so it would be great to customize user here. See [DataSourceInitializationConfiguration](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/sql/init/DataSourceInitializationConfiguration.java) for reference

`BatchProperties.Jdbc`, `IntegrationProperties.Jdbc`, `JdbcSessionProperties` and `QuartzProperties.Jdbc` do not support username/password.

linarkou:
Thank you! Yes, you right, it is not about alignment, it is just a feature. But I think it can be useful in some usecases.
Created #3018 for it

quaff:
Close it as duplicate of https://github.com/spring-projects/spring-ai/pull/2662.

Issue #2667, opened 2025-04-08
springBeanToolCallbackResolver  want to create a bean which is not existed


daicidemeigui:
I only used the MCP client to create methods and did not use Bean to construct them. However, when calling a method that does not exist in the MCP client, I will use the resolve method of the SpringBeanToolFHIR Parser created in ToolCallingAutoConfiguration to create the Bean. At this point, the program will directly report an error:
java.lang.IllegalArgumentException: Functional bean with name mcp_client_filesystem_delete_file does not exist in the context.

tzolov:
Hey @daicidemeigui , can you please elaborate on this issue. Is it possible to reproduce it?

yvannh:
Hey, I can tell that I got the same issue yesterday and I dont think it can be reproduced (maybe just in my case though), so I'll be just describing the architecture and what went wrong. I have two servers exposing tools (one is for translation - Translation MCP - and one for a database operations - Movie MCP, also acting as a client). When the Movie MCP client was asking for some translation, I got the issue on one of them, it did well but at this step no. The error is that the tool was misnamed with some random uppercase character (from tool calling trace : Executing tool call: spring_ai_mcp_client_translator_mCP_translate, instead of just spring_ai_mcp_client_translator_mcp_translate, mcp with all lower case). When I reran, the issue happened on the same translation I don't know why, there was no explanation on that. To "fix" the issue, I just reloaded the translation MCP server, the issue disappeared, but in my opinion it can reappear again.

daicidemeigui:
> Hey [@daicidemeigui](https://github.com/daicidemeigui) , can you please elaborate on this issue. Is it possible to reproduce it?

Maybe it's a model illusion, because I deployed a low parameter model locally, and he may have created a non-existent tool based on the provided tool, but I think this is also a bug

daicidemeigui:
> Hey [@daicidemeigui](https://github.com/daicidemeigui) , can you please elaborate on this issue. Is it possible to reproduce it?

The scene at that time was like this:
My question is' help me to delete a file named 'test. txt' under'D: '/ 006765/Desktop/claude mcp examples main '", then using the following two open-source MCP services, configured as follows:
{
  "mcpServers": {
    "playwright": {
      "command": "node",
      "args": [
        "-e",
        "require('child_process').execSync('npx -y @executeautomation/playwright-mcp-server', {stdio: 'inherit'})"
      ]
    },
    "filesystem": {
      "command": "node",
      "args": [
[...]

PR #3019, opened 2025-05-07
fix(tool): handle non-existing tool callbacks gracefully

tzolov:
- Add error handling to SpringBeanToolCallbackResolver to return null instead of throwing exceptions when resolving non-existing tools. This make is consitent with other tool resolver implementations and centralize the error handling in the tool calling manager.
- Add test to verify that the ToolCallbackResolver correctly returns null for non-existing tools.

Related to #2667


ilayaperumalg:
Rebased and merged as e010899fb

PR #3020, opened 2025-05-07
Fix mutating global RestClient and WebClient builders

chemicL:
Since the builders for HTTP clients are mutable and shared, they should only be configured with globally applicable settings. The current use leaks specific details into other usages and affects newly instantiated clients. This PR applies the clone() method right before mutation happens as it probably is the strategy that avoids multiple unnecessary copies.

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/AnthropicApi.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/AnthropicApi.java
@@ -119,12 +119,14 @@ public class AnthropicApi {
 
 		this.completionsPath = completionsPath;
 
-		this.restClient = restClientBuilder.baseUrl(baseUrl)
+		this.restClient = restClientBuilder.clone()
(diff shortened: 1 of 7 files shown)


chemicL:
The build fails due to unrelated change (b167f21) in the main branch.

tzolov:
Rebased, Updated with 

    feat: improve RestClient and WebClient instantiation and configuration across AI modules
    
    - Ensure RestClient.Builder and WebClient.Builder are properly cloned before setting base URLs and headers in Mistral, OpenAI, and Ollama API classes to prevent side effects.
    - Add WebClientAutoConfiguration to auto-configuration classes for Anthropic and Mistral AI modules.
    - Update Mistral AI auto-configuration to inject and use WebClient.Builder.
    - Clean up test and annotation usage (remove unnecessary @Nullable, fix test builder usage in AnthropicApiIT).
    - Improve consistency and reliability of HTTP client configuration across model integrations.
    
and merged at 97f90b19d62076b0c1249c89c7dc3a2869cc367a

PR #3025, opened 2025-05-07
Polish `Neo4jChatMemoryRepository`.

michael-simons:
This change turns all the labels into parameters, avoiding the possibility of Cypher injection as the config does not do any sanitization.
In addition, the interaction with the driver is changed so that it uses transactional functions, which are retried when any communication with the Neo4j DBMS fails.
We can do this here as the repository is not subject to application wide transactions.

An alternative to the parameters for labels would be using Cypher-DSL as we did in other parts of this project to sanitize labels proper.


michael-simons:
Wonderful, thank you @ilayaperumalg and nice to see you again!

PR #3031, opened 2025-05-07
Support DocumentPostProcessors in RAG Advisor

ThomasVitale:
The DocumentPostProcessor is one of the modular RAG components introduce in M8. You can now use this API from within the RetrievalAugmentationAdvisor to post-process the retrieved documents before passing them to the model. For example, you can use such an interface to perform re-ranking of the retrieved documents based on their relevance to the query, remove irrelevant or redundant documents, or compress the content of each document to reduce noise and redundancy.


markpollack:
merged in 2ea518686f33046daeeb541515c94fb9031a114a

PR #3034, opened 2025-05-07
Improve StTemplateRenderer

markpollack:
Rename supportStFunctions to validateStFunctions and improve variable extraction

- Renamed all occurrences of supportStFunctions to validateStFunctions for clarity.
- Updated default field, constructor, and builder method to use new naming.
- Improved getInputVariables logic to better distinguish variables, function calls, and property access.
- Ensured built-in functions accessed as properties are only skipped when validateStFunctions is true.
- Enhanced builder usage to reflect new flag and naming.
- More tests added


sobychacko:
LGTM - Merged via https://github.com/spring-projects/spring-ai/commit/0e15197298c0848b78a746f3d740191e6a6aee7a.

PR #3040, opened 2025-05-08
Fix Gemini Tool Calling for texts returned from MethodToolCallback

nlinhvu:
`MethodToolCallback` can return a `String`(which is **Textual Json/Json String**) that is not supported by `jsonToStruct` method. => Update `jsonToStruct` method with **Textual Json/Json String**.

`MethodToolCallback` relies on `DefaultToolCallResultConverter` which is not always return **Textual Json/Json String**, in case of methodReturnType is `void`, it simply returns simple `String`: "Done" that cannot be parsed by `jsonToStruct` => Update `DefaultToolCallResultConverter` always return **Textual Json/Json String**.

### DIFF
--- a/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatModel.java
+++ b/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatModel.java
@@ -340,7 +340,10 @@ public class VertexAiGeminiChatModel implements ChatModel, DisposableBean {
 
 			Struct.Builder structBuilder = Struct.newBuilder();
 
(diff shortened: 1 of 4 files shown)


tzolov:
@nlinhvu thank you for catching this

tzolov:
Rebased and merged at f2df87b5672339f0dbac8a2896bbbf88e99c8d5d

cho-thinkfree-com:
The issue I found is slightly different from this one.

When creating an exception processor like this:
```
DefaultToolExecutionExceptionProcessor(false)
```

The exception message is passed into the ToolResponse as a plain String in the responseData field.
Since it is delivered as a plain string rather than a JSON string, it causes errors when processed by the Vertex AI Chat Model.

I have submitted a fix for this in [PR #4521](https://github.com/spring-projects/spring-ai/pull/4521)

PR #3046, opened 2025-05-08
Add JsonIgnoreProperties to Model Response objects

### LINKED ISSUE #3026: [EPIC] - Add @JsonIgnoreProperties(ignoreUnknown = true) to all response object
Be more flexibility in what we receive, allowing for extra fields that may be added in model responses.  These are records in the various `*Api.java` classes.

- [x] OpenAI
- [x] Anthropic

... etc.

ilayaperumalg:
- For all the model response API objects, add `@JsonIgnoreProperties(ignoreUnknown = true)` which provides flexibility to ignore any unknown properties which come as part of the response

Resolves #3026


sobychacko:
Merged via https://github.com/spring-projects/spring-ai/commit/53019c69d16f6e4b048765a39bb62d92b6f33c7a.

Issue #2995, opened 2025-05-05
NumberFormatException when model returns scientific notation (e.g., “1.5009969E7”) for int parameter in tool method


Oren-Peer:
**Bug description**
When using the @Tool annotation to define a method with an int parameter, and the model returns a numeric value in scientific notation (e.g., "1.5009969E7"), Spring AI’s internal JsonParser attempts to parse it using Integer.parseInt, which fails with a NumberFormatException.

```
java.lang.NumberFormatException: For input string: \"1.5009969E7\"
	 java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67) ~[?:?]
	 java.base/java.lang.Integer.parseInt(Integer.java:588) ~[?:?]
	 java.base/java.lang.Integer.parseInt(Integer.java:685) ~[?:?]
	 org.springframework.ai.util.json.JsonParser.toTypedObject(JsonParser.java:131) ~[spring-ai-model-1.0.0-M8.jar:1.0.0-M8]
	 org.springframework.ai.tool.method.MethodToolCallback.buildTypedArgument(MethodToolCallback.java:148) ~[spring-ai-model-1.0.0-M8.jar:1.0.0-M8]
	 org.springframework.ai.tool.method.MethodToolCallback.lambda$buildMethodArguments$1(MethodToolCallback.java:139) ~[spring-ai-model-1.0.0-M8.jar:1.0.0-M8]
	 java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:215) ~[?:?]
	 java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:1024) ~[?:?]
	 java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:570) ~[?:?]
	 java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:560) ~[?:?]
	 java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:636) ~[?:?]
	 java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:291) ~[?:?]
	 java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:656) ~[?:?]
	 java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:662) ~[?:?]
[...]

PR #3051, opened 2025-05-08
fix(jsonparser): support scientific notation parsing for int/long

BeomSeogKim:
Fixes: #2995

Replaces the use of `parseInt` / `parseLong` in `JsonParser.toTypedObject()` 
with `BigDecimal.intValueExact()` / `longValueExact()` to support scientific notation 
(e.g., "1.5E7") safely and accurately.

- Added tests for scientific notation in `JsonParserTests`


ilayaperumalg:
@BeomSeogKim Thanks for the PR. Rebased and merged as 69c081c77

Issue #1924, opened 2024-12-13
[Observability] Add observability support for function calling.


Cirilla-zmh:
**Expected Behavior**

The observability of Spring AI is quite impressive, and it has helped us resolve numerous issues. Thank you so much!

However, as a crucial component of the Spring AI, function callings lack support for observability. In fact, our agent application contains a significant number of function calling invocations, and we would like to see them as spans in the trace. This visibility would help us understand how the LLM organizes these function calls and allow us to see the input and output during the invocation process (it's important data for evaluate LLM correctness).

We have conducted some preliminary research and believe that adding observability instrumentation to the `org.springframework.ai.chat.model.AbstractToolCallSupport#executeFunctions` method would be a good solution. If you have plans to support this but are currently not available, we would be more than happy to contribute.

**Current Behavior**

Function callings lack support for observability.

**Context**

No more context. If you have more you need to know, feel free to ask. :)

chickenlj:
I have encountered exactly the same issue when using Spring Ai, it would be great if the function calling branch could be displayed in the tracing result.

Cirilla-zmh:
By the way, we are also attempting to use the `opentelemetry-java-instrumentation` to export trace data generated by Spring AI to an OTLP-compatible observability backend. However, we have not yet found a suitable implementation approach. Are there any best practices or similar solutions that you could recommend? I would greatly appreciate any guidance.

Additionally, I have already raised this issue in the OpenTelemetry community:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/12878

xiaohai-78:
Observability for function calls would help me pinpoint performance bottlenecks more accurately. Looking forward to this feature!

ThomasVitale:
@Cirilla-zmh thanks for submitting this issue. I've started working on an implementation for this.

It's worth noting that Spring AI is instrumented the same way the rest of the Spring portfolio, based on the Micrometer Observation API. You can export traces and metrics via OTLP to an OpenTelemetry backend by adding the needed Micrometer dependencies. 

Using `opentelemetry-java-instrumentation` would be an alternative to Spring-own instrumentation. You can get the metrics and traces supported by the OpenTelemetry Java Instrumentation, but you won't get the ones defined by Spring itself (including Spring AI).

You can find full observability examples for models and vector stores here: https://github.com/ThomasVitale/llm-apps-java-spring-ai/tree/main/observability

Cirilla-zmh:
[...]

PR #3066, opened 2025-05-09
Micrometer instrumentation for tool calling

ThomasVitale:
Introduce instrumentation for tool calling using the Micrometer Observation API.

By default, metadata about tool calling are exported as metrics and traces. Optionally, the actual tool call input and result can be exported as well by enabling the dedicated feature flag.

Fixes gh-1924

### DIFF
--- a/auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
+++ b/auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
@@ -16,17 +16,17 @@
 
 package org.springframework.ai.model.tool.autoconfigure;
 
-import java.util.ArrayList;
-import java.util.List;
-
 import io.micrometer.observation.ObservationRegistry;
-
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.ai.chat.model.ChatModel;
 import org.springframework.ai.model.tool.ToolCallingManager;
 import org.springframework.ai.tool.ToolCallback;
 import org.springframework.ai.tool.ToolCallbackProvider;
 import org.springframework.ai.tool.execution.DefaultToolExecutionExceptionProcessor;
 import org.springframework.ai.tool.execution.ToolExecutionExceptionProcessor;
+import org.springframework.ai.tool.observation.ToolCallingContentObservationFilter;
+import org.springframework.ai.tool.observation.ToolCallingObservationConvention;
 import org.springframework.ai.tool.resolution.DelegatingToolCallbackResolver;
 import org.springframework.ai.tool.resolution.SpringBeanToolCallbackResolver;
 import org.springframework.ai.tool.resolution.StaticToolCallbackResolver;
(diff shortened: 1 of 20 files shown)


ThomasVitale:
FYI @shakuzen @jonatan-ivanov Standard Micrometer-based instrumentation following the patterns we used for rest of the project.

Some of the shared attributes come from naming conventions originally based on the OTel SemConv. If it's gonna be decided to rename them, then this new instrumentation will use those new names automatically since they are based on enums (i.e. not much refactoring needed).

auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
```diff
@@ -76,12 +84,27 @@ ToolExecutionExceptionProcessor toolExecutionExceptionProcessor() {
...
 			.build();
+
+		observationConvention.ifAvailable(toolCallingManager::setObservationConvention);
+
+		return toolCallingManager;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnProperty(prefix = ToolCallingProperties.CONFIG_PREFIX + ".observations", name = "include-content",
```
codefromthecrypt, line 102: fyi: in otel this I believe would be the same ENV var used for normal content. `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` as there is no difference in toggle whether it is a message or a tool. So, for folks using otel they need to map two properties to that.

codefromthecrypt:
@ThomasVitale curious is the schema capture limited to certain inference platforms or for all of them. I like that you have the ability to capture schema because there are some models that act differently depending on how schema are translated.

fyi on the otel genai semconv issue on schema capture https://github.com/open-telemetry/semantic-conventions/issues/1760

Another consumer of this data could be openinference (also otel export) which has a stable schema that includes tool .. schema ;) https://github.com/Arize-ai/openinference/blob/076bb7966d44fccdb2ab94e6f379ef4ae22c39b1/spec/semantic_conventions.md?plain=1#L45 cc @mikeldking @axiomofjoy

ThomasVitale:
@codefromthecrypt thanks for reviewing this! The schema captured is part of the ToolDefinition in Spring AI. It's defined consistently across providers, and most of them use JSON Schema. There are exceptions, though. For example, Vertex uses some OpenAPI variant instead of JSON Schema. Some providers might also support a subset of JSON Schema, so having this info part of the telemetry should be useful to troubleshoot those scenarios.

I'm aware that some platforms have defined their own naming for such attributes (and at some point, also OTel SemConv might have something). They can all be supported by customising the `ToolCallingObservationConvention`. The naming provided here out-of-the-box will be guaranteed by Spring AI to be stable while still allowing the adoption of different sets of conventions (e.g. to support OpenInference).

auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
ThomasVitale: Thanks for pointing this out! Yeah, I guess that will be needed.

codefromthecrypt:
@ThomasVitale thanks for the code and the commentary. Looking forward to it!

tzolov:
Rebased and merged at 30eb3ce3f2690482481e317d5ab396415c035eec

PR #3067, opened 2025-05-09
chore: Add IgnoredInToolInputSchema annotation to ignore the parameter when generating the tool's input schema.

He-Pin:
Motivation: 

Ignore the parameter when generating the input schema.

instead of https://github.com/spring-projects/spring-ai/pull/3062

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/model/ToolContext.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/model/ToolContext.java
@@ -16,11 +16,12 @@
 
 package org.springframework.ai.chat.model;
 
+import org.springframework.ai.chat.messages.Message;
+import org.springframework.ai.tool.annotation.IgnoredInToolInputSchema;
+
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-
-import org.springframework.ai.chat.messages.Message;
 
 /**
  * Represents the context for tool execution in a function calling scenario.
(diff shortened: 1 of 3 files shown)


markpollack:
Can you explain the motivation for this?  If you don't want a method in your @Tool annotated method to be used as an argument, it just shouldn't be there in the first place?  At least that is how I'm reading this...  Thanks

He-Pin:
@markpollack Thanks for looking into this. We are adding `TmcpContext` to the `call` method, so we want to ignore that type to be generated as the input schema.

```java
public interface TmcpToolCallback extends ToolCallback {

    /**
     * @param toolInput        the input to the tool
     * @param toolContext       tool context
     * @param tmcpInputContext tmcp input context
     */
    default String call(final String toolInput,
                        final @Nullable ToolContext toolContext,
                        final TmcpInputContext tmcpInputContext) {
        if (tmcpInputContext != null && !tmcpInputContext.getContext().isEmpty()) {
            throw new UnsupportedOperationException("tmcp context is not supported!");
        }
        return call(toolInput, toolContext);
    }
}
```

sobychacko:
Thanks for the PR!

Closing this one. It's a new public API rather than a fix, and we'd rather agree the shape before adding an annotation for this. That's a decision on the design, not the patch.

The branch is also well behind current code. If you still want the feature, a fresh issue with the use cases is the right place.

Thanks again.

PR #3082, opened 2025-05-10
feat: Add support for generic argument types in tool callbacks

### LINKED ISSUE #2462: MethodToolCallback: Incorrect Deserialization of List<T> with Class<?>
**Bug description**
A clear and concise description of what the bug is about.
Within the MethodToolCallback class, in the buildMethodArguments method, specifically the following implementation:

Image

The current implementation in buildTypedArgument uses Class<?> to determine the target type for deserialization. This approach can lead to deserialization failures for generic types like List<T> because the specific type parameter T is lost. The implementation here should use Jackson's JavaType instead of Class<?>. Using JavaType would allow Jackson to retain the complete generic type information, enabling the correct deserialization of generic collections like List<T>, where the specific type T can be properly identified.

**Environment**
JDK21 Spring-AI 1.0.0-M6

**Steps to reproduce**
Define tool method like this，items is actually of the type List\<LinkedHashMap\>.

Image

**Expected behavior**
items should be List<SourcePoItemDO>

tzolov:
- Enhance MethodToolCallback to properly handle generic types by using parameterized types
- Add unit tests for generic type handling (List, Map<String,Integer>, nested generics)
- Add integration tests for both Anthropic and OpenAI clients to verify tool calls with generic argument types

Resolves #2462

### DIFF
--- a/models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/client/ChatClientToolsWithGenericArgumentTypesIT.java
+++ b/models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/client/ChatClientToolsWithGenericArgumentTypesIT.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2023-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


markpollack:
merged in 5f6c618f7629ba28c9742c36eba0ea46a40387f6

PR #3085, opened 2025-05-10
Fix typo in PromptTemplateTest

jamespud:
The comments should follow the same pattern as those in PromptTemplate, such as:
```java
/**
 * A template for creating prompts. It allows you to define a template string with
 * placeholders for variables, and then render the template with specific values for those
 * variables.
 */
```


markpollack:
merged in 08814be3f151663c7d1236e2904b5ddc24d9b253

Thanks!

PR #3089, opened 2025-05-10
Chatclient developer message

andresssantos:
- Included DEVELOPER in the MessageType enum with value "developer"
- Created DeveloperMessage class with constructors, copy(), mutate(), and builder methods
- Updated AbstractMessage constructor to accept DEVELOPER messages without validation errors
- Updated ChatModel and ChatClient methods to support DEVELOPER message type
- Developed unit tests for DeveloperMessage, validating exception cases and proper inclusion in prompts


dev-jonghoonpark:
related pr : https://github.com/spring-projects/spring-ai/pull/2864

markpollack:
hi folks.  I do appreciate the PR, but I made a mistake in my comments on the other related one, saying that was incomplete.  This PR adds non portable 'developer' fluent api methods to the portable chatclient interface and that goes against the portability design idea.  These issues require more up front design as they touch large areas of the code. 

This issue as well as other would be addressed by a 'capability' feature, see https://github.com/spring-projects/spring-ai/issues/3103  so that can have more information on how to make correct adaptations to various features for specific models.  Likely this will live in a sort of options like manner, but again, needs desgin.

I'm going to leave this PR open now under 'to discuss` and move it to the post GA time frame.  I really love the capabilities (ha ha - no pun intended) of the reasoning models and want to get support for them in a a priority after GA.  Too bad it didn't work out before GA.

andresssantos:
Thank you for your feedback @markpollack. I agree that implementing a more general solution, such as a "capability" feature, makes more sense in the long run.

Issue #3118, opened 2025-05-12
NullPointerException in OpenAiChatModel during streaming chat completion (.stream()) with version 1.0.0-M8


Oussemasahbeni:
**Bug description**
When using spring-ai-openai version 1.0.0-M8, invoking the streaming chat API via chatClient.prompt(...).stream().content() results in a NullPointerException. The non-streaming equivalent (chatClient.prompt(...).call().content()) works correctly without errors.

![Image](https://github.com/user-attachments/assets/8addd4f0-26e7-4049-82c0-abdf5651c083)

**Environment**
Spring AI Version: 1.0.0-M8
Spring Boot Version: 3.4.5
Java Version:  JDK 21
 Model: gemini-2.5-flash-preview-04-17

**Steps to reproduce**
Create a Spring Boot application using spring-boot-starter-webflux and spring-ai-openai-spring-boot-starter.
Set the Spring AI version: <spring-ai.version>1.0.0-M8</spring-ai.version>
Configure OpenAI credentials (e.g., spring.ai.openai.api-key).
Inject ChatClient into a Spring component (e.g., a RestController).
Implement two endpoints: one for standard chat and one for streaming chat, similar to the examples below.
Call the /stream endpoint (e.g., GET /stream?message=hello).

**Expected behavior**
The /stream endpoint should return a Flux<String> that emits the streamed content chunks from the OpenAI API without any server-side exceptions.

**Minimal Complete Reproducible example**
[Please provide a failing test or a minimal complete verifiable example that reproduces the issue.](https://github.com/Oussemasahbeni/spring-ai-streaming)

dev-jonghoonpark:
The NPE occurs in the following code of `OpenAiChatModel.java` because `id` is null.

```java
Flux<ChatResponse> chatResponse = completionChunks.map(this::chunkToChatCompletion)
	.switchMap(chatCompletion -> Mono.just(chatCompletion).map(chatCompletion2 -> {
		try {
			@SuppressWarnings("null")
			String id = chatCompletion2.id();

			List<Generation> generations = chatCompletion2.choices().stream().map(choice -> { // @formatter:off
				if (choice.message().role() != null) {
					roleMap.putIfAbsent(id, choice.message().role().name()); // <-- NPE occurs in here
```

dev-jonghoonpark:
I will modify it so that when an ID is not provided, it is set to `NO_ID` instead of null, test the changes, and then create a PR.

PR #3120, opened 2025-05-12
Fix NPE in compatible streaming api with Vertex AI Gemini

dev-jonghoonpark:
related issue: https://github.com/spring-projects/spring-ai/issues/3118

---

In the OpenAI API, the streaming response includes an `id`, but in the Gemini Compatible API, the streaming response does not contain an `id`.

As a result, NPE occurs in the following section.

https://github.com/spring-projects/spring-ai/blob/2d517eec5cd7ce5f88149b876ed57a06ad353e11/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java#L304-L316

To address this, I modified the implementation to assign `NO_ID` if the `id` is null.

Additionally, the **GEMINI_API_KEY** environment variable is required to run the tests.
The key can be generated from the Google Cloud Console.


sobychacko:
@dev-jonghoonpark Merged via https://github.com/spring-projects/spring-ai/commit/0c95cc396f0e739c52ca07cb0106c90f630c2a99 after some minor polishing.

Issue #1949, opened 2024-12-16
Investigate consistency in Cassandra Autoconfiguration and Builder


markpollack:
see https://github.com/spring-projects/spring-ai/pull/1912#discussion_r1882915205

PR #3122, opened 2025-05-12
GH-1949: Align CassandraVectorStore API naming with other vector stores

sobychacko:
Fixes: #1949

- Rename 'disallowSchemaChanges(boolean)' to 'initializeSchema(boolean)' for consistency with other vector store implementations
- Maintain semantic meaning by inverting the default value (from disallowSchemaChanges=false to initializeSchema=true) in the CassandraVectorStore implementation
- Keep default behavior in auto-configuration consistent with other vector stores
- Remove unused 'returnEmbeddings' functionality and related code
- Update test cases to use the new initialization parameter
- Ref docs and javadocs updates

This change improves API consistency across Spring AI vector stores while preserving the same behavior in the Cassandra implementation.


markpollack:
merged in 868e288e01f232be60d8eea0ecb38abbfec3fab8

PR #3129, opened 2025-05-13
Improve docs working with multiple ChatModels

markpollack:
- Add introduction explaining common scenarios where multiple chat models are useful.
- Restructuring the content into clear subsections:
   - Multiple ChatClients with a Single Model Type
   - ChatClients for Different Model Types
   - Multiple OpenAI-Compatible API Endpoints


tzolov:
LGTM

Issue #2954, opened 2025-04-30
[OpenSearch] Bug: Cannot connect to multiple OpenSearch hosts with credentials


linarkou:
Please do a quick search on GitHub issues first, there might be already a duplicate issue for the one you are about to create.
If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

**Bug description**
`spring-ai-opensearch-store` cannot connect to multiple hosts with basic authentication

**Environment**
Java 17, Spring AI 1.0.0-SNAPSHOT
**Steps to reproduce**
pom.xml
```
        <dependency>
            <groupId>org.springframework.ai</groupId>
            <artifactId>spring-ai-starter-vector-store-opensearch</artifactId>
        </dependency>
        <dependency>
            <groupId>org.springframework.ai</groupId>
            <artifactId>spring-ai-advisors-vector-store</artifactId>
        </dependency>
```

application.yml
```
spring:
  ai:
    vectorstore:
      opensearch:
        uris: http://localhost:9200,localhost:9201
        username: ${USERNAME}
        password: ${PASSWORD}
        index-name: ai
        initialize-schema: true
```

Main class
```
@SpringBootApplication
public class MainPlatformVSearchVectorStoreRagExample {

    public static void main(String[] args) {
        new SpringApplicationBuilder(MainPlatformVSearchVectorStoreRagExample.class)
                .web(WebApplicationType.NONE)
                .run(args)
                .close();
    }

    @Value("classpath:/rag/rag.txt")
    Resource ragDocument;

    @Bean
    public CommandLineRunner runner(ChatClient.Builder builder, VectorStore vectorStore) {
        return args -> {
            ChatClient chatClient = builder.defaultAdvisors(new QuestionAnswerAdvisor(vectorStore))
                    .build();

            List<String> documentIds = initializeVectorStore(vectorStore);

            String question = "What university John has finished?";
            String answer = chatClient.prompt().user(question).call().content();

            System.out.println(" - " + question);
            System.out.println(" - " + answer);
        };
    }

    @SneakyThrows
    private List<String> initializeVectorStore(VectorStore vectorStore) {
        String sourceText = ragDocument.getContentAsString(StandardCharsets.UTF_8);
        List<Document> documents =
                sourceText.lines().filter(s -> !s.isBlank()).map(Document::new).toList();
        vectorStore.add(documents);
        return documents.stream().map(Document::getId).toList();
    }
}
```
**Expected behavior**
[...]

PR #3137, opened 2025-05-13
HttpHost regression issues in OpenSearchVectorStoreAutoConfiguration

sobychacko:
Changes introduced for https://github.com/spring-projects/spring-ai/issues/2954 introduced a regression in the way OpenSearch works with test containers.

Restore the retrieval of HttpHost from OpenSearchConnectionDetails insted of getting them directly from properties.


markpollack:
merged in f3460926487286a7351f7fc98b0cab90b2d1a93b

PR #3007, opened 2025-05-06
[OpenSearch] Updated transport settings, fix auth to multiple hosts

linarkou:
Fixes #2954 

* fixed authorization to multiple hosts
* ability to provide SSL certs via `SslBundles`
* ability to set connectTimeout/readTimeout

PR #3140, opened 2025-05-13
Updated docs for new OpenSearch vector-store properties

linarkou:
Added docs for #3007


ilayaperumalg:
@linarkou Thanks for the PR!

PR #3156, opened 2025-05-14
Fix outdated configuration key `include-completion` and `include-query-response`

quaff:
Follow-up of commit ca843e85887aa1da6300c77550c379c103500897,`includeCompletion` is renamed to `logCompletion`.


tzolov:
Thank you @quaff 
Rebased and merged at 0bcb2e138b1ae4138790275d0e58751bad76e060

PR #3158, opened 2025-05-14
Added transaction support for saveAll in JdbcChatMemoryRepository

sunyuhan1998:
In the saveAll method of JdbcChatMemoryRepository, all historical messages of the current conversation are first deleted, and then the latest list of messages is saved. This operation should be transactional. This PR adds transaction support for it.

### DIFF
--- a/memory/repository/spring-ai-model-chat-memory-repository-jdbc/src/main/java/org/springframework/ai/chat/memory/repository/jdbc/JdbcChatMemoryRepository.java
+++ b/memory/repository/spring-ai-model-chat-memory-repository-jdbc/src/main/java/org/springframework/ai/chat/memory/repository/jdbc/JdbcChatMemoryRepository.java
@@ -20,7 +20,6 @@ import java.sql.*;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Optional;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.springframework.ai.chat.memory.ChatMemoryRepository;
@@ -33,8 +32,9 @@ import org.springframework.ai.chat.messages.UserMessage;
 import org.springframework.jdbc.core.BatchPreparedStatementSetter;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.RowMapper;
-import org.springframework.jdbc.datasource.DataSourceUtils;
+import org.springframework.jdbc.datasource.DataSourceTransactionManager;
 import org.springframework.lang.Nullable;
+import org.springframework.transaction.support.TransactionTemplate;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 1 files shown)


memory/repository/spring-ai-model-chat-memory-repository-jdbc/src/main/java/org/springframework/ai/chat/memory/repository/jdbc/JdbcChatMemoryRepository.java
```diff
@@ -83,9 +82,31 @@ public void saveAll(String conversationId, List<Message> messages) {
 		Assert.hasText(conversationId, "conversationId cannot be null or empty");
 		Assert.notNull(messages, "messages cannot be null");
 		Assert.noNullElements(messages, "messages cannot contain null elements");
-		this.deleteByConversationId(conversationId);
-		this.jdbcTemplate.batchUpdate(dialect.getInsertMessageSql(),
-				new AddBatchPreparedStatement(conversationId, messages));
+
+		Connection connection = null;
```
quaff, line 86: I think it's better to use `TransacationTemplate` here.

sunyuhan1998: Of course, using a TransactionTemplate would be better, but this seems to introduce additional constructor parameter requirements for JdbcChatMemoryRepository — would that be considered a significant change?

quaff: What about `new TransactionTemplate(new DataSourceTransactionManager(jdbcTemplate.getDataSource()))` ?

sunyuhan1998: I think your point makes sense. Although there is some additional overhead caused by repeatedly creating these transaction management objects every time we save, considering the current situation, this should be a relatively balanced implementation. I have made the adjustments. Thank you!

quaff: `TransacationTemplate` should be created in constructor.

sunyuhan1998: Done

memory/repository/spring-ai-model-chat-memory-repository-jdbc/src/main/java/org/springframework/ai/chat/memory/repository/jdbc/JdbcChatMemoryRepository.java
```diff
@@ -83,9 +86,19 @@ public void saveAll(String conversationId, List<Message> messages) {
 		Assert.hasText(conversationId, "conversationId cannot be null or empty");
 		Assert.notNull(messages, "messages cannot be null");
 		Assert.noNullElements(messages, "messages cannot contain null elements");
-		this.deleteByConversationId(conversationId);
-		this.jdbcTemplate.batchUpdate(dialect.getInsertMessageSql(),
-				new AddBatchPreparedStatement(conversationId, messages));
+
+		transactionTemplate.execute(status -> {
+			try {
```
quaff, line 91: try-catch is not required here.

sunyuhan1998: You're right. I checked the source code of TransactionTemplate, and it automatically rolls back the transaction when an exception occurs. Thank you!

markpollack:
This approach won't enable participation in a surrounding outer scope transaction, I'll work on a solution.

quaff:
> This approach won't enable participation in a surrounding outer scope transaction, I'll work on a solution.

Indeed, but do we really need participating in outer scope transaction here?

markpollack:
Who knows what people will do in the wild, it is best practice in any case I feel.

markpollack:
Thanks @sunyuhan1998 merged in 0e1bb52ab79c54a7b12d7a5f63934314a8326d37

PR #3159, opened 2025-05-14
Fixed the incorrect SQL in getSelectMessagesSql of MysqlChatMemoryRepositoryDialect

sunyuhan1998:
The getSelectMessagesSql in MysqlChatMemoryRepositoryDialect seemed to incorrectly add a LIMIT at the end. This PR fixes that issue.


sunyuhan1998:
Hi @markpollack 
What do you think? I believe this was likely an oversight when the Dialect was introduced for JdbcChatMemoryRepository in the previous change.

Boone1997:
哥们 你传到哪个版本了

xiaosenho (approved):

markpollack:
Thanks.  I've added some tests.  Also ASC is default sort order so removed that.

PR #3169, opened 2025-05-15
Fixing miscellaneous checkstyle errors

sobychacko:


advisors/spring-ai-advisors-vector-store/src/main/java/org/springframework/ai/chat/client/advisor/vectorstore/VectorStoreChatMemoryAdvisor.java
```diff
@@ -50,7 +50,7 @@
  * @author Mark Pollack
  * @since 1.0.0
  */
-public class VectorStoreChatMemoryAdvisor implements BaseChatMemoryAdvisor {
+public final class VectorStoreChatMemoryAdvisor implements BaseChatMemoryAdvisor {
```
ilayaperumalg, line 53: Is this by design that we don't want to allow any customization?

ilayaperumalg:
Rebased and merged as 368be3a04

PR #3177, opened 2025-05-15
fix(memory): update message ordering in MySQL chat memory repository

Ahoo-Wang:
- Change the message ordering from descending to ascending in the MySQL chat memory repository
- Remove the LIMIT clause to return all messages for a conversation


markpollack:
Thanks, this has been fixed in this commit 31da8f3dbb63909bab52c9855ea854a553a3349d

PR #3188, opened 2025-05-15
refactor: remove redundant Builder.samples() method in StabilityAiImageOptions

JHZLO:
### Summary

This PR removes the `samples(Integer samples)` method from the `StabilityAiImageOptions.Builder` class, as it duplicates the existing `N(Integer n)` method.

### Motivation

- Both `samples()` and `N()` internally call `setN()`, which sets the same field (`samples` JSON property).
- Having both methods introduces unnecessary redundancy and may cause confusion for users.
- `N()` is already consistent with the `ImageOptions.getN()` interface method and commonly used throughout the codebase.

### Additional Notes

If needed, developers can still set the `samples` value using the `N()` method:

```java
StabilityAiImageOptions options = StabilityAiImageOptions.builder()
    .N(3)
    .build();
```


sobychacko:
Merged via https://github.com/spring-projects/spring-ai/commit/8a5f68d655608ff4adf11e659c78f61e118c9820.

PR #3194, opened 2025-05-15
Fix VectorStoreDocumentRetriever to handle Filter.Expression objects

### LINKED ISSUE #3179: `VectorStoreDocumentRetriever` does not use the `Filter.Expression` object
**Current Behavior**

When supplying a `Filter.Expression` to a ChatClient in a prompt like the code below:
```
chatClient.prompt()
          .user(userQuery)
          .advisors(retrievalAugmentationAdvisor)
          .advisors(a -> a.param(VectorStoreDocumentRetriever.FILTER_EXPRESSION, new Filter.Expression(...)))
          .call();
```

Inside `VectorStoreDocumentRetriever`, in line 92 method `computeRequestFilterExpression`
```
private Filter.Expression computeRequestFilterExpression(Query query) {
	var contextFilterExpression = query.context().get(FILTER_EXPRESSION);
	if (contextFilterExpression != null && StringUtils.hasText(contextFilterExpression.toString())) {
		return new FilterExpressionTextParser().parse(contextFilterExpression.toString());
	}
	return this.filterExpression.get();
}
```
It finds the contextFilterExpression but assumes that it should be a string and tries to parse it to a FilterExpression using `FilterExpressionTextParser`, while in fact it could already be a `Filter.Expression` and no conversion is needed, and now it is throwing an exception:
`org.springframework.ai.vectorstore.filter.FilterExpressionTextParser$FilterExpressionParseException: Source: <unknown>, Line: 1:11, Error: no viable alternative at input 'Group['`

**Expected Behavior**
It would be nice if it checks if `contextFilterExpression` is already a `Filter.Expression` and skips parsing and just use that object.

markpollack:
- Updated computeRequestFilterExpression to check if the context value is already a Filter.Expression object before attempting to parse it as a string
- Added docs for FILTER_EXPRESSION key that it accepts both String and Filter.Expression
- Added test

Fixes #3179

### DIFF
--- a/spring-ai-rag/src/main/java/org/springframework/ai/rag/retrieval/search/VectorStoreDocumentRetriever.java
+++ b/spring-ai-rag/src/main/java/org/springframework/ai/rag/retrieval/search/VectorStoreDocumentRetriever.java
@@ -45,6 +45,11 @@ import org.springframework.util.StringUtils;
  * List<Document> documents = retriever.retrieve(new Query("example query"));
  * }</pre>
  *
+ * <p>
+ * The {@link #FILTER_EXPRESSION} context key can be used to provide a filter expression
+ * for a specific query. This key accepts either a string representation of a filter
+ * expression or a {@link Filter.Expression} object directly.
+ *
  * @author Thomas Vitale
  * @since 1.0.0
  */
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
Rebased and merged as c0b9240c8

Issue #3160, opened 2025-05-14
CassandraFilterExpressionConverter bug with `ExpressionType.IN` with collections


soheilrahsaz:
When using a filter expression on a collection like this
`new Filter.Expression(Filter.ExpressionType.IN, new Filter.Key("currencies"), new Filter.Value(List.of("BTC")));`

In CassandraFilterExpressionConverter, it will go through `doListValue` and then tries to encode each value using `doValue`, But in `doValue` method, it is trying to get the codec for the column itself, which is a list, and then fails to encode the string value in the list and throws exception:
`java.lang.ClassCastException: class java.lang.String cannot be cast to class java.util.List (java.lang.String and java.util.List are in module java.base of loader 'bootstrap')`

It seems that for collections, it should use the `elementType` of the `dataType` and not the dataType of the column itself. (CassandraFilterExpressionConverter line 124)

![Image](https://github.com/user-attachments/assets/6247755d-851d-4584-8e52-5f2da306fd2b)

sobychacko:
@soheilrahsaz PR submitted to address this specific issue here: https://github.com/spring-projects/spring-ai/pull/3197.

Just wanted to clarify a limitation we're dealing with in Cassandra's query language. When we tried using the IN operator on collection columns (like currencies IN ('BTC')), Cassandra throws an error because it fundamentally doesn't support this syntax. In Cassandra, you can't use IN to check if a collection contains specific values - instead, you'd need to use the CONTAINS operator with the syntax flipped around ('BTC' IN currencies). Our fix here prevents Java exceptions when building the query, but to fully support filtering on collection columns, we'd need to implement proper support for the [CONTAINS operator in the Spring AI Cassandra vector store](https://github.com/spring-projects/spring-ai/blob/main/vector-stores/spring-ai-cassandra-store/src/main/java/org/springframework/ai/vectorstore/cassandra/CassandraFilterExpressionConverter.java#L64). For now, Cassandra will correctly reject these queries with a clear error message rather than causing application errors.

soheilrahsaz:
Well, it needs to be `currencies CONTAINS ('BTC')` as the documents say: https://cassandra.apache.org/doc/latest/cassandra/developing/cql/indexing/sai/sai-working-with.html
And the `Contains_key` would be for map types. =

sobychacko:
Ya, thanks for that clarification. Regardless, that needs to wait until we properly implement the `CONTAINS` operator in `CassandraVectorStore`.

PR #3197, opened 2025-05-16
GH-3160: Handle element types correctly in CassandraFilterExpressionC…

sobychacko:
…onverter.doValue

Fixes: 3160

https://github.com/spring-projects/spring-ai/issues/3160

When using a filter expression with IN operator on a collection field in CassandraVectorStore.similaritySearch, a ClassCastException was thrown because the code attempted to format individual collection elements using the collection's codec instead of the element type's codec.

This fix modifies doValue to detect when we are formatting elements inside a collection type and use the appropriate element type codec. While Cassandra does not support using the IN operator directly on collection columns, this fix ensures we generate syntactically correct CQL rather than throwing a Java exception.

The change specifically addresses ListType collections by using the element type codec for individual elements within the list.


ilayaperumalg:
Rebased and merged as fa8f24633

PR #3204, opened 2025-05-16
Introduce default methods for JdbcChatMemoryRepositoryDialect

quaff:


markpollack:
Thanks, we will address this in 1.0.x release.

memory/repository/spring-ai-model-chat-memory-repository-jdbc/src/main/java/org/springframework/ai/chat/memory/repository/jdbc/HsqldbChatMemoryRepositoryDialect.java
```diff
@@ -18,27 +18,10 @@
...
  * HSQLDB-specific SQL dialect for chat memory repository.
+ *
+ * @author Mark Pollack
+ * @author Yanming Zhou
  */
 public class HsqldbChatMemoryRepositoryDialect implements JdbcChatMemoryRepositoryDialect {
 
-	@Override
-	public String getSelectMessagesSql() {
-		return "SELECT content, type FROM SPRING_AI_CHAT_MEMORY WHERE conversation_id = ? ORDER BY timestamp ASC";
```
essien, line 26: I noticed the table name `SPRING_AI_CHAT_MEMORY` is in uppercase. Is there a specific reason for this, or would it be okay to change it to lowercase (e.g., `spring_ai_chat_memory`) to improve readability?

memory/repository/spring-ai-model-chat-memory-repository-jdbc/src/main/java/org/springframework/ai/chat/memory/repository/jdbc/JdbcChatMemoryRepositoryDialect.java
```diff
@@ -20,29 +20,49 @@
...
  */
 public interface JdbcChatMemoryRepositoryDialect {
 
 	/**
 	 * Returns the SQL to fetch messages for a conversation, ordered by timestamp, with
 	 * limit.
 	 */
-	String getSelectMessagesSql();
+	default String getSelectMessagesSql() {
+		return "SELECT content, type FROM SPRING_AI_CHAT_MEMORY WHERE conversation_id = ? ORDER BY "
```
essien, line 34: I noticed the table name `SPRING_AI_CHAT_MEMORY` is in uppercase. Is there a specific reason for this, or would it be okay to change it to lowercase (e.g., `spring_ai_chat_memory`) to improve readability?

ericbottard:
After careful consideration, we've decided not to merge this PR.
Default methods are useful for expressing implementations in terms of other (abstract) methods.
This PR tries to over-generalize the concept of Dialect a bit too much (it's ok to have some repetition between implementations, and as a matter of fact if users need to customize they can (and need) to provide their own implementation).

Issue #2146, opened 2025-01-29
Allow BedrockProxyChatModel to use AWS default region configuration


yargna:
Currently, `BedrockProxyChatModel` in `spring-ai-bedrock-converse` does not automatically pick up the AWS region from standard AWS configuration mechanisms, such as:

* Environment variable (`AWS_REGION`)
* AWS shared config file (`~/.aws/config`)
* System properties (`-Daws.region=us-west-2`)
* Instance metadata service (IMDS) (EC2/ECS)

Instead, it defaults to `us-east-1`, unless explicitly overridden via `.region(Region.of(...))`.

Please update AWS region retrieval in `BedrockProxyChatModel` to ensure consistency with other AWS SDK clients. This is especially important for services that are already running in AWS environment.

**Expected Behavior**

The following code should properly pick up the region from available sources similar to other AWS clients:

```java
@Bean
public BedrockProxyChatModel bedrockChatModel() {
    return BedrockProxyChatModel.builder().build();
}
```

**Current Behavior**

`BedrockProxyChatModel.builder().build()` overrides the region with `us-east-1`.

**Context**

I want to integrate AWS Bedrock seamlessly, just like other AWS services, with minimal configuration. Other AWS SDK clients pick up the configuration automatically from the environment, which is crucial when running in AWS environments where overriding variables is discouraged. This ensures better interoperability and allows smooth transitions between regions or accounts without requiring manual adjustments. The expectation is that services should work out of the box with the existing AWS configuration, improving ease of deployment and consistency across different environments.

Currently I have to emulate it like this:

```java
@Bean
public BedrockProxyChatModel bedrockChatModel() {
    var defaultAwsRegionProviderChain = DefaultAwsRegionProviderChain.builder().build();

    return BedrockProxyChatModel
            .builder()
            .withRegion(defaultAwsRegionProviderChain.getRegion())
            .build();
}
```

markpollack:
@maschnetwork

markpollack:
Duplicate of https://github.com/spring-projects/spring-ai/issues/823

tzolov:
[...]

PR #3217, opened 2025-05-16
Fix: default AWS region #2146

ashakirin:
fix: #2146 getting default AWS region using DefaultAwsRegionProvider


tzolov:
Thanks @ashakirin 
Rebased and merged at 0bbccf8ca058d882ecb3a31f98fa600c5b2b3d42

Issue #3127, opened 2025-05-13
Document how to configure prompt and completion logging in the observability section


asaikali:
The docs at https://docs.spring.io/spring-ai/reference/observability/index.html don't explain that log levels on the observation handler need to be set to debug to see the completion and prompt show up in the logs. 

Example below shows the collection of settings needed to print out the chats and prompts. 

```yaml
spring:
   ai:
    chat:
      observations:
        log-prompt: true
        log-completion: true
logging:
  level:
    org:
      springframework:
        ai:
          chat:
            observation: DEBUG
            client:
              observation: DEBUG
```

Example application at https://github.com/asaikali/spring-ai-zero-to-hero/tree/main/applications/provider-openai

ThomasVitale:
I wonder if we should log at the `info` level instead of `debug`. If I explicitly configure the inclusions of prompts and completions in the logs, it's a bit counter-intuitive that I don't see any logs until I change the logger level. What do you think?

quaff:
> I wonder if we should log at the `info` level instead of `debug`. If I explicitly configure the inclusions of prompts and completions in the logs, it's a bit counter-intuitive that I don't see any logs until I change the logger level. What do you think?

I agree, I tried but didn't see any log.

asaikali:
@ThomasVitale I agree, I think we should change this to log at the info level, it so much easier that way. Also we are leaking internal implementation details when we use the logger for the observation handlers.

jonatan-ivanov:
I think if this feature would live in a vacuum, `DEBUG` would be the right level for it because of the nature of data that it logs out. On the other hand, I agree that since this is explicitly enabled by the users, so probably it makes more sense to log on `INFO` level so that users don't need to change the log level if they use the default behavior of Spring Boot.

I created https://github.com/spring-projects/spring-ai/pull/3218 for this.

jonatan-ivanov:
Superseded by https://github.com/spring-projects/spring-ai/pull/3218

PR #3218, opened 2025-05-16
Change log level from DEBUG to INFO for ObservationHandlers

jonatan-ivanov:
The features are disabled by default and users need to explicitly enable them. Probably it makes more sense to log on INFO level so that users don't need to change the log level if they use the default behavior of Spring Boot.

See gh-3003
See gh-3127


tzolov (approved):
LGTM, thanks @jonatan-ivanov

Issue #886, opened 2024-06-18
Ability to specify Opensearch `pathPrefix` using configuration properties


quachc:
**Expected Behavior**
Ability to specify `pathPrefix` using configuration properties (similar to `spring.elasticsearch.path-prefix`)

**Current Behavior**
Have to configure a `OpenSearchClient` bean to set `pathPrefix`

**Context**
Opensearch server is running behind a reverse proxy with a path other then `/`

PR #3221, opened 2025-05-17
GH-886: feat(opensearch): Add path prefix configuration support

sobychacko:
Fixes: #886

Add ability to configure path prefix for OpenSearch API endpoints via properties. This allows connecting to OpenSearch instances running behind a reverse proxy with a non-root path, similar to Spring Elasticsearch's path-prefix property.

- Add pathPrefix property to OpenSearchVectorStoreProperties
- Apply pathPrefix to OpenSearchClient when configured
- Add documentation and unit tests


ilayaperumalg:
Rebased and merged as 8b8fb4f02

Issue #3223, opened 2025-05-17
Change the BedrockAwsConnectionProperties  region property to be empty by default


tzolov:
If the BedrockAwsConnectionProperties#region by default the auto-configuration will try to pick the default region using the `DefaultAwsRegionProviderChain.builder().build()`.

Related to #2146 and #3217

Issue #823, opened 2024-06-04
SpringAI does not follow the AWS SDK's default region resolution rules


kennyk65:
**Bug description**
When using any of the AWS / Bedrock models, the documentation states "The region property is compulsory."  But this should not be the case since SpringAI uses the AWS SDK for Java, and it provides a default region resolver.  This will automatically look in AWS's configuration files such as ~.aws/config, which is where AWS developers would often set the region, especially if profiles are involved. 
 See https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/region-selection.html

**Environment**
Currently observed in SpringAI 1.0.0, any model.

**Steps to reproduce**
Configure a local environment for AWS usage with the "aws configure" command, setting any region other than us-east-1.  Write code to use any Bedrock model, do not set the region using "spring.ai.bedrock.aws.region".  Enable your bedrock models in the same region configured above.  Run and get a permission error, because SpringAI has not followed the AWS Java SDK's default behavior.

**Expected behavior**
I would expect SpringAI, using the AWS SDK for Java, to follow the same (or similar) region resolution rules, certainly not omitting the shared config file at ~.aws/config

**Minimal Complete Reproducible example**
See above for steps to reproduce

maxjiang153:
You can use Spring AI and  [Spring Cloud AWS ](https://spring.io/projects/spring-cloud-aws) together to manage IAM identity and region selection.

Spring AI is fully compatible with Spring Cloud AWS; besides region resolution, it also provides a variety of IAM authentication mechanisms, such as Instance profile, STS assume roles, etc.

markpollack:
We should add more information to the documentation.

markpollack:
This issue is also in https://github.com/spring-projects/spring-ai/issues/2146  closing this one to help with issue mgmt.

PR #3225, opened 2025-05-17
Make `BedrockAwsConnectionProperties#region` empty by default

therepanic:
If our region is empty, then `DefaultAwsRegionProviderChain` will be used, which will search for the region in every possible way. This should be better for us, instead of setting the default field ourselves.

Resolves: #3223


ilayaperumalg:
@therepanic Thanks for the PR! Could you update the documentation for the change (removing the default)?

therepanic:
Thanks, @ilayaperumalg. Sorry, I accidentally formatted the wrong changes and closed the PR.

I changed the documentation of the region field (to make it clear that if it is empty, `DefaultAwsRegionProviderChain` will figure out which region to use).

Also, I have now specified the region in the tests to make the tests pass. Everything should be fine now.

ilayaperumalg:
@therepanic Thank you for the quick follow up!

markpollack:
@ashakirin what do you think?

tzolov:
Mind, that removing the default region is a breaking change. It has to come at 1.1.x

ashakirin:
@markpollack, @tzolov: basically this change makes sense in combination with #823 fix, but it can break existing code based on fixed default region (Bedrock "us-east-1" has the most comprehensive model selection). Plan for next minor release?

markpollack:
I think this PR is asking for the opposite of #823 and #2146.

To clarify the current behavior in the code base.  The `BedrockProxyChatModel.Builder` uses `DefaultAwsRegionProviderChain` to get the region.  If an exception is thrown, then we log it (poorly) and default the region to be us-east-1.

That seems like reasonable behavior.  What is wrong here?

@ashakirin @therepanic

Issue #3234, opened 2025-05-18
Why toolCallArguments cannot be null or empty?


lanyuanxiaoyao:
**Bug description**
A simple toolcalling for asking current datetime, but i receive `toolCallArguments cannot be null or empty`

**Environment**
1.0.0-RC1
spring-ai-starter-model-openai

**Steps to reproduce**
```
@Slf4j
public class DatetimeTools {
  private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");

  @Tool(description = "Query current date and time")
  public String getCurrentDatetime() {
    String currentDatetime = LocalDateTime.now().format(formatter);
    return currentDatetime;
  }
}
```
```
client.prompt()
        .user("What time is it now? ")
        .tools(new DatetimeTools())
        .stream()
        .content()
        .subscribe(log::info);
```

**Expected behavior**
It's ok when i use sync calling, but fail on stream calling. Anything i missed?

**Minimal Complete Reproducible example**

ThomasVitale:
@lanyuanxiaoyao I tried running your same example, but it works correctly for me. Still, in order to be resilient to situations where a model implementation uses null `toolCallArguments` instead of `{}`, I relaxed the constraint and handled it internally in `ToolCallingObservationContext`. See https://github.com/spring-projects/spring-ai/pull/3236

Thanks for reporting this issue!

lanyuanxiaoyao:
> [@lanyuanxiaoyao](https://github.com/lanyuanxiaoyao) I tried running your same example, but it works correctly for me. Still, in order to be resilient to situations where a model implementation uses null `toolCallArguments` instead of `{}`, I relaxed the constraint and handled it internally in `ToolCallingObservationContext`. See [#3236](https://github.com/spring-projects/spring-ai/pull/3236)我尝试运行你的相同示例，但它对我来说工作正常。尽管如此，为了能够适应模型实现使用 null `toolCallArguments` 而不是 `{}` 的情况，我放宽了约束并在 `ToolCallingObservationContext` 内部处理了它。参见 [#3236](https://github.com/spring-projects/spring-ai/pull/3236)
> 
> Thanks for reporting this issue!感谢您报告此问题！

Wow, thanks for your quick response to this issue. I originally thought this issue had nothing to do with the model. In fact, I am using vLLM + Qwen3 0.6B for test.

ThomasVitale:
[...]

PR #3236, opened 2025-05-18
Relaxed toolCallArguments validation

ThomasVitale:
There seems to be some models not handling correctly the non-nullability of toolCallArguments when defining a tool call. For example, the Ollama implementation is always null-safe (if no tool argument, then '{}' is used). Other implementations might produce a null value.

This PR relaxes the toolCallArguments validation to support null toolCallArguments within ToolCallingObservationContext.

Fixes gh-3234


markpollack:
merged in 46491c042686cc2aa7063d447efcabbe735ea4f0

PR #3241, opened 2025-05-18
RAG - Document joiner should sort by score

ThomasVitale:
The current implementation of `ConcatenationDocumentJoiner` should sort the final document list by score in descending order, so to keep the most relevant documents at the front of the list. This PR fixes that.


markpollack:
merged in f9f71a805da1efbc43b546b6e2afc4ab9283f8d2

PR #3244, opened 2025-05-18
MCP ToolCallbackProvider autoconfig should use concrete class

ThomasVitale:
Currently, the MCP Tool autoconfiguration defines the `SyncMcpToolCallbackProvider` and `AsyncMcpToolCallbackProvider` beans returning the `ToolCallbackProvider` interface. That generates autowiring warnings in IDEs when trying to inject the specific class.

Screenshot 2025-05-18 at 19 25 37

The Spring Boot project recommends being specific:

> When declaring a @Bean method, provide as much type information as possible in the method’s return type. For example, if your bean’s concrete class implements an interface the bean method’s return type should be the concrete class and not the interface. Providing as much type information as possible in @Bean methods is particularly important when using bean conditions as their evaluation can only rely upon to type information that is available in the method signature.

See https://docs.spring.io/spring-boot/reference/features/developing-auto-configuration.html#features.developing-auto-configuration.condition-annotations.bean-conditions

This PR ensures concrete classes are used. That's also in line with what included in the [documentation](https://docs.spring.io/spring-ai/reference/api/mcp/mcp-client-boot-starter-docs.html#_usage_example):

Screenshot 2025-05-18 at 19 35 47


tzolov:
Thanks @ThomasVitale 
There is small risk with List beans but lets take it

PR #3246, opened 2025-05-18
feat(openai): add support for pdf files as media in completions api

jharnack2:
Add support for media with mime type `application/pdf`, which is sent as content with type `file` in openai completions api.
Relevant openai docs: https://platform.openai.com/docs/guides/pdf-files?api-mode=chat#base64-encoded-files


ilayaperumalg:
@jharnack2 Thanks for the PR! Rebased and merged as 1f2bd0878. Also, back-ported into 1.0.x via a33045871

Issue #3168, opened 2025-05-14
1.0.0-RC1 + native image = scheduler cannot be null


mpalourdio:
Hi,

I have moved from moving from `1.0.0-M8` to `1.0.0-RC1`. A small app that successfully compiles and runs with GraalVM.

From `1.0.0-RC1`, it fails at *runtime* with 
```log
Caused by: java.lang.IllegalArgumentException: scheduler cannot be null
	at org.springframework.util.Assert.notNull(Assert.java:181)
	at org.springframework.ai.chat.client.advisor.MessageChatMemoryAdvisor.<init>(MessageChatMemoryAdvisor.java:61)
	at org.springframework.ai.chat.client.advisor.MessageChatMemoryAdvisor$Builder.build(MessageChatMemoryAdvisor.java:179)
	at com.mpalourdio.projects.mpalourd_ai.controller.OpenAiController.<init>(OpenAiController.kt:53)
	at com.mpalourdio.projects.mpalourd_ai.controller.OpenAiController__BeanDefinitions.lambda$getOpenAiControllerInstanceSupplier$0(OpenAiController__BeanDefinitions.java:25)
	at org.springframework.util.function.ThrowingBiFunction.apply(ThrowingBiFunction.java:68)
	at org.springframework.util.function.ThrowingBiFunction.apply(ThrowingBiFunction.java:54)
	at org.springframework.beans.factory.aot.BeanInstanceSupplier.lambda$get$2(BeanInstanceSupplier.java:225)
	at org.springframework.util.function.ThrowingSupplier.get(ThrowingSupplier.java:58)
	at org.springframework.util.function.ThrowingSupplier.get(ThrowingSupplier.java:46)
	at org.springframework.beans.factory.aot.BeanInstanceSupplier.invokeBeanSupplier(BeanInstanceSupplier.java:258)
	at org.springframework.beans.factory.aot.BeanInstanceSupplier.get(BeanInstanceSupplier.java:225)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.obtainInstanceFromSupplier(DefaultListableBeanFactory.java:1031)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.obtainFromSupplier(AbstractAutowireCapableBeanFactory.java:1249)
```

Forcing the scheduler works, but it should not be necessary as it's already a default in `MessageChatMemoryAdvisor#Builder` . Here is the faulty code :

```kotlin
[...]

PR #3256, opened 2025-05-19
Fix NPE caused by uninitialized static variables in interfaces in native images

dev-jonghoonpark:
related issue: https://github.com/spring-projects/spring-ai/issues/3168

The `DEFAULT_SCHEDULER` variable in the `BaseAdvisor` interface was being initialized to null when running a native image.

To address this issue, a separate `SchedulerHolder` class was created to handle the variable initialization properly and ensure it is initialized as expected.


markpollack:
I don't understand what is going on here,currently in `BaseAdvisor` there is this code 

```
public interface BaseAdvisor extends CallAdvisor, StreamAdvisor {

	Scheduler DEFAULT_SCHEDULER = Schedulers.boundedElastic();
```

Why would this field be set to null when creating a native image?

joshlong:
> hi - could you tell us against what version of Spring AI you're reporting this? and against what version of GraalVM? Which OS?
> 
> ive attempted to reproduce the bug here [1] on macOS using graalvm for java 24 and haven't yet seen the issue...
> 
> I wonder if having some sort of proxy / AOP logic is causing that reference to be `null`? Are you defining the ChatMemoryAdvisor as a bean? Is it proxied at some point? What happens if you instantiate it inline, as i have?
> 
> Alternatively: is it possible you don't have the Reactor libraries on the class path? I don't know..
> 
> 1. https://github.com/joshlong-attic/2025-05-29-spring-ai-graalvm-advisor-npe-3256

the good news is, as I just realized, is that most of the common BaseAdvisors are also `final` so they're likely not being proxied in the first place. which leaves me to wonder about Reactor, and all the other variables.

mpalourdio:
hi,

@joshlong trick works too, but forcing the scheduler as I have done as a workaround in the original issue does the trick for the moment (at least for me).  Encapsulating all those builders in `if` defeats the purpose of those fluent interfaces IMO.

I am also a bit meh by the fact that those factories are stored in the constants, but it works on the jvm so why not.

joshlong:
thanks. the only thing we're encapsulating is the non constant default value on the interface, which is initialized at build time but has to be null since there's nothing useful at build time. 

I agree - it's weird that the default is stored in an interface constant :/

PR #3259, opened 2025-05-20
Provides a Jinja2 implementation for TemplateRenderer

sunyuhan1998:
Implements TemplateRenderer with Jinjava to support Jinja2 syntax.

This can help users who are more familiar with Jinja2 syntax to adapt more easily, and also allows prompts that already use Jinja2-style templates in Python to be smoothly migrated into Spring AI.

### DIFF
--- a/spring-ai-template-jinja/src/main/java/org/springframework/ai/template/jinja/JinjaTemplateRenderer.java
+++ b/spring-ai-template-jinja/src/main/java/org/springframework/ai/template/jinja/JinjaTemplateRenderer.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


spring-ai-template-jinja/src/main/java/org/springframework/ai/template/jinja/JinjaTemplateRenderer.java
```diff
@@ -0,0 +1,172 @@
...
+	private static final ValidationMode DEFAULT_VALIDATION_MODE = ValidationMode.THROW;
+
+	private final ValidationMode validationMode;
+
+	/**
+	 * Constructs a new {@code JinjaTemplateRenderer} with the specified validation mode.
+	 * @param validationMode the mode to use for template variable validation; must not be
+	 * null
+	 */
+	public JinjaTemplateRenderer(ValidationMode validationMode) {
```
ericbottard: This constructor should be made protected

sunyuhan1998: done

PR #3264, opened 2025-05-20
feat(tools): Add configurable exception handling for tool execution

tzolov:
Adds a new configuration property `spring.ai.tools.throw-exception-on-error` that controls how tool execution errors are handled:
- When false (default): errors are converted to messages and sent back to the AI model
- When true: errors are thrown as exceptions for the caller to handle

The implementation:
- Adds the property to ToolCallingProperties
- Updates ToolCallingAutoConfiguration to use the property
- Improves error handling in MCP tool callbacks to use ToolExecutionException
- Adds tests to verify both behaviors
- Updates documentation with the new property


ilayaperumalg:
LGTM. In the ToolCallingAutoConfiguration, we can perhaps use the builder method in `DefaultToolExecutionExceptionProcessor` to set the `alwaysThrow` boolean value instead of setting at the constructor level. But this can be addressed separately as we may also need to decide whether to deprecate the existing public constructor for `DefaultToolExecutionExceptionProcessor`. Merging.

ilayaperumalg:
Rebased and merged as 6f61fee77.

ilayaperumalg:
Back ported to 1.0.x via 3068c04dc

Issue #3266, opened 2025-05-21
Prompt.augmentSystemMessage will result in adding two system prompts.


leooolee:
Please do a quick search on GitHub issues first, there might be already a duplicate issue for the one you are about to create.
If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

**Bug description**
A clear and concise description of what the bug is about.

![Image](https://github.com/user-attachments/assets/ca13550e-42c3-4e3b-8823-20c440aea9dd)

**Environment**
Please provide as many details as possible: Spring AI version, Java version, which vector store you use if any, etc
JDK21 + Spring ai 1.0.0 

**Steps to reproduce**
Steps to reproduce the issue.
1. There is a system prompt, and the system prompt is not the first message.
2. Augment the system message

**Expected behavior**
Add only one system prompt.

sunyuhan1998:
I have reproduced your issue, and it is indeed as you described. I have submitted a PR to fix this issue: https://github.com/spring-projects/spring-ai/pull/3268

ilayaperumalg:
Closing this as this was addressed via #3268. Thanks @sunyuhan1998 for the resolution.

PR #3268, opened 2025-05-21
Fixed a bug in the augmentSystemMessage method of the Prompt.

sunyuhan1998:
As mentioned in issue: https://github.com/spring-projects/spring-ai/issues/3266, when executing the `augmentSystemMessage` method of the Prompt, an extra system message would be incorrectly added if the system message was not the first one in the message list. This PR fixes that issue.


sunyuhan1998:
Hi @ThomasVitale , what do you think? Is this a bug?

ilayaperumalg:
Squashed, rebased and merged as 902fc0963955aec2338018e2925f6d0c72b10d99. Also, back-ported into 1.0.x via 623e7053366480ab64a0be47c7c8c14a27b2fe94

PR #3271, opened 2025-05-21
Bump org.apache.pdfbox to 3.0.4

dafriz:
Resolves dependency convergence error with tika version of pdfbox

Fixes 3265

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -280,7 +280,7 @@
 		<protobuf-java.version>3.25.2</protobuf-java.version>
 
 		<!-- readers/writer/stores dependencies-->
-		<pdfbox.version>3.0.3</pdfbox.version>
+		<pdfbox.version>3.0.4</pdfbox.version>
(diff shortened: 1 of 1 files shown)


dafriz:
Note that the update to PDF Box 3.0.4 also needs the guard against empty unicode strings change otherwise the existing test case fails.

pom.xml
```diff
@@ -280,7 +280,7 @@
 		<protobuf-java.version>3.25.2</protobuf-java.version>
 
 		<!-- readers/writer/stores dependencies-->
-		<pdfbox.version>3.0.3</pdfbox.version>
+		<pdfbox.version>3.0.4</pdfbox.version>
```
GrantGochnauer, line 283: Is it worth trying to go to 3.0.5?

dafriz: 3.0.4 resolves the current maven convergence issue.

Tika 3.2.0 is due any day now which uses 3.0.5 so would reccomend waiting a while and then updating them both together.

PR #3279, opened 2025-05-21
Update mcp-server-boot-starter-docs.adoc

yangbuyiya:
Older STDIO transports can't be pulled, this should be a milestone version and should use `spring-ai-starter-mcp-server`.


ilayaperumalg:
Back ported the changes into 1.0.x via ea38612fe

PR #3284, opened 2025-05-21
fix: correct visibility scope issues in Spring AI components

sobychacko:
- Change return types from implementation to interface in AutoConfiguration beans
  - OllamaConnectionDetails in OllamaApiAutoConfiguration
  - WeaviateConnectionDetails in WeaviateVectorStoreAutoConfiguration

- Fix visibility modifiers for:
  - Make Aws class public in OpenSearchVectorStoreProperties
  - Set OpenAIAutoConfigurationUtil.resolveConnectionProperties to package-private
  - Change static NumericBoundary constants to private in RedisFilterExpressionConverter
  - Explicitly mark metadataFields as final in RedisFilterExpressionConverter

These changes ensure components are properly encapsulated and follow API design best practices by exposing only what's necessary through interfaces rather than implementations.


auto-configurations/models/spring-ai-autoconfigure-model-ollama/src/main/java/org/springframework/ai/model/ollama/autoconfigure/OllamaApiAutoConfiguration.java
```diff
@@ -42,7 +42,7 @@ public class OllamaApiAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean(OllamaConnectionDetails.class)
-	public PropertiesOllamaConnectionDetails ollamaConnectionDetails(OllamaConnectionProperties properties) {
```
sobychacko, line 45: `PropertiesOllamaConnectionDetails` is a package-protected inner class that was marked as the return type from a public API method. Although the compiler works with that, the IDE throws a visibility issue warning.

auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-opensearch/src/main/java/org/springframework/ai/vectorstore/opensearch/autoconfigure/OpenSearchVectorStoreProperties.java
```diff
@@ -144,7 +144,7 @@ public void setAws(Aws aws) {
 		this.aws = aws;
 	}
 
-	static class Aws {
```
sobychacko, line 147: We provide public accessor/setter for this although this is a package-protected inner class. Since we can't change those public API's, we should make this inner `AWS` class as public.

vector-stores/spring-ai-redis-store/src/main/java/org/springframework/ai/vectorstore/redis/RedisFilterExpressionConverter.java
```diff
@@ -38,11 +38,11 @@
  */
 public class RedisFilterExpressionConverter extends AbstractFilterExpressionConverter {
 
-	public static final NumericBoundary POSITIVE_INFINITY = new NumericBoundary(Double.POSITIVE_INFINITY, true);
```
sobychacko, line 41: These public constants cause visibility issues since the underlying types are package-protected.

sobychacko:
Closing this PR as these changes may not be relevant since the codebase has evolved further since the PR.

PR #3292, opened 2025-05-22
Fix tools docs

eddumelendez:
I updated project from 1.0.0-M7 to 1.0.0 and noticed I should be using `toolNames` instead of `tools`. See [commit](https://github.com/eddumelendez/spring-ai-dmr/commit/7830892df4bf179991d7ad4dc03115893e723ad9)


ilayaperumalg:
Back ported the changes into 1.0.x as cc5379344

PR #3298, opened 2025-05-23
fix(deepseek): prevent incorrect ToolCall merging caused by empty id and name strings

ggndnn:
Fix tool call merging issue during deepseek stream processing, same as #2589.


mxsl-gr:
Hi, @ggndnn Thanks for the PR!
I’ve done quite a few tests, and both `deepseek-chat` and `ZhiPuAI` in stream mode don’t seem to return an empty string for the tool call id.
It seems like this issue might be specific to `OpenAI`, If you have a minimal reproducible scenario, that would be even better!

mxsl-gr:
> @mxsl-gr Hi, In order to obtain the reasoning_content while also utilizing the tool calling feature, I believe the DeepSeekApi is capable of doing so, so I used it to access aliyun models such as `qwen3-235b-a22b`, then got the issue.

You’re right! I tested it with Qwen3 and saw the same issue.
It related to the model’s default return values. From a defensive programming perspective, I think this merge change doesn’t hurt and can be helpful.

ilayaperumalg:
@ggndnn Thanks for the PR. Rebased and merged as d61192a63. Back-ported into 1.0.x as well.

ilayaperumalg:
Related fix at the OpenAI side: https://github.com/spring-projects/spring-ai/pull/3608

PR #3304, opened 2025-05-23
Fix PropmptTemplate typo in structured-output-api.jpg

### LINKED ISSUE #3295: Diagram structured-output-api PropmptTemplate Typo
Hello everyone, 
I would like to point out a spelling mistake on the [structured-output-api.jpg](https://github.com/spring-projects/spring-ai/blob/fbfc87e814c6d5d2382d7e5f9b045246731a10dd/spring-ai-docs/src/main/antora/modules/ROOT/images/structured-output-api.jpg) image. There the ``PromptTemplate`` is called ``PropmptTemplate``.

Image

Skillkiller:
Since I couldn't find a source file, I rearranged the existing letters with Gimp and re-centered the text. I don't think the result is pixel perfect but I don't think it's visible.

Fix #3295


ilayaperumalg:
@Skillkiller Thanks for fixing the image.

PR #3305, opened 2025-05-23
Add support for max_completion_tokens in AzureOpenAI chat options request and update the test and document.

iAMSagar44:
The maxTokens value is now deprecated in favor of max_completion_tokens, and is not compatible with o1 series models.

Reference - https://learn.microsoft.com/en-us/java/api/com.azure.ai.openai.models.chatcompletionsoptions?view=azure-java-preview#com-azure-ai-openai-models-chatcompletionsoptions-setmaxtokens(java-lang-integer)


markpollack:
Hi, thanks for the PR.  We can't introduce a breaking change in our public API (e.g. by removing `getMaxTokens()`  and similar accessors.)   I think this should be handled in a more subtle way, keeping the current API but under the covers passing the right value into the azure sdk.

PR #3308, opened 2025-05-23
build: Use spring-web instead of spring-webflux in spring-ai-retry

### LINKED ISSUE #3307: spring-ai-retry should not require spring-webflux dependency
**Bug description**
`spring-ai-retry/pom.xml` contains `spring-webflux`. This is an invasive library (conflicts with spring-webmvc) and should not be added as a dependency without a good reason. There is no reason: no class from spring-webflux is imported. There is one import of `org.springframework.web.client.ResponseErrorHandler`, but it's provided by much smaller `spring-web`. So the dependency should be on spring-web instead.

**Environment**
Spring AI 1.0.0 Release

**Steps to reproduce**
Check spring-ai-retry/pom.xml

**Expected behavior**
No spring-webflux dependency

N/A (dependencies issue)

pkubowicz:
The dependency on spring-webflux was unnecessary, there was no code inside spring-ai-retry that made use of it.

Also remove 'optional' from slf4j-api dependency (it's not optional, if you don't have it on classpath, RetryUtils will fail to load).

Fixes #3307

### DIFF
--- a/models/spring-ai-anthropic/pom.xml
+++ b/models/spring-ai-anthropic/pom.xml
@@ -78,6 +78,11 @@
 		</dependency>
 
 		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-webflux</artifactId>
+		</dependency>
+
+		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
--- a/models/spring-ai-deepseek/pom.xml
+++ b/models/spring-ai-deepseek/pom.xml
@@ -41,6 +41,11 @@
 		</dependency>
 
 		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-webflux</artifactId>
+		</dependency>
+
+		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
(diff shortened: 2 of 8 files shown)


bog-sb (approved):

markpollack:
Thanks, great catch.

ilayaperumalg:
@pkubowicz Thanks for the PR! As you can see from the compilation issues with the build, the dependency web-flux is being used by a few other modules via spring-ai-retry. It appears that we need a better strategy in terms of defining common vs specific dependencies for the individual modules.

pkubowicz:
I don't think 'retry' module should be the place to decide on common dependencies. Especially, the decision on the web client library used should not be in a helper/tooling module like 'retry'.

Being explicit about dependencies is a well-known rule to build reusable publicly available libraries. Is it worth a discussion? To me, the fix is straightforward, and is to make the dependency on webflux explicit in each place using it. I was surprised the rule is not consequently applied to this repository, hence the compilation failure. But the fix should be easy, and if you don't have anything against, I will add it to my PR.

ilayaperumalg:
> I don't think 'retry' module should be the place to decide on common dependencies. Especially, the decision on the web client library used should not be in a helper/tooling module like 'retry'.

I agree as that's exactly what you fixed in this. 

> Being explicit about dependencies is a well-known rule to build reusable publicly available libraries. Is it worth a discussion?
I marked this to "to discuss" for the team to identify any other similar dependencies and potentially move these common dependencies to their respective common/shared module. 

Feel free to update the PR with your suggestion.

pkubowicz:
I rebased on the current `main` and added the explicit dependency on spring-webflux where compilation failed without it.

ilayaperumalg:
Rebased and merged as cecc046c2

PR #3312, opened 2025-05-24
Add unit tests for OpenAiStreamFunctionCallingHelper

essien:
Added unit tests for [OpenAiStreamFunctionCallingHelper](https://github.com/spring-projects/spring-ai/blob/39192049ff98114602212395f37e7ada31e92ba1/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiStreamFunctionCallingHelper.java#L46).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,4 +45,3 @@ shell.log
 
 **/.claude/settings.local.json
 .devcontainer
-/models/spring-ai-openai/nbproject/


.gitignore
```diff
@@ -45,3 +45,4 @@ shell.log
 
 **/.claude/settings.local.json
 .devcontainer
+/models/spring-ai-openai/nbproject/
```
sobychacko, line 48:
```suggestion
```
There is a conflict, and we are removing this line.

Issue #3253, opened 2025-05-19
Model incompatible request argument supplied: temperature


hstaudacher:
**Bug description**
When using the OpenAI search models the following error is returned: Model incompatible request argument supplied: temperature

There is no way to remove the `temperature` parameter from the OpenAiChatOptions programmatically without changing the global temperature. 

**Environment**
Spring AI latest snapshot

**Steps to reproduce**

```java
 var options = OpenAiChatOptions.builder()
      .model(MODEL)
      .build();
    Prompt prompt = new Prompt(messages, options);
    return client.prompt(prompt)
      .call()
      .chatResponse();
```

**Expected behavior**
Either allow setting the temperature to null or remove it when a search model is used.

LEEJaeHyeok97:
Hi, I'd like to work on this issue. Currently, in `OpenAiChatOptions`, there's no way to explicitly remove the `temperature` parameter, which leads to incompatibility when using search models.

As a solution, I propose:

- Changing the `temperature` field in `OpenAiChatOptions` to `Optional<Double>` to allow explicit absence of the parameter.
- Modifying the `ModelOptionsUtils.merge()` method to avoid merging default values when the field is `Optional.empty()`.
- Updating `OpenAiChatModel` to exclude the `temperature` parameter from the request when it's `Optional.empty()`.

Does this approach align with the project's direction? If so, I'll proceed to fork the repository, implement the changes, and submit a PR. Thank you!

therepanic:
Hi, @LEEJaeHyeok97.

I think I discovered a deeper issue (which I mentioned on this issue) that involves more flexible scenarios when we don't want the data from the properties to merge with internal data. That would solve probably many problems, including this one.

I think your solution solves this problem quite well, however it seems to me that labeling each field as Optional would be redundant for us and we can look at this problem from another angle as I described in #3254. It's not for me to decide either way though.

LEEJaeHyeok97:
Thanks @therepanic for the thoughtful perspective!

I agree that introducing a global flag like `isPreventDefaultMerge` can provide broader flexibility, especially for advanced use cases where merging should be suppressed entirely.

That said, I’d like to move forward with the more localized fix for the `temperature` issue first. Since this is a concrete problem with a well-defined cause, making `temperature` explicitly skippable via `Optional` seems like a practical and low-risk change for now.

Of course, if a more general solution (like the flag) is implemented later on, we could revisit and refactor to adopt that approach uniformly across all options.

[...]

PR #3315, opened 2025-05-24
fix: omit temperature for search models and add tests

LEEJaeHyeok97:
This PR fixes issue #3253 by ensuring that the `temperature` field is omitted from the request when using OpenAI search models (e.g. `text-search-*`). For chat models, it preserves the default temperature unless the user explicitly overrides it.


tzolov:
@LEEJaeHyeok97 are you truing to use the ChatModel (e.g. completion ) API for accessing Embedding Mdoles?
https://github.com/spring-projects/spring-ai/issues/3253#issuecomment-2940554696

sobychacko:
@LEEJaeHyeok97 Please see the latest comment on the issue: https://github.com/spring-projects/spring-ai/issues/3253#issuecomment-2993254977

markpollack:
I think the better solution that is overdue is to remove us setting the temperature as a default.  this way it won't be sent on the wire to the model unless the user explicitly sets it.  I'm also a bit afraid of hardcoding the model name, it's brittle and we need better metadata to handle these cases, but it maybe all we can do in the 1.0.x line.

PR #3323, opened 2025-05-26
Fix MCP protocol, related commands cannot be executed under Windows, such as npx

twelvet-s:
I cannot run bat related commands such as npx properly under Windows. I hope to be able to start it more friendly, otherwise it may be troublesome or catastrophic for cross platform development, which has caused me to be unable to configure it properly. Therefore, I hope to get a fix. Thank you.

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client/src/main/java/org/springframework/ai/mcp/client/autoconfigure/properties/McpStdioClientProperties.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client/src/main/java/org/springframework/ai/mcp/client/autoconfigure/properties/McpStdioClientProperties.java
@@ -16,9 +16,7 @@
 
 package org.springframework.ai.mcp.client.autoconfigure.properties;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
@@ -44,6 +42,10 @@ import org.springframework.core.io.Resource;
 public class McpStdioClientProperties {
 
 	public static final String CONFIG_PREFIX = "spring.ai.mcp.client.stdio";
+
+	private static final List<String> WIN_COMMAND = List.of("npx");
+
+	private static final String OS_NAME = System.getProperty("os.name").toLowerCase();
 
 	/**
 	 * Resource containing the MCP servers configuration.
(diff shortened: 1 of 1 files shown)


auto-configurations/mcp/spring-ai-autoconfigure-mcp-client/src/main/java/org/springframework/ai/mcp/client/autoconfigure/properties/McpStdioClientProperties.java
```diff
@@ -16,9 +16,7 @@
 
 package org.springframework.ai.mcp.client.autoconfigure.properties;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
markpollack, line 19: fyi - no star imports

auto-configurations/mcp/spring-ai-autoconfigure-mcp-client/src/main/java/org/springframework/ai/mcp/client/autoconfigure/properties/McpStdioClientProperties.java
```diff
@@ -128,6 +130,11 @@ public record Parameters(
 			@JsonProperty("env") Map<String, String> env) {
 
 		public ServerParameters toServerParameters() {
+			if (OS_NAME.contains("win") && WIN_COMMAND.contains(command)) {
+				List<String> winArgs = new LinkedList<>(Arrays.asList("/c", this.command()));
+				winArgs.addAll(this.args);
+				return ServerParameters.builder("cmd.exe").args(winArgs).env(this.env()).build();
```
markpollack, line 136: I believe we need to add `/c` for windows and perhaps simliar `-c` for the unix one to be more robust.   look at https://github.com/spring-attic/spring-cli/blob/efa39cd922aedcfd3b172a4bb9bfd1721c29d490/src/main/java/org/springframework/cli/command/SpecialCommands.java#L59

auto-configurations/mcp/spring-ai-autoconfigure-mcp-client/src/main/java/org/springframework/ai/mcp/client/autoconfigure/properties/McpStdioClientProperties.java
```diff
@@ -45,6 +43,10 @@ public class McpStdioClientProperties {
 
 	public static final String CONFIG_PREFIX = "spring.ai.mcp.client.stdio";
 
+	private static final List<String> WIN_COMMAND = List.of("npx");
```
ericbottard, line 46: This looks like a code smell to me.
There shouldn't be specifically hardcoded list of programs that are supported (and hence some that are not)

markpollack: yes indeed.  There can be any command not just npx (though that is common) that needs to be passed into the windows (and unix ) shell.

markpollack:
I've verfied how to run this on windows without any code changes.  you can add 'cmd.exe' into the json config or programmatically.  If we want this to be more portable, with 'cmd.exe' baked into the json it will need to change in the java mcp sdk.  @tzolov thoughts?

https://github.com/spring-projects/spring-ai-examples/commit/57add8153f134bbefb43f582243b3d6434c40db2  shows the changes in a sample app for linux/windows, programmatic/declarative usage in the mcp client app.  ref docs have been updated.  see https://github.com/spring-projects/spring-ai/commit/e4642663c87dd90f338359d01d96b79543147efc

thanks for pointing this out @twelvet-s

PR #3325, opened 2025-05-26
Add metadata support for vector store types

m1ngyuan:
Add metadata support for vector store types


ilayaperumalg:
@m1ngyuan Thanks for the PR!

PR #3329, opened 2025-05-26
Remove redundant validation call in getToolCallbacks method

sunghyuk:
Remove redundant `validateToolCallbacks()` call as validation is already performed in the constructor


ilayaperumalg:
Hi @sunghyuk Thanks for the PR. But, the validation logic you removed checks for duplicate tools with the same name. Please see the test case in MethodToolCallbackProviderTests#whenMultipleToolObjectsWithSameToolNameThenThrow .

Closing this PR now. But, please re-open if you have a different point of view here. Thanks!

sunghyukatkurly:
@ilayaperumalg Thanks for the review. 
It seems unnecessary to validate again in the getToolCallbacks() method since validation has already been performed in the constructor.
Similarly, the MethodToolCallbackProviderTests#whenMultipleToolObjectsWithSameToolNameThenThrow test code also passes the constructor's validation.

PR #3340, opened 2025-05-27
support setting a similarity threshold in VectorStoreChatMemoryAdvisor

scionaltera:
The `VectorStoreChatMemoryAdvisor` did not support setting a similarity threshold, so it was fetching everything it found including messages that were not relevant to the user's message in the case where the user has asked about a different topic.

I experimented with setting a threshold higher than 0 and liked the results so I thought I'd submit a PR. The default is still 0 so as not to change the original behavior.


scionaltera:
Closing to squash the additional commit.

PR #3341, opened 2025-05-27
support for similarity threshold in VectorStoreChatMemoryAdvisor

scionaltera:
This advisor did not have support for setting the similarity threshold, so it defaulted to returning every document regardless of similarity. In a conversation where the user has changed to a different topic the result was returning a bunch of documents with no relevance to the user's latest input.

I experimented with setting a value greater than 0 and liked the results because it would begin to include results more relevant to the conversation. I thought it would be nice to contribute the change back.

I left the default at 0 so as not to change the earlier behavior.


sobychacko:
Can you follow the DCO guidelines outlined here and resubmit the PR? https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring. Thanks!

scionaltera:
Closing and resubmitting this to comply with new DCO policy.

Issue #3339, opened 2025-05-27
ToolResponseMessage results in null content, causing SQLIntegrityConstraintViolationException in JDBC repository


astor-dev:
**Bug description**
When a ChatResponse from the model includes a ToolCall (i.e., finishReason=TOOL_CALLS), the corresponding AssistantMessage has textContent = null. 
When this message is added to the JdbcChatMemoryRepository, the implementation attempts to insert it into the database without null-checking the content field. This causes a SQL integrity error due to a NOT NULL constraint on the content column.

The ToolResponseMessage (or AssistantMessage with only toolCalls and no content) is not handled correctly in JdbcChatMemoryRepository. This leads to the following exception:

```sh
java.sql.SQLIntegrityConstraintViolationException: (conn=498) Column 'content' cannot be null
```

In contrast, the default in-memory implementation (MessageWindowChatMemory) **tolerates null content** and saving such messages with null textContent.

**Environment**
Please provide as many details as possible: Spring AI version, Java version, which vector store you use if any, etc
- spring-ai 1.0.0
- mariadb:10.11

**Steps to reproduce**
1. Use a ChatModel that returns a response containing only toolCalls (e.g., OpenAI with function/tool calling).
2. Capture that AssistantMessage and attempt to save it to JdbcChatMemoryRepository.
3. Since textContent == null, the following SQL fails:
```
INSERT INTO SPRING_AI_CHAT_MEMORY (conversation_id, content, type, timestamp)
VALUES (?, NULL, 'ASSISTANT', ?)
```
.

**Expected behavior**
The repository should store the response with null value

**Minimal Complete Reproducible example**
Here is the code that caused the bug in my project.
```
fun call(conversationId: String, prompt: Prompt, toolCallingChatOption: ToolCallingChatOptions?): ChatResponse {
        val systemMessage = prompt.systemMessage
        chatMemory.add(conversationId, prompt.userMessage)
        val promptWithMemory = Prompt(chatMemory.get(conversationId) + prompt.systemMessage, toolCallingChatOption)
        var chatResponse = chatModel.call(promptWithMemory) ?: throw ErrorCode.CHAT_NULL_RESPONSE.toException()
        chatMemory.add(conversationId, chatResponse.result.output) // throw Exception
}
```

csjbebetter:
[...]

PR #3343, opened 2025-05-27
feat: Add tool_calls support to JdbcChatMemoryRepository

### LINKED ISSUE #3342: Support persisting tool_calls in JdbcChatMemoryRepository
**Expected Behavior**

`JdbcChatMemoryRepository` should persist and restore the `tool_calls` field for `AssistantMessage`s that use tool calling (i.e., function-calling).

Specifically:
- Add a `tool_calls TEXT NULL` column to the schema
- Serialize the list of `ToolCall` objects to JSON on insert
- Deserialize and rehydrate `AssistantMessage(null, toolCalls)` when loading from the database

This ensures that responses such as:

```java
new AssistantMessage(null, toolCalls)
````

...can be stored and later reused without violating OpenAI's API requirements.

---

**Current Behavior**

Currently, `JdbcChatMemoryRepository` only stores the `content` field. When an `AssistantMessage` has `content = null` and non-empty `toolCalls`, the following happens:

* The `toolCalls` are silently discarded during persistence
* Upon loading, the message becomes: `content = null`, `toolCalls = []`
* This leads to a malformed payload when used in a follow-up prompt:

```json
{ "role": "assistant", "content": null }
```

Which violates the OpenAI API spec:

> If `content` is null, a non-empty `tool_calls` array must be present

Resulting in:

```json
  "error": {
    "message": "Invalid value for 'content': expected a string, got null.",
    "type": "invalid_request_error",
    "param": "messages.[3].content",
    "code": null
  }
```

---

**Context**

Tool calling is a core feature of OpenAI and Azure OpenAI LLMs. Without persisting `tool_calls`, JDBC-based chat memory implementations cannot reconstruct valid assistant messages, causing:

* 400 API errors
* Broken multi-turn flows
* Prompt history loss

This makes `JdbcChatMemoryRepository` incompatible with modern LLM usage patterns.

---

**Alternatives considered**

* In-memory chat memory works correctly (retains toolCalls)
* Custom JDBC memory with JSON serialization (invasive)
* Skipping such messages entirely (causes context loss)

---

**Workaround**

Avoid using `JdbcChatMemoryRepository` when tool calling is enabled. Use in-memory fallback instead.

---

**Related**

* [https://github.com/spring-projects/spring-ai/issues/3339](https://github.com/spring-projects/spring-ai/issues/3339)

astor-dev:
## Description

This PR implements support for persisting `tool_calls` in `JdbcChatMemoryRepository` to resolve issues with tool calling functionality in JDBC-based chat memory implementations.

## Issues Addressed
Fixes #3342 - Support persisting tool_calls in JdbcChatMemoryRepository

It can also serve as a preliminary step to fix the issues below:
#3339 - ToolResponseMessage results in null content, causing SQLIntegrityConstraintViolationException

### Database Schema
- Added `tool_calls` column to all database schemas (PostgreSQL, MySQL, SQL Server, HSQLDB)
- PostgreSQL uses `JSONB` type with explicit `::jsonb` casting
- Other databases use `TEXT`/`LONGVARCHAR` for JSON storage

### Implementation
- Serialize `List<AssistantMessage.ToolCall>` to JSON on insert
- Deserialize tool calls when loading from database
- Updated all dialect classes to include `tool_calls` in SELECT and INSERT queries
- Added error handling for malformed JSON with warning logs

### Database Dialects Updated
- `PostgresChatMemoryRepositoryDialect`
- `MysqlChatMemoryRepositoryDialect`
- `SqlServerChatMemoryRepositoryDialect`
- `HsqldbChatMemoryRepositoryDialect`


markpollack:
Thanks for this, I need to spend more time reviewing this, but in the 1.0.x branch we won't be able to merge any schema breaking changed, we can for 1.1.x  No firm timeline on the 1.1.x release date yet, just starting to review what we can do and publish a public roadmap.

memory/repository/spring-ai-model-chat-memory-repository-jdbc/src/main/java/org/springframework/ai/chat/memory/repository/jdbc/JdbcChatMemoryRepository.java
```diff
@@ -140,10 +151,24 @@ private static class MessageRowMapper implements RowMapper<Message> {
...
 			var content = rs.getString(1);
 			var type = MessageType.valueOf(rs.getString(2));
+			var toolCallsJson = rs.getString(3);
 
 			return switch (type) {
 				case USER -> new UserMessage(content);
-				case ASSISTANT -> new AssistantMessage(content);
+				case ASSISTANT -> {
+					List<AssistantMessage.ToolCall> toolCalls = List.of();
+					if (toolCallsJson != null && !toolCallsJson.trim().isEmpty()) {
```
essien, line 160: `trim()` is not aware of unicode whitespace characters and so `strip()` is the preferred way to remove whitespace characters since Java 11.

For this specific case, you can use `isBlank()` which is available since Java 11 and also unicode aware.

Suggested modification:
```diff
- !toolCallsJson.trim().isEmpty()
+ !toolCallsJson.isBlank()
```

astor-dev: Got it, updated to isBlank() as suggested. Thanks!

suzukua:
The current implementation of the memory module lacks flexibility. It should provide developers with more control, such as the ability to manage conversation history more granularly. In many cases, it's essential to support long-term memory retention rather than simply performing physical deletion.
This PR further restricts the flexibility of the implementation, which might not be ideal for extensibility or future customization needs.

S-Dragon0302:
The version 2.0.0-M4 also has this issue

PR #3347, opened 2025-05-27
fix(ai-rag): Fix query expansion result handling logic, same as #3345

### LINKED ISSUE #3345: When using `MultiQueryExpander`, if the model's response contains empty newlines, the query expansion fails and returns the original query unchanged. The current implementation doesn't properly handle or filter out empty lines from the model's response.
**Bug description**
When using `MultiQueryExpander`, if the model's response contains empty newlines, the query expansion fails and returns the original query unchanged. The current implementation doesn't properly handle or filter out empty lines from the model's response.

**Environment**
- Spring AI version: 1.0.0
- Java version: 17
- Model/API used: deepseek-reasoner
- - Spring Boot version: 3.4.6

**Steps to reproduce**
1. Configure a `MultiQueryExpander` with `numberOfQueries(3)`
2. Call `expand()` with any query
3. Have the AI model return a response that includes empty newlines between valid queries
4. Observe that the original query is returned instead of the expanded queries

**Expected behavior**
The expander should:
1. Filter out empty lines from the model's response
2. Return the valid expanded queries as long as there are enough non-empty variants
3. Only fall back to the original query if there aren't enough valid expanded queries

**Actual behavior**  
The expander fails when encountering empty newlines in the response, even when there are enough valid query variants present.

```java
MultiQueryExpander queryExpander = MultiQueryExpander.builder()
        .chatClientBuilder(this.chatClient.mutate())
        .includeOriginal(false)
        .numberOfQueries(3)
        .build();
return queryExpander.expand(new Query("How to run a Spring Boot app?"));
```

**Proposed solution​**
The split("\n") operation should be followed by filtering out empty strings. Here's the suggested fix:
```java
var queryVariants = Arrays.stream(response.split("\n"))
        .filter(StringUtils::hasText)
        .toList();

if (CollectionUtils.isEmpty(queryVariants) || this.numberOfQueries > queryVariants.size()) {
    logger.warn(
            "Query expansion result does not contain the requested {} variants. Returning the input query unchanged.",
            this.numberOfQueries);
    return List.of(query);
}
```

**Additional context​**
This is particularly important because:

1. LLMs often include formatting newlines in their responses
2. The current behavior causes valid expansions to be discarded unnecessarily
3. The fix would make the expander more robust to normal model output variations

kaijunyang:
## Description:
This PR addresses an issue where the MultiQueryExpander fails to properly handle empty newlines in the model's response, causing it to return the original query even when valid expanded queries are present.

## Related Issue:
Fixes #3345 (Bug description: MultiQueryExpander fails on empty newlines in response)

### DIFF
--- a/.monkeycode/specs/sensitive-word-filter/design.md
+++ b/.monkeycode/specs/sensitive-word-filter/design.md
@@ -1,0 +1,192 @@
+# 敏感词过滤（Sensitive Word Filter）
+
+Feature Name: sensitive-word-filter
+Updated: 2026-08-25
+
+## Description
+
+增强 Spring AI 现有 `SafeGuardAdvisor`，在 ChatClient Advisor 链上对模型输入与输出执行敏感词过滤。
+
+- **输入检查（既有能力，保留并默认开启）**：用户输入命中敏感词时，在链入口直接返回失败响应，不调用底层模型。
+- **输出检查（新增能力）**：模型响应命中敏感词时，以失败响应替换原始响应并停止回答。
+- **匹配策略（新增）**：引入公开的 `SensitiveWordMatcher` 接口，默认实现执行忽略大小写与空白的子串匹配。
+- **兼容性**：既有公开构造方法、默认失败响应文本、输入检查默认开启行为均保持不变。
+
+## Architecture
+
+### 非流式（call）执行流程
+
+```mermaid
+graph TD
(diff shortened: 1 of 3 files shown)


markpollack:
@ThomasVitale thoughts?

ilayaperumalg:
@ThomasVitale Just a reminder ping here. Could you share your thoughts on this fix? Thanks

markpollack:
can add a test in MultiQueryExpanderTests

PR #3348, opened 2025-05-27
Fix: Avoid double-serialization of JSON strings in JsonParser.toJson

ohMaruf:
Passing a JSON string like "[]" to toJson would result in "\"[]\"", which breaks deserialization for tool parameters expecting structured types such as List<Integer>. This causes exceptions like:

```
java.lang.IllegalStateException: Conversion from JSON to java.util.List<java.lang.Integer> failed
...
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `java.util.ArrayList<java.lang.Integer>` from String value (token `JsonToken.VALUE_STRING`)
 at [Source: (String)""[]""; line: 1, column: 1]
```

This PR updates the JsonParser.toJson method to avoid unnecessary double-serialization when the input is a String that already represents valid JSON (e.g. "[1,2,3]" or {"key":"value"}).

### DIFF
--- a/spring-ai-model/src/test/java/org/springframework/ai/util/json/JsonParserTests.java
+++ b/spring-ai-model/src/test/java/org/springframework/ai/util/json/JsonParserTests.java
@@ -255,6 +255,13 @@ class JsonParserTests {
 		assertThat(value).isEqualTo(1_500_000_000_000L);
 	}
 
+	@Test
+	void doesNotDoubleSerializeValidJsonString() {
+	    String input = "[1,2,3]";
+	    String result = JsonParser.toJson(input);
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
@ohMaruf Thanks for the PR! 

Fixed the formatting issues, changed the test assertion to use `org.assertj.core.api.Assertions.assertThat`, fixed the test failures at MethodToolCallbackGenericTypesTest due to the change in JSON string parsing with this fix, rebased and merged as 18ae16e6a. Back-ported the change into 1.0.x as cfa4128d5

Issue #2399, opened 2025-03-06
SpringBeanToolCallbackResolver has a problem with bean cache inside


Forest-Fairy:
This resolver will resolve and cache a tool but what will happen if a tool bean has been unregistered?
`
	@Override
	public ToolCallback resolve(String toolName) {
		Assert.hasText(toolName, "toolName cannot be null or empty");

		logger.debug("ToolCallback resolution attempt from Spring application context");

		ToolCallback resolvedToolCallback = toolCallbacksCache.get(toolName);

		if (resolvedToolCallback != null) {
			return resolvedToolCallback;
		}

		ResolvableType toolType = TypeResolverHelper.resolveBeanType(applicationContext, toolName);
		ResolvableType toolInputType = (ResolvableType.forType(Supplier.class).isAssignableFrom(toolType))
				? ResolvableType.forType(Void.class) : TypeResolverHelper.getFunctionArgumentType(toolType, 0);

		String toolDescription = resolveToolDescription(toolName, toolInputType.toClass());
		Object bean = applicationContext.getBean(toolName);

		resolvedToolCallback = buildToolCallback(toolName, toolType, toolInputType, toolDescription, bean);

		toolCallbacksCache.put(toolName, resolvedToolCallback);

		return resolvedToolCallback;
	}
`
I think it has better to implement ApplicationEventListener interface in order to remove the cache when bean has been removed in application context

PR #3356, opened 2025-05-28
fix issue@ #2399

Forest-Fairy:
resolved the cache problem with tcb by identity the hash of tool bean

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/resolution/SpringBeanToolCallbackResolver.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/resolution/SpringBeanToolCallbackResolver.java
@@ -45,6 +45,8 @@ import org.springframework.core.ResolvableType;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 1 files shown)


markpollack:
You mentioned in the issue that using an ApplicationEventListener would be a more appropriate way to handle this.  Thoughts?

Forest-Fairy:
> You mentioned in the issue that using an ApplicationEventListener would be a more appropriate way to handle this. Thoughts?

yes, I had sent a pr to resolve this problem with that, but i found another problem that if the resolver is not a bean in spring then we should remove cache manually.  this will complicate the resolver's using. however getting bean every time may cost more time than before but it simplifies the using.

Forest-Fairy:
> > You mentioned in the issue that using an ApplicationEventListener would be a more appropriate way to handle this. Thoughts?
> 
> yes, I had sent a pr to resolve this problem with that, but i found another problem that if the resolver is not a bean in spring then we should remove cache manually. this will complicate the resolver's using. however getting bean every time may cost more time than before but it simplifies the using.

the prototype tool bean will also be supported in this way, and probably the time that it costs more is still in a range that user is acceptable.

PR #3364, opened 2025-05-28
Resolve OpenAI ApiKey for every request

filiphr:
I am working on adding some dynamic support for using the `OpenAiApi` from Spring AI. Reading the `ApiKey` javadoc, it looks like a great fit for us, since we want to support dynamic resolving of it.

However, I realized that it's value is not resolved on every request, but only once in the constructor when the headers are set.

While adding this to the `OpenAiApi` I realized that the rest of the Open AI API classes were also not resolving it so I added it there as well. 

I did apply the spring format, but I am not sure that the imports are correct. Please let me know if I need to adjust something

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -128,10 +128,6 @@ public class OpenAiApi {
 
 		// @formatter:off
 		Consumer<HttpHeaders> finalHeaders = h -> {
-			if (!(apiKey instanceof NoopApiKey)) {
-				h.setBearerAuth(apiKey.getValue());
-			}
-
 			h.setContentType(MediaType.APPLICATION_JSON);
 			h.addAll(headers);
 		};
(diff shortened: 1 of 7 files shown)


markpollack:
This is a super important fix!  We had some internal users at vmware that were supposed to have an implementation with rotating keys ,but that never materialized in time.  Sorry for the oversight and we will get this bug fixed asap.

ilayaperumalg:
Also, back-ported the changes into 1.0.x via be79b892a

filiphr:
Thanks a lot @ilayaperumalg

ilayaperumalg:
@filiphr Sorry, I had to revert the changes as the support for setting the http header via OpenAI chat option is broken with this change. 

While your PR introduced the right approach to set the API keys for the client requests, we need to investigate and decide how to proceed to support the issue of supporting case which can be replicated via `OpenAiChatModelAdditionalHttpHeadersIT#additionalApiKeyHeader`

filiphr:
Thanks for letting me know @ilayaperumalg. Thanks for letting me know @ilayaperumalg. Those IT tests are a bit tricky to run without an API Key. In any case. I have provided #3398, which should solve that problem as well.

Issue #3369, opened 2025-05-29
ollamaModel  stream+tools : NullPointerException "evalDuration" is null


xulisha123:
**Bug description**
Ollama has released support for Stream+tools: https://ollama.com/blog/streaming-tool
But today I encountered an error during testing:

```

java.lang.NullPointerException: Cannot invoke "java.time.Duration.plus(java.time.Duration)" because "evalDuration" is null
	at org.springframework.ai.ollama.OllamaChatModel.from(OllamaChatModel.java:174) ~[spring-ai-ollama-1.0.0.jar:1.0.0]
	at org.springframework.ai.ollama.OllamaChatModel.lambda$internalStream$5(OllamaChatModel.java:332) ~[spring-ai-ollama-1.0.0.jar:1.0.0]
	at reactor.core.publisher.FluxMap$MapSubscriber.onNext(FluxMap.java:106) ~[reactor-core-3.6.5.jar:3.6.5]
	at reactor.core.publisher.FluxHandle$HandleSubscriber.onNext(FluxHandle.java:129) ~[reactor-core-3.6.5.jar:3.6.5]
	at reactor.core.publisher.FluxFlatMap$FlatMapMain.tryEmit(FluxFlatMap.java:547) ~[reactor-core-3.6.5.jar:3.6.5]
	at reactor.core.publisher.FluxFlatMap$FlatMapInner.onNext(FluxFlatMap.java:988) ~[reactor-core-3.6.5.jar:3.6.5]
	at reactor.core.publisher.Operators$BaseFluxToMonoOperator.completePossiblyEmpty(Operators.java:2097) ~[reactor-core-3.6.5.jar:3.6.5]
	at reactor.core.publisher.MonoReduceSeed$ReduceSeedSubscriber.onComplete(MonoReduceSeed.java:163) ~[reactor-core-3.6.5.jar:3.6.5]
	at reactor.core.publisher.FluxWindowPredicate$WindowFlux.checkTerminated(FluxWindowPredicate.java:768) ~[reactor-core-3.6.5.jar:3.6.5]
	at reactor.core.publisher.FluxWindowPredicate$WindowFlux.drainRegular(FluxWindowPredicate.java:662) ~[reactor-core-3.6.5.jar:3.6.5]
	at reactor.core.publisher.FluxWindowPredicate$WindowFlux.drain(FluxWindowPredicate.java:748) ~[reactor-core-3.6.5.jar:3.6.5]
[...]

PR #3372, opened 2025-05-29
Fixed an issue where tool calls could not function properly when using the new version of Ollama.

sunyuhan1998:
As stated in the issue: https://github.com/spring-projects/spring-ai/issues/3369, when using the new version of Ollama and a tool call is made, a `NullPointerException` occurs involving `"evalDuration"`. 

This is because, according to the official Ollama API documentation: https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-chat-completion, fields like `eval_duration` are only populated when the model response is completed. However, during a tool call, the `from` method of `org.springframework.ai.ollama.OllamaChatModel` fails to handle null values properly, leading to this issue.

I'm not entirely sure whether it's due to any changes Ollama may have made to the response interface. However, by adding more robust null checks, we can make this functionality work correctly.

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
@@ -84,6 +84,7 @@ import org.springframework.util.StringUtils;
  * @author Jihoon Kim
  * @author Alexandros Pappas
  * @author Ilayaperumal Gopinathan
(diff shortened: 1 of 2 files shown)


sunyuhan1998:
Hi @tzolov , would you please have time to review this PR? Thank you.

ilayaperumalg:
@sunyuhan1998 Thanks for fixing the bug! Rebased and merged as 72fe3bfd5. Also, back-ported the changes into 1.0.x via 1eecd1752

Issue #3333, opened 2025-05-26
Max tool calls / Max internalCall limit


aichemzee:
Please do a quick search on GitHub issues first, the feature you are about to request might have already been requested.

**Expected Behavior**

When using multiple tools, the model can sometimes get caught in an infinite loop due to the nature of the prompt. This can be avoided by limiting the max number of tools calls. 

**Current Behavior**

There is no way to put a limit on max recursion depth on internal call or provide a limit on tool calls.

**Context**

lambochen:
I'm also encountering the same issue and am trying to resolve it. Wait a moment, I'll submit my solution and PR.

TonyJeans:
I use a manual method to call the tools in order to control the maximum number of times.

```
ToolCallingManager toolCallingManager = DefaultToolCallingManager.builder().build();
ChatMemory chatMemory = MessageWindowChatMemory.builder().build();
String conversationId = UUID.randomUUID().toString();

ChatOptions chatOptions = ToolCallingChatOptions.builder()
    .toolCallbacks(ToolCallbacks.from(new MathTools()))
    .internalToolExecutionEnabled(false)
    .build();
Prompt prompt = new Prompt(
        List.of(new SystemMessage("You are a helpful assistant."), new UserMessage("What is 6 * 8?")),
        chatOptions);
chatMemory.add(conversationId, prompt.getInstructions());

Prompt promptWithMemory = new Prompt(chatMemory.get(conversationId), chatOptions);
ChatResponse chatResponse = chatModel.call(promptWithMemory);
chatMemory.add(conversationId, chatResponse.getResult().getOutput());
int count= 0;
while (chatResponse.hasToolCalls() && count< 3) {
    ToolExecutionResult toolExecutionResult = toolCallingManager.executeToolCalls(promptWithMemory,
            chatResponse);
    chatMemory.add(conversationId, toolExecutionResult.conversationHistory()
        .get(toolExecutionResult.conversationHistory().size() - 1));
    promptWithMemory = new Prompt(chatMemory.get(conversationId), chatOptions);
    chatResponse = chatModel.call(promptWithMemory);
    chatMemory.add(conversationId, chatResponse.getResult().getOutput());
    count++;
}

UserMessage newUserMessage = new UserMessage("What did I ask you earlier?");
chatMemory.add(conversationId, newUserMessage);

ChatResponse newResponse = chatModel.call(new Prompt(chatMemory.get(conversationId)));
```

[...]

PR #3380, opened 2025-05-29
ToolCallingChatOptions support internalToolExecutionMaxAttempts, to limit the maximum number of tool calls and prevent infinite recursive calls to LLM in special cases

lambochen:
This PR primarily addresses the issue of infinite recursive calls to LLM under certain special conditions, with internalToolExecutionEnabled=true in place, ref issue: https://github.com/spring-projects/spring-ai/issues/3333.

By introducing support for internalToolExecutionMaxAttempts in ToolCallingChatOptions and performing call count checks in each ChatModel, interrupt the process when the maximum call count is reached.

[x] Sign the [contributor license agreement](https://cla.pivotal.io/sign/spring)
[x] Rebase your changes on the latest `main` branch and squash your commits
[ ] Add/Update unit tests as needed
[ ] Run a build and make sure all tests pass prior to submission

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
@@ -33,6 +33,7 @@ import org.springframework.util.Assert;
  * Default implementation of {@link ToolCallingChatOptions}.
  *
  * @author Thomas Vitale
+ * @author lambochen
  * @since 1.0.0
  */
 public class DefaultToolCallingChatOptions implements ToolCallingChatOptions {
@@ -116,6 +117,16 @@ public class DefaultToolCallingChatOptions implements ToolCallingChatOptions {
 	@Override
 	public void setInternalToolExecutionEnabled(@Nullable Boolean internalToolExecutionEnabled) {
 		this.internalToolExecutionEnabled = internalToolExecutionEnabled;
+	}
+
+	@Override
+	public Integer getInternalToolExecutionMaxIterations() {
+		return 0;
+	}
+
+	@Override
+	public void setInternalToolExecutionMaxIterations(Integer internalToolExecutionMaxIterations) {
+
 	}
 
 	@Override
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
@@ -48,7 +48,7 @@ public class DefaultToolCallingChatOptions implements ToolCallingChatOptions {
 	private Boolean internalToolExecutionEnabled;
 
 	@Nullable
-	private Integer internalToolExecutionMaxAttempts;
+	private Integer internalToolExecutionMaxAttempts = ToolCallingChatOptions.DEFAULT_TOOL_EXECUTION_MAX_ATTEMPTS;
 
 	@Nullable
 	private String model;
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
@@ -293,7 +293,8 @@ public class DefaultToolCallingChatOptions implements ToolCallingChatOptions {
 		}
 
 		@Override
-		public ToolCallingChatOptions.Builder internalToolExecutionMaxAttempts(Integer internalToolExecutionMaxAttempts) {
+		public ToolCallingChatOptions.Builder internalToolExecutionMaxAttempts(
+				Integer internalToolExecutionMaxAttempts) {
 			this.options.setInternalToolExecutionMaxAttempts(internalToolExecutionMaxAttempts);
 			return this;
 		}
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
@@ -48,6 +48,9 @@ public class DefaultToolCallingChatOptions implements ToolCallingChatOptions {
 	private Boolean internalToolExecutionEnabled;
 
 	@Nullable
+	private Integer internalToolExecutionMaxAttempts;
+
+	@Nullable
 	private String model;
 
 	@Nullable
@@ -120,13 +123,13 @@ public class DefaultToolCallingChatOptions implements ToolCallingChatOptions {
 	}
 
 	@Override
-	public Integer getInternalToolExecutionMaxIterations() {
-		return 0;
+	public Integer getInternalToolExecutionMaxAttempts() {
+		return this.internalToolExecutionMaxAttempts;
 	}
 
 	@Override
-	public void setInternalToolExecutionMaxIterations(Integer internalToolExecutionMaxIterations) {
-
+	public void setInternalToolExecutionMaxAttempts(Integer internalToolExecutionMaxAttempts) {
+		this.internalToolExecutionMaxAttempts = internalToolExecutionMaxAttempts;
 	}
 
 	@Override
@@ -217,6 +220,7 @@ public class DefaultToolCallingChatOptions implements ToolCallingChatOptions {
 		options.setToolNames(getToolNames());
 		options.setToolContext(getToolContext());
 		options.setInternalToolExecutionEnabled(getInternalToolExecutionEnabled());
+		options.setInternalToolExecutionMaxAttempts(getInternalToolExecutionMaxAttempts());
 		options.setModel(getModel());
 		options.setFrequencyPenalty(getFrequencyPenalty());
 		options.setMaxTokens(getMaxTokens());
@@ -285,6 +289,12 @@ public class DefaultToolCallingChatOptions implements ToolCallingChatOptions {
 		public ToolCallingChatOptions.Builder internalToolExecutionEnabled(
 				@Nullable Boolean internalToolExecutionEnabled) {
 			this.options.setInternalToolExecutionEnabled(internalToolExecutionEnabled);
+			return this;
+		}
+
+		@Override
+		public ToolCallingChatOptions.Builder internalToolExecutionMaxAttempts(Integer internalToolExecutionMaxAttempts) {
+			this.options.setInternalToolExecutionMaxAttempts(internalToolExecutionMaxAttempts);
 			return this;
 		}
 
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
@@ -96,7 +96,7 @@ public interface ToolCallingChatOptions extends ChatOptions {
 	 * Set the maximum number of attempts for tool execution.
 	 * @param internalToolExecutionMaxAttempts the maximum number of attempts.
 	 */
-	void setInternalToolExecutionMaxAttempts(Integer internalToolExecutionMaxAttempts);
+	void setInternalToolExecutionMaxAttempts(@Nullable Integer internalToolExecutionMaxAttempts);
 
 	/**
 	 * Get the configured tool context.
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
@@ -48,6 +48,7 @@ public interface ToolCallingChatOptions extends ChatOptions {
 	 * No limit for tool execution attempts.
 	 */
 	int TOOL_EXECUTION_NO_LIMIT = Integer.MAX_VALUE;
+
 	int DEFAULT_TOOL_EXECUTION_MAX_ATTEMPTS = TOOL_EXECUTION_NO_LIMIT;
 
 	/**
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
@@ -37,6 +37,7 @@ import org.springframework.util.CollectionUtils;
  *
  * @author Thomas Vitale
  * @author Ilayaperumal Gopinathan
+ * @author lambochen
  * @since 1.0.0
  */
 public interface ToolCallingChatOptions extends ChatOptions {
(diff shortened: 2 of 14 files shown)


markpollack:
thanks, one comment is that we need to preserve api compatability and 

```
public ChatResponse internalCall(Prompt prompt, ChatResponse previousChatResponse, int attempts) {
```

would need to be an overload vs adding a new method parameter.

lambochen:
> thanks, one comment is that we need to preserve api compatability and
> 
> ```
> public ChatResponse internalCall(Prompt prompt, ChatResponse previousChatResponse, int attempts) {
> ```
> 
> would need to be an overload vs adding a new method parameter.

Your suggestion is great, I'll optimize it, thanks

lambochen:
Hi, @markpollack  @tzolov , the PR is ready, please review it, thank you.

If there are any issues or if additional content is needed, please to comment, and I will promptly correct or supplement it. thank you

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
```diff
@@ -46,6 +47,9 @@ public class DefaultToolCallingChatOptions implements ToolCallingChatOptions {
 	@Nullable
 	private Boolean internalToolExecutionEnabled;
 
+	@Nullable
+	private Integer internalToolExecutionMaxAttempts = ToolCallingChatOptions.DEFAULT_TOOL_EXECUTION_MAX_ATTEMPTS;
```
ThomasVitale: Consider renaming this to something like `toolExecutionMaxIterations`. 

I'm suggesting dropping the `internal` prefix since this option can also be used as part of the _user-controller tool execution_, which is executed externally from the `ChatModel`.

Also, I'm suggesting replacing `attempts` with `iterations` since attempts would make seem related to the concept of retrying tool calls, whereas here we want to limit the number of iterations in the loop supporting the tool call requests from the model.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
```diff
@@ -37,12 +37,20 @@
...
  * @since 1.0.0
  */
 public interface ToolCallingChatOptions extends ChatOptions {
 
 	boolean DEFAULT_TOOL_EXECUTION_ENABLED = true;
 
+	/**
+	 * No limit for tool execution attempts.
+	 */
+	int TOOL_EXECUTION_NO_LIMIT = Integer.MAX_VALUE;
```
ThomasVitale, line 50: I would say it's a mistake that no limit was set in the original implementation. I would consider having a lower value as the default limit. We should think of a good default value here that avoids infinite loops.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
```diff
@@ -109,6 +131,21 @@ static boolean isInternalToolExecutionEnabled(ChatOptions chatOptions) {
 		return internalToolExecutionEnabled;
 	}
 
+	static boolean isInternalToolExecutionEnabled(ChatOptions chatOptions, int attempts) {
```
ThomasVitale: In general, I would suggest replacing `attempts` with `iterations`. In this case, I would make it more specific and name the argument `toolExecutionIterations` or `toolExecutionCompletedIterations` or something like that.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
hongling-dev: Hi, ThomasVitale, thank you for your attention and comments.

Regarding removing the `internal` prefix: I agree. In the past，considering consistency with `internalToolExecutionEnabled`, I think we should reconsider this.

Regarding replacing `attempts` with `iterations`: Yes, I accept your suggestion, actually, I initially considered `iterations` as well.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
hongling-dev: Yes, I agree with your point. To ensure downward compatibility, I have used Integer.MAX_VALUE (a very large number) to indicate no limit.

I tend to prefer setting a smaller value to limit the number of executions and the scope of impact on the program, but I don't have specific comparative data. If you have suggestions, please feel free to share them. Thank you

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
hongling-dev: Thank you for your suggestion, I will rename `attempts` to use `iterations`.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
hongling-dev: Of course, I also agree with the naming of the previous developer. The `internal` can express that it is executed internally by the Spring AI framework, which is a preset behavior.

Regarding the `internal` prefix, I support not using the prefix, as I believe this is a capability of the framework, a feature. The `internal` prefix makes users think this is a field only for internal use of the framework, and using it may come with some "concerns."

Discussion：If the `internal` prefix is removed, should the original `internalToolExecutionEnabled` field also be deprecated and a new `toolExecutionEnabled` created to replace it?

hongling-dev:
@ThomasVitale  Hi, I have optimized the code and replied to your suggestion. Please re-review this PR. Thank you for your hard work.

aichemzee:
Thank you @lambochen for picking this up. This would be a very helpful change with tool calling.

tzolov:
Thanks @lambochen! 

I agreed on handling the infinite loop risk, but, IMO, should consider: 

1. What should happen when max interactions are reached? Currently it silently returns the last tool response. I think we should throw an error instead - maybe ToolExecutionException?

2. Could we introduce a `ToolExecutionRequest` (like `ToolExecutionResult`) containing the prompt, chat response, and iteration counter? We could also extend `ToolExecutionResult` to include the call count. This would let us centralize the check logic in `executeToolCalls`.

@lambochen , @ThomasVitale what do you think?

hongling-dev:
Thanks! @tzolov 

> 1. What should happen when max interactions are reached? Currently it silently returns the last tool response. I think we should throw an error instead - maybe ToolExecutionException?

I agree, should we also consider other strategies?
1. Silently return the last tool response
2. Throw a `ToolExecutionException`
3. Custom handler

> 2. Could we introduce a `ToolExecutionRequest` (like `ToolExecutionResult`) containing the prompt, chat response, and iteration counter? We could also extend `ToolExecutionResult` to include the call count. This would let us centralize the check logic in `executeToolCalls`.

That's great, introducing ToolExecutionRequest will make the logic much cleaner.

tzolov:
@lambochen 
> I agree, should we also consider other strategies?

I don't think we need this flexibility right now. I can't see a use case where "silently return the last tool response" would be helpful. Also, this change needs to work with `internalToolExecutionEnabled=false`, and returning a random tool response could cause issues there.

Let's save strategy support for a future PR if we find compelling reasons for it.

ThomasVitale:
I'll have a look tomorrow at the updated PR and share my comments then. Thanks!

hongling-dev:
> @lambochen
> 
> > I agree, should we also consider other strategies?
> 
> I don't think we need this flexibility right now. I can't see a use case where "silently return the last tool response" would be helpful. Also, this change needs to work with `internalToolExecutionEnabled=false`, and returning a random tool response could cause issues there.
> 
> Let's save strategy support for a future PR if we find compelling reasons for it.

@tzolov  Thank you for your reply, I agree with your point, I will optimize the code and resubmit, thank you

hongling-dev:
> 2. Could we introduce a `ToolExecutionRequest` (like `ToolExecutionResult`) containing the prompt, chat response, and iteration counter? We could also extend `ToolExecutionResult` to include the call count. This would let us centralize the check logic in `executeToolCalls`.

Hi, @tzolov  
I didn't quite understand your suggestion. Are you expecting to centralize all tool execution logic into ToolCallingManager.executeToolCalls, including isToolExecutionRequired and isLimitExceeded? 
I hope you can clarify my confusion. Thank you.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
```diff
@@ -76,6 +84,23 @@ public interface ToolCallingChatOptions extends ChatOptions {
...
 
+	/**
+	 * Get the maximum number of iteration for tool execution. If the number of iterations
+	 * exceeds the limit, an {@link ToolExecutionLimitExceededException} will be thrown.
+	 * @return the maximum number of iteration.
+	 * @see #getInternalToolExecutionEnabled()
+	 * @see ToolExecutionLimitExceededException
+	 */
+	@Nullable
+	Integer getInternalToolExecutionMaxIterations();
```
ThomasVitale: I still recommend changing the name of this. The "internal" prefix in the other option is to enable/disable having ChatModel perform the tool execution _internally_ as part of an inference request (see [framework-controlled tool execution](https://docs.spring.io/spring-ai/reference/api/tools.html#_framework_controlled_tool_execution)). Arguably, not the best naming choice, but that's what we have now.

This new option about the max iterations is more generic than that, just like tool context, tool callbacks, and tool names. It shouldn't have the "internal" prefix since it can be used even when relying on [user-controlled tool execution](https://docs.spring.io/spring-ai/reference/api/tools.html#_user_controlled_tool_execution) (i.e. _not internal_). Especially since the max iterations are now part of the public API of ToolExecutionEligibilityPredicate and Checker which can be used for user-controlled tool execution.

@tzolov thoughts?

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
```diff
@@ -46,6 +47,9 @@ public class DefaultToolCallingChatOptions implements ToolCallingChatOptions {
 	@Nullable
 	private Boolean internalToolExecutionEnabled;
 
+	@Nullable
+	private Integer internalToolExecutionMaxIterations = ToolCallingChatOptions.DEFAULT_TOOL_EXECUTION_MAX_ITERATIONS;
```
ThomasVitale, line 51: Consider moving the default value definition to this class rather than the interface, since this is the class implementing the defaults.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
ThomasVitale: @tzolov what do you think? Should we keep it unbounded (i.e infinite loop) or find a good default value?

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolExecutionEligibilityChecker.java
```diff
@@ -74,4 +92,36 @@ default boolean isInternalToolExecutionEnabled(ChatOptions chatOptions) {
...
+		return !isLimitExceeded(chatOptions, toolExecutionIterations);
+	}
+
+	/**
+	 * Determines if the tool execution limit has been exceeded.
+	 * @param promptOptions The options from the prompt
+	 * @param toolExecutionIterations The number of toolExecutionIterations
+	 * @return true if the tool execution limit has been exceeded, false otherwise
+	 */
+	default boolean isLimitExceeded(ChatOptions promptOptions, int toolExecutionIterations) {
```
ThomasVitale, line 118: Should this be a public API or a utility method?

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolExecutionEligibilityPredicate.java
```diff
@@ -43,4 +43,39 @@ default boolean isToolExecutionRequired(ChatOptions promptOptions, ChatResponse
...
+		return !isLimitExceeded(promptOptions, toolExecutionIterations);
+	}
+
+	/**
+	 * Determines if the tool execution limit has been exceeded.
+	 * @param promptOptions The options from the prompt
+	 * @param toolExecutionIterations The number of toolExecutionIterations
+	 * @return true if the tool execution limit has been exceeded, false otherwise
+	 */
+	default boolean isLimitExceeded(ChatOptions promptOptions, int toolExecutionIterations) {
```
ThomasVitale, line 72: Same question here as in `ToolExecutionEligibilityChecker`. Should this be a public API or a utility method?

I would suggest moving this to `ToolCallingChatOptions` as a static method, similar to other utilities we have there to deal with tool execution.

hongling-dev: Hi, @ThomasVitale ，Thank you for your comment, your considerations are very valuable.

Regarding this issue, I think both are acceptable.

As a public API, it opens up possibilities for users to extend, allowing users to be more flexible when implementing ToolExecutionEligibilityPredicate, of course, this is on the premise that users expect to implement isLimitExceeded.

As a utility method, it's also a good design that can meet the business's customization for ToolExecutionEligibilityPredicate (by implementing the isToolExecutionRequired method). It's equivalent to users only needing to be aware of the isToolExecutionRequired SPI method, which would be more concise.

What do you think?

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
hongling-dev: Hi，@ThomasVitale 

> Of course, I also agree with the naming of the previous developer. The internal can express that it is executed internally by the Spring AI framework, which is a preset behavior.
> 
> Regarding the internal prefix, I support not using the prefix, as I believe this is a capability of the framework, a feature. The internal prefix makes users think this is a field only for internal use of the framework, and using it may come with some "concerns."
> 
> Discussion：If the internal prefix is removed, should the original internalToolExecutionEnabled field also be deprecated and a new toolExecutionEnabled created to replace it?

As in the previous comment, I agree with your point and strongly support removing the "internal" prefix. If we can reach a consensus, I will optimize the naming and resubmit the code, thank you

hongling-dev: Renamed to toolExecutionMaxIterations

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolCallingChatOptions.java
ThomasVitale: Should we go with 20?

hongling-dev: > Should we go with 20?

Hi, @ThomasVitale  , Thank you for your reply. Could you explain the reason for the specific value of 20?

markpollack: I was thinking more like 3...  three strikes and you are out!

PR #3381, opened 2025-05-29
Upgrade Spring Boot to 3.5.0

ilayaperumalg:
- Update Spring Boot version to 3.5.0
 - Enforce elasticsearch-java client to use the explicit version 8.18.1 - This is because of the incompatibility between Spring Boot 3.4.x and 3.5.0 in using the elastic search java client versions 8.15.1 vs 8.18.1


markpollack:
merged in 2be1e42505307a7b71ed2bcc0fa1758d933d295b and backported to 1.0.x

Issue #3383, opened 2025-05-30
Accessing the New "Thought" Field from Ollama in the Spring AI Framework


Dudu0831:
![Image](https://github.com/user-attachments/assets/ece30474-ab5c-4067-8574-303ac3cc1b2e)
Ollama has updated with a new "thought" field. I want to know how to retrieve this field within the Spring AI framework.

sunyuhan1998:
Hi @Dudu0831 Can you please provide the URL of the document in the picture?I can't seem to find it in the ollama repository.

sunyuhan1998:
> https://github.com/ollama/ollama/releases

Thank you very much!I will check out this URL!

Dudu0831:
@sunyuhan1998 你好，我发现了一个新的问题，不知道是不是ollama升级的原因。我升级到了0.8.0。
只有调用工具时会出现以下情况
1.报空指针异常错误。
报错阶段发生在模型调用完工具后进行回答时，如下图
![Image](https://github.com/user-attachments/assets/474f85f6-9217-43ee-bbe5-78954bb36fc6)
当我恢复程序时发生如下错误
![Image](https://github.com/user-attachments/assets/2bd263fb-7305-4bc5-8f29-867f3261bca8)

sunyuhan1998:
> [@sunyuhan1998](https://github.com/sunyuhan1998) 你好，我发现了一个新的问题，不知道是不是ollama升级的原因。我升级到了0.8.0。 只有调用工具时会出现以下情况 1.报空指针异常错误。 

This is the same issue as https://github.com/spring-projects/spring-ai/issues/3369, I've pushed a PR to fix the issue, which should be due to an interface tweak in ollama.

sunyuhan1998:
> [@sunyuhan1998](https://github.com/sunyuhan1998) Alright, thanks a lot for your explanation.

By the way, regarding the new version of Ollama that now supports the `think` field, since this is a recent update, Spring AI hasn't supported it yet. I've already submitted a PR to try to add support for it: https://github.com/spring-projects/spring-ai/pull/3386.

sharan245:
I was waiting for [1.1.0.M4](https://github.com/spring-projects/spring-ai/milestone/28) release so that i can use the Thinking feature,  hope will get [1.1.0.RC1](https://github.com/spring-projects/spring-ai/milestone/18) soon.

shuanggegege:
[...]

PR #3386, opened 2025-05-30
Added support for the "think" for Ollama

sunyuhan1998:
As mentioned in issue: https://github.com/spring-projects/spring-ai/issues/3383 , Ollama added support for "think" in its latest 0.9.0 version: 
https://github.com/ollama/ollama/releases
https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-chat-completion. 

This PR implements support for that attribute and includes the following key changes: 
1. Added the `think` field to Ollama's `ChatRequest`
2. Added the `thinking` field to Ollama's `Message`
3. Added the `think` property to `OllamaOptions`, allowing users to specify whether to enable or disable thinking

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
@@ -460,7 +460,8 @@ public class OllamaChatModel implements ChatModel {
 		OllamaApi.ChatRequest.Builder requestBuilder = OllamaApi.ChatRequest.builder(requestOptions.getModel())
 			.stream(stream)
 			.messages(ollamaMessages)
-			.options(requestOptions);
+			.options(requestOptions)
+			.think(requestOptions.getThink());
 
 		if (requestOptions.getFormat() != null) {
 			requestBuilder.format(requestOptions.getFormat());
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
@@ -51,6 +51,7 @@ import org.springframework.web.reactive.function.client.WebClient;
  * @author Christian Tzolov
  * @author Thomas Vitale
  * @author Jonghoon Park
+ * @author Sun Yuhan
  * @since 0.8.0
  */
 // @formatter:off
@@ -251,6 +252,7 @@ public final class OllamaApi {
 	 *
 	 * @param role The role of the message of type {@link Role}.
 	 * @param content The content of the message.
+	 * @param thinking The thinking of the model.
 	 * @param images The list of base64-encoded images to send with the message.
 	 * 				 Requires multimodal models such as llava or bakllava.
 	 * @param toolCalls The relevant tool call.
@@ -260,6 +262,7 @@ public final class OllamaApi {
 	public record Message(
 			@JsonProperty("role") Role role,
 			@JsonProperty("content") String content,
+			@JsonProperty("thinking") String thinking,
 			@JsonProperty("images") List<String> images,
 			@JsonProperty("tool_calls") List<ToolCall> toolCalls) {
 
@@ -321,6 +324,7 @@ public final class OllamaApi {
 
 			private final Role role;
 			private String content;
+			private String thinking;
 			private List<String> images;
 			private List<ToolCall> toolCalls;
 
@@ -330,6 +334,11 @@ public final class OllamaApi {
 
 			public Builder content(String content) {
 				this.content = content;
+				return this;
+			}
+
+			public Builder thinking(String thinking) {
+				this.thinking = thinking;
 				return this;
 			}
 
@@ -344,7 +353,7 @@ public final class OllamaApi {
 			}
 
 			public Message build() {
-				return new Message(this.role, this.content, this.images, this.toolCalls);
+				return new Message(this.role, this.content, this.thinking, this.images, this.toolCalls);
 			}
 		}
 	}
@@ -359,6 +368,7 @@ public final class OllamaApi {
 	 * @param keepAlive Controls how long the model will stay loaded into memory following this request (default: 5m).
 	 * @param tools List of tools the model has access to.
 	 * @param options Model-specific options. For example, "temperature" can be set through this field, if the model supports it.
+	 * @param think The model should think before responding, if the model supports it.
 	 * You can use the {@link OllamaOptions} builder to create the options then {@link OllamaOptions#toMap()} to convert the options into a map.
 	 *
 	 * @see <a href=
@@ -375,7 +385,8 @@ public final class OllamaApi {
 			@JsonProperty("format") Object format,
 			@JsonProperty("keep_alive") String keepAlive,
 			@JsonProperty("tools") List<Tool> tools,
-			@JsonProperty("options") Map<String, Object> options
+			@JsonProperty("options") Map<String, Object> options,
+			@JsonProperty("think") Boolean think
 	) {
 
 		public static Builder builder(String model) {
@@ -448,6 +459,7 @@ public final class OllamaApi {
 			private String keepAlive;
 			private List<Tool> tools = List.of();
 			private Map<String, Object> options = Map.of();
+			private boolean think;
 
 			public Builder(String model) {
 				Assert.notNull(model, "The model can not be null.");
@@ -492,8 +504,13 @@ public final class OllamaApi {
 				return this;
 			}
 
+			public Builder think(boolean think) {
+				this.think = think;
+				return this;
+			}
+
 			public ChatRequest build() {
-				return new ChatRequest(this.model, this.messages, this.stream, this.format, this.keepAlive, this.tools, this.options);
+				return new ChatRequest(this.model, this.messages, this.stream, this.format, this.keepAlive, this.tools, this.options, this.think);
 			}
 		}
 	}
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
@@ -23,6 +23,7 @@ import org.springframework.ai.model.ChatModelDescription;
  *
  * @author Siarhei Blashuk
  * @author Thomas Vitale
+ * @author Sun Yuhan
  * @since 1.0.0
  */
 public enum OllamaModel implements ChatModelDescription {
@@ -31,6 +32,16 @@ public enum OllamaModel implements ChatModelDescription {
 	 * Qwen 2.5
 	 */
 	QWEN_2_5_7B("qwen2.5"),
+
+	/**
+	 * Qwen3
+	 */
+	QWEN_3_8B("qwen3"),
+
+	/**
+	 * Qwen3 4b
+	 */
+	QWEN_3_4B("qwen3:4b"),
 
 	/**
 	 * QwQ is the reasoning model of the Qwen series.
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaOptions.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaOptions.java
@@ -44,6 +44,7 @@ import org.springframework.util.Assert;
  * @author Christian Tzolov
  * @author Thomas Vitale
  * @author Ilayaperumal Gopinathan
+ * @author Sun Yuhan
  * @since 0.8.0
  * @see <a href=
  * "https://github.com/ollama/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values">Ollama
@@ -318,6 +319,14 @@ public class OllamaOptions implements ToolCallingChatOptions, EmbeddingOptions {
 	@JsonProperty("truncate")
 	private Boolean truncate;
 
+	/**
+	 * The model should think before responding, if supported.
+	 * If this value is not specified, it defaults to null, and Ollama will return
+	 * the thought process within the `content` field of the response, wrapped in `&lt;thinking&gt;` tags.
+	 */
+	@JsonProperty("think")
+	private Boolean think;
+
 	@JsonIgnore
 	private Boolean internalToolExecutionEnabled;
 
@@ -365,6 +374,7 @@ public class OllamaOptions implements ToolCallingChatOptions, EmbeddingOptions {
 				.format(fromOptions.getFormat())
 				.keepAlive(fromOptions.getKeepAlive())
 				.truncate(fromOptions.getTruncate())
+				.think(fromOptions.getThink())
 				.useNUMA(fromOptions.getUseNUMA())
 				.numCtx(fromOptions.getNumCtx())
 				.numBatch(fromOptions.getNumBatch())
@@ -704,6 +714,14 @@ public class OllamaOptions implements ToolCallingChatOptions, EmbeddingOptions {
 		this.truncate = truncate;
 	}
 
+	public Boolean getThink() {
+		return this.think;
+	}
+
+	public void setThink(Boolean think) {
+		this.think = think;
+	}
+
 	@Override
 	@JsonIgnore
 	public List<ToolCallback> getToolCallbacks() {
@@ -804,7 +822,8 @@ public class OllamaOptions implements ToolCallingChatOptions, EmbeddingOptions {
 				&& Objects.equals(this.repeatPenalty, that.repeatPenalty)
 				&& Objects.equals(this.presencePenalty, that.presencePenalty)
 				&& Objects.equals(this.frequencyPenalty, that.frequencyPenalty)
-				&& Objects.equals(this.mirostat, that.mirostat) && Objects.equals(this.mirostatTau, that.mirostatTau)
+				&& Objects.equals(this.think, that.think) && Objects.equals(this.mirostat, that.mirostat)
+				&& Objects.equals(this.mirostatTau, that.mirostatTau)
 				&& Objects.equals(this.mirostatEta, that.mirostatEta)
 				&& Objects.equals(this.penalizeNewline, that.penalizeNewline) && Objects.equals(this.stop, that.stop)
 				&& Objects.equals(this.toolCallbacks, that.toolCallbacks)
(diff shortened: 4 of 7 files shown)


sunyuhan1998:
@tzolov @ilayaperumalg @markpollack Could you please help review this PR? Thank you.

markpollack:
Yes, we will review. Thanks

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
```diff
@@ -460,7 +460,8 @@ else if (message instanceof ToolResponseMessage toolMessage) {
 		OllamaApi.ChatRequest.Builder requestBuilder = OllamaApi.ChatRequest.builder(requestOptions.getModel())
 			.stream(stream)
 			.messages(ollamaMessages)
-			.options(requestOptions);
+			.options(requestOptions)
+			.think(requestOptions.getThink());
```
markpollack, line 464: This seems awkward, should be part of the `.options` call.  Similar to my other comment in the Builder class.

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaOptions.java
```diff
@@ -318,6 +319,14 @@ public class OllamaOptions implements ToolCallingChatOptions, EmbeddingOptions {
 	@JsonProperty("truncate")
 	private Boolean truncate;
 
+	/**
+	 * The model should think before responding, if supported.
+	 * If this value is not specified, it defaults to null, and Ollama will return
+	 * the thought process within the `content` field of the response, wrapped in `&lt;thinking&gt;` tags.
+	 */
+	@JsonProperty("think")
```
markpollack, line 362: It seems like 'think' is not a part of the options map in ollama, but a 'top level' field in the request object.
In https://ollama.com/blog/thinking there is the example
```
curl http://localhost:11434/api/chat -d '{
  "model": "deepseek-r1",
  "messages": [
    {
      "role": "user",
      "content": "how many r in the word strawberry?"
    }
  ],
  "think": true,
  "stream": false
}'
```

and the golang type supporting this feature also shows the same structure.

https://github.com/ollama/ollama/blob/45f56355d557b7130c7c07bbd6e1b634a758d946/api/types.go#L91

So it shouldn't be added to the options map.

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
```diff
@@ -260,6 +262,7 @@ public Flux<ProgressResponse> pullModel(PullModelRequest pullModelRequest) {
 	public record Message(
 			@JsonProperty("role") Role role,
 			@JsonProperty("content") String content,
+			@JsonProperty("thinking") String thinking,
```
markpollack: We should propagate the 'thinking' response back into the `ChatGenerationMetadata` so that it is accessible in the response when using OllamaChatModel.  This would be inside the `internalCall` method of `OllamaChatModel` and add add a test for that in `OllamaChatModelIT`

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaOptions.java
markpollack: The comment in the code 

```
and Ollama will return
	 * the thought process within the `content` field of the response, wrapped in `&lt;thinking&gt;` tags.
```

seems to contradict what was documented on the ollama web site that shows the 'think' response as a separate field from 'content', and not nested inside the 'content' field.

markpollack: Maybe we want to also expose it in `OllamaOptions` as that right now would be the only way to pass in this feature flag when making calls via `ChatModel` or  `ChatClient`.   I think if the feature of enabling thinking mode is implemented from a `ChatModel` or `ChatClient` level the right solution will present itself.  Can you improve this PR to handle this scenario please?

sunyuhan1998: > Maybe we want to also expose it in `OllamaOptions` as that right now would be the only way to pass in this feature flag when making calls via `ChatModel` or `ChatClient`. I think if the feature of enabling thinking mode is implemented from a `ChatModel` or `ChatClient` level the right solution will present itself. Can you improve this PR to handle this scenario please?

Of course, no problem. I will continue improving this PR and update my progress here in a timely manner.

Before proceeding, I want to confirm whether I’ve correctly understood your point: Are you suggesting that we should not add the `think` flag in `OllamaOptions`, but instead make adjustments at the `OllamaChatModel` level? Or do you mean that we should implement support for `think` at the `ChatModel` or `ChatClient` level? If it's the latter, I think we would also need to adjust the implementations of different `ChatModel`s to support this option (of course, depending on whether the underlying model actually supports it).

sunyuhan1998: > seems to contradict what was documented on the ollama web site that shows the 'think' response as a separate field from 'content', and not nested inside the 'content' field.


Firstly, the content in the comment is actually a summary I derived through practical testing.

I think this is a form of backward compatibility Ollama implemented for users still employing the old parameter-passing method. Before Ollama supported the "think" flag, if we made a request to a model that supports thinking like this:

```
curl http://localhost:11434/api/chat -d '{
  "model": "qwen3:4b",
  "messages": [
    { "role": "user", "content": "why is the sky blue?" }
  ]
}'
```

Ollama would enable thinking by default and return the thought process wrapped in `<thinking>` tags within the `content` field of the response. The response would look something like this:

```
{"model":"qwen3:4b","created_at":"2025-06-12T12:17:08.385341Z","message":{"role":"assistant","content":"\u003cthink\u003e"},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:17:08.402756Z","message":{"role":"assistant","content":"\n"},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:17:08.420851Z","message":{"role":"assistant","content":"Okay"},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:17:08.439825Z","message":{"role":"assistant","content":","},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:17:08.457618Z","message":{"role":"assistant","content":" the"},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:17:08.474711Z","message":{"role":"assistant","content":" user"},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:17:08.491833Z","message":{"role":"assistant","content":" is"},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:17:08.509124Z","message":{"role":"assistant","content":" asking"},"done":false}
```

This remains the case in the latest version of Ollama: if the `think` flag is not specified, the behavior of Ollama remains unchanged from before. This is what I intended to express in the comment:

> If this value is not specified, it defaults to null, and Ollama will return the thought process within the `content` field of the response, wrapped in `<thinking>` tags.

Only when we specify the `think` flag will Ollama return the thought process in the `thinking` field of the response:

```
curl http://localhost:11434/api/chat -d '{
  "model": "qwen3:4b",
  "messages": [
    { "role": "user", "content": "why is the sky blue?" }
  ],
  "think": true
}'
```

Response:

```
{"model":"qwen3:4b","created_at":"2025-06-12T12:22:48.135211Z","message":{"role":"assistant","content":"","thinking":"Okay"},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:22:48.152511Z","message":{"role":"assistant","content":"","thinking":","},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:22:48.169911Z","message":{"role":"assistant","content":"","thinking":" the"},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:22:48.187023Z","message":{"role":"assistant","content":"","thinking":" user"},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:22:48.204039Z","message":{"role":"assistant","content":"","thinking":" is"},"done":false}
{"model":"qwen3:4b","created_at":"2025-06-12T12:22:48.221233Z","message":{"role":"assistant","content":"","thinking":" asking"},"done":false}
```
I think it's the same for Spring AI. We should maintain compatibility with users who are using older versions, meaning that if the "think" flag is not specified, the returned format should remain unchanged.

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
sunyuhan1998: No problem, I will add the implementation for this part.

sunyuhan1998: I have added the logic to propagate the `thinking` back to `ChatGenerationMetadata`, along with the corresponding unit tests. Could you please help review it? Thank you.

https://github.com/spring-projects/spring-ai/pull/3386/commits/b188adcab68373c2ee57f4134f4e2b87b04913ed

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaOptions.java
sunyuhan1998: > Maybe we want to also expose it in `OllamaOptions` as that right now would be the only way to pass in this feature flag when making calls via `ChatModel` or `ChatClient`. I think if the feature of enabling thinking mode is implemented from a `ChatModel` or `ChatClient` level the right solution will present itself. Can you improve this PR to handle this scenario please?

Of course, there's no problem. I will continue improving this PR and keep you updated on my progress here. 

I'd like to first confirm with you whether I've correctly understood your point: Are you suggesting that we implement support for `think"` at the `ChatModel` level — that is, first support `think` through the `ChatOptions` interface, and then inherit and implement support for `think` in `OllamaOptions`?  

I've made a commit based on this understanding. Please help me review it, and if I misunderstood, feel free to point it out — I'll gladly make further adjustments.

TonyReggae:
可以考虑加一个可自由扩展参数的字段，可以灵活应对兼容参数并避免延迟于ollama的更新。

sunyuhan1998:
Hi @markpollack , I hope you're doing well!

I just wanted to check in and see if there are any further changes needed for this PR. Let me know if there's anything else I should address — happy to make adjustments if needed!

Thanks again for your time and review!

hbsjz-swl:
I would like to ask: If this PR is merged, will this feature be included in the snapshot version library? Then how can developers who are currently using the GA version use this feature?

sunyuhan1998:
> I would like to ask: If this PR is merged, will this feature be included in the snapshot version library? Then how can developers who are currently using the GA version use this feature?

It seems to me that this feature should be included in 1.1.x, as well as being backport to 1.0.x. By the time the 1.0.x version is officially released, users currently using the 1.0.0 GA version will be able to use it by upgrading to the 1.0.x version

markpollack:
yes, I will backport it.  Sorry about the delay.  Only now getting my project mgmt hat on full time @sunyuhan1998

sunyuhan1998:
> yes, I will backport it. Sorry about the delay. Only now getting my project mgmt hat on full time @sunyuhan1998

no problem at all! Please feel free to let me know if any changes are needed.

markpollack:
i'm still working through this.  exploring thinking more generally as an overall feature.

sunyuhan1998:
> i'm still working through this. exploring thinking more generally as an overall feature.

No problem. If there's any part I can contribute to, please feel free to let me know. I'd be happy to get involved.

hbsjz-swl:
@sunyuhan1998 Currently, there is no proposal that fully supports custom parameters. I have noticed that the same type of parameters provided by some models or framework providers are not entirely standardized. For example, Qwen3 uses "enable_thinking" for the thinking parameter, while Ollama uses "thinking." We should provide parameters that are entirely user-defined, such as using a Map, rather than defining fixed keys with a POJO

ilayaperumalg:
@sunyuhan1998 Sorry about the delay in getting this in. Could you rebase this PR from the latest main please? Thanks

sunyuhan1998:
> @sunyuhan1998 Sorry about the delay in getting this in. Could you rebase this PR from the latest main please? Thanks

Hi, I've rebased this branch onto the latest main branch and properly resolved all conflicts. 

I sincerely apologize for the messy commit history in my previous submissions. This was caused by an earlier merge of the main branch into the current branch, which led to some issues during the subsequent rebase process. However, everything has now been cleaned up, and the final code has been restored to a proper state.

Additionally, I'd like to clarify: the changes originally included in this PR to add `think` functionality support in `OllamaApi` have since been implemented and merged via #3969. As a result, after this rebase, the current PR retains only the changes related to `Options` and `OllamaChatModel`.

The PR has now been updated to its latest state and is ready for review. Thank you!

markpollack:
Hi.  Thank so much @sunyuhan1998   Regarding sending the previous thinking back, this comes up in other models as well now.  We can perhaps stick it in the metadata of the message for now.  The design of the messaging has changed, i think we can improve the model in 2.0, most leading models have 'thinking'.

spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/ChatOptions.java
```diff
@@ -83,6 +83,15 @@ public interface ChatOptions extends ModelOptions {
 	@Nullable
 	Double getTopP();
 
+	/**
+	 * Returns the think flag to use for the chat.
+	 * @return the think flag to use for the chat
+	 */
+	@Nullable
+	default Boolean getThink() {
+		return false;
```
markpollack, line 92: We can't add 'think' to default chat option as a simple boolean is not sufficient to capture the full range of options across models, e.g. there is often a thinking token budget or other parameters.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
```diff
@@ -198,6 +201,16 @@ public void setTopP(@Nullable Double topP) {
 		this.topP = topP;
 	}
 
+	@Override
+	@Nullable
+	public Boolean getThink() {
+		return this.think;
```
markpollack, line 207: As with ChatOptions, the boolean itself isn't sufficient across all model providers.

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaApiIT.java
```diff
@@ -146,4 +148,88 @@ public void think() {
 		assertThat(response.message().thinking()).isNotEmpty();
```
markpollack, line 148: this assertion isn't passing.

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaApiIT.java
```diff
@@ -146,4 +148,88 @@ public void think() {
...
+		Flux<ChatResponse> response = getOllamaApi().streamingChat(request);
+
+		List<ChatResponse> responses = response.collectList().block();
+		System.out.println(responses);
+
+		assertThat(responses).isNotNull();
+		assertThat(responses.stream()
+			.filter(r -> r.message() != null)
+			.map(r -> r.message().thinking())
+			.collect(Collectors.joining(System.lineSeparator()))).contains("Sofia");
```
markpollack, line 171: this assertion is failing for me.

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaApiIT.java
```diff
@@ -146,4 +148,88 @@ public void think() {
...
+		Flux<ChatResponse> response = getOllamaApi().streamingChat(request);
+
+		List<ChatResponse> responses = response.collectList().block();
+		System.out.println(responses);
+
+		assertThat(responses).isNotNull();
+		assertThat(responses.stream()
+			.filter(r -> r.message() != null)
+			.map(r -> r.message().thinking())
+			.collect(Collectors.joining(System.lineSeparator()))).contains("solar");
```
markpollack, line 197: this assertion is failing for me.

markpollack:
Moving to 1.1 RC1 due to the need to move 'think' out of the high level chat options interfaces and the debug the test failures.

shuanggegege:
、、

> Moving to 1.1 RC1 due to the need to move 'think' out of the high level chat options interfaces and the debug the test failures.

Hello, may I ask if this feature will no longer exist in version 1.1.0-M4

spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/ChatOptions.java
sunyuhan1998: Indeed, initially this attribute was designed solely to enable or disable Ollama's thinking. If we consider it to represent the entire content of "thinking" and intend to apply it across all models, it should be a complex object. I'm currently considering what properties we should include in it. Do you have any more specific suggestions?

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaApiIT.java
sunyuhan1998: I'm using the Qwen model, and it works fine on my side. Perhaps we need to adjust this assertion test, as the model's output has some randomness. I'll modify it soon.

ilayaperumalg:
Hi @shuanggegege , yes, this feature is targeted to get merged in 1.1.0-RC1 hopefully.

shuanggegege:
OK,I understand, thank you @ilayaperumalg

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
```diff
@@ -266,6 +266,7 @@ private ChatResponse internalCall(Prompt prompt, ChatResponse previousChatRespon
 				if (ollamaResponse.promptEvalCount() != null && ollamaResponse.evalCount() != null) {
 					generationMetadata = ChatGenerationMetadata.builder()
 						.finishReason(ollamaResponse.doneReason())
+						.metadata("thinking", ollamaResponse.message().thinking())
```
liugddx, line 269: Currently, metadata is only constructed when both promptEvalCount and evalCount exist, along with "thinking". This results in some responses containing thinking not being recorded due to missing count fields.

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
```diff
@@ -505,7 +506,8 @@ else if (message.getMessageType() == MessageType.TOOL) {
 		OllamaApi.ChatRequest.Builder requestBuilder = OllamaApi.ChatRequest.builder(requestOptions.getModel())
 			.stream(stream)
 			.messages(ollamaMessages)
-			.options(requestOptions);
+			.options(requestOptions)
+			.think(requestOptions.getThink());
```
liugddx, line 510: Enter null, ensure it is not serialized into a JSON field to trigger the server's default behavior. It is recommended to set it only when non-empty think
```
Boolean think = requestOptions.getThink();
if (think != null) requestBuilder.think(think);
```

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
liugddx: Placing "thinking" directly as a top-level key in ChatGenerationMetadata may conflict with future keys of the same name from other providers. It is recommended to consider using a namespace prefix (e.g., "ollama.thinking") or defining the key name in a public constant to avoid collisions.

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
```diff
@@ -51,6 +52,16 @@ public enum OllamaModel implements ChatModelDescription {
...
+	/**
+	 * Qwen3 1.7b
+	 */
+	QWEN_3_1_7_B("qwen3:1.7b"),
+
+	/**
+	 * Qwen3 0.6b
+	 */
+	QWEN_3_06B("qwen3:0.6b"),
+
```
liugddx, line 64: Among the existing constants, there is QWEN3_4B, while the newly added ones are QWEN_3_1_7_B and QWEN_3_06B. The naming styles are inconsistent (such as the use of underscores and the separation of numbers and units). It is recommended to adopt a unified style (for example, using QWEN3_1_7B / QWEN3_0_6B consistently or applying a uniform underscore scheme) to ensure readability and searchability.

spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/ChatOptions.java
liugddx: The semantics of Ollama are: when not explicitly specified (null) → the server embeds the thought process within the <thinking> tags in the content; explicitly false → thinking is disabled; explicitly true → separated into message.thinking.
The current ChatOptions interface provides a default implementation: `default Boolean getThink() { return false; }`. Although specific implementation classes have overridden this to return nullable values, if any implementation fails to override it in the future, "not set" will be treated as false, thereby altering the default behavior.

markpollack:
This PR has been reworked quite a bit, there are also thinking options for the openai open source model that aren't boolean in nature.  See the commit 0b8293ec969b87e049b806203af25d200d16c044 and the PR https://github.com/spring-projects/spring-ai/pull/4821   I hope it meets the requirements.  Thanks for all the feedback and patience.

PR #3390, opened 2025-05-30
Fix CustomerTools example to use existing methods

filiphr:
The `ToolContext` does not have a `get(String)` method. This PR is fixing the documentation to show the methods that can be used.

However, I think that it would be better if `ToolContext` had a `get(String)` method that would look like:

```java
public Object get(String key) {
    return this.context.get(key);
}
```

It could also be made generic:

```java
public <T> T get(String key) {
    return (T) this.context.get(key);
}
```

If you prefer to add the method, I can go ahead and create a PR for that instead


ilayaperumalg:
@filiphr Thanks for fixing the documentation to use the right method for ToolContext!

ilayaperumalg:
Back-ported into 1.0.x as cd364e39d

PR #3391, opened 2025-05-30
Run PR Check GHA on every branch and allow it in Forks

filiphr:
This allows contributors to verify that everything is running fine before doing a PR

I saw that in 6df84845a248a525b251450d1fcb4f6f1f6b30f1 there were some adjustments done to make sure that GHA are skipped on forks. I think that it could be useful if we adjust the pr-check GHA and allow it to run on every branch and for forks. Perhaps it needs a bit more dynamic one in order to avoid that it runs for Spring AI committers on every branch, or perhaps we can add a new action that would be like the pr-check, but it'll run when the repo owner is not spring-projects.


sobychacko:
@filiphr We don't usually enable PR GH action in other projects on the forks. We usually expect the PR authors to make sure that everything is clean locally before issuing the PR. People might not even notice that the action ran on the forks. If you think there is a valid reason for enabling this, we can merge it, but I was curious if we need to do so. Thanks!

filiphr:
>We usually expect the PR authors to make sure that everything is clean locally before issuing the PR

I agree with you about this one. I was doing some contributions in different areas and was switching between branches and of course I made some mistakes with the code format. So I enabled the Actions on my fork so I can check before doing a PR. However, I do understand your sentiment that it is good that people first run this locally. 

>People might not even notice that the action ran on the forks.

Actions are by default disabled when a fork is created. People have to explicitly enable them.

-----

In any case, if you are doing it differently for other projects, then let's keep that as is and close this PR without merging it.

sobychacko:
@filiphr Thanks for the feedback. We will close this PR, but monitor the community to see if it's a persistent issue, and if so, we will revisit later. Thanks again!

PR #3395, opened 2025-05-30
fix: Remove unreachable else-if branch in PromptChatMemoryAdvisor

liqingdong:
## Problem
In the `after` method of `PromptChatMemoryAdvisor`, there was an issue with the conditional logic where the `else-if` branch was unreachable. The original code structure had redundant null checks and didn't properly prioritize the single-result case check.

## Solution
Reordered the conditional statements to:
1. First check for single-result case with complete null safety
2. Then handle the multi-result case as the fallback

This approach is more efficient and maintains better null safety while handling both cases appropriately.

## Changes
```diff
- if (chatClientResponse.chatResponse() != null) {
-     assistantMessages = chatClientResponse.chatResponse()
-         .getResults()
-         .stream()
-         .map(g -> (Message) g.getOutput())
-         .toList();
- }
- else if (chatClientResponse.chatResponse() != null && 
-          chatClientResponse.chatResponse().getResult() != null &&
-          chatClientResponse.chatResponse().getResult().getOutput() != null) {
-     assistantMessages = List.of((Message) chatClientResponse.chatResponse().getResult().getOutput());
- }

+ if (chatClientResponse.chatResponse() != null && 
+     chatClientResponse.chatResponse().getResult() != null &&
+     chatClientResponse.chatResponse().getResult().getOutput() != null) {
+     assistantMessages = List.of((Message) chatClientResponse.chatResponse().getResult().getOutput());
+ }
+ else if (chatClientResponse.chatResponse() != null) {
+     assistantMessages = chatClientResponse.chatResponse()
+         .getResults()
+         .stream()
+         .map(g -> (Message) g.getOutput())
+         .toList();
+ }


ilayaperumalg:
@liqingdong Thanks for fixing the conditional logic!

PR #3396, opened 2025-05-30
Avoid looping through all the page tree if there is no destination page

torakiki:
Added a simple null check to avoid looping through all the page tree if currentPage is null and there is nothing to be found


ilayaperumalg:
@torakiki Thanks for the PR fixing the bug. Could you check the format issue in the build failure and update the PR?

PR #3397, opened 2025-05-30
Create new ToolChatOptions if input request has tool configuration without options

filiphr:
This fixes a problem where the tool information is not provided to the underlying client if the prompt does not have any chat options set.

e.g.

```java
chatClient.prompt().toolNames("myTool").call()
```

The example above will not provide the tool name to the underlying chat model through the request.

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
@@ -22,10 +22,14 @@ import java.net.URL;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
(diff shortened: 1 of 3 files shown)


srijit219 (approved):

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
```diff
@@ -685,7 +689,7 @@ public List<Media> getMedia() {
 			return this.media;
 		}
 
-		public List<String> getToolNames() {
```
markpollack, line 769: Set maybe more appropriate ,but at this point since it is GA, we can't change the API without doing a deprecation first.  Probably isn't worth the hastle to change to `Set`

filiphr: Is the `DefaultChatClientRequestSpec` part of the API? I changed it to `Set` since the public API is actually using a `Set`. e.g. in `DefaultToolCallingChatOptions` also the `ToolCallingChatOptions#toolNames`.

In any case, you are right. I have reverted that change

filiphr:
Seems like the failing test is failing on main as well.

filiphr:
Closing this since it has been resolved as part of #4748

PR #3402, opened 2025-05-31
fix for Mongo vectorStore.similaritySearch failing due to org.bson.ty…

Kota-SH:
…pes.ObjectId cannot be cast to class java.lang.String: issue#3318


ilayaperumalg:
@Kota-SH I couldn't see the issue with the use case posted in the original issue reported. Also, the `MongoDBAtlasVectorStoreIT` fails with this change. Closing this PR for now until we can replicate the issue. Thank you for the contribution!

Issue #3415, opened 2025-06-03
ResponseErrorHandler used in `RetryUtils` fails with Spring Framework 7.x


sobychacko:
## Problem
Our code breaks when users upgrade to Spring 7 because we override the deprecated `handleError(ClientHttpResponse)` method that gets completely removed.

**Current code:**
```java
@Override
public void handleError(@NonNull ClientHttpResponse response) throws IOException {
    // our error handling logic
}
```

**What changed:**
- Spring 6.2.x: Has both `handleError(response)` and `handleError(url, method, response)` 
- Spring 7.x: Removes the deprecated `handleError(response)` method entirely

## Solution
Implement both methods, but don't use `@Override` on the old one:

```java
@Override
public void handleError(URI url, HttpMethod method, @NonNull ClientHttpResponse response) throws IOException {
    handleError(response);  // delegate to existing logic
}

// No @Override annotation - works in both versions
public void handleError(@NonNull ClientHttpResponse response) throws IOException {
    // existing error handling logic stays here
}
```

**How it works:**
- When using Spring 6.2.x, we call the old method directly (no overhead)
- When using Spring 7.x, we call the new method, which delegates to our logic
- Same behavior, works with both versions

PR #3416, opened 2025-06-03
GH-3415: Add Spring Framework 7.x compatibility to RetryUtils#Respons…

sobychacko:
…eErrorHandler

Fixes: #3415

- Remove @Override annotation from handleError(ClientHttpResponse) method to support compilation against both Spring 6.2.x and 7.x
- Add delegation from new handleError(URI, HttpMethod, ClientHttpResponse) method to maintain backward compatibility
- Preserve existing error handling logic

This change allows Spring AI to work seamlessly across Spring Framework versions 6.x and 7.x without requiring reflection or version detection at runtime.


ilayaperumalg:
Cherry-picked into 1.0.x via 5ecfcce16

Issue #3418, opened 2025-06-03
Implement Guardrail Functionality for Content Validation


karsonto:
Description

We need to implement a Guardrail mechanism to intercept and block restricted user inputs before they reach core processing agents. This system should provide a flexible framework for defining validation rules and handling policy violations.

PR #3419, opened 2025-06-03
feat: Guardrail for input validation.

karsonto:
As mentioned in issue  #3418 This PR implements a guardrail mechanism to intercept restricted inputs before they reach core processing agents

### DIFF
--- a/spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/advisor/GuardrailAdvisorTests.java
+++ b/spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/advisor/GuardrailAdvisorTests.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


apappascs:
This is a breaking change—the constructor and builder now require `Predicate<String>` instead of `List<String>`
May consider overloading or deprecating the old methods?

karsonto:
May I implement a new GuardrailAdvisor to replace SafeGuardAdvisor without modifying the existing class?

apappascs:
> May I implement a new GuardrailAdvisor to replace SafeGuardAdvisor without modifying the existing class?

Thanks for the suggestion! I think this may require a bit more than just adding a predicate filter. Let me think a bit more on this and get back to you soon.

markpollack:
circling back and doing triage.  thoughts @apappascs ?

apappascs:
> circling back and doing triage.  thoughts @apappascs ?

Thanks for circling back. I had already shared my proposal with you and @tzolov, but let me summarize it more clearly here.  

My suggestion is to leverage the existing advisor chain so that guardrails can be executed as part of the normal sequence. In addition, I propose supporting a **parallel (grouped) execution model** for cases where multiple advisors/guardrails can run independently of each other.  

In practice, this would allow chain definitions like:

```
Sequential flow (today):

advisor1  -->  advisor2  -->  advisor3  -->  advisor4  -->  advisor5  -->  advisor6

Proposed flow with grouped/parallel advisors:

advisor1  -->  advisor2  -->  [ grouped_advisor ]
|---- advisor3
|---- advisor4
|---- advisor5
(all executed in parallel, results merged)

-->  advisor6  -->  ...

```

This way:
- We keep compatibility with the current sequential advisor model.  
- We gain flexibility to group and parallelize advisors where it makes sense (e.g. independent guardrails or validations).  
- The grouped advisor can collect results from its children and then pass a combined outcome to the next element in the chain.

PR #3433, opened 2025-06-03
chore: move speech1.mp3 to src/test/resources

jharnack2:
Moved a test mp3 file from src/main/resources to src/test/resources so its not included in the jar


sobychacko:
Merged via https://github.com/spring-projects/spring-ai/commit/ab88fea83e662c09904f42f2bf3096069c8f8852.

PR #3443, opened 2025-06-04
Fix: MySQL/MariaDB TIMESTAMP precision for message ordering (#3442)

### LINKED ISSUE #3442: MySQL/MariaDB schema TIMESTAMP precision causes message ordering issues in JdbcChatMemoryRepository
### Bug description

The `saveAll()` method in `JdbcChatMemoryRepository` attempts to guarantee message ordering using millisecond precision, but fails due to MySQL and MariaDB's default TIMESTAMP precision of 0 (seconds only).

Specifically, in the `AddBatchPreparedStatement` class:

1. `Instant.now().toEpochMilli()` initializes the current time in milliseconds
2. `AtomicLong.getAndIncrement()` attempts to guarantee ordering by incrementing each message by 1 millisecond
3. However, MySQL/MariaDB's default `TIMESTAMP` has precision 0, causing millisecond information to be lost
4. As a result, all messages stored within the same second have identical timestamps, causing random ordering

Message ordering must be accurately guaranteed for maintaining conversation context, but this is impossible with the current MySQL/MariaDB schema.

**saveAll**
```java
@Override
public void saveAll(String conversationId, List<Message> messages) {
	Assert.hasText(conversationId, "conversationId cannot be null or empty");
	Assert.notNull(messages, "messages cannot be null");
	Assert.noNullElements(messages, "messages cannot contain null elements");

	this.transactionTemplate.execute(status -> {
		deleteByConversationId(conversationId);
		this.jdbcTemplate.batchUpdate(this.dialect.getInsertMessageSql(),
				new AddBatchPreparedStatement(conversationId, messages));
		return null;
	});
}
```

**AddBatchPreparedStatement**
```java
private record AddBatchPreparedStatement(String conversationId, List<Message> messages,
		AtomicLong instantSeq) implements BatchPreparedStatementSetter {

	private AddBatchPreparedStatement(String conversationId, List<Message> messages) {
		this(conversationId, messages, new AtomicLong(Instant.now().toEpochMilli()));
	}

	@Override
	public void setValues(PreparedStatement ps, int i) throws SQLException {
		var message = this.messages.get(i);

		ps.setString(1, this.conversationId);
		ps.setString(2, message.getText());
		ps.setString(3, message.getMessageType().name());
		ps.setTimestamp(4, new Timestamp(this.instantSeq.getAndIncrement()));
	}

	@Override
	public int getBatchSize() {
		return this.messages.size();
	}
}
```

**schema-mysql**
```sql
CREATE TABLE IF NOT EXISTS SPRING_AI_CHAT_MEMORY (
    `conversation_id` VARCHAR(36) NOT NULL,
    `content` TEXT NOT NULL,
    `type` ENUM('USER', 'ASSISTANT', 'SYSTEM', 'TOOL') NOT NULL,
    `timestamp` TIMESTAMP NOT NULL,

    INDEX `SPRING_AI_CHAT_MEMORY_CONVERSATION_ID_TIMESTAMP_IDX` (`conversation_id`, `timestamp`)
);
```

**schema-mariadb**
```sql
CREATE TABLE IF NOT EXISTS SPRING_AI_CHAT_MEMORY (
    conversation_id VARCHAR(36) NOT NULL,
    content TEXT NOT NULL,
    type VARCHAR(10) NOT NULL,
    `timestamp` TIMESTAMP NOT NULL,
    CONSTRAINT TYPE_CHECK CHECK (type IN ('USER', 'ASSISTANT', 'SYSTEM', 'TOOL'))
);

CREATE INDEX IF NOT EXISTS SPRING_AI_CHAT_MEMORY_CONVERSATION_ID_TIMESTAMP_IDX
ON SPRING_AI_CHAT_MEMORY(conversation_id, `timestamp`);
```

### Proposed solution
```sql
-- Current (schema-mysql.sql, schema-mariadb.sql):
`timestamp` TIMESTAMP NOT NULL

-- Proposed fix:
`timestamp` TIMESTAMP(3) NOT NULL
```

seungwone:
Fixes #3442

The `AddBatchPreparedStatement` class uses `AtomicLong.getAndIncrement()` to create sequential millisecond timestamps, but MySQL/MariaDB's default TIMESTAMP precision (0) truncates this information, causing message ordering issues.

Added fractional seconds precision (TIMESTAMP(3)) to MySQL and MariaDB schemas to support millisecond-based message ordering in `JdbcChatMemoryRepository`.

### Changes
- `schema-mysql.sql`: Changed `TIMESTAMP` to `TIMESTAMP(3)`
- `schema-mariadb.sql`: Changed `TIMESTAMP` to `TIMESTAMP(3)`


sunyuhan1998 (approved):
I think this PR itself is fine, but as @markpollack previously mentioned: 

https://github.com/spring-projects/spring-ai/pull/3154#issuecomment-2884724357

this field is actually being used as an auto-incrementing sequence to ensure message ordering, rather than representing time in a temporal sense. Perhaps we should consider renaming this field in the future.

YunKuiLu:
There is a common belief that `TIMESTAMP` can only be used until 2038. Perhaps it should be changed to `BIGINT` or `DATETIME`. 

https://stackoverflow.com/questions/409286/should-i-use-the-datetime-or-timestamp-data-type-in-mysql

markpollack:
yea, I'm not sure what to do here in a few ways.

1. I think the 1.1 line should drop the timestamp completely.
2. if we change the schema 'just like that' in a 1.0.1 release, stuff breaks - though one can argue of course it is already broken.

As for point 2, a migration script can be provided to run to upgrade to 1.0.1 and advertised in the release notes etc.

I'm thinking for the 1.0.x line, we provide the migration script.

thoughts @YunKuiLu @sunyuhan1998 @seungwone  @ThomasVitale   ?

YunKuiLu:
> yea, I'm not sure what to do here in a few ways.
> 
> 1. I think the 1.1 line should drop the timestamp completely.
> 2. if we change the schema 'just like that' in a 1.0.1 release, stuff breaks - though one can argue of course it is already broken.
> 
> As for point 2, a migration script can be provided to run to upgrade to 1.0.1 and advertised in the release notes etc.
> 
> I'm thinking for the 1.0.x line, we provide the migration script.
> 
> thoughts @YunKuiLu @sunyuhan1998 @seungwone @ThomasVitale ?

Maybe we can do something like this:

1. In the `1.0.1-SNAPSHOT` version, provide a migration script and support both schemas at the same time. When constructing the `JdbcChatMemoryRepository`, we can check the current schema. If the table in the DB still uses `TIMESTAMP` as the column type, we log a warn, and include a link to the upgrade guide.

2. In the `1.0.x` release version, we do the same check. If `TIMESTAMP` is detected, we log an error instead, and again include a link to the migration guide. Also, users would need to explicitly set the flag `-Dspring.ai.chat.memory.repository.jdbc.compatible-version=1.0.0` to bypass the error — otherwise, the app won't start.

3. In `1.1.x`, if the schema still has the timestamp column with type `TIMESTAMP`, we simply block startup and show an error message with a link to the upgrade guide.

These are just my initial thoughts though — not sure if we really need this many steps.

markpollack:
I think I found an easier way, since the values of the timestamps are meaningless, they just need to increase, we make sure that when we are assigning the timestamp, it is always increases at an appropriate level of granulatirty - eg. one second or something like that.

sobychacko:
@seungwone Thanks for working on this! After reviewing the discussion and based on Mark's suggestion about fixing the granularity in code, here is a more fine-grained approach that doesn't require any schema breaking changes in `1.1.x` or `1.0.x`:

  Plan:
  - Implement a code-only fix using second-level granularity (as Mark suggested)
  - Apply this non-breaking change to both `1.1.x` and `1.0.x`
  - No schema changes required
  - We'll implement the fix in `1.1.x` first, then backport to `1.0.x`

This approach avoids migration complexity for existing users while solving the ordering issue across all databases.

Here is the new PR: https://github.com/spring-projects/spring-ai/pull/4739

Also closing this PR in preference to the new one.

I created an [issue](https://github.com/spring-projects/spring-ai/issues/4740) for tracking all the breaking changes we discussed above (dropping the timestamp column altogether and using a sequence_id column) for the 2.0.x timeline.

PR #3448, opened 2025-06-05
perf(logging): optimize debug logging in PromptChatMemoryAdvisor

Ahoo-Wang:
- Enclosed debug logging statements within an isDebugEnabled() check
- Reduced unnecessary method calls and object creations when debug logging is disabled


ilayaperumalg:
@Ahoo-Wang Thanks for the code cleanup!

PR #3454, opened 2025-06-05
add similarity threshold capability to VectorStoreChatMemoryAdvisor

scionaltera:
This advisor did not have support for setting the similarity threshold, so it defaulted to returning every document regardless of similarity. In a conversation where the user has changed to a different topic the result was returning a bunch of documents with no relevance to the user's latest input.

I experimented with setting a value greater than 0 and liked the results because it would begin to include results more relevant to the conversation. I thought it would be nice to contribute the change back.

I left the default at 0 so as not to change the earlier behavior.


advisors/spring-ai-advisors-vector-store/src/main/java/org/springframework/ai/chat/client/advisor/vectorstore/VectorStoreChatMemoryAdvisor.java
```diff
@@ -88,14 +94,17 @@ public final class VectorStoreChatMemoryAdvisor implements BaseChatMemoryAdvisor
 	private final VectorStore vectorStore;
 
 	private VectorStoreChatMemoryAdvisor(PromptTemplate systemPromptTemplate, int defaultTopK,
-			String defaultConversationId, int order, Scheduler scheduler, VectorStore vectorStore) {
+			double defaultSimilarityThreshold, String defaultConversationId, int order, Scheduler scheduler,
+			VectorStore vectorStore) {
 		Assert.notNull(systemPromptTemplate, "systemPromptTemplate cannot be null");
 		Assert.isTrue(defaultTopK > 0, "topK must be greater than 0");
+		Assert.isTrue(defaultSimilarityThreshold >= 0, "similarityThreshold must be equal to or greater than 0");
```
YunKuiLu: According to the assertion of method`org.springframework.ai.vectorstore.SearchRequest.Builder#similarityThreshold`.
The `defaultSimilarityThreshold` should be `defaultSimilarityThreshold >= 0 && defaultSimilarityThreshold <= 1`
```suggestion
		Assert.isTrue(defaultSimilarityThreshold >= 0 && defaultSimilarityThreshold <= 1, "similarityThreshold must be in [0,1] range");
```

advisors/spring-ai-advisors-vector-store/src/main/java/org/springframework/ai/chat/client/advisor/vectorstore/VectorStoreChatMemoryAdvisor.java
```diff
@@ -257,6 +275,17 @@ public Builder defaultTopK(int defaultTopK) {
 			return this;
 		}
 
+		/**
+		 * Set the similarity threshold for retrieving relevant documents.
+		 * @param defaultSimilarityThreshold the required similarity for documents to
+		 * retrieve
+		 * @return this builder
+		 */
+		public Builder defaultSimilarityThreshold(Double defaultSimilarityThreshold) {
```
YunKuiLu, line 284: Maybe we don't need to use the Double wrapper class here?

advisors/spring-ai-advisors-vector-store/src/main/java/org/springframework/ai/chat/client/advisor/vectorstore/VectorStoreChatMemoryAdvisor.java
```diff
@@ -221,6 +237,8 @@ public static class Builder {
 
 		private Integer defaultTopK = DEFAULT_TOP_K;
 
+		private Double defaultSimilarityThreshold = DEFAULT_SIMILARITY_THRESHOLD;
```
YunKuiLu, line 240: Same comment as above.

YunKuiLu (commented):
LGTM! It would be great if these minor issues could be fixed.

scionaltera:
> LGTM! It would be great if these minor issues could be fixed.

Sure thing. I've been sick the past couple of days but I can get these changes in soon.

PR #3470, opened 2025-06-08
Ollama: Support custom Ollama API paths

lambochen:

---

Hi, This PR is a feature enhancement change that supports custom Ollama API paths.

Background: When implementing Spring AI in my project, I made some adjustments to the Ollama API, and the API provided externally is a custom path. Therefore, I expect the Spring AI Ollama Model to support custom paths.

Current behavior: The Ollama API path is fixed in the hard-coded code, for example:
```java
public ListModelResponse listModels() {
	return this.restClient.get()
			.uri("/api/tags")
			.retrieve()
			.body(ListModelResponse.class);
}
```

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-ollama/src/main/java/org/springframework/ai/model/ollama/autoconfigure/OllamaConnectionDetails.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-ollama/src/main/java/org/springframework/ai/model/ollama/autoconfigure/OllamaConnectionDetails.java
@@ -22,9 +22,24 @@ import org.springframework.boot.autoconfigure.service.connection.ConnectionDetai
  * Connection details for an Ollama service.
  *
  * @author Eddú Meléndez
+ * @author lambochen
  */
 public interface OllamaConnectionDetails extends ConnectionDetails {
 
 	String getBaseUrl();
+
+	String getChatPath();
+
+	String getEmbedPath();
+
+	String getListModelsPath();
+
+	String getShowModelPath();
+
+	String getCopyModelPath();
+
+	String getDeleteModelPath();
+
+	String getPullModelPath();
 
 }
--- a/auto-configurations/models/spring-ai-autoconfigure-model-ollama/src/test/java/org/springframework/ai/model/ollama/autoconfigure/OllamaApiBuilderTests.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-ollama/src/test/java/org/springframework/ai/model/ollama/autoconfigure/OllamaApiBuilderTests.java
@@ -1,0 +1,68 @@
+package org.springframework.ai.model.ollama.autoconfigure;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.ai.ollama.api.OllamaApi;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
+
+/**
+ * @author lambochen
+ */
+public class OllamaApiBuilderTests {
+
+	@Test
+	void fullBuilder() {
+		OllamaApi ollamaApi = OllamaApi.builder()
+			.baseUrl("http://localhost:11434")
+			.chatPath("/api/chat")
+			.embedPath("/api/embeddings")
+			.listModelsPath("/api/tags")
+			.showModelPath("/api/show")
+			.copyModelPath("/api/copy")
+			.deleteModelPath("/api/delete")
+			.pullModelPath("/api/pull")
+			.build();
+
+		assertThat(ollamaApi).isNotNull();
+	}
+
+	@Test
+	void defaultBuilder() {
+		OllamaApi ollamaApi = OllamaApi.builder().build();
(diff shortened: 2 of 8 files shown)


hongling-dev:
ref: https://github.com/spring-projects/spring-ai/issues/3471

hongling-dev:
@ThomasVitale @dev-jonghoonpark  
Hi, please review the PR, thank you.

auto-configurations/models/spring-ai-autoconfigure-model-ollama/src/test/java/org/springframework/ai/model/ollama/autoconfigure/OllamaApiBuilderTests.java
```diff
@@ -0,0 +1,68 @@
+package org.springframework.ai.model.ollama.autoconfigure;
```
yuluo-yx, line 17: Pls add code license header

dev-jonghoonpark:
Just out of personal curiosity, Can you be more specific with your example?
I've seen the OpenAI API compatible API many times, but this is the first time I've seen the
Ollama API compatible format.

hongling-dev:
> but this is the first time I've seen the Ollama API compatible format.

Thank you for your interest in this capability.
I am working on an experimental service where ollama is deployed remotely (non-local environment), providing platform services to developers. Due to the numerous APIs exposed by the platform, there has been an issue of path conflicts, so it is necessary to support customizing the Ollama API path.

> I've seen the OpenAI API compatible API many times

Yes, for similar solutions, we also have custom path requirements for OpenAI and other models (such as DeepSeek).

auto-configurations/models/spring-ai-autoconfigure-model-ollama/src/main/java/org/springframework/ai/model/ollama/autoconfigure/OllamaConnectionDetails.java
```diff
@@ -22,9 +22,24 @@
...
+
+	String getListModelsPath();
+
+	String getShowModelPath();
+
+	String getCopyModelPath();
+
+	String getDeleteModelPath();
+
+	String getPullModelPath();
```
eddumelendez, line 43: I don't think the connection details should be updated. The connection details to connect to ollama is the baseUrl only.

yuluo-yx: yep, these should be user-defined, but it would be helpful if the sub path definitions were added at the framework level, although this may cause configuration redundancy.

hongling-dev: Hi, @eddumelendez  I agree with your point, it is indeed possible to run the process through baseUrl, even though there might be conflicts with other API interface paths as I mentioned earlier.

Actually, I'm using it this way right now, as @yuluo-yx said, it will add more redundant configuration. The specific implementation method is: add a special prefix (custom_ollama_path) to all ollama interfaces, that is baseUrl={xxx}/{custom_ollama_path}/{ollama_methods}.

For this PR, supporting user-defined various paths eliminates redundant configuration issues. The default values remain consistent with previous versions, and users without custom requirements will not face additional burdens.

Finally, special thanks for your attention and review. thank you.  @eddumelendez  @yuluo-yx

ThomasVitale:
@lambochen thanks for raising this issue. If I understood the situation correctly, this would seem more a platform issue than a Spring AI issue. Since the Ollama service is provided on-premises by an internal platform, I wonder if the path conflict issue might be fixed more easily by exposing Ollama via a subpath on the platform? I assume the Ollama endpoints will not change (e.g. `/api/tags`). But the base URL can be customised based on how the platform serves Ollama. 

For example, it could be served by the platform from `https://my-company-platform.example.net/ollama` (which would be configured in the application via `spring.ai.ollama.base-url`). The full endpoint would then be `https://my-company-platform.example.net/ollama/api/tags` in the case of tags, which would already work with the current version of Spring AI.

Could that work?

hongling-dev:
> @lambochen thanks for raising this issue. If I understood the situation correctly, this would seem more a platform issue than a Spring AI issue. Since the Ollama service is provided on-premises by an internal platform, I wonder if the path conflict issue might be fixed more easily by exposing Ollama via a subpath on the platform? I assume the Ollama endpoints will not change (e.g. `/api/tags`). But the base URL can be customised based on how the platform serves Ollama.
> 
> For example, it could be served by the platform from `https://my-company-platform.example.net/ollama` (which would be configured in the application via `spring.ai.ollama.base-url`). The full endpoint would then be `https://my-company-platform.example.net/ollama/api/tags` in the case of tags, which would already work with the current version of Spring AI.
> 
> Could that work?

@ThomasVitale  Yes, your understanding is correct, it is indeed a platform issue (or rather, the platform's configuration does not match the existing Ollama path well).
The suggestion you proposed to add a unified prefix on the platform side is also feasible and can work properly.

If Spring AI supports custom paths, it can better integrate with existing platforms from a user's perspective when the platform is not compatible with Ollama's existing paths.

PR #3478, opened 2025-06-09
Fixed the logic error in the validateToolContextSupport method of MethodToolCallback

### LINKED ISSUE #3466: Wrong implementation of MethodToolCallback.validateToolContextSupport
Consider the current implementation of validateToolContextSupport compared from 1.0.0 and 1.0.0-M6

```
	private void validateToolContextSupport(@Nullable ToolContext toolContext) {
		var isNonEmptyToolContextProvided = toolContext != null && !CollectionUtils.isEmpty(toolContext.getContext());
		var isToolContextAcceptedByMethod = Stream.of(this.toolMethod.getParameterTypes())
			.anyMatch(type -> ClassUtils.isAssignable(type, ToolContext.class));
		if (isToolContextAcceptedByMethod && !isNonEmptyToolContextProvided) {
			throw new IllegalArgumentException("ToolContext is required by the method as an argument");
		}
	}
```

M6
```
	private void validateToolContextSupport(@Nullable ToolContext toolContext) {
		var isToolContextRequired = toolContext != null && !CollectionUtils.isEmpty(toolContext.getContext());
		var isToolContextAcceptedByMethod = Stream.of(toolMethod.getParameterTypes())
			.anyMatch(type -> ClassUtils.isAssignable(type, ToolContext.class));
		if (isToolContextRequired && !isToolContextAcceptedByMethod) {
			throw new IllegalArgumentException("ToolContext is not supported by the method as an argument");
		}
	}
```

1. Both have the same serious bug also refered in #3355 , it seems that the writer does not know the semantics of **ClassUtils.isAssignable(left, right)** => right is assinable to left, so left is the "super" type. All this code has the arguments in incorrect order.

2. More, in the case of a toolContext not being provided, the change in the validation logic now make it fail due to error 1
**M6** : No context 
 => isToolContextRequired  = false  
=> isToolContextRequired && !isToolContextAcceptedByMethod = false => **no exception**
**1.0.0** : No context 
=> isNonEmptyToolContextProvided = false + isNonEmptyToolContextProvided = true (because of error 1) 
=> isToolContextAcceptedByMethod && !isNonEmptyToolContextProvided = true 
=> **exception is thrown !**

sunyuhan1998:
Fixes #3466 

As mentioned in the PR, there were some issues with the implementation of the `validateToolContextSupport` method in `MethodToolCallback`, which led to validation failures. This PR makes the following adjustments:  

1. Adjusted the parameter order when calling `ClassUtils.isAssignable` to ensure correct semantics.  
2. Modified the condition for throwing exceptions to align with the method's intended functionality.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallback.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallback.java
@@ -118,8 +118,8 @@ public final class MethodToolCallback implements ToolCallback {
 	private void validateToolContextSupport(@Nullable ToolContext toolContext) {
 		var isNonEmptyToolContextProvided = toolContext != null && !CollectionUtils.isEmpty(toolContext.getContext());
 		var isToolContextAcceptedByMethod = Stream.of(this.toolMethod.getParameterTypes())
-			.anyMatch(type -> ClassUtils.isAssignable(type, ToolContext.class));
-		if (isToolContextAcceptedByMethod && !isNonEmptyToolContextProvided) {
+			.anyMatch(type -> ClassUtils.isAssignable(ToolContext.class, type));
+		if (isNonEmptyToolContextProvided && !isToolContextAcceptedByMethod) {
 			throw new IllegalArgumentException("ToolContext is required by the method as an argument");
 		}
 	}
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
Rebased and merged as 59d26610e. Also, back-ported the fix into 1.0.x via 2ca1be2b8

Issue #3420, opened 2025-06-03
Qdrant vector source unsupport null value in metadata


lanyuanxiaoyao:
**Bug description**
NullPointerException when null value in metadata.
`toObjectMap`use `Collectors.toMap` to create map, but it unsupport null value. Qdrant allow null value in metadata.
[it may help this](https://stackoverflow.com/questions/24630963/nullpointerexception-in-collectors-tomap-with-null-entry-values)

https://github.com/spring-projects/spring-ai/blob/094eeab405ff9518ea9722226f18faeef33c7f3f/vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantVectorStore.java#L280

https://github.com/spring-projects/spring-ai/blob/094eeab405ff9518ea9722226f18faeef33c7f3f/vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantObjectFactory.java#L44

**Environment**
Spring AI 1.0.0

WOONBE:
I'd like to contribute to this!

ilayaperumalg:
Addressed via #3483

PR #3483, opened 2025-06-09
Refactor QdrantObjectFactory to allow null metadata values

WOONBE:
related issue : #3420 
When retrieving documents from a Qdrant vector store, a NullPointerException was being thrown if the metadata payload for any document contained a field with a null value.

I fix the issue by replacing the stream-based Collectors.toMap() implementation with a more robust and explicit for-each loop.

I have confirmed the fix by performing tests in my local environment.

### DIFF
--- a/vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantObjectFactory.java
+++ b/vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantObjectFactory.java
@@ -16,6 +16,7 @@
 
 package org.springframework.ai.vectorstore.qdrant;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.stream.Collectors;
 
@@ -30,6 +31,7 @@ import org.springframework.util.Assert;
  * Utility methods for building Java objects from io.qdrant.client.grpc.JsonWithInt.Value.
  *
  * @author Anush Shetty
+ * @author Heonwoo Kim
  * @since 0.8.1
  */
 final class QdrantObjectFactory {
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
@WOONBE Thanks for the PR! Could you add test(s) to validate the fix please?

WOONBE:
> @WOONBE Thanks for the PR! Could you add test(s) to validate the fix please?

@ilayaperumalg Regarding the unit test for this change, I couldn't find an existing test class for QdrantObjectFactory where this new test case would naturally fit.
Please let me know if I should create a new test file, QdrantObjectFactoryTests.java, or if there's a more suitable existing file you'd prefer I add it to. I'm happy to add the test case accordingly.

ilayaperumalg:
@WOONBE Please feel free to add a new test class for the unit test. You could also add a test case in our ITs. Thanks again for looking into this!

WOONBE:
> @WOONBE Please feel free to add a new test class for the unit test. You could also add a test case in our ITs. Thanks again for looking into this!

@ilayaperumalg Thank you for the guidance!

I've pushed a new commit that adds a test class with a unit test to cover this fix.

Please let me know if there are any improvements or if the format needs any adjustments. I'd be happy to make further changes.

WOONBE:
@ilayaperumalg My apologies, it seems I missed running the code formatter before submitting the initial PR.

I saw the "Formatting violations" error in the PR checks for the test file I added. I have now run mvn spring-javaformat:apply and pushed the updated code to resolve this.

Aside from this, please let me know if there's anything else I might have missed or any other feedback you have. 
Thank you!

ilayaperumalg:
Rebased and merged as 8b45f4113. Also, cherry-picked into 1.0.x via 8053ae39c

ilayaperumalg:
@WOONBE Thank you once again for the PR and following it up with the test update.

Issue #3486, opened 2025-06-09
Documentation: The example code for Vision in the OpenAI reference documentation is incorrect.


NanKisu:
**Bug description**
![Image](https://github.com/user-attachments/assets/698065d0-e5dd-4d8c-918f-59b7cf0c9102)
[https://docs.spring.io/spring-ai/reference/api/chat/openai-chat.html#_vision](https://docs.spring.io/spring-ai/reference/api/chat/openai-chat.html#_vision)
In the 1.0.0 version of the documentation, the Vision example shown is actually from version 1.1.0-SNAPSHOT. Shouldn't the example below be included instead if it's truly the 1.0.0 documentation?

	var imageData = new ClassPathResource("/test.png");

	var userMessage = UserMessage.builder()
		.text("Explain what do you see on this picture?")
		.media(List.of(new Media(MimeTypeUtils.IMAGE_PNG, imageData)))		
		.build();

	var response = this.chatModel
		.call(new Prompt(List.of(userMessage), OpenAiChatOptions.builder().model(modelName).build()));

dev-jonghoonpark:
@NanKisu
It seems that the PR has been merged.
If that's right, please close the issue. : )

NanKisu:
@dev-jonghoonpark Thank you for notice.
https://docs.spring.io/spring-ai/reference/api/chat/openai-chat.html#_vision
The PR has been merged. But, the page still show incorrect example codes. 
Is there anything else I should do?

dev-jonghoonpark:
You're right.
I thought it had been completed since it appeared as `merged`.

It seems the changes were lost after the commit was merged.
We'll have to wait for the maintainer's response.

PR #3487, opened 2025-06-09
docs: update OpenAI page Vision section

NanKisu:
**📄 Documentation Bug**

**Problem**:
#3486 
In the 1.0.0 version of the documentation, the Vision example shown is actually from version 1.1.0-SNAPSHOT. Shouldn't the example below be included instead if it's truly the 1.0.0 documentation?

**Suggested Fix:**
Update OpenAI page Vision section

**Version:** Spring AI 1.0.0


dev-jonghoonpark:
There is a link in the body, but the markdown seems to be handled incorrectly.
So I rewrite the link: https://github.com/spring-projects/spring-ai/issues/3486 (so that the PR record is reflected on the issue page as well).

NanKisu:
I need to update the commit author information, so I'll force-push a revised version. Sorry for the inconvenience!

PR #3489, opened 2025-06-09
Remove `Object...` varargs overloads in `FilterExpressionBuilderTests`

### LINKED ISSUE #3482: Generic problems with FilterExpressionBuilder
**Bug description**
```java
List<String> params =List.of("v1","v2");
new FilterExpressionBuilder().in("param", params ).build();
```
I don't know if this is considered a bug,When I do create a filter this way,Parameters of type List are treated as Object parameters.

Image
I think it is somewhat misleading
**Environment**
java 17.x 
windows11 pro
spring AI 1.0.0

therepanic:
When we swipe `List` where in the generic we don't have `Object`, we hit `Object...` which is wrong, and we make a double list. That is, one element is added in this case, which is fundamentally wrong.

There are two ways to fix this. Either remove overloading with `Object...` (as I did), or make `FilterExpressionBuilderTests` generic and make `List<T>` values in the argument of acceptance. However, both changes **are breaking**.

If the second option is better for you to fix the problem I am ready to discuss it as I like both options I have given.

_The test I gave will not work without changes in the class itself._

Fix: #3482

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionBuilder.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


Matthiasvanderhallen:
Isn't the better solution to have `List<?>` as the argument type for the List overload instead? You can keep the varargs `Object` overload, but any single list passed in hits the `List<?>` overload as (likely) intended:

```java
    static void test(Object... values) {
        System.out.println (1);
    }

    static void test(List<?> values) {
        System.out.println (2);
    }

    public static void main() {
        List<String> v = List.of("a","b","c");
        test(v);//prints 2
    }
```

PR #3498, opened 2025-06-10
build: Use swagger-annotations-jakarta instead of swagger-annotations

pkubowicz:
Functionality is the same, but modern applications won't use javax variant of Swagger JARs, so the dependency on swagger-annotations caused duplicate classes on the classpath.

spring-ai is built against Spring Boot 3.4, which is jakarta-only.

See:
* https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Getting-started
* modern applications will avoid https://mvnrepository.com/artifact/io.swagger.core.v3/swagger-core/2.2.25 and use https://mvnrepository.com/artifact/io.swagger.core.v3/swagger-core-jakarta/2.2.25 instead

### DIFF
--- a/models/spring-ai-huggingface/pom.xml
+++ b/models/spring-ai-huggingface/pom.xml
@@ -47,11 +47,11 @@
 			<version>${project.parent.version}</version>
 		</dependency>
 
-		<!-- https://mvnrepository.com/artifact/io.swagger.core.v3/swagger-annotations -->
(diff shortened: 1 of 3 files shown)


yuluo-yx (approved):

markpollack:
Thanks, this was in fact an oversight!

markpollack:
merged in 3e02edc2dbcf8846a09b0457f37cdb8e8067caa2 and backported.

Issue #3499, opened 2025-06-10
Is it supported to expose MCP tools by adding spring-ai-starter-mcp-server-webflux to a legacy WebMVC app? No endpoint GET /sse error


openai0229:
**Bug description**
When adding the spring-ai-starter-mcp-server-webflux starter to an existing Spring WebMVC application to expose MCP tools, accessing the /see or /sse endpoints results in an error. The actual exception is:
```
org.springframework.web.servlet.NoHandlerFoundException: No endpoint GET /sse.
    at org.springframework.web.servlet.DispatcherServlet.noHandlerFound(DispatcherServlet.java:1305)
    at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1067)
    at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
    at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
    at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
    at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:564)
    at org.springframework.web.servlet.FrameworkServlet.service(HttpServlet.java:885)
    at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:658)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:195)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:140)
    at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:51)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:164)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:140)
```
I am not sure if this is a bug or a misconfiguration on my side. Any help or guidance would be appreciated.
Is this kind of integration supported? Or is there any official recommendation for integrating MCP into an existing WebMVC application?

**Environment**
- spring-ai-starter-mcp-server-webflux version: 1.0.0
- Spring Boot version: 3.4.5
- Java version: 17

**Steps to reproduce**
1. Add the spring-ai-starter-mcp-server-webflux dependency to an existing Spring WebMVC application.
2. Start the project.
3. Visit /see or /sse in a browser.
[...]

Issue #2732, opened 2025-04-14
我用spring ai实现一个mcp server.服务起来后，一直无法访问/sse端点


shirehappy:
Please do a quick search on GitHub issues first, there might be already a duplicate issue for the one you are about to create.
If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

**Bug description**
我用spring ai 实现 mcp server ，如果采用stdio的方式
打成jar包，然后用charry studio 配置，能够正常访问

但是通过spring ai 实现mcp server ，采用sse方式
服务正常启动，但是浏览器死活访问不到sse节点
localhost:8080/sse
通过postman也无法访问（控制台没有任何报错信息）

问题的原因是什么？

**Environment**
pom.xml:

    <dependency>
      <groupId>org.springframework.ai</groupId>
      <artifactId>spring-ai-mcp-server-webmvc-spring-boot-starter</artifactId>
      <version>1.0.0-M6</version>
    </dependency>

    <dependency>
      <groupId>org.springframework.ai</groupId>
      <artifactId>spring-ai-starter-mcp-server-webmvc</artifactId>
      <version>1.0.0-M7</version>
    </dependency>

**Steps to reproduce**
项目结构非常简单，服务也是正常启动
![Image](https://github.com/user-attachments/assets/4149dde8-b31a-42ec-86aa-6de82c01d1e4)
但是页面上就是一直无法访问：
localhost:8080/sse
通过postman也无法访问

**Expected behavior**

**Minimal Complete Reproducible example**
Please provide a failing test or a minimal complete verifiable example that reproduces the issue.
Bug reports that are reproducible will take priority in resolution over reports that are not reproducible.

shirehappy:
![Image](https://github.com/user-attachments/assets/76eeecf1-b11d-4841-9307-c1e9edd0692c)
配置也很简单

就是页面请求不到

![Image](https://github.com/user-attachments/assets/219d5cc4-07b5-42c2-9df3-a2494ad8daa9)

hymmyh:
为什么 1.0.0-M6 和 1.0.0-M7 混用，只用1.0.0-M7不就可以？

mengyanb:
sse use this spring-ai-starter-mcp-server-webflux

fireman-ace:
sse use this <dependency>
            <groupId>org.springframework.ai</groupId>
            <artifactId>spring-ai-starter-mcp-server-webflux</artifactId>
[...]

PR #3511, opened 2025-06-11
Improved the documentation in the MCP Server Boot Starter regarding Starter selection

sunyuhan1998:
Spring AI provides three starters for MCP servers: `spring-ai-starter-mcp-server`, `spring-ai-starter-mcp-server-webmvc`, and `spring-ai-starter-mcp-server-webflux`, among which `spring-ai-starter-mcp-server-webflux` is designed for reactive projects. 

However, due to Spring Boot's default behavior, if both `org.springframework.web.servlet.DispatcherServlet` and `org.springframework.web.reactive.DispatcherHandler` are present in the classpath, Spring Boot will prioritize the former:

https://github.com/spring-projects/spring-boot/blob/0daf3f5897a9bad169ad7ebad6d79e4260a24822/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/WebApplicationType.java#L60-L71

This means that if users mistakenly use `spring-ai-starter-mcp-server-webflux` in a project built with `spring-boot-starter-web`, MCP endpoints such as `/sse` will become non-functional.

Of course, this is not an issue with Spring AI itself. However, considering that it has already caused considerable confusion :

* https://github.com/spring-projects/spring-ai/issues/3499
* https://github.com/spring-projects/spring-ai/issues/2732

 I believe adding appropriate documentation would help users better understand what they are doing and avoid such pitfalls.

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/mcp/mcp-server-boot-starter-docs.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/mcp/mcp-server-boot-starter-docs.adoc
@@ -80,7 +80,12 @@ The starter activates the `McpWebFluxServerAutoConfiguration` and `McpServerAuto
 * Optional `STDIO` transport (enabled by setting `spring.ai.mcp.server.stdio=true`)
 * Included `spring-boot-starter-webflux` and `mcp-spring-webflux` dependencies
 
-== Configuration Properties
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


sunyuhan1998:
Hi @ilayaperumalg what do you think of this?

sunyuhan1998:
> BTW, I think this can solve the problem you are facing now, or we can also solve it through configuration?
> 
> such as `spring.main.web-application-type=reactive`. Should this be added to the documentation?

Yes, we can indeed solve this issue through this configuration. I had also considered documenting this configuration, but ultimately, I thought that if users are using other components in their projects that depend on `WebMvc`, applying this configuration might lead to some unintended side effects. Therefore, the best approach may be to guide users to use the appropriate `Starter`, rather than using `spring-ai-starter-mcp-server-webflux` in a `WebMvc` project, since doing so would be meaningless.

PR #3520, opened 2025-06-12
fix(MessageChatMemoryAdvisor): Fix incorrect prompt position for system messages

Ahoo-Wang:
- First position of prompt to insert system message
- Prioritize system messages in conversation history

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-elevenlabs/pom.xml
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-elevenlabs/pom.xml
@@ -1,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xmlns="http://maven.apache.org/POM/4.0.0"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.springframework.ai</groupId>
+		<artifactId>spring-ai-parent</artifactId>
+		<version>1.1.0-SNAPSHOT</version>
+		<relativePath>../../../pom.xml</relativePath>
+	</parent>
+	<artifactId>spring-ai-autoconfigure-model-elevenlabs</artifactId>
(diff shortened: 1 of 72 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
```diff
@@ -83,8 +86,18 @@ public ChatClientRequest before(ChatClientRequest chatClientRequest, AdvisorChai
...
 		// 2. Advise the request messages list.
-		List<Message> processedMessages = new ArrayList<>(memoryMessages);
-		processedMessages.addAll(chatClientRequest.prompt().getInstructions());
+		SystemMessage systemMessage = chatClientRequest.prompt().getSystemMessage();
+		List<Message> processedMessages = new ArrayList<>();
+		if (StringUtils.hasText(systemMessage.getText())) {
+			processedMessages.add(systemMessage);
+		}
+		processedMessages.addAll(memoryMessages);
+		List<Message> instructions = chatClientRequest.prompt()
```
apappascs: this part could be :

```
// Add non-system messages from instructions
for (Message message : instructions) {
    if (message.getMessageType() != MessageType.SYSTEM) {
        processedMessages.add(message);
    }
}
```

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
```diff
@@ -83,9 +86,18 @@ public ChatClientRequest before(ChatClientRequest chatClientRequest, AdvisorChai
 		List<Message> memoryMessages = this.chatMemory.get(conversationId);
 
 		// 2. Advise the request messages list.
-		List<Message> processedMessages = new ArrayList<>(memoryMessages);
-		processedMessages.addAll(chatClientRequest.prompt().getInstructions());
-
+		List<Message> processedMessages = new ArrayList<>();
```
yuluo-yx, line 89: Why put system in the first place? In my opinion, system indicates LLMS, and all subsequent conversations are based on system.

And from `https://openai.com/index/gpt-4-api-general-availability/`, system and user message do not have the same function.

PR #3521, opened 2025-06-12
Fixed GH-3466 part 2: Fixed the issue where validation failed when the model invoked different `MethodToolCallback` instances

sunyuhan1998:
This PR is used to address an issue that was raised in #3466: 

https://github.com/spring-projects/spring-ai/issues/3466#issuecomment-2965311698 

i.e. when a model tries to call back multiple `MethodToolCallback`, if some of the `MethodToolCallback` accept `ToolContext` as a parameter and some don't, and the `ToolContext` is not null, the `validateToolContextSupport` method validation fails.

The main changes made in this PR are as follows: 

* Added a `getMethod` method to `MethodToolCallback` to get the `toolMethod`
* When executing a tool call in `DefaultToolCallingManager`, based on the information about the tool that the model expects to be called back, it decides which `call` method to execute for the call 
* Added some corresponding unit test

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
@@ -16,11 +16,13 @@
 
 package org.springframework.ai.model.tool;
 
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import io.micrometer.observation.ObservationRegistry;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
Hi @sunyuhan1998, Thanks for the PR. The fix you made would still work with Spring AI 1.0 (before the changes made via https://github.com/spring-projects/spring-ai/pull/3478. We discussed this issue and we want to revert the original changes from #3478(except fixing ClassUtils.isAssignable(ToolContext.class, type)). 

In terms of fixing #3466, we don't want to break the behaviour in Spring AI 1.0 and perhaps, we need to change the exception message to be precise.

ilayaperumalg:
@sunyuhan1998 Reverted the changes from the PR https://github.com/spring-projects/spring-ai/pull/3478 and updated the fix to use the right order of usage in ClassUtils.isAssignable via 76c70189e2de7b8fb872ea0b01604258a2290e1f

sunyuhan1998:
> @sunyuhan1998 Reverted the changes from the PR #3478 and updated the fix to use the right order of usage in ClassUtils.isAssignable via [76c7018](https://github.com/spring-projects/spring-ai/commit/76c70189e2de7b8fb872ea0b01604258a2290e1f)

Ok, I see what you mean, do we need to merge the unit test part of this current PR?It seems like it could be targeted to test this scenario we are discussing

ilayaperumalg:
@sunyuhan1998 sure, I agree the unit test will be a good addition and will get it merged.

ilayaperumalg:
Merged the unit test into main as b9a683460 and back-ported into 1.0.x as well. Thanks once again!

PR #1852, opened 2024-12-02
feat(ollama): add retry template integration to OllamaChatModel

apappascs:
- Integrated RetryTemplate for API calls in OllamaChatModel.
- Ensured consistent retry logic across all model classes.

---
Run a build and make sure all tests pass prior to submission
```
./mvnw spring-javaformat:apply
./mvnw clean install -DskipTests -Dmaven.javadoc.skip=true
./mvnw verify -Pintegration-tests -pl spring-ai-core 
./mvnw verify -Pintegration-tests -pl models/spring-ai-ollama
./mvnw verify -Pintegration-tests -pl spring-ai-spring-boot-autoconfigure 
./mvnw javadoc:javadoc -Pjavadoc
```

Issue #1193, opened 2024-08-08
Add proper retry implementation for OpenAiChatModel streaming


tzolov:

csterwa:
@tzolov does the PR referenced here resolve the retry implementation enhancement?

ThomasVitale:
Adding some context to this issue: a `RetryTemplate` is used to add retry capabilities to the blocking model calls. At first, the same strategy was used for the streaming calls. After realising that it wasn't working as expected in a reactive stream, the usage of `RetryTemplate` streaming calls has been removed. In this issue, we would need to come up with a design for a retry feature to adopt in all streaming implementations across the different model providers.

ilayaperumalg:
@chemicL Could you check the above? Thanks

chemicL:
@ilayaperumalg sure, if you meant the proposal from @apappascs, my comment follows.

@apappascs thanks for taking the time to look into it. In my view it is not necessary to mimic the `RetryTemplate`. The particular `Flux`/`Mono` can have retry defined on top using the `retryWhen` spec directly without the builder in between. What is probably more important is whether the caller of the SpringAI API should take care of retries or whether that functionality is built in. If it's built-in then some specification in the API can be provided that maps to the retryWhen usage internally. Otherwise, the user should be free to define a retry logic on top of the streaming responses themselves. E.g. in case of `WebClient` use typically it's the user defining the retry on the returned `Mono`/`Flux` types from the API.

@tzolov can you offer some guidance what your expectations are?

PR #3524, opened 2025-06-12
fix(streaming): Remove retry wrapper from chat model stream methods

apappascs:
Removes the `retryTemplate.execute` wrapper from the `chatCompletionStream` calls in MiniMax, MistralAI, and ZhiPuAI chat models.

following up on this comment https://github.com/spring-projects/spring-ai/pull/1852#discussion_r2141004106 from @markpollack

### DIFF
--- a/models/spring-ai-minimax/src/main/java/org/springframework/ai/minimax/MiniMaxChatModel.java
+++ b/models/spring-ai-minimax/src/main/java/org/springframework/ai/minimax/MiniMaxChatModel.java
@@ -323,8 +323,7 @@ public class MiniMaxChatModel implements ChatModel {
 		return Flux.deferContextual(contextView -> {
 			ChatCompletionRequest request = createRequest(requestPrompt, true);
 
-			Flux<ChatCompletionChunk> completionChunks = this.retryTemplate
-				.execute(ctx -> this.miniMaxApi.chatCompletionStream(request));
+			Flux<ChatCompletionChunk> completionChunks = this.miniMaxApi.chatCompletionStream(request);
 
 			// For chunked responses, only the first chunk contains the choice role.
 			// The rest of the chunks with same ID share the same role.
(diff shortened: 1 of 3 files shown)


sobychacko:
@apappascs The PR build fails - can you please take a look? https://github.com/spring-projects/spring-ai/actions/runs/15634414758/job/44072094143

apappascs:
> [sobychacko](/sobychacko)

thank you @sobychacko I am working on that. need to disable a few tests.

sobychacko:
@apappascs Curious why you disabled those tests. Did your changes make those tests fail? In that case, we need to investigate why that is the case and address it accordingly. If we leave these tests disabled, we risk missing regressions that would normally be caught by these tests, including failures unrelated to the current changes.

apappascs:
> @apappascs Curious why you disabled those tests. Did your changes make those tests fail? In that case, we need to investigate why that is the case and address it accordingly. If we leave these tests disabled, we risk missing regressions that would normally be caught by these tests, including failures unrelated to the current changes.

Very valid points @sobychacko . I sould have linked this issue
https://github.com/spring-projects/spring-ai/issues/1193
for clarity. The retry method used for blocking calls isn't working as expected for the streaming calls. The tests are related to the code changes I did and have been disabled or removed for the other models as well.

markpollack:
@chemicL thoughts?

chemicL:
> @chemicL thoughts?

https://github.com/spring-projects/spring-ai/issues/1193#issuecomment-2610324696
I am not sure whether this PR tackles any real problem? The retry template should not have any effects on reactive chains, unless their implementations throw, which should be a programming error. So removing them should not have a significant effect, the bulk of the work is the strategy for retries - whether the framework implements it or it's the user burden.

apappascs:
Thank you for the input @chemicL . I think this alone, as you mentioned, does nt really do much. We discussed last year about removing the retry from streamable calls because it didn't work properly as we expected.

I followed the same pattern we have for the non streamable calls:

https://github.com/spring-projects/spring-ai/issues/1193#issuecomment-2517586047

And your comment added to the context:
https://github.com/spring-projects/spring-ai/issues/1193#issuecomment-2610324696

@markpollack @tzolov This is still open.

PR #3529, opened 2025-06-13
Fixed GH-3526, Fixed the error in using the Builder in SystemPromptTemplate.

### LINKED ISSUE #3526: Invalid message type created when using SystemPromptTemplate builder
**Bug description**
When using `SystemPromptTemplate` builder, it's expected that an instance of `SystemPromptTemplate` is being built, not `PromptTemplate`. However, currently an instance of `PromptTemplate` is returned. As a result, when `createMessage` is called, it creates `UserMessage` instead of `SystemMessage`:

**Environment**
Spring AI 1.0.0

**Steps to reproduce**
Here's how the issue can be reproduced:

```java
var promptTemplate = SystemPromptTemplate
        .builder()
        .template("template")
        .build();

System.out.println(promptTemplate.getClass());
System.out.println(promptTemplate.createMessage().getClass());
```

**Expected behavior**
It is expected that a builder creates an instance of `SystemPromptTemplate` so that when `createMessage` is called, an instance of `SystemMessage` is created.

sunyuhan1998:
As mentioned in the issue, currently when using the `Builder` in `SystemPromptTemplate`, it returns a `Builder` of `PromptTemplate`. This causes subsequent operations like `createMessage` based on the `Builder` class to actually return a `UserMessage` instead of a `SystemMessage`.

The key point is that `SystemPromptTemplate` inherits the `Builder` from `PromptTemplate`.

I can think of two ways to address this problem, but in my opinion, neither is perfect:

1. Add a `Builder` inner class in `SystemPromptTemplate` that extends the `Builder` class of `PromptTemplate`, and override all of its methods. However, this approach would require updating all subclasses' `Builder` classes whenever any new properties are added to `PromptTemplate` in the future.

2. Make the `Builder` class of `PromptTemplate` generic and use the “Simulated Self-Type Idiom” so that each subclass can return its own `Builder` type. However, this would involve significant changes to the current `Builder` class of `PromptTemplate`, and all existing code that uses it would also need to support generics.

Currently, I have adopted the first approach because it has a smaller impact scope compared to the second one. 

This PR includes the following changes:

1. Adjusted the visibility of some fields and methods in the `Builder` of `PromptTemplate` to support inheritance in the `Builder` of `SystemPromptTemplate`.  
2. Added a `Builder` for `SystemPromptTemplate`.  
3. Added relevant unit tests.

Fixes #3526

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java
@@ -209,17 +209,17 @@ public class PromptTemplate implements PromptTemplateActions, PromptTemplateMess
 		return new Builder();
 	}
 
-	public static final class Builder {
+	public static class Builder {
 
-		private String template;
+		protected String template;
 
-		private Resource resource;
+		protected Resource resource;
 
-		private Map<String, Object> variables = new HashMap<>();
+		protected Map<String, Object> variables = new HashMap<>();
 
-		private TemplateRenderer renderer = DEFAULT_TEMPLATE_RENDERER;
+		protected TemplateRenderer renderer = DEFAULT_TEMPLATE_RENDERER;
 
-		private Builder() {
+		protected Builder() {
 		}
 
 		public Builder template(String template) {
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
Since this PR fixes the underlying issue with the Builder in SystemPromptTemplate, we can back port the changes into 1.0.x as well.

ilayaperumalg:
Rebased and merged as dbaa608be. Also, back-ported into 1.0.x

PR #3532, opened 2025-06-13
fix(memory): Replace the outdated manual way of creating the ChatMemoryRepository

Cyanty:
According to the documentation, some outdated methods have emerged from using Chat Memory:

- Update the document information and replace the unavailable creation method of ChatMemoryRepository.
- Increase the create function to create Neo4jChatMemoryRepository, this way to manually create the way (Instead of the way of "new Object").


memory/repository/spring-ai-model-chat-memory-repository-neo4j/src/main/java/org/springframework/ai/chat/memory/repository/neo4j/Neo4jChatMemoryRepository.java
```diff
@@ -56,6 +56,10 @@ public Neo4jChatMemoryRepository(Neo4jChatMemoryRepositoryConfig config) {
 		this.config = config;
 	}
 
+	public static Neo4jChatMemoryRepository create(Neo4jChatMemoryRepositoryConfig conf) {
```
yuluo-yx: IMO, wouldn't it be better to use Builder.builder()?

Cyanty: Ok, let me modify that

sobychacko:
Thanks for the PR!

Closing as unrebasable. The chat memory modules were reorganised, so this can't be brought forward without a rewrite.

The underlying gap is still real and we'll track it separately. One note if you revisit it: making the existing public constructor private is a breaking change and would need a deprecation.

PR #3553, opened 2025-06-16
fix: handle root exceptions in McpToolCallback

Kehrlann:
Sync and Async McpToolCallback can now handle exceptions where cause == null instead of throwing a NPE.

This is the case when we do OAuth2 flows with Spring Security in MCP tools, because Spring Security relies on exceptions as signals to trigger "fetch an access_token" behavior.


tzolov:
Thanks @Kehrlann 
Rebased and merged at d6924b668c13ef03e1eb78b02dce8199a59fe25f

Issue #3535, opened 2025-06-13
Enable custom configuration of more properties for weaviate vector store


sp-yang:
Please do a quick search on GitHub issues first, the feature you are about to request might have already been requested.

**Expected Behavior**

Different instances of weaviate may have different properties, such as CONTENT_FIELD_NAME, OBJECT_CLASS, COLLECTION_NAME. It would be great if more properties can be customized to fit for existing weaviate instances, like those did for NEO4J vector store, which is more customizable.

**Current Behavior**

Currently only very few properties like OBJECT_CLASS and CONSISTENCY_LEVEL can be customized, many properties are hard-coded in the implementation of weaviate vector store, which is difficult to be reused. 

**Context**

As many properties are hard-coded in the current implementation of weaviate vector store, it's very difficult to fit for existing weaviate instances, unless re-writting the implementation.

dev-jonghoonpark:
@sp-yang 

I'd like to try implementing this. 
but first, I have a few questions about the implementation.

---

## Question 1.

Could you list the specific properties that are needed?
Should all of the following variables be configurable?

https://github.com/spring-projects/spring-ai/blob/384478edbf99e28c0a89cd5740d13a4e3f8483e7/vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStore.java#L98-L110

---

## Question 2.

Should `COLLECTION_NAME` be configurable?

Currently, the implementation of weaviateVectorStore uses objectClass as the value for collectionName.

https://github.com/spring-projects/spring-ai/blob/384478edbf99e28c0a89cd5740d13a4e3f8483e7/vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStore.java#L432-L438

sp-yang:
@dev-jonghoonpark : Many thanks for your prompt response. 
Regarding the 1st question, only CONTENT_FIELD_NAME need to be configurable at the moment.
Regarding the 2nd question, using ObjectClass as the value for CollectionName would be OK for us, too.

![Image](https://github.com/user-attachments/assets/09087b21-c281-4fe0-89af-8a586af0b822)

jPhy:
[...]

PR #3555, opened 2025-06-16
Make the `content` field configurable in Weaviate vector store properties

dev-jonghoonpark:
related issue: https://github.com/spring-projects/spring-ai/issues/3535

Currently, the content field in the Weaviate vector store cannot be modified.
This PR has been improved to make it configurable, and relevant tests have been added.

## changes
- Added a new property: `spring.ai.vectorstore.weaviate.content-field-name`

### DIFF
--- a/auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-weaviate/src/main/java/org/springframework/ai/vectorstore/weaviate/autoconfigure/WeaviateVectorStoreProperties.java
+++ b/auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-weaviate/src/main/java/org/springframework/ai/vectorstore/weaviate/autoconfigure/WeaviateVectorStoreProperties.java
@@ -27,6 +27,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
  * Configuration properties for Weaviate Vector Store.
  *
  * @author Christian Tzolov
+ * @author Jonghoon Park
  */
 @ConfigurationProperties(WeaviateVectorStoreProperties.CONFIG_PREFIX)
 public class WeaviateVectorStoreProperties {
@@ -40,6 +41,8 @@ public class WeaviateVectorStoreProperties {
 	private String apiKey = "";
 
 	private String objectClass = "SpringAiWeaviate";
+
+	private String contentFieldName = "content";
 
 	private ConsistentLevel consistencyLevel = WeaviateVectorStore.ConsistentLevel.ONE;
 
@@ -76,6 +79,20 @@ public class WeaviateVectorStoreProperties {
 
 	public String getObjectClass() {
 		return this.objectClass;
+	}
+
+	/**
+	 * @since 1.1.0
+	 */
+	public String getContentFieldName() {
+		return contentFieldName;
+	}
+
+	/**
+	 * @since 1.1.0
+	 */
+	public void setContentFieldName(String contentFieldName) {
+		this.contentFieldName = contentFieldName;
 	}
 
 	public void setObjectClass(String indexName) {
--- a/vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStore.java
+++ b/vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStore.java
@@ -75,7 +75,7 @@ import org.springframework.util.StringUtils;
  * <pre>{@code
  * // Create the vector store with builder
  * WeaviateVectorStore vectorStore = WeaviateVectorStore.builder(weaviateClient, embeddingModel)
- *     .objectClass("CustomClass")                // Optional: Custom class name (default: SpringAiWeaviate)
+ *     .options(options)                     	  // Optional: use custom options
  *     .consistencyLevel(ConsistentLevel.QUORUM)  // Optional: Set consistency level (default: ONE)
  *     .filterMetadataFields(List.of(             // Optional: Configure filterable metadata fields
  *         MetadataField.text("country"),
@@ -89,6 +89,7 @@ import org.springframework.util.StringUtils;
  * @author Josh Long
  * @author Soby Chacko
  * @author Thomas Vitale
+ * @author Jonghoon Park
  * @since 1.0.0
  */
 public class WeaviateVectorStore extends AbstractObservationVectorStore {
@@ -96,8 +97,6 @@ public class WeaviateVectorStore extends AbstractObservationVectorStore {
 	private static final Logger logger = LoggerFactory.getLogger(WeaviateVectorStore.class);
 
 	private static final String METADATA_FIELD_PREFIX = "meta_";
-
-	private static final String CONTENT_FIELD_NAME = "content";
 
 	private static final String METADATA_FIELD_NAME = "metadata";
 
@@ -111,9 +110,9 @@ public class WeaviateVectorStore extends AbstractObservationVectorStore {
 
 	private final WeaviateClient weaviateClient;
 
-	private final ConsistentLevel consistencyLevel;
+	private final WeaviateVectorStoreOptions options;
 
-	private final String weaviateObjectClass;
+	private final ConsistentLevel consistencyLevel;
 
 	/**
 	 * List of metadata fields (as field name and type) that can be used in similarity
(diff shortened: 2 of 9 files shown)


auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-weaviate/src/main/java/org/springframework/ai/vectorstore/weaviate/autoconfigure/WeaviateVectorStoreProperties.java
```diff
@@ -78,6 +81,14 @@ public String getObjectClass() {
 		return this.objectClass;
 	}
 
+	public String getContentFieldName() {
```
sobychacko, line 87: New public API's could benefit a ``@since`` tag as javadoc - `@since 1.1.0`.

auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-weaviate/src/main/java/org/springframework/ai/vectorstore/weaviate/autoconfigure/WeaviateVectorStoreProperties.java
```diff
@@ -78,6 +81,14 @@ public String getObjectClass() {
...
+		return contentFieldName;
+	}
+
+	public void setContentFieldName(String contentFieldName) {
```
sobychacko, line 94: New public API's could benefit a `@since` tag as javadoc - `@since 1.1.0`.

vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStore.java
```diff
@@ -548,14 +548,14 @@ private Builder(WeaviateClient weaviateClient, EmbeddingModel embeddingModel) {
...
 		/**
-		 * Configures the Weaviate object class.
-		 * @param objectClass the object class to use
+		 * Configures the Weaviate vector store option.
+		 * @param options the vector store options to use
 		 * @return this builder instance
-		 * @throws IllegalArgumentException if objectClass is null or empty
+		 * @throws IllegalArgumentException if options is null or empty
 		 */
-		public Builder objectClass(String objectClass) {
```
sobychacko, line 556: We cannot remove a publicly available method in a GA release in a minor version (`1.1.x`). We should deprecate this method and encourage users to use the new API in the javadocs using the `@deprecated` tag.

sobychacko: That also suggests that we may not be able to backport this PR since it adds a new API method, deprecates the existing one, etc.

vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStore.java
```diff
@@ -548,14 +548,14 @@ private Builder(WeaviateClient weaviateClient, EmbeddingModel embeddingModel) {
...
-			Assert.hasText(objectClass, "objectClass must not be empty");
-			this.weaviateObjectClass = objectClass;
+		public Builder options(WeaviateVectorStoreOptions options) {
```
sobychacko, line 573: New public APIs could benefit from a `@since` tag via javadoc.

vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStoreOptions.java
```diff
@@ -0,0 +1,50 @@
...
+
+package org.springframework.ai.vectorstore.weaviate;
+
+import org.springframework.util.Assert;
+
+/**
+ * Provided Weaviate vector option configuration.
+ *
+ * @author Jonghoon Park
+ */
```
sobychacko, line 26: `@since 1.1.0`.

sobychacko: Also, having some Javadoc on all the new public API methods is a good practice, IMHO.

sobychacko (changes requested):

vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStore.java
sobychacko: Also, since `objectClass` looks like a top-level concept in Weaviate vector store, it is also probably fine to leave it as is, and only provide the option to configure the content field. What do you think?

sobychacko (changes requested):

vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStore.java
dev-jonghoonpark: @sobychacko 
Thank you for your feedback 👍

From my perspective, I don't have a strong preference either way.

The reason I did it this way is that `ElasticsearchVectorStoreOptions` has `indexName` and `embeddingFieldName`.
To maintain consistency, I set it up so that `WeaviateVectorStoreOptions` has `objectClass` and `contentFieldName`.

dev-jonghoonpark: I have applied all feedback except for this discussion.

vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStore.java
```diff
@@ -179,7 +179,7 @@ private Field[] buildWeaviateSimilaritySearchFields() {
 
 		List<Field> searchWeaviateFieldList = new ArrayList<>();
 
-		searchWeaviateFieldList.add(Field.builder().name(CONTENT_FIELD_NAME).build());
+		searchWeaviateFieldList.add(Field.builder().name(this.options.getContentFieldName()).build());
 		searchWeaviateFieldList.add(Field.builder().name(METADATA_FIELD_NAME).build());
 		searchWeaviateFieldList.addAll(this.filterMetadataFields.stream()
 			.map(mf -> Field.builder().name(METADATA_FIELD_PREFIX + mf.name()).build())
```
jPhy, line 185: To be able to use this class, we would need to define metadata without the "meta_"-prefix here.
Our Database has metadata, but they don't have a special prefix.

```suggestion
			.map(mf -> Field.builder().name(mf.name()).build())
```

Having `METADATA_FIELD_PREFIX` customizable just like `CONTENT_FIELD_NAME` would also be fine for us, we could set it to an empty string.

dev-jonghoonpark: @jPhy 
Thank you for sharing your thoughts.
1. What exactly does `this class` refer to in your comment?
2. It seems that your comment is suggesting additional functionality. Is that correct? If so, perhaps it might be more appropriate to address this in a separate PR.

jPhy: `this class` refers to `WeaviateVectorStore`.
I saw it as answering [Question 1](https://github.com/spring-projects/spring-ai/issues/3535#issuecomment-2974993656)  slightly differently compared to @sp-yang: We would need, in addition to `CONTENT_FIELD_NAME`, also `METADATA_FIELD_PREFIX` to be configurable.
I can also open a separate issue if you think that would be better.

dev-jonghoonpark: Depending on the maintainer’s decision, the structure might change slightly,
so I think it’s better to proceed after this PR is completed.

After that, I will also try to work on your suggestion.
So, it’s not necessary to open an issue,
but feel free to open one if you think it's needed.

sobychacko:
@dev-jonghoonpark Merged the PR via https://github.com/spring-projects/spring-ai/commit/24de06ebebb5657e742ea2f52c30e8615b91dd2b. I didn't backport the PR due to the deprecation and the new API in the builder. Thanks for the PR!

vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStore.java
dev-jonghoonpark: The configurable `METADATA_FIELD_PREFIX` implemented in https://github.com/spring-projects/spring-ai/pull/3585 @jPhy

Issue #3557, opened 2025-06-16
Transcription API File Extension Issue


juandaco:
**Bug description**
Spring AI overrides the file name with `audio.webm` when using the Transcription API of Open AI. [See code in here](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiAudioApi.java#L163). Sending an `*.mp3` file with the file name as `*.webm` is not supported by the new Open AI transcription models `gpt-4o-transcribe` and `gpt-4o-mini-transcribe`. These models will fail with an unsupported error while `whisper-1` will allow it.

**Environment**
Java 21, Spring Boot 3.5.0, Spring AI 1.0 GA

**Steps to reproduce**
This problem can be tested with `curl` using the [Open AI Transcription API](https://platform.openai.com/docs/api-reference/audio/createTranscription). 

1. Grab a mp3 file with a speech
2. Rename the file as `audio.webm`
3. Send a request with `curl` providing the API key and making sure that the model selected is one of the `gpt-4o-transcribe` (not `whisper-1`).
4. Observe the returned error

Make sure to send a renamed mp3 file as `audio.webm`
```sh
curl https://api.openai.com/v1/audio/transcriptions \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -H "Content-Type: multipart/form-data" \
  -F file="@/path/to/file/audio.webm" \
  -F model="gpt-4o-transcribe"
```

**Expected behavior**
Spring AI should not be changing the file name and failing silently. It should send the correct file name (at least the correct file extension).

sunyuhan1998:
I've pushed a PR: #3558  to try to fix this issue. Thank you for reporting it!

ilayaperumalg:
@juandaco Thanks for reporting the issue. @sunyuhan1998 Thanks for the PR!

PR #3558, opened 2025-06-17
Fixed GH-3557, Fixed the issue where the filename was lost when requesting the OpenAI audio processing interface.

sunyuhan1998:
As mentioned in the issue, there is currently a problem where the filename gets overwritten when requesting OpenAI's `/audio/translations` and `/audio/transcriptions` interfaces, leading to request failures. 

The root cause lies in the code prior to this PR, specifically in the `org.springframework.ai.openai.OpenAiAudioTranscriptionModel#createRequest` method. When converting the `Resource` to a `byte[]`, the original filename information was lost, resulting in a hardcoded filename "audio.webm" being used when sending requests to the OpenAI API, which is clearly unreasonable. 

This PR fixes the issue by preserving and passing the original filename to the REST interface.


ilayaperumalg:

ilayaperumalg:
@sunyuhan1998 Thanks for the PR fixing the issue. Rebased and merged as af07517b5

PR #3562, opened 2025-06-17
fixed #3559 Modify the DefaultToolCallResultConverter.convert method …

### LINKED ISSUE #3559: Modify the DefaultToolCallResultConverter.convert method and add filtering code
If the method returns a value of Mono type, the calling effect cannot be achieved.  
So the modifications are as follows:

```

       // This is the newly added filtering code used to obtain the real values in Mono
       **if (result instanceof Mono<?>) {
            result = ((Mono<?>) result).block();
        }**

            return JsonParser.toJson("Done");
        } else if (result instanceof RenderedImage) {
            ByteArrayOutputStream buf = new ByteArrayOutputStream(4096);

            try {
                ImageIO.write((RenderedImage)result, "PNG", buf);
            } catch (IOException var5) {
                return "Failed to convert tool result to a base64 image: " + var5.getMessage();
            }

            String imgB64 = Base64.getEncoder().encodeToString(buf.toByteArray());
        } else {
            logger.debug("Converting tool result to JSON.");
            return JsonParser.toJson(result);
        }
    }
```

TheEterna:
…and add the Mono processing code


spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
```diff
@@ -30,6 +30,7 @@
 
 import org.springframework.ai.util.json.JsonParser;
 import org.springframework.lang.Nullable;
+import reactor.core.publisher.Mono;
```
dev-jonghoonpark, line 33: Please place this line right under the `import org.slf4j.LoggerFactory;`.

dev-jonghoonpark:
🙌 Please fix the DCO check by signing your commit. `git commit --amend --signoff`

PR #3564, opened 2025-06-17
Reuse DocumentRowMapper instance to reduce redundant object creation

dev-jonghoonpark:
This PR addresses the issue of creating a new `DocumentRowMapper` instance on every call to the `doSimilaritySearch()` method.

`DocumentRowMapper` is a stateless RowMapper and does not need to be created repeatedly.
In this PR, `DocumentRowMapper` has been declared as a final field of the class, initialized once during instance creation, and reused across subsequent calls.

Additionally, the unused variable `COLUMN_EMBEDDING`, defined in the `DocumentRowMapper` class,  has been removed.  
This variable became unused as a result of https://github.com/spring-projects/spring-ai/issues/1826 and the corresponding https://github.com/spring-projects/spring-ai/commit/ebd29e0959086ea950d505a8aefb9e32ca7732a7.


ilayaperumalg:
@dev-jonghoonpark Thanks for the PR cleaning up the code.

PR #3568, opened 2025-06-17
fixed #3559 Modify the DefaultToolCallResultConverter.convert method …

### LINKED ISSUE #3559: Modify the DefaultToolCallResultConverter.convert method and add filtering code

TheEterna:
…and add the Mono processing code

Especially for MCP asynchronous servers, this bug is fatal, request merge.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
@@ -27,6 +27,7 @@ import javax.imageio.ImageIO;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import reactor.core.publisher.Mono;
 
 import org.springframework.ai.util.json.JsonParser;
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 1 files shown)


tzolov:
Hi @TheEterna the DefaultToolCallResultConverter is related to the only to the MethodToolCallback (and FunctionToolCallback). Currently neither of them support reactive handing (e.g. Mono<?> ), because they implement the ToolCallback interface (later is imperative).

If you are building MCP Server you can use the new `@McpTool` annotations on the server side that support reactive method handlers at the server level.  
On the client side (e.g. Spring AI), both the AsyncMcpToolCallback and the SyncMcpToolCallback are non-reactive because the ToolCallback interface is non-reactive. 
As referred [here](https://github.com/spring-projects/spring-ai/issues/3559#issuecomment-3003600746) the DefaultToolCallResultConverter is not meant for reactive use.

PR #3582, opened 2025-06-18
chore: Fix order of getter and setter methods in `WeaviateVectorStoreProperties`

dev-jonghoonpark:
Fix order of getter and setter methods in `WeaviateVectorStoreProperties`

related pr: https://github.com/spring-projects/spring-ai/pull/3555

I apologize for the mistake.
I mistakenly placed the getContentFieldName and setContentFieldName methods between the getter and setter for objectClass. I've corrected the order to align with the conventional grouping of getter and setter methods for the same property.

## current 

```java
	public String getObjectClass() {
		return this.objectClass;
	}

	/**
	 * @since 1.1.0
	 */
	public String getContentFieldName() {
		return contentFieldName;
	}

	/**
	 * @since 1.1.0
	 */
	public void setContentFieldName(String contentFieldName) {
		this.contentFieldName = contentFieldName;
	}

	public void setObjectClass(String indexName) {
		this.objectClass = indexName;
	}
```

## changed

```java
	public String getObjectClass() {
		return this.objectClass;
	}
	
	public void setObjectClass(String indexName) {
		this.objectClass = indexName;
	}

	/**
	 * @since 1.1.0
	 */
	public String getContentFieldName() {
		return contentFieldName;
	}

	/**
	 * @since 1.1.0
	 */
	public void setContentFieldName(String contentFieldName) {
		this.contentFieldName = contentFieldName;
	}
```

PR #3585, opened 2025-06-18
Make the `meta field prefix` configurable in Weaviate vector store properties

dev-jonghoonpark:
related issue: https://github.com/spring-projects/spring-ai/issues/3535

Currently, the `meta field prefix` in the Weaviate vector store cannot be modified.
This PR has been improved to make it configurable, and relevant tests have been added.

## changes
- Added a new property: `spring.ai.vectorstore.weaviate.meta-field-prefix`
  - Note: Whitespace is also allowed in this property.

### DIFF
--- a/vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateFilterExpressionConverter.java
+++ b/vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateFilterExpressionConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStoreOptions.java
```diff
@@ -48,4 +50,13 @@ public void setContentFieldName(String contentFieldName) {
...
 	}
 
+	public String getMetaFieldPrefix() {
+		return metaFieldPrefix;
+	}
+
+	public void setMetaFieldPrefix(String metaFieldPrefix) {
+		Assert.notNull(metaFieldPrefix, "metaFieldPrefix can be empty but not null");
+		this.metaFieldPrefix = metaFieldPrefix;
+	}
```
dev-jonghoonpark, line 60: These methods are newly created, but since the class Javadoc already includes the `@since 1.1.0` tag, the `@since` tag was not added to the methods.

dev-jonghoonpark:
fixed conflict with https://github.com/spring-projects/spring-ai/pull/3582

vector-stores/spring-ai-weaviate-store/src/main/java/org/springframework/ai/vectorstore/weaviate/WeaviateFilterExpressionConverter.java
```diff
@@ -35,19 +35,44 @@
...
+	private final String metaFieldPrefix;
+
+	/**
+	 * Constructs a new instance of the {@code WeaviateFilterExpressionConverter} class.
+	 * This constructor uses the default meta field prefix
+	 * ({@link #DEFAULT_META_FIELD_PREFIX}).
+	 * @param allowedIdentifierNames A {@code List} of allowed identifier names.
+	 */
 	public WeaviateFilterExpressionConverter(List<String> allowedIdentifierNames) {
 		Assert.notNull(allowedIdentifierNames, "List can be empty but not null.");
```
sobychacko: Why don't you just call the other constructor from here?

vector-stores/spring-ai-weaviate-store/src/test/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStoreIT.java
```diff
@@ -359,6 +361,77 @@ public void addAndSearchWithCustomContentFieldName() {
 		});
 	}
 
+	@ParameterizedTest(name = "{0} : {displayName} ")
+	@ValueSource(strings = { "custom_", "" })
+	public void addAndSearchWithCustomMetaFieldPrefix(String metaFieldPrefix) {
```
sobychacko, line 366: Could you help me understand where the configurable metadata prefix is used in this test and then assert that it takes effect?

sobychacko (changes requested):

sobychacko:
@dev-jonghoonpark Some minor comments. In your description above, I see a note that the white spaces are also allowed with the metadata prefix. I don't see that in the code. Could you point me to where that is addressed? Thanks!

vector-stores/spring-ai-weaviate-store/src/test/java/org/springframework/ai/vectorstore/weaviate/WeaviateVectorStoreIT.java
dev-jonghoonpark: @sobychacko 

In this test, the following behavior is expected:

1. Verify that data is stored and retrieved correctly using a `customVectorStore` with a modified prefix.
2. Confirm that data cannot be retrieved using a `vectorStore` with the default prefix (as the prefixes differ, retrieval should fail).
3. Remove the data from the `customVectorStore` to ensure it does not affect subsequent tests.

If there’s a better approach, I’m open to suggestions for improvement. 👍

dev-jonghoonpark:
@sobychacko 

In `WeaviateVectorStoreOptions`, validation in setMetaFieldPrefix was implemented to use `hasNull` instead of `hasText`.

To verify that this works as expected, a test case with an empty metaFieldPrefix (`""`) has been added to `addAndSearchWithCustomMetaFieldPrefix` in `WeaviateVectorStoreIT`.

Is there anything I might have missed?

sobychacko (approved):

PR #3589, opened 2025-06-18
Fixed GH-3586, Add logging for sensitive word matches in `SafeGuardAdvisor`.

### LINKED ISSUE #3586: SafeGuardAdvisor lacks logs
version: 1.1.0-SNAPSHOT
SafeGuardAdvisor lacks any logs
It would be nice if the log informs which amongst the specific sensitiveWords causes the failure.
I guess it would be essential for production logs.
At least there should be traces that it acted upon the request and response.

sunyuhan1998:
As mentioned in the issue, currently when `SafeGuardAdvisor` matches sensitive words, there is a lack of logs indicating which sensitive words were matched. This PR adds debug logging for that purpose.

Fixes #3586

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/SafeGuardAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/SafeGuardAdvisor.java
@@ -18,7 +18,11 @@ package org.springframework.ai.chat.client.advisor;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 
 import org.springframework.ai.chat.client.ChatClientRequest;
@@ -43,6 +47,8 @@ import org.springframework.util.CollectionUtils;
  * @since 1.0.0
  */
 public class SafeGuardAdvisor implements CallAdvisor, StreamAdvisor {
+
+	private static final Logger log = LoggerFactory.getLogger(SafeGuardAdvisor.class);
 
 	private static final String DEFAULT_FAILURE_RESPONSE = "I'm unable to respond to that due to sensitive content. Could we rephrase or discuss something else?";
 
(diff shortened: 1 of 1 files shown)


YunKuiLu:
Other than logging the sensitive words, can we also put the matched sensitive words into the `context` of `ChatClientResponse`, so that we can get them?

sunyuhan1998:
Hi @ilayaperumalg , Can you help review this PR?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/SafeGuardAdvisor.java
```diff
@@ -76,22 +82,28 @@ public String getName() {
 
 	@Override
 	public ChatClientResponse adviseCall(ChatClientRequest chatClientRequest, CallAdvisorChain callAdvisorChain) {
-		if (!CollectionUtils.isEmpty(this.sensitiveWords)
-				&& this.sensitiveWords.stream().anyMatch(w -> chatClientRequest.prompt().getContents().contains(w))) {
-			return createFailureResponse(chatClientRequest);
+		if (!CollectionUtils.isEmpty(this.sensitiveWords)) {
+			String content = chatClientRequest.prompt().getContents();
+			Set<String> hitWords = this.sensitiveWords.stream().filter(content::contains).collect(Collectors.toSet());
+			if (!hitWords.isEmpty()) {
```
yuluo-yx, line 88: hi, for `if (!hitWords.isEmpty())`. 

IMO, If you enter this `if (!CollectionUtils.isEmpty(this.sensitiveWords))` branch, the following conditions must be true is. means `hitWords` is not null.I think this if case can be deleted.

ilayaperumalg:
@sunyuhan1998 We don't want to log any sensitive information as they are indeed sensitive. Perhaps, you can create a custom advisor for your use case.

sunyuhan1998:
> @sunyuhan1998 We don't want to log any sensitive information as they are indeed sensitive. Perhaps, you can create a custom advisor for your use case. 

Oh, okay, I see. I think I misunderstood its function before, but now I completely understand. Thank you for the clarification!

PR #3595, opened 2025-06-18
Support OAuth2-based flows in Tool calling

Kehrlann:
Spring Security relies on specific exceptions (based on `ClientAuthorizationException`) to signal that an OAuth2 token should be requested.

With the recent changes in Spring AI, those exceptions were now swallowed and sent to the LLM, or, with the appropriate property, rethrown after being wrapped in another exception. This broke OAuth2 flows. With this PR, we change the default behavior to rethrow spring security OAuth2 exceptions in the `DefaultToolExecutionExceptionProcessor`.

The `DefaultToolExecutionExceptionProcessor` now has an allowlist of exceptions that will be rethrown instead of being wrapped.


tzolov:
Thanks @Kehrlann 
LGTM

tzolov:
Rebased, squashed and merged at beb1d05d756f3814d913dc981b09349b541a5a2d
Back ported to 1.0.x

PR #3600, opened 2025-06-19
Refactor: Replace `System.err.println` with log output

sunyuhan1998:
Replace `System.err.println` with log output


ilayaperumalg:
@sunyuhan1998 Thanks for fixing the log!

PR #3608, opened 2025-06-19
add a verification when using tools

XuGongming:
When deploying qwen3 using vllm version 0.9.1, the ‘current. toolCalls()’ will result in an empty array. If only verifying whether it is null, an exception will be thrown


ilayaperumalg:
@XuGongming Thanks for the PR. Rebased and merged as a221a3784.

PR #3610, opened 2025-06-19
fix(elevenlabs): correct autoconfiguration class name and add ElevenLabs speech doc link

apappascs:
This commit updates the ElevenLabs autoconfiguration import to use the correct class name and adds a navigation link for ElevenLabs speech documentation.


ilayaperumalg:
@apappascs Thanks for the fix.

ilayaperumalg:
Removed the back-porting as the ElevenLabs feature is an enhancement which will be available only from 1.1.x

Issue #3151, opened 2025-05-14
ChatObservationProperties.class missing logCompletion field and corresponding setter method.


iAMSagar44:
The ChatObservationProperties.class is missing the logCompletion field and its corresponding setter method.

As per the documentation, the following properties have changed in RC1 -

```
spring.ai.chat.observations.include-prompt → spring.ai.chat.observations.log-prompt
spring.ai.chat.observations.include-completion → spring.ai.chat.observations.log-completion
```

This is resulting in the following error -

_**"message": "'spring.ai.chat.observations.log-completion' is an unknown property. Did you mean 'spring.ai.chat.observations.log-prompt'?"**_

A similar error is noticed for the below property too, due to the missing field (logQueryResponse) in VectorStoreObservationProperties.class
`spring.ai.vectorstore.observations.log-query-response`

Spring AI version - 1.0.0-RC1

quaff:
It should be fixed by https://github.com/spring-projects/spring-ai/pull/3156

iAMSagar44:
Thanks @quaff. I also noticed that with RC-1 release the chat prompts are no longer shown in a tracing service (like Zipkin). Prior to the RC-1 release, I was able to see the prompts in tracing services like Zipkin, but with the RC-1 release, they have switched observability from tracing to logging. 
Is there a way I can switch back to tracing?

tzolov:
Thanks for catching this.
IMO, the issue isn't as critical as it appear. You can continue using the properties as documented (e.g., log-xyz).
The discrepancy exists because ChatObservationProperties isn't used programmatically. The affected properties are only used for conditional auto-configurations ([like this](https://github.com/spring-projects/spring-ai/blob/1ede66492fceb21fbf49edb3d99d6db8b939632b/auto-configurations/models/chat/observation/spring-ai-autoconfigure-model-chat-observation/src/main/java/org/springframework/ai/model/chat/observation/autoconfigure/ChatObservationAutoConfiguration.java#L130)), where the values are correctly implemented (e.g., log-completion vs. include-completion).
[...]

PR #3612, opened 2025-06-19
feat(observability): Refactor content observation mechanism to support both logging and tracing

traburiss:
This commit refactors the content observation mechanism by adding support for trace recording without altering the original logging functionality. The motivation for this change is to ensure complete context availability in scenarios such as integration with Langfuse, which is essential for proper functionality.

Main changes:

1. New configuration options:
   1. `trace-prompt`: Enables recording prompts to trace
   2. `trace-completion` : Enables recording completions to trace
   3. `trace-prompt-size` : Limits the length of prompt context
   4. `content-formatter` : Handles content formatting, currently supporting both 'text' and 'langfuse' modes
2. Added two handlers, ChatModelCompletionObservationTraceHandler and ChatModelPromptContentObservationTraceHandler, to support recording context to trace.
3. Introduced the MessageFormatter interface and its subclasses to support formatting for prompt and completion content.
4. Rolled back parts of the code and dependencies from commit ca843e85.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ You can find more details in the [Reference Documentation](https://docs.spring.i
   - [Moderation](https://docs.spring.io/spring-ai/reference/api/index.html#api/moderation)
 * Portable API support across AI providers for both synchronous and streaming options. Access to [model-specific features](https://docs.spring.io/spring-ai/reference/api/chatmodel.html#_chat_options) is also available.
 * [Structured Outputs](https://docs.spring.io/spring-ai/reference/api/structured-output-converter.html) - Mapping of AI Model output to POJOs.
-* Support for all major [Vector Database providers](https://docs.spring.io/spring-ai/reference/api/vectordbs.html) such as *Apache Cassandra, Azure Vector Search, Chroma, Milvus, MongoDB Atlas, MariaDB, Neo4j, Oracle, PostgreSQL/PGVector, PineCone, Qdrant, Redis, and Weaviate*.
+* Support for all major [Vector Database providers](https://docs.spring.io/spring-ai/reference/api/vectordbs.html) such as *Apache Cassandra, Azure Vector Search, Chroma, Elasticsearch, Milvus, MongoDB Atlas, MariaDB, Neo4j, Oracle, PostgreSQL/PGVector, PineCone, Qdrant, Redis, and Weaviate*.
 * Portable API across Vector Store providers, including a novel SQL-like [metadata filter API](https://docs.spring.io/spring-ai/reference/api/vectordbs.html#metadata-filters).
 * [Tools/Function Calling](https://docs.spring.io/spring-ai/reference/api/tools.html) - permits the model to request the execution of client-side tools and functions, thereby accessing necessary real-time information as required.
 * [Observability](https://docs.spring.io/spring-ai/reference/observability/index.html) - Provides insights into AI-related operations.
--- a/models/spring-ai-openai/pom.xml
+++ b/models/spring-ai-openai/pom.xml
@@ -88,27 +88,8 @@
 		</dependency>
 
 		<dependency>
-			<groupId>org.springframework.ai</groupId>
-			<artifactId>spring-ai-qdrant-store</artifactId>
-			<version>${project.version}</version>
-			<exclusions>
-				<exclusion>
-					<groupId>org.springframework.ai</groupId>
-					<artifactId>spring-ai-openai</artifactId>
-				</exclusion>
-			</exclusions>
-			<scope>test</scope>
-		</dependency>
-
-		<dependency>
 			<groupId>io.micrometer</groupId>
 			<artifactId>micrometer-observation-test</artifactId>
-			<scope>test</scope>
-		</dependency>
-
-		<dependency>
-			<groupId>org.testcontainers</groupId>
-			<artifactId>qdrant</artifactId>
 			<scope>test</scope>
 		</dependency>
 
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
@@ -167,8 +167,10 @@ public class OpenAiAudioTranscriptionModel implements Model<AudioTranscriptionPr
 			}
 		}
 
+		Resource instructions = transcriptionPrompt.getInstructions();
 		return OpenAiAudioApi.TranscriptionRequest.builder()
-			.file(toBytes(transcriptionPrompt.getInstructions()))
+			.file(toBytes(instructions))
+			.fileName(instructions.getFilename())
 			.responseFormat(options.getResponseFormat())
 			.prompt(options.getPrompt())
 			.temperature(options.getTemperature())
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiAudioApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiAudioApi.java
@@ -160,7 +160,7 @@ public class OpenAiAudioApi {
 
 			@Override
 			public String getFilename() {
-				return "audio.webm";
+				return requestBody.fileName();
 			}
 		});
 		multipartBody.add("model", requestBody.model());
@@ -206,7 +206,7 @@ public class OpenAiAudioApi {
 
 			@Override
 			public String getFilename() {
-				return "audio.webm";
+				return requestBody.fileName();
 			}
 		});
 		multipartBody.add("model", requestBody.model());
@@ -496,6 +496,7 @@ public class OpenAiAudioApi {
 	 * Transcription</a>
 	 *
 	 * @param file The audio file to transcribe. Must be a valid audio file type.
+	 * @param fileName The audio file name.
 	 * @param model ID of the model to use. Only whisper-1 is currently available.
 	 * @param language The language of the input audio. Supplying the input language in
 	 * ISO-639-1 format will improve accuracy and latency.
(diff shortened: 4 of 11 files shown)


ilayaperumalg:
@jonatan-ivanov Could you take a look a this PR? Thanks!

shakuzen:
It would help in reviewing this if you could share a link to the relevant parts of Langfuse's documentation that specifies what they're expecting for the desired integration. How likely is it to change? This is reintroducing using the deprecated API on OTel Spans to add Event Attributes, which was removed because it's deprecated and the suggested alternative was Log Events.

traburiss:
@shakuzen  

https://langfuse.com/docs/integrations/spring-ai

Langfuse doc still uses `include-prompt` and `include-completion`, but these parameters are now deprecated. However, the new parameters `log-prompt` and `log-completion` cannot work with Langfuse. 
Langfuse need both prompt and completion data included within spans.

auto-configurations/models/chat/observation/spring-ai-autoconfigure-model-chat-observation/src/main/java/org/springframework/ai/model/chat/observation/autoconfigure/ChatObservationAutoConfiguration.java
```diff
@@ -104,6 +111,30 @@ TracingAwareLoggingObservationHandler<ChatModelObservationContext> chatModelComp
 			return new TracingAwareLoggingObservationHandler<>(new ChatModelCompletionObservationHandler(), tracer);
 		}
 
+		@Bean
+		@ConditionalOnMissingBean(value = ChatModelPromptContentObservationTraceHandler.class,
+				name = "chatModelPromptContentObservationTraceHandler")
+		@ConditionalOnProperty(prefix = ChatObservationProperties.CONFIG_PREFIX, name = "trace-prompt",
+				havingValue = "true")
+		TracingAwareLoggingObservationHandler<ChatModelObservationContext> chatModelPromptContentObservationTraceHandler(
```
jonatan-ivanov, line 119: Why are these wrapped into `TracingAwareLoggingObservationHandler`?
Please check what it does, it is used to make the tracing data available for log entries (log correlation). If I understand correctly, this is not needed in this use-case.

spring-ai-commons/pom.xml
```diff
@@ -61,7 +61,7 @@
 
 		<dependency>
 			<groupId>io.micrometer</groupId>
-			<artifactId>micrometer-tracing</artifactId>
+			<artifactId>micrometer-tracing-bridge-otel</artifactId>
```
jonatan-ivanov, line 64: spring-ai-commons should not depend on a bridge (implementation instead of the abstraction).

spring-ai-commons/src/main/java/org/springframework/ai/observation/tracing/TracingHelper.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public final class TracingHelper {
+
+	private static final Logger logger = LoggerFactory.getLogger(TracingHelper.class);
+
+	private TracingHelper() {
+	}
+
+	@Nullable
+	public static Span extractOtelSpan(@Nullable TracingObservationHandler.TracingContext tracingContext) {
```
jonatan-ivanov, line 42: I think we should depend on the abstraction (tracing API) instead of a single implementation (OTel API/SDK).

spring-ai-model/pom.xml
```diff
@@ -60,7 +60,7 @@
 
 		<dependency>
 			<groupId>io.micrometer</groupId>
-			<artifactId>micrometer-tracing</artifactId>
+			<artifactId>micrometer-tracing-bridge-otel</artifactId>
```
jonatan-ivanov, line 63: spring-ai-model should not depend on a bridge (implementation instead of the abstraction).

spring-ai-model/src/main/java/org/springframework/ai/chat/observation/ChatModelCompletionObservationTraceHandler.java
```diff
@@ -0,0 +1,88 @@
...
+			.filter(generation -> generation.getOutput() != null
+					&& StringUtils.hasText(generation.getOutput().getText()))
+			.map(generation -> this.messageFormatter.format(generation.getOutput()))
+			.toList();
+
+		TracingObservationHandler.TracingContext tracingContext = context
+			.getRequired(TracingObservationHandler.TracingContext.class);
+		Span currentSpan = TracingHelper.extractOtelSpan(tracingContext);
+		if (currentSpan != null) {
+			currentSpan.addEvent(AiObservationEventNames.CONTENT_COMPLETION.value(),
```
jonatan-ivanov, line 78: It seems this API will be deprecated (maybe already is?), see: https://github.com/micrometer-metrics/micrometer/issues/5238#issuecomment-2706051527

jonatan-ivanov:
>Langfuse doc still uses include-prompt and include-completion, but these parameters are now deprecated. However, the new parameters log-prompt and log-completion cannot work with Langfuse.

This makes me a bit confused. Why are we doing this if one is already deprecated and another is not supported?

markpollack:
Basically my view is that our observability needs to support products like langfuse.  How to achieve this goal.

wrt to this PR

>>> Rolled back parts of the code and dependencies from commit https://github.com/spring-projects/spring-ai/commit/ca843e85887aa1da6300c77550c379c103500897.

that change in behavior can't occur in 1.0.x - and not sure about the strategy for 1.1 on this front.  Hoping to use this issue to drive a strategy.

I thought there is flexibility to override behavior - we can perhaps create ready to use repos for products that customize observability in our https://github.com/orgs/spring-ai-community/repositories org

@ThomasVitale suggestions on how to structure the convo and go forward - a working google doc would seem to be a good start.

markpollack:
Also related discussion is here - https://github.com/spring-projects/spring-ai/issues/3151#issuecomment-2880245614

auto-configurations/models/chat/observation/spring-ai-autoconfigure-model-chat-observation/src/main/java/org/springframework/ai/model/chat/observation/autoconfigure/ChatObservationAutoConfiguration.java
traburiss: Yes, the wrapping of `TracingAwareLoggingObservationHandler` should be removed.

spring-ai-commons/src/main/java/org/springframework/ai/observation/tracing/TracingHelper.java
traburiss: got it

spring-ai-model/src/main/java/org/springframework/ai/chat/observation/ChatModelCompletionObservationTraceHandler.java
traburiss: I got it, it's better to use Micrometer Observation.Event instead of OTel Span Event.

But I noticed that the KeyValues feature on Observation.Event (https://github.com/micrometer-metrics/micrometer/issues/5238#issuecomment-2706051527) hasn’t been released yet. Are we required to wait for this functionality to become available?

I wonder if it would be possible to write directly to Observation.tag instead. As I’ve tested, Langfuse properly captures the data when using this approach.

traburiss: I wonder if it would be possible to write directly to `Span.Tag`. 

like

```java
tracingContext.getSpan().tagOfStrings(AiObservationAttributes.CONTENT_COMPLETION.value(), completion);
```

As I’ve tested, Langfuse properly captures the data when using this approach.

traburiss:
> Basically my view is that our observability needs to support products like langfuse. How to achieve this goal.
> 
> wrt to this PR
> 
> > > > Rolled back parts of the code and dependencies from commit [ca843e8](https://github.com/spring-projects/spring-ai/commit/ca843e85887aa1da6300c77550c379c103500897).
> 
> that change in behavior can't occur in 1.0.x - and not sure about the strategy for 1.1 on this front. Hoping to use this issue to drive a strategy.
> 
> I thought there is flexibility to override behavior - we can perhaps create ready to use repos for products that customize observability in our https://github.com/orgs/spring-ai-community/repositories org
> 
> @ThomasVitale suggestions on how to structure the convo and go forward - a working google doc would seem to be a good start.

@markpollack 

In my opinion, if the intention is to prevent large amount of data within span, constraints should be imposed through parameters to limit the length/byte size rather than directly removing these two fields from the span.

shakuzen:
Given the lack of stability around what to send in what format to what observability backend, I think it would be most maintainable and usable to make separate modules that customize the obserability data in the way that a specific use case needs for well-known use cases, such as Langfuse. I think this is what @markpollack suggested in his previous comment. From the user's perspective this looks like adding a dependency on a Spring AI module when they want to ship data in the way Langfuse expects. From a maintainability perspective, it means we don't need to keep making changes to core modules when things change in Langfuse. This should all be possible with auto-configuring different ObservationHandler implementations, which is what this PR does. In fact, any user can configure an implementation of ObservationHandler that does this or anything else. So it's already possible for users to achieve what this PR is doing without changes in Spring AI, right? I agree it would be best if it were easier to achieve, and that's the reason for a module that does what's needed that people can use.

I suspect there are other integrations out there for AI observability and what they expect may be different than Langfuse and may change at different times and in different ways than other integrations. Trying to make the observability workable with everything from core modules sounds like a losing proposition. If there were standardization and stability around this, it would be feasible to do this from common modules, but it seems we're not there yet. Perhaps an alternative is to make the separate module based around OTel semantic conventions for Gen AI and put the onus on observability backends to support the latest version of that (I would suspect there's some pressure for this anyway).

That said, my reading of the linked documentation to Langfuse is that they intend to support what Spring AI produces, so we may not need to try so hard to send observability data in the way we used to just because that is what Langfuse supported (because that's what we produced... notice the chicken and egg problem).

traburiss:
I agree with @shakuzen  suggested in his previous comment, it is a better solution..

In fact, this is exactly how I've implemented it in my production code
I developed a dedicated autoconfiguration and starter, which provide additional configuration options and a custom handler to support Langfuse

I’ll create an issue in the spring-ai-community repository at https://github.com/orgs/spring-ai-community/repositories.

markpollack:
Yes, I will create a new issue for this approach @shakuzen

PR #3618, opened 2025-06-19
fix(bedrock): Handle streaming tool calling with no input arguments

### LINKED ISSUE #1878: Bedrock/Converse Chat Model Does not Properly Stream Tools When Using Suppliers vs Functions

**Bug description**
When attempting to use a `tools` along with streaming, I am using a java.util.function.Supplier as my FunctionCallback (vs a java.util.function.Function); I am using BedrockProxyChatModel as my ChatModel.

The initial bedrock LLM response correctly selects my Supplier as the tool choice, however to tool is never called by Spring AI.  This appears to only happen with Suppliers;  java.util.function.Function works as expected.

While debugging, the issue a manifested in the `isToolCall` of the org.springframework.ai.chat.model.AbstractToolCallSupport class when after the following debug message.
```
DEBUG 34272 ---  [tyEventLoop-2-2] o.s.a.b.converse.BedrockProxyChatModel   : Received converse stream output:MessageStopEvent(StopReason=tool_use)
```

The `chatResponse` contains no `generations`, therefor the tool is not executed.
```
	protected boolean isToolCall(ChatResponse chatResponse, Set<String> toolCallFinishReasons) {
		Assert.isTrue(!CollectionUtils.isEmpty(toolCallFinishReasons), "Tool call finish reasons cannot be empty!");
		if (chatResponse == null) {
			return false;
		}
		var generations = chatResponse.getResults();
		if (CollectionUtils.isEmpty(generations)) {
			return false;
		}
		return generations.stream().anyMatch(g -> isToolCall(g, toolCallFinishReasons));
	}
```

**Environment**
This was encountered using spring-ai M4 along with Java 21.

**Steps to reproduce**
The following is an excerpt from a trivial SpringBoot application that can reproduce the problem when Spring AI is configured use the Bedrock/Converse chat models.  It includes a very simple System.out to view the output.

```
	@Bean
	public CommandLineRunner commandLineRunner(ChatClient chatClient) {
		
		return (args) -> {
			
            FunctionCallback.Builder builder =
                    FunctionCallback.builder();
            
            FunctionCallback functionCallBack = builder.description("Get the weather")
            .function("weather", new WeatherService())
            .build();
            
			
			ChatClientRequestSpec spec = chatClient.prompt();
			
			spec.functions(functionCallBack).user("What's the weather like").stream().chatResponse()
			.doOnNext(resp -> {
				if (resp.getResult() != null)
					System.out.println(resp.getResult().getOutput().getContent());
				}).collectList().block();			

		};
	}
	
	public static class WeatherService implements Supplier<WeatherService.Response> {

		public record Response(double temp) {}
	
		public Response get() {
			return new Response(30.0);
		}
		
	}
```

**Expected behavior**
I expect an output something similar to "The current weather is 30 degrees".

See the `steps to reproduce` for a code sample.

tzolov:
- Fix isEmpty() check in ConverseApiUtils to properly handle null partialJson
- Add workaround for streaming tool calls with empty arguments by defaulting to {}
- Add parameterized tests for multiple Bedrock models (Nova Pro and Claude Sonnet)
- Update AWS SDK and Bedrock runtime versions to 2.31.65

Fixes #1878

### DIFF
--- a/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/api/ConverseApiUtils.java
+++ b/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/api/ConverseApiUtils.java
@@ -421,8 +421,7 @@ public final class ConverseApiUtils {
 		}
 
 		public boolean isEmpty() {
-			return (this.index == null || this.id == null || this.name == null
-					|| !StringUtils.hasText(this.partialJson));
+			return (this.index == null || this.id == null || this.name == null || this.partialJson == null);
 		}
 
 		ToolUseAggregationEvent withIndex(Integer index) {
@@ -451,7 +450,9 @@ public final class ConverseApiUtils {
 		}
 
 		void squashIntoContentBlock() {
-			this.toolUseEntries.add(new ToolUseEntry(this.index, this.id, this.name, this.partialJson, this.usage));
+			// Workaround to handle streaming tool calling with no input arguments.
+			String json = StringUtils.hasText(this.partialJson) ? this.partialJson : "{}";
+			this.toolUseEntries.add(new ToolUseEntry(this.index, this.id, this.name, json, this.usage));
 			this.index = null;
 			this.id = null;
 			this.name = null;
--- a/models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/client/BedrockNovaChatClientIT.java
+++ b/models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/client/BedrockNovaChatClientIT.java
@@ -212,28 +212,6 @@ public class BedrockNovaChatClientIT {
 		assertThat(content).contains("20 degrees");
 	}
 
-	@Test
-	void streamingToolCallingWithArgumentlessToolSonnet() {
-
-		ChatClient chatClient = ChatClient.builder(this.chatModel).build();
-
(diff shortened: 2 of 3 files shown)


models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/api/ConverseApiUtils.java
```diff
@@ -451,7 +450,9 @@ ToolUseAggregationEvent appendPartialJson(String partialJson) {
 		}
 
 		void squashIntoContentBlock() {
-			this.toolUseEntries.add(new ToolUseEntry(this.index, this.id, this.name, this.partialJson, this.usage));
+			// Workaround to handle streaming tool calling with no input arguments.
+			String json = StringUtils.hasText(this.partialJson) ? this.partialJson : "{}";
```
markpollack, line 454: maybe we should report this as a bug?

tzolov:
merged at 81076126482aef6815482c08b5be4e4b0c083ded

models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/api/ConverseApiUtils.java
tzolov: Not sure. this appears to be an Anthropic Model's inconsistency b/w the call and stream json payload for parameterless tools. 
The behavior is is consistent for all Anthropic models.

PR #3622, opened 2025-06-20
fix(anthropic): Add streaming parameter-less tool support

tzolov:
- Remove partialJson text check from isEmpty() method in AnthropicApi
- Add parameterized test for streaming parameter-less tools across Claude models
- Addresses issue #1878 with parameter-less tool streaming functionality


ilayaperumalg:
Rebased and merged as f5750517b. Back-ported into 1.0.x as well.

PR #3627, opened 2025-06-20
Cleanup: Cleaned up redundant assignments in TextReader.

sunyuhan1998:
In the `org.springframework.ai.reader.TextReader#get` method of `TextReader`, there was a redundant assignment for the `SOURCE_METADATA` key in `customMetadata`. 

In addition, the null checks for the results of `resource.getURI()` and `resource.getURL()` in `getResourceIdentifier` were also unnecessary.


ilayaperumalg:

PR #3638, opened 2025-06-23
docs: correct `SimpleLoggerAdvisor` documentation

dev-jonghoonpark:
## Issue

In https://docs.spring.io/spring-ai/reference/api/chatclient.html,
The `SimpleLoggerAdvisor` class documentation incorrectly shows a constructor signature:

```java
SimpleLoggerAdvisor(
    Function<AdvisedRequest, String> requestToString,
    Function<ChatResponse, String> responseToString
)
```
However, the actual constructor includes an additional parameter, `int order`.

Additionally, the `AdvisedRequest` class does not include a variable named `userText`.

## Correction
Update the documentation to reflect the correct constructor:

```java
SimpleLoggerAdvisor(
    Function<AdvisedRequest, String> requestToString,
    Function<ChatResponse, String> responseToString,
    int order
)
```

And Update the example to replace the incorrect `userText` reference with `request.prompt().getUserMessage()`

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chatclient.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chatclient.adoc
@@ -593,8 +593,9 @@ You can customize what data from `AdvisedRequest` and `ChatResponse` is logged b
 [source,java]
 ----
 SimpleLoggerAdvisor(
-    Function<AdvisedRequest, String> requestToString,
-    Function<ChatResponse, String> responseToString
+    Function<ChatClientRequest, String> requestToString,
+    Function<ChatResponse, String> responseToString,
+    int order
 )
 ----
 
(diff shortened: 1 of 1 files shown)


spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chatclient.adoc
```diff
@@ -594,7 +594,8 @@ You can customize what data from `AdvisedRequest` and `ChatResponse` is logged b
 ----
 SimpleLoggerAdvisor(
     Function<AdvisedRequest, String> requestToString,
```
ilayaperumalg: AdvisedRequest should be ChatClientRequest

dev-jonghoonpark: @ilayaperumalg 
Thanks for pointing out what I missed.
I've fixed it

PR #3640, opened 2025-06-23
Propagate reactive Context to AsyncMcpToolCallback

Kehrlann:
- When calling tools while using ChatModel#stream, store the reactive context in a thread-local, so it can be used by downstream reactive tools.
- In AsyncMcpToolCallback, restore the reactive context so it can be accessed by the tool. This will be useful for Spring Security OAuth2 support in reactive scenarios, because it relies on the context.


tzolov:
Thanks @Kehrlann 

TODO: optimize the repeatable set context logic across all models.

rebased and merged at 9a521827277f509803b19628696ef550b8b277df

PR #3653, opened 2025-06-24
Fix NPE for media-only Document in QdrantVectorStore (GH-3609)

### LINKED ISSUE #3609: QdrantVectorStore throws NPE when Document contains only media

**Bug description**
`QdrantVectorStore` throws a **NullPointerException** when adding a `Document`
that contains **media only** (no text).

**Environment**
spring-ai-bom:1.0.0
spring-ai-starter-vector-store-qdrant
Java : 24
springboot version : 3.5.0
spring dependency management : version 1.1.7

**Steps to reproduce**
1. Run Qdrant locally (`docker run -p 6333:6333 -p 6334:6334 qdrant/qdrant`).
2. Create a Spring Boot app with `spring-ai-starter-vector-store-qdrant`.
3. Inside a test or service method:

```java
   Media media = new Media(
           MimeType.valueOf("image/png"),
           new byte[] { 0x00 });                  // 1×1 transparent pixel

   Document imgDoc = Document.builder()
           .media(media)                          // ⚠️ no text
           .metadata(Map.of("fileName", "pixel.png"))
           .build();

   vectorStore.add(List.of(imgDoc));             // ← NPE here
 ```

Image

 ```
Observe the stack trace:
Caused by: java.lang.RuntimeException: java.lang.NullPointerException
	at org.springframework.ai.vectorstore.qdrant.QdrantVectorStore.toPayload(QdrantVectorStore.java:304)
	at org.springframework.ai.vectorstore.qdrant.QdrantVectorStore.lambda$doAdd$0(QdrantVectorStore.java:186)
 ```

**Expected behavior**
QdrantVectorStore should accept media-only documents and store their vector plus metadata without requiring doc_content to be present.

When a Document instance contains only media (no text) and is passed to QdrantVectorStore.add(...), the call fails with a NullPointerException.
The failure originates in QdrantVectorStore.toPayload(Document):

```java
var payload = QdrantValueFactory.toValueMap(document.getMetadata());
payload.put(CONTENT_FIELD_NAME,
            io.qdrant.client.ValueFactory.value(document.getText())); // ← getText() is null
```
Because Document was refactored to be either text or media (never both), getText() legitimately returns null for image-only documents.
The method still unconditionally adds the "doc_content" field, so a media document triggers the NPE during ValueFactory.value(null).

Typical flow that exposes the bug:

My custom EmbeddingModel converts each incoming image to a float-array vector via a FastAPI CLIP endpoint.

I build a list of media-only Document objects, one per uploaded file.

I inject VectorStore (backed by QdrantVectorStore) and call vectorStore.add(documents).

doAdd(...) calls toPayload(document) for every item; the very first image document crashes with the NPE, aborting the entire batch.

The problem is independent of the embedding model: it happens before the vectors are sent to Qdrant, purely because the payload builder assumes text content is always present.

devMtn30:
### Problem
Storing a Document that contains only media currently results in a NPE because
`ValueFactory.value(null)` is invoked.

### Solution
* Check `document.isText()`.  
* If false, throw an `IllegalArgumentException` with a clear message.

### Issue
Closes #3609

### DIFF
--- a/vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantVectorStore.java
+++ b/vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantVectorStore.java
@@ -296,6 +296,13 @@ public class QdrantVectorStore extends AbstractObservationVectorStore implements
 	 */
 	private Map<String, Value> toPayload(Document document) {
 		try {
+			if(!document.isText()) {
+				throw new IllegalArgumentException("""
+						QdrantVectorStore supports only text-based Document for now –
+						received media-only Document; see issue #3609.
+						""");
+			}
+
 			var payload = QdrantValueFactory.toValueMap(document.getMetadata());
 			payload.put(CONTENT_FIELD_NAME, io.qdrant.client.ValueFactory.value(document.getText()));
 			return payload;


vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantVectorStore.java
dev-jonghoonpark, line 1:
```java
if (!document.isText()) {
    throw new IllegalArgumentException("""
            QdrantVectorStore supports only text-based Document for now –
            received media-only Document; see issue #3609.
            """);
}

var payload = QdrantValueFactory.toValueMap(document.getMetadata());
payload.put(CONTENT_FIELD_NAME, io.qdrant.client.ValueFactory.value(document.getText()));
return payload;
```

How about this code?

It can prevent unnecessary execution of `QdrantValueFactory.toValueMap`.
And it's more intuitive.

devMtn30: > ```java
> if (!document.isText()) {
>     throw new IllegalArgumentException("""
>             QdrantVectorStore supports only text-based Document for now –
>             received media-only Document; see issue #3609.
>             """);
> }
> 
> var payload = QdrantValueFactory.toValueMap(document.getMetadata());
> payload.put(CONTENT_FIELD_NAME, io.qdrant.client.ValueFactory.value(document.getText()));
> return payload;
> ```
> 
> How about this code?
> 
> It can prevent unnecessary execution of `QdrantValueFactory.toValueMap`. And it's more intuitive.

Thanks for the suggestion-updated!

vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantVectorStore.java
```diff
@@ -296,8 +296,16 @@ private Document toDocument(ScoredPoint point) {
 	 */
 	private Map<String, Value> toPayload(Document document) {
 		try {
+			String documentContent = document.getText();
```
dev-jonghoonpark: I was wondering if there’s a specific reason for declaring a separate variable named `documentContent`.
If you believe it’s necessary, feel free to keep it. However, in my opinion, it might not be needed.

Additionally, it might be better to move it either before or after the declaration of `payload` (line 307).

devMtn30: Removed the extra variable, inlined `document.getText()`, and force pushed. Thanks for the spot!

yuluo-yx (approved):

ilayaperumalg:
@devMtn30 Thank you for the PR. I have submitted a PR which addresses this issue across the vector stores: https://github.com/spring-projects/spring-ai/pull/3687. Could you review the PR please?

ilayaperumalg:
Closing this PR in favour of #3687

PR #3654, opened 2025-06-24
[Refactor] Remove redundant null check in OpenAiApi.Builder#apiKey(String)

KoreaNirsa:
### Description

This pull request removes the redundant `Assert.notNull(...)` check from the `OpenAiApi.Builder#apiKey(String)` method.

### Motivation

The `SimpleApiKey` constructor already performs a null check and throws an `IllegalArgumentException` if the value is null:

```
java
public SimpleApiKey(String value) {
    Assert.notNull(value, "API key value must not be null or empty");
    this.value = value;
}
```

Thus, the additional null check inside the builder is unnecessary and violates the DRY principle (Don't Repeat Yourself).
By relying on the constructor's validation, the code becomes cleaner and easier to maintain.

Impact
No behavioral change; exception will still be thrown with the same message if the input is null.

Improves code clarity and separation of responsibility.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -1909,7 +1909,6 @@ public class OpenAiApi {
 		}
 
 		public Builder apiKey(String simpleApiKey) {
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
@KoreaNirsa Thanks for the PR!

PR #3663, opened 2025-06-25
Refactor: Add null check, optimize string joining, and add JavaDocs

KoreaNirsa:
### Summary
This PR improves the `getTextContent` method in `OpenAiApi` with better performance, input validation, and complete documentation.

### Changes
- `Assert.notNull(content, "content cannot be null")` to prevent potential `NullPointerException`
- `.reduce("", (a, b) -> a + b)` with `.collect(Collectors.joining())` for O(n) string concatenation
- JavaDocs with parameter/return/exception descriptions for better maintainability

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -21,15 +21,7 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
-
-import com.fasterxml.jackson.annotation.JsonFormat;
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonInclude.Include;
-import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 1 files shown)


sunyuhan1998 (approved):
LGTM!

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -50,6 +42,16 @@
...
 
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
```
dev-jonghoonpark: Is there a particular reason for changing the order of imports?
Based on the context, it seems to be an unnecessary change, so it would be better to revert it.

---

Typically, imports are organized in the following order:

1. `java.*`
2. other imports
3. `org.springframework.*`
4. static imports

KoreaNirsa: It looks like the import order was unintentionally changed.
I've reverted it as suggested.

Thank you for pointing it out!

ilayaperumalg:
@KoreaNirsa Thanks for the cleanup and documentation!

PR #3672, opened 2025-06-26
Add link to javadoc on Content class

enimiste:
The javadoc on top of the class Content mention a reference to the Message class but there is non {@link } around it.


sunyuhan1998 (approved):

ilayaperumalg:
@enimiste Thanks for the PR!

PR #3687, opened 2025-06-26
Validate documents for text/media check when adding to vector store

### LINKED ISSUE #3609: QdrantVectorStore throws NPE when Document contains only media

ilayaperumalg:
- Since the non-text documents aren't supported, we need a validation to throw exception if non text documents are encountered
 - Add tests
    - Add test to QdrantVectorStoreIT with the usecase provided by @devMtn30
    - Add simple vector store test for the validation 
   
 Fixes #3609

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/observation/AbstractObservationVectorStore.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/observation/AbstractObservationVectorStore.java
@@ -74,7 +74,7 @@ public abstract class AbstractObservationVectorStore implements VectorStore {
 	 */
 	@Override
 	public void add(List<Document> documents) {
-
+		validateNonTextDocuments(documents);
 		VectorStoreObservationContext observationContext = this
 			.createObservationContextBuilder(VectorStoreObservationContext.Operation.ADD.value())
 			.build();
@@ -83,6 +83,17 @@ public abstract class AbstractObservationVectorStore implements VectorStore {
 			.observation(this.customObservationConvention, DEFAULT_OBSERVATION_CONVENTION, () -> observationContext,
 					this.observationRegistry)
 			.observe(() -> this.doAdd(documents));
+	}
+
+	private void validateNonTextDocuments(List<Document> documents) {
+		if (documents == null)
+			return;
+		for (Document document : documents) {
+			if (document != null && !document.isText()) {
+				throw new IllegalArgumentException(
+						"Only text documents are supported for now. One of the documents contains non-text content.");
+			}
+		}
 	}
 
 	@Override
--- a/spring-ai-vector-store/src/test/java/org/springframework/ai/vectorstore/SimpleVectorStoreTests.java
+++ b/spring-ai-vector-store/src/test/java/org/springframework/ai/vectorstore/SimpleVectorStoreTests.java
@@ -32,13 +32,18 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.CleanupMode;
 import org.junit.jupiter.api.io.TempDir;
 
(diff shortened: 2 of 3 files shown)


dev-jonghoonpark:
I also think this approach is more appropriate. 👍

However, I have a personal question.
Is there a specific reason why `testNonTextDocuments` test was added in `QdrantVectorStoreIT.java`?
Was it intended to verify the implementation's behavior? Or was it added because it wasn’t included in [PR #3653](https://github.com/spring-projects/spring-ai/pull/3653)?

devMtn30:
Hi @ilayaperumalg, the changes in PR #3687 look excellent! You’ve neatly applied the shared validation logic across all VectorStores and added the automated NPE regression test, so I’m happy to merge it as-is. Thanks again for your thorough review and suggestions! 🙏

ilayaperumalg:
@dev-jonghoonpark @devMtn30 Thank you for the review! 

@dev-jonghoonpark 
> Is there a specific reason why testNonTextDocuments test was added in QdrantVectorStoreIT.java?
Was it intended to verify the implementation's behavior? Or was it added because it wasn’t included in https://github.com/spring-projects/spring-ai/pull/3653?

I intended to use one of the vector store ITs and since the report was made from QdrantVectorStore, I ended up choosing one.

Issue #3702, opened 2025-06-28
spring ai doc is wrong


sinaweibo:
![Image](https://github.com/user-attachments/assets/b20e5af4-784e-4ba7-accd-0a609f2174d3)
it should be    spring.ai.vectorstore.milvus.auto-id

dev-jonghoonpark:
Nice Catch 👍

PR #3704, opened 2025-06-28
docs: correct property name from `is-auto-id` to `auto-id` in milvus.adoc

dev-jonghoonpark:
correct property name from `is-auto-id` to `auto-id` in milvus.adoc

close https://github.com/spring-projects/spring-ai/issues/3702


sunyuhan1998 (approved):
LGTM！

PR #3705, opened 2025-06-28
improve schema initialization logic and update deprecated code for Milvus vectorstore

dev-jonghoonpark:
### improve schema initialization logic

In the current logic, since `createCollection` is executed without following it up with `createIndex`, the following error occurs during `describeIndex`:

```
20:36:07.664 [main] ERROR io.milvus.client.AbstractMilvusGrpcClient -- DescribeIndexRequest collectionName:test_vector_store_custom_fields, indexName: failed! Exception:{}
io.milvus.exception.ServerException: index not found[collection=test_vector_store_custom_fields]
	at io.milvus.client.AbstractMilvusGrpcClient.handleResponse(AbstractMilvusGrpcClient.java:399)
	at io.milvus.client.AbstractMilvusGrpcClient.describeIndex(AbstractMilvusGrpcClient.java:1423)
	at io.milvus.client.MilvusServiceClient.lambda$describeIndex$32(MilvusServiceClient.java:611)
	at io.milvus.client.MilvusServiceClient.retry(MilvusServiceClient.java:326)
	at io.milvus.client.MilvusServiceClient.describeIndex(MilvusServiceClient.java:611)
	at org.springframework.ai.vectorstore.milvus.MilvusVectorStore.createCollection(MilvusVectorStore.java:449)
```

However, it is natural for an index to not exist immediately after `createCollection`.
Thus, the logic was modified to attempt `createIndex` right after `createCollection`.

**This prevents unnecessary error logs from being recorded.**

current example log: https://github.com/spring-projects/spring-ai/actions/runs/15934060747/job/44949781008#step:7:10851

### update deprecated code 

`addFieldType` has been deprecated.
It is recommended to use `withSchema` instead for adding field types, so the implementation has been updated accordingly.

### DIFF
--- a/vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/milvus/MilvusVectorStore.java
+++ b/vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/milvus/MilvusVectorStore.java
@@ -38,6 +38,7 @@ import io.milvus.param.MetricType;
 import io.milvus.param.R;
 import io.milvus.param.R.Status;
 import io.milvus.param.RpcStatus;
+import io.milvus.param.collection.CollectionSchemaParam;
 import io.milvus.param.collection.CreateCollectionParam;
 import io.milvus.param.collection.DropCollectionParam;
 import io.milvus.param.collection.FieldType;
@@ -443,6 +444,8 @@ public class MilvusVectorStore extends AbstractObservationVectorStore implements
 		if (!isDatabaseCollectionExists()) {
 			createCollection(this.databaseName, this.collectionName, this.idFieldName, this.isAutoId,
 					this.contentFieldName, this.metadataFieldName, this.embeddingFieldName);
+			createIndex(this.databaseName, this.collectionName, this.embeddingFieldName, this.indexType,
+					this.metricType, this.indexParameters);
 		}
 
 		R<DescribeIndexResponse> indexDescriptionResponse = this.milvusClient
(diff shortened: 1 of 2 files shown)


sobychacko:
@dev-jonghoonpark, could you verify these changes via a test?

dev-jonghoonpark:
@sobychacko 

Thank you for the feedback.

I have added test using the fact that no logs are output when the operation is successful.

Before applying the PR, errors occur, so the tests fail.
After applying the PR, no errors occur, so the tests pass.

---

For reference, these logs are not output by Spring AI but rather by the Milvus client internally. The errors are only logged and not thrown.

Due to this, I adopted a validation method based on checking the logs.


AbstractMilvusGrpcClient.java :
```java
} catch (Exception e) {
    logError("{} failed! Exception:{}", title, e);
    return R.failed(e);
}
```

sobychacko:
@dev-jonghoonpark What did you mean by the following? 
> However, it is natural for an index to not exist immediately after createCollection.

How about a more straightforward approach to fix this issue? For example, what if you add the initial `describeIndex` call to a `try/catch` block, and then if it causes an exception and the index data is null, create the index? That way, you can avoid the subsequent index creation call (from `if (indexDescriptionResponse.getData() == null) {...}`). 

Here is what I am thinking.

```
if (!isDatabaseCollectionExists()) {
			createCollection(this.databaseName, this.collectionName, this.idFieldName, this.isAutoId,
					this.contentFieldName, this.metadataFieldName, this.embeddingFieldName);
		}

             try {

		R<DescribeIndexResponse> indexDescriptionResponse = this.milvusClient
			.describeIndex(DescribeIndexParam.newBuilder()
				.withDatabaseName(this.databaseName)
				.withCollectionName(this.collectionName)
				.build());
              }
              catch (Exception e) {
                     // if the index is null, then create the index.
              }

             //At this point, index must be created. you can add some assert of something similar to verify that. 

```

What do you think about that? 

Also, the `LogAppender` maybe a bit too much for this test. Why don't you simply assert on the existence of the index and no exception thrown etc?

Thanks!

dev-jonghoonpark:
@sobychacko 

The current schema initialization logic works without any issues.
However, I aimed to improve it by eliminating unnecessary log outputs.

As mentioned in the comments, the logs are generated not by Spring AI but by the internal file `AbstractMilvusGrpcClient.java` in the Milvus client, and they are output without being thrown.

Therefore, even if describeIndex is wrapped in a try-catch block, unnecessary logs will still be output.

Whether `describeIndex` method succeeds or fails, an `R<DescribeIndexResponse>` is returned.
The returned result is then checked to determine whether it contains data or an exception.

```
public class R<T> {
    private Exception exception;
    private Integer status;
    private T data;
```

dev-jonghoonpark:
> What did you mean by the following?
>
> > However, it is natural for an index to not exist immediately after createCollection.

A collection does not have an index immediately after it is created.

Therefore, after calling

https://github.com/spring-projects/spring-ai/blob/1f605ccf0f4ecaadff6147e5b965a76d3ed9d01b/vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/milvus/MilvusVectorStore.java#L445

I added

https://github.com/spring-projects/spring-ai/blob/1f605ccf0f4ecaadff6147e5b965a76d3ed9d01b/vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/milvus/MilvusVectorStore.java#L447

to create the index.

Issue #3604, opened 2025-06-19
PromptTemplate should configure error reporting in String Template


pkubowicz:
**Expected Behavior**

I would like to see details about what is wrong with my prompt template in my logs (that go to LogStash or a similar system), not in standard error.

So if I have code:

```java
PromptTemplate.builder().template("bad {template").build().create()
```
there should be something like
```
11:12:13 org.springframework.ai.template.st.StTemplateRenderer ERROR 1:13: premature EOF
```

**Current Behavior**

org.stringtemplate.v4.misc.ErrorManager#DEFAULT_ERROR_LISTENER is used, which prints using `System.err.println`. So the message is unstructured message in stderr:

```
1:13: premature EOF
```

**Context**

I think StTemplateRenderer should pass a custom implementation of STErrorListener that uses SLF4J instead of System.err.

sunyuhan1998:
Good catch! The default `STErrorListener` outputs exception information through `System.err.println`, which can cause some critical rendering exception details to be lost. 

I've submitted a PR: https://github.com/spring-projects/spring-ai/pull/3606 to try redirecting the exceptions to the log output. Thanks for reporting this issue!

pkubowicz:
Thank you for the quick reaction!

PR #3606, opened 2025-06-19
Fixed GH-3604, Add `STErrorListener` to `StTemplateRenderer`.

sunyuhan1998:
As mentioned in the issue, `StTemplateRenderer` uses `StringTemplate` as the template rendering engine. 

By default, `StringTemplate` use `org.stringtemplate.v4.misc.ErrorManager#DEFAULT_ERROR_LISTENER` as the error listener for rendering exceptions, which outputs errors using `System.err.println`. 

```
public static STErrorListener DEFAULT_ERROR_LISTENER = new STErrorListener() {
        public void compileTimeError(STMessage msg) {
            System.err.println(msg);
        }

        public void runTimeError(STMessage msg) {
            if (msg.error != ErrorType.NO_SUCH_PROPERTY) {
                System.err.println(msg);
            }

        }

        public void IOError(STMessage msg) {
            System.err.println(msg);
        }

        public void internalError(STMessage msg) {
            System.err.println(msg);
        }

        public void error(String s) {
            this.error(s, (Throwable)null);
        }

        public void error(String s, Throwable e) {
            System.err.println(s);
            if (e != null) {
                e.printStackTrace(System.err);
            }

        }
}
```

This can lead to users losing important information about rendering errors. 

This PR introduces a custom `STErrorListener` that logs the exceptions to the application's logging system instead.

PR #3708, opened 2025-06-29
feat(template-st): log StringTemplate compile/runtime errors via SLF4J

renechoi:
### Why

StringTemplate v4 defaults to `ErrorManager.DEFAULT_ERROR_LISTENER`, which
writes compile-time and runtime diagnostics directly to *stderr*.  In most
Spring-based deployments the process stderr stream is either discarded or
treated as an unstructured blob that never reaches Logstash/Grafana, making
template failures virtually invisible at run-time.

Issue **#3604 “PromptTemplate should configure error reporting in String
Template”** highlighted this gap and requested a way to surface those
messages through the conventional SLF4J pipeline.  This commit fulfils that
request and aligns error handling with the rest of the framework logging
strategy.

### What

* **`Slf4jStErrorListener`** – new `STErrorListener` implementation that
  delegates `compileTimeError`, `runTimeError`, `IOError`, and
  `internalError` events to the `Logger` of `StTemplateRenderer`.
* **`StTemplateRenderer`**
  * Allocates a single `STGroup` per render operation and registers the
    SLF4J listener so every template instance receives structured logging.
  * Keeps constructor/builder API unchanged; no public surface breakage.
* **Unit tests**
  * Added `malformedTemplateShouldLogErrorViaSlf4j` which
    - Injects a Logback `ListAppender` to capture log events.
    - Verifies that a malformed template triggers an `ERROR` log entry and
      that nothing is emitted to *stderr*.
  * Existing tests updated to work with the new code path but remain
    behaviour-identical.
* **Docs / Javadoc**
  * Updated class‐level Javadoc in `StTemplateRenderer` to mention the new
    error-reporting mechanism.

### How

When `createST(String template)` is invoked, the renderer now:

1. Creates an `STGroup` with the configured delimiters.
2. Registers the shared `Slf4jStErrorListener` instance on the group.
3. Instantiates the `ST` template from that group.

Because the listener is stateless and threads only interact with it via the
SLF4J facade, the change is thread-safe and produces no additional
allocation hot-spots beyond the lifetime of the template rendering call.

### Impact

* **Operational visibility** – Template compile/runtime errors now appear in
  application logs with timestamps, context, and stack traces, making them
  searchable and monitorable.
* **No breaking changes** – The public API, default delimiters, and
  validation semantics remain untouched.  Projects depending on
  `StTemplateRenderer` require no code changes.
* **Minimal footprint** – Adds ~90 logical lines, zero external
  dependencies.

### Migration Notes

No migration steps are required.  Applications will automatically benefit
from structured error logging after updating to the version that includes
this commit.

### References

* Inspired by **GH-3604** “PromptTemplate should configure error reporting in
  String Template”.
* Supersedes ad-hoc stderr printing by adopting the framework’s SLF4J
  logging conventions.

Signed-off-by: renchoi90@gmail.com

### DIFF
--- a/spring-ai-template-st/src/main/java/org/springframework/ai/template/st/StTemplateRenderer.java
+++ b/spring-ai-template-st/src/main/java/org/springframework/ai/template/st/StTemplateRenderer.java
@@ -25,6 +25,8 @@ import org.antlr.runtime.TokenStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.stringtemplate.v4.ST;
+import org.stringtemplate.v4.STErrorListener;
+import org.stringtemplate.v4.STGroup;
 import org.stringtemplate.v4.compiler.Compiler;
 import org.stringtemplate.v4.compiler.STLexer;
 
@@ -73,6 +75,8 @@ public class StTemplateRenderer implements TemplateRenderer {
 
 	private final boolean validateStFunctions;
 
+	private final STErrorListener stErrorListener = new Slf4jStErrorListener();
+
 	/**
 	 * Constructs a new {@code StTemplateRenderer} with the specified delimiter tokens,
 	 * validation mode, and function validation flag.
@@ -112,12 +116,15 @@ public class StTemplateRenderer implements TemplateRenderer {
 
 	private ST createST(String template) {
 		try {
-			return new ST(template, this.startDelimiterToken, this.endDelimiterToken);
+			STGroup group = new STGroup(this.startDelimiterToken, this.endDelimiterToken);
+			group.setListener(this.stErrorListener);
+			return new ST(group, template);
 		}
 		catch (Exception ex) {
 			throw new IllegalArgumentException("The template string is not valid.", ex);
 		}
 	}
+
 
 	/**
 	 * Validates that all required template variables are provided in the model. Returns
--- a/spring-ai-template-st/src/test/java/org/springframework/ai/template/st/StTemplateRendererTests.java
+++ b/spring-ai-template-st/src/test/java/org/springframework/ai/template/st/StTemplateRendererTests.java
@@ -16,16 +16,23 @@
 
 package org.springframework.ai.template.st;
 
+import static org.assertj.core.api.Assertions.*;
+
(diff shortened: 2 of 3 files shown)


spring-ai-template-st/src/test/java/org/springframework/ai/template/st/StTemplateRendererTests.java
dev-jonghoonpark, line 1: Please reorganize the imports in the following order:

1. java.*
2. other imports
3. org.springframework.*
4. static imports

Also, please avoid using wildcard imports `.*`

---

For reference, you can refer to the example in this file:
[OpenAiApiIT.java](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/api/OpenAiApiIT.java).

dev-jonghoonpark: Additionally, unnecessary tabs have been added. It would be great if you could remove those as well.

spring-ai-template-st/src/main/java/org/springframework/ai/template/st/StTemplateRenderer.java
```diff
@@ -112,13 +116,16 @@ public String apply(String template, Map<String, Object> variables) {
...
+			STGroup group = new STGroup(this.startDelimiterToken, this.endDelimiterToken);
+			group.setListener(this.stErrorListener);
+			return new ST(group, template);
 		}
 		catch (Exception ex) {
 			throw new IllegalArgumentException("The template string is not valid.", ex);
 		}
 	}
 
+
```
dev-jonghoonpark, line 128: This line appears to have an unnecessary line break.

sunyuhan1998:
Hi @renechoi , thank you for your contribution! I think this is a duplicate PR, as the issue is already being addressed through PR #3606 , which is currently under the 1.1.x milestone. I noticed that your PR includes unit tests for the log output, and I think this is a great addition!

renechoi:
Closing this: GH-3604 was fixed upstream by CommonsLoggingStErrorListener, so this is no longer needed.

PR #3716, opened 2025-07-01
Fix annotation package

ngocnhan-tran1996:


sobychacko:
@ngocnhan-tran1996 You need to add the `package-info.java` file to those packages and include the packages there. Ensure that `NonNullApi` remains the default. Once you do that, we won't need the explicit `NonNull` annotation in those modified classes, as that is the default.

ngocnhan-tran1996:
@sobychacko 

I've updated `package-info.java`.

PR #3717, opened 2025-07-02
fix: GH-3713 Fixed outdated code examples in the documentation.

### LINKED ISSUE #3713: Outdated new PromptTemplate(...) calls in Documentation
On the following documentation page:
https://github.com/spring-projects/spring-ai/blob/1.0.x/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/structured-output-converter.adoc the code examples still use the outdated API with direct instantiation via new PromptTemplate(...). This should be updated to use the builder pattern.

sunyuhan1998:
As mentioned in the issue, the code example in the documentation is outdated — the constructor for `PromptTemplate` no longer exists in its previous form. This PR corrects that by updating it to use the Builder pattern.

Fixes #3713


ilayaperumalg:

PR #3751, opened 2025-07-06
docs(mcp): add dynamic MCP client selection example

paoxiaomooo:
- demonstrate how to filter injected MCP clients by server name
  using getServerInfo().name()

This helps users dynamically choose which MCP services to expose
based on context (e.g. permissions, intent), instead of registering
all available tools unconditionally.


liuhongwei1998:
good

ilayaperumalg:
@paoxiaomooo Thanks for the documentation update!

PR #3623, opened 2025-06-20
GH 3421 : Fix silent failures in PDF outline processing

WOONBE:
As identified in issue #3421, the `ParagraphPdfDocumentReader` is not resilient to common imperfections in PDF outline structures. This can lead to an `IndexOutOfBoundsException` or, more critically, silent failures where valid sections of the document are skipped, resulting in an empty or incomplete list of documents.

This PR includes the following changes:

* Corrects the iteration logic in the `get()` method by replacing the buggy iterator-based loop, which failed to process the last outline item, with an indexed `for` loop that guarantees every paragraph is processed correctly.

* Adds a defensive fallback for page range calculation in `getTextBetweenParagraphs()`.

* Refactors the text extraction area calculation, replacing the flawed logic that produced zero-height rectangles. 

* Adds a unit test that reproduces the exact condition that caused the IndexOutOfBoundsException.

Fixed #3421

PR #3756, opened 2025-07-07
 Fix string return conversion in DefaultToolCallResultConverter

JM-Lab:
Return String tool results directly (or "null" if null) when returnType is String.class, instead of serializing them to JSON. This removes unnecessary formatting while preserving existing behavior for other types.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
@@ -35,6 +35,7 @@ import org.springframework.lang.Nullable;
  * A default implementation of {@link ToolCallResultConverter}.
  *
  * @author Thomas Vitale
+ * @author Jemin Huh
  * @since 1.0.0
  */
 public final class DefaultToolCallResultConverter implements ToolCallResultConverter {
@@ -45,7 +46,11 @@ public final class DefaultToolCallResultConverter implements ToolCallResultConve
 	public String convert(@Nullable Object result, @Nullable Type returnType) {
 		if (returnType == Void.TYPE) {
 			logger.debug("The tool has no return type. Converting to conventional response.");
-			return JsonParser.toJson("Done");
+			return "Done";
+		}
+		if (returnType instanceof Class<?> cls && cls.equals(String.class)) {
+			logger.debug("Tool return type is String. Returning result as is.");
+			return String.valueOf(result);
 		}
 		if (result instanceof RenderedImage) {
 			final var buf = new ByteArrayOutputStream(1024 * 4);
--- a/spring-ai-model/src/test/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverterTests.java
+++ b/spring-ai-model/src/test/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverterTests.java
@@ -38,6 +38,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  * Unit tests for {@link DefaultToolCallResultConverter}.
  *
  * @author Thomas Vitale
+ * @author Jemin Huh
  */
 class DefaultToolCallResultConverterTests {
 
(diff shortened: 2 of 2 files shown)


spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
```diff
@@ -45,7 +47,11 @@ public final class DefaultToolCallResultConverter implements ToolCallResultConve
 	public String convert(@Nullable Object result, @Nullable Type returnType) {
 		if (returnType == Void.TYPE) {
 			logger.debug("The tool has no return type. Converting to conventional response.");
-			return JsonParser.toJson("Done");
+			return "Done";
+		}
+		if (returnType instanceof Class<?> cls && cls.equals(String.class)) {
+			logger.debug("Tool return type is String. Returning result as is.");
+			return Optional.ofNullable((String) result).orElse("null");
```
valery1707: What about more simple variant?
```suggestion
			return String.valueOf(result);
```
This will return `"null"` for null `result` and string representation for non-null `result`, same as suggested by PR but more efficiently.

JM-Lab: Thanks for the feedback.
This is implemented intentionally to maintain backward compatibility with version 1.0.0. The previous JsonParser.toJson(result) method returned the literal string "null" for a null input, and this code preserves that exact behavior.

valery1707: And suggested `String.valueOf(result)` will preserve same behavior but without creating instance of `Optional` and less operations.

JM-Lab: Thank you for pointing that out.
You’re absolutely right that String.valueOf(result) exactly matches the behavior I intended: it returns the literal string "null" when result is null, without needing an Optional.

spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
```diff
@@ -22,6 +22,7 @@
 import java.lang.reflect.Type;
 import java.util.Base64;
 import java.util.Map;
+import java.util.Optional;
```
valery1707: This import is needed now?

JM-Lab: Thanks for the review! I've squashed the optimization commits into one and pushed the update.

spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
```diff
@@ -16,25 +16,24 @@
 
 package org.springframework.ai.tool.execution;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.ai.util.json.JsonParser;
+import org.springframework.lang.Nullable;
+
+import javax.imageio.ImageIO;
```
valery1707: As far as I can see in neighboring PRs ([1](https://github.com/spring-projects/spring-ai/pull/3708#discussion_r2173600504), [2](https://github.com/spring-projects/spring-ai/pull/3663#discussion_r2166604335), [3](https://github.com/spring-projects/spring-ai/pull/3623#discussion_r2158913904)), it is recommended to maintain the order of imports in a certain structure:
1. java.*
1. other imports
1. org.springframework.*
1. static imports

Also, please avoid using wildcard imports .*

JM-Lab: Fixed imports, squashed commits, and pushed. Thanks for the review!

valery1707 (approved):

markpollack:
What is wrong serializing it to JSON?  This could maybe go into the 1.1 main branch, but I suspect would be considered a breaking change for 1.0.1    

Thoughts @JM-Lab  ?

JM-Lab:
The change addresses an issue where return values like "\"Done\"" or "\"test\"" cause unnecessary escape handling on the consumer side.
I encountered this while working on a project based on version 1.0.0 and applied a workaround to simplify processing.

I agree this is a breaking change and should be included in version 1.1 to give users a chance to adapt.

Thanks, @markpollack — let me know if there's anything else.

ilayaperumalg:
@JM-Lab Thanks for the PR. Unfortunately, returning the String value for the "String" type response has the following downsides:

1) It is by design that the return value of the tool response is expected to be a valid JSON. This sets the contract for the tool response which gets returned to LLM or the caller (when the tool definition enforces "returnDirect").
2) This change breaks the current experience even if we get this in just for 1.1.x.

While this fix is definitely a convenience, it's better to keep the interchange contract as is. 

I am closing this PR without merging. Please feel free to re-open if you still want to continue the discussion. Thanks once again.

JM-Lab:
@ilayaperumalg Got it, thanks for the clarification! Makes sense to keep the JSON contract consistent.  
Appreciate the review and your time.

PR #3779, opened 2025-07-10
fix: update default sse-message-endpoint from /mcp/message to /mcp/messages

JM-Lab:
This pull request updates the default sse-message-endpoint in the Spring AI MCP server from /mcp/message to /mcp/messages, aligning it with the official MCP specification and the common practice used by MCP Framework (which defaults to /messages) to improve interoperability and avoid confusion.

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server/src/main/java/org/springframework/ai/mcp/server/autoconfigure/McpServerProperties.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server/src/main/java/org/springframework/ai/mcp/server/autoconfigure/McpServerProperties.java
@@ -36,6 +36,7 @@ import org.springframework.util.Assert;
  * All properties are prefixed with {@code spring.ai.mcp.server}.
  *
  * @author Christian Tzolov
+ * @author Jemin Huh
  * @since 1.0.0
  * @see McpServerAutoConfiguration
  */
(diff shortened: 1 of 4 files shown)


markpollack:
i can't find reference to this in the spec, can you provide a link?

JM-Lab:
Thank you for pointing this out. I initially thought the official MCP specification mandated a specific path for the SSE message endpoint, but after rechecking, it turns out it does not. According to [the official MCP specification](https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse), servers should send an `endpoint` event to inform clients of the URI to use for posting messages.

The motivation for this PR is to align with common practice observed in the MCP ecosystem, where `/messages` is frequently used for SSE message endpoints:

* [MCP Official TypeScript SDK examples](https://github.com/modelcontextprotocol/typescript-sdk/blob/c54309598505007c4467f78fa7dfb3e7c763679c/src/examples/server/simpleSseServer.ts#L117) - Uses `/messages` in SSE examples
* [MCP Official Python SDK](https://github.com/modelcontextprotocol/python-sdk/blob/35a9ccdb4229198a3db64585acf85486ec4bec02/src/mcp/server/sse.py#L9) - `SseServerTransport` constructor accepts an endpoint parameter, commonly set to `/messages`

It improves **out-of-the-box interoperability** and reduces confusion. 

Since SSE is **already a deprecated transport** in the MCP spec, if keeping `/mcp/message` is preferred for Spring AI conventions or backward compatibility, I'm happy to close this PR or discuss alternatives.

ilayaperumalg:
@JM-Lab We discussed and agree that at this point we don't want to break anything but just keep the current default "mcp/message".  Thanks for your inputs! Closing this PR without merging.

PR #3784, opened 2025-07-10
fix: Handling default implementation of ToolCallback#call(String,ToolContext)

### LINKED ISSUE #3389: ToolCallback fails if tool context is provided
I have few `ToolCallback`(s)  and one of them is taking use of the `ToolContext`, whereas the others aren't.

e.g.

```java
public class CustomersToolCallback implements ToolCallback {

    @Override
    public ToolDefinition getToolDefinition() {
        return ToolDefinition.builder()
                .name("getCustomerInfo")
                .description("Retrieve customer information")
                .build();

    }

    @Override
    public String call(String toolInput, ToolContext tooContext) {
        if (tooContext == null) {
            throw new UnsupportedOperationException("ToolContext cannot be null");
        }
        Object tenantId = tooContext.getContext().get("tenant");
        // Fetch customer by their id
    }

    @Override
    public String call(String toolInput) {
        throw new UnsupportedOperationException("This method is not supported. Use call(String toolInput, ToolContext toolContext) instead.");
    }
}
```

And another one

```java
public class MockCustomerToolCallback implements ToolCallback {

    @Override
    public ToolDefinition getToolDefinition() {
        return ToolDefinition.builder()
                .name("mockCustomer")
                .description("Retrieve mock customer information")
                .build();

    }

    @Override
    public String call(String toolInput) {
        // Some implementation
    }
}
```

If I have this tool callback registered with the `ChatModel` and I do 

```java
ChatModel chatModel = ...

String response = ChatClient.create(chatModel)
        .prompt("How does a mock customer look like")
        .toolContext(Map.of("tenantId", "acme"))
        .call()
        .content();
```

You can imagine that the call is a bit more dynamic than this. So if this happens, and my `MockCustomerToolCallback` needs to be called it'll fail due to the 
```java
if (tooContext != null && !tooContext.getContext().isEmpty()) {
    throw new UnsupportedOperationException("Tool context is not supported!");
}
```

In the `ToolCallback`.

What exactly is the reason for this `UnsupportedOperationException` on the interface level? It is a bit inconvenient if all of my `ToolCallback` implementations have to implement the method with the `ToolContext` even if they do not need it.

## Fix proposal

I would propose the following:

### Add new interface

Similar to how it is done for Spring Kafka [`GenericMessageListener`](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/GenericMessageListener.java) and its different implementations, or Spring AMQP [`MessageListener`](https://github.com/spring-projects/spring-amqp/blob/main/spring-amqp/src/main/java/org/springframework/amqp/core/MessageListener.java) and its different implementation. Spring AI could  add a `ToolContextAwareToolCallback` which can look something like

```java
public interface ToolContextAwareToolCallback extends ToolCallback {

	@Override
	default String call(String toolInput) {
		throw new UnsupportedOperationException("This method is not supported. Use call(String toolInput, ToolContext toolContext) instead.");
	}

	@Override
	String call(String toolInput, @NonNull ToolContext toolContext);
}
```

This would make it easier to implement, since I won't need to override `ToolCallback#cal(String)`. The exception can be adjusted though.

### Relax the constraint in the `ToolCallback#call(String, ToolContext)`

The `ToolCallback.call(String, ToolContext)` would need to be adjusted to allow empty tool context.

## Potentially similar issues

* https://github.com/spring-projects/spring-ai/issues/2868
* https://github.com/spring-projects/spring-ai/issues/2714 - I believe that this issue is the same, with the only difference being the fact that this is the one for the `@Tool`

ilayaperumalg:
Context: Currently, the default implementation of ToolCallback#call(String,ToolContext) throws exception when an non-empty ToolContext is passed. This gives bad user experience when the application has multiple tool callbacks with one of them doesn't use toolcontext but the chat client passes non-empty toolcontext. For more information #https://github.com/spring-projects/spring-ai/issues/3389

  - This fix removes the explicit exception when the toolcontext is non empty for the default implementation of call(String, ToolContext)
    - Instead, log debug message that tool context will be ignored

  - Also, add explicit null check at the DefaultToolCallingManager when invoking ToolCallback's methods

Resolves #3389

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
@@ -217,7 +217,12 @@ public final class DefaultToolCallingManager implements ToolCallingManager {
 				.observe(() -> {
 					String toolResult;
 					try {
-						toolResult = toolCallback.call(toolInputArguments, toolContext);
+						if (toolContext != null) {
+							toolResult = toolCallback.call(toolInputArguments, toolContext);
+						}
+						else {
+							toolResult = toolCallback.call(toolInputArguments);
+						}
 					}
 					catch (ToolExecutionException ex) {
 						toolResult = this.toolExecutionExceptionProcessor.process(ex);
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/ToolCallback.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/ToolCallback.java
@@ -16,6 +16,9 @@
 
 package org.springframework.ai.tool;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import org.springframework.ai.chat.model.ToolContext;
 import org.springframework.ai.tool.definition.ToolDefinition;
 import org.springframework.ai.tool.metadata.ToolMetadata;
@@ -28,6 +31,8 @@ import org.springframework.lang.Nullable;
  * @since 1.0.0
  */
 public interface ToolCallback {
+
+	Logger logger = LoggerFactory.getLogger(ToolCallback.class);
 
 	/**
 	 * Definition used by the AI model to determine when and how to call the tool.
@@ -53,7 +58,8 @@ public interface ToolCallback {
 	 */
 	default String call(String toolInput, @Nullable ToolContext toolContext) {
 		if (toolContext != null && !toolContext.getContext().isEmpty()) {
-			throw new UnsupportedOperationException("Tool context is not supported!");
+			logger.debug(
+					"Using default implementation of ToolCallback#call(String,ToolContext) and toolContext will be ignored.");
 		}
 		return call(toolInput);
 	}


spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
```diff
@@ -217,7 +217,12 @@ private InternalToolExecutionResult executeToolCall(Prompt prompt, AssistantMess
 				.observe(() -> {
 					String toolResult;
 					try {
-						toolResult = toolCallback.call(toolInputArguments, toolContext);
+						if (toolContext != null) {
```
filiphr: This can never be `null`. The `toolContext` cromes from `buildToolContext`, which never returns `null`.

filiphr (commented):
Thanks for this changes @ilayaperumalg. I've left a comment. The `toolContext` can never be `null`, which means the `if` check is entirely obsolete. Additionally, this only solves #3389 partially.

As a user it is still cumbersome to implement a `ToolCallback` which needs to use the `ToolContext`. I need to override the `call(String)` method and the `call(String, ToolContext)` method. Is there a reason why you are against providing a new interface `ToolContextAwareToolCallback` that would look like:

```java
public interface ToolContextAwareToolCallback extends ToolCallback {

	@Override
	default String call(String toolInput) {
		throw new UnsupportedOperationException("This method is not supported. Use call(String toolInput, ToolContext toolContext) instead.");
	}

	@Override
	String call(String toolInput, @NonNull ToolContext toolContext);
}
```

This interface is using the same pattern as other Spring ecosystem implementations.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
ilayaperumalg: You are right. This can never be `null` and I will remove this unnecessary check.

ilayaperumalg:
> Is there a reason why you are against providing a new interface ToolContextAwareToolCallback

We want to have a solution that can work for 1.0.x as well. Obviously, the new interface can only exist in 1.1.x. 

The simplest option is to ignore the ToolContext in the default implementation of call(String,ToolContext).

filiphr:
>We want to have a solution that can work for 1.0.x as well. Obviously, the new interface can only exist in 1.1.x

Yes indeed the the new interface is not for 1.0.x. The fix (to ignore the `ToolContext` in the default implementation of the `call(String, ToolContext)` is the right one for the 1.0.x release.

If I understood it correctly, you would be open to the new interface in a separate PR for 1.1.x, right? This is in tangent to this PR and fixing the original problem I was facing.

ilayaperumalg:
> If I understood it correctly, you would be open to the new interface in a separate PR for 1.1.x, right? This is in tangent to this PR and fixing the original problem I was facing.

Introducing a new interface in 1.1.x would be fine. But, we are still discussing the need for one more ToolCallback interface and also the default implementation handles this in the PR now.

markpollack:
merged in 3e17e1684b933568871c85bd700df5016d3e5532

backported to 1.0.x

Issue #3481, opened 2025-06-09
@Tool returnDirect Attribute Ignored


mylater:
@Tool returnDirect Attribute Ignored by DefaultToolCallingManager in MCP Integration

**Bug description**
When integrating an MCP service with Spring AI, the `returnDirect` attribute in `@Tool` annotations is being ignored during execution. The `DefaultToolCallingManager` fails to retain this parameter in method metadata, causing MCP services to ​**always perform an extra LLM call** even when `returnDirect = true`.

sunyuhan1998:
In my opinion, this attribute currently cannot take effect on the MCP server side, for the following reasons:

The normal workflow of the `returnDirect` attribute is as follows:

1. The model parses the list of tools and decides whether to invoke a tool and which one to invoke based on the user's prompt.
2. Execute the tool call.
3. All model implementations in Spring AI support the `returnDirect` attribute. If this attribute is set to `true`, the tool call will not be sent to the LLM, but instead returned directly to the user; otherwise, the result of the tool execution will be sent to the model.

This workflow presents some issues in the MCP scenario:

All tools are defined on the MCP server side, and the MCP client obtains the tool definitions from the server and requests the server to invoke the corresponding tool.  
However, according to the MCP specification for tools: [listing-tools](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#listing-tools), the tool itself does not have an attribute like `returnDirect`, It has only a few attributes, such as `name`, `description`, and `inputSchema`:

```
{
  "name": "get_weather",
  "description": "Get current weather information for a location",
  "inputSchema": {
    "type": "object",
    "properties": {
      "location": {
        "type": "string",
        "description": "City name or zip code"
      }
    },
    "required": ["location"]
  }
}
```

 As a result, the client has no way of knowing whether the tool should be returned directly to the user without being sent to the LLM.

In the following code, we can see that when the MCP server exposes the tool definitions, it converts `ToolCallback` into the MCP-standard `McpSchema.Tool`, at which point the `returnDirect` property is lost:

https://github.com/spring-projects/spring-ai/blob/161c437556a3f79761c8d8991f1fa22f68cbc78e/mcp/common/src/main/java/org/springframework/ai/mcp/McpToolUtils.java#L167-L188

[...]

PR #3781, opened 2025-07-10
Upgrade io.modelcontextprotocol 0.10.0 -> 0.11.0-SNAPSHOT

Kehrlann:
For the next development version, we use MCP v0.11.0-SNAPSHOT.

Since this is a SNAPSHOT, it will be the Central Portal Snapshots repo, which you can add to your build file:

```xml
<repositories>
	<repository>
		<name>Central Portal Snapshots</name>
		<id>central-portal-snapshots</id>
		<url>https://central.sonatype.com/repository/maven-snapshots/</url>
		<releases>
			<enabled>false</enabled>
		</releases>
		<snapshots>
			<enabled>true</enabled>
		</snapshots>
	</repository>
	<!-- Spring snapshots & milestones repo ... -->
</repositories>
```

PR #3787, opened 2025-07-11
fix: Fixed the issue where the returnDirect attribute was not effective on the MCP server side.

sunyuhan1998:
### Background

Before this PR, when using the `@Tool` annotation to define tools on the MCP server side and specifying the `returnDirect` attribute, the attribute would not be properly propagated. As a result, the client was unable to retrieve this information. For a detailed discussion and root cause analysis of this issue, please refer to: #3481 .

Thanks to the merge of PR #3781 , we now have the foundation to fix this problem.

---

### Changes in This PR

1. **Preserve the `returnDirect` Attribute**  
   Adjusted the logic in the `McpToolUtils` class that constructs tool declarations, ensuring that the `returnDirect` attribute from the `@Tool` annotation is correctly retained and passed through.

2. **Implemented `getToolMetadata` Method**  
   Added implementations of the `getToolMetadata` method in both `AsyncMcpToolCallback` and `SyncMcpToolCallback` classes. This enables clients to retrieve tool metadata, including the `returnDirect` flag.

3. **Added Unit Tests**  
   Included corresponding unit tests to verify the correctness and stability of the above changes.

Fixes: #3481

### DIFF
--- a/mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallback.java
+++ b/mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallback.java
@@ -17,6 +17,7 @@
 package org.springframework.ai.mcp;
 
 import io.modelcontextprotocol.client.McpSyncClient;
+import io.modelcontextprotocol.spec.McpSchema;
 import io.modelcontextprotocol.spec.McpSchema.CallToolRequest;
 import io.modelcontextprotocol.spec.McpSchema.CallToolResult;
 import io.modelcontextprotocol.spec.McpSchema.Tool;
@@ -30,6 +31,7 @@ import org.springframework.ai.tool.ToolCallback;
 import org.springframework.ai.tool.definition.DefaultToolDefinition;
 import org.springframework.ai.tool.definition.ToolDefinition;
 import org.springframework.ai.tool.execution.ToolExecutionException;
+import org.springframework.ai.tool.metadata.ToolMetadata;
 
 /**
  * Implementation of {@link ToolCallback} that adapts MCP tools to Spring AI's tool
@@ -57,6 +59,7 @@ import org.springframework.ai.tool.execution.ToolExecutionException;
  * }</pre>
  *
  * @author Christian Tzolov
+ * @author Sun Yuhan
  * @see ToolCallback
  * @see McpSyncClient
  * @see Tool
@@ -65,9 +68,13 @@ public class SyncMcpToolCallback implements ToolCallback {
 
 	private static final Logger logger = LoggerFactory.getLogger(SyncMcpToolCallback.class);
 
+	private static final ToolMetadata DEFAULT_TOOL_METADATA = ToolMetadata.builder().build();
+
 	private final McpSyncClient mcpClient;
 
 	private final Tool tool;
+
+	private final ToolMetadata toolMetadata;
 
 	/**
 	 * Creates a new {@code SyncMcpToolCallback} instance.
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
@sunyuhan1998 Thanks for the quick follow up on this!

mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallback.java
```diff
@@ -65,10 +68,14 @@ public class SyncMcpToolCallback implements ToolCallback {
 
 	private static final Logger logger = LoggerFactory.getLogger(SyncMcpToolCallback.class);
 
+	private static final ToolMetadata DEFAULT_TOOL_METADATA = ToolMetadata.builder().build();
```
rlagyu0, line 53: The default value of returnDirect is false.

sunyuhan1998: Yes, factually its default value is indeed `false`, but I think we shouldn't manually set it to `false` here; instead, we should let it follow the default implementation of the `returnDirect()` method in `ToolMetadata`.

KallivdH:
Will this fix be added to the 1.1.0.M3 pre-release @markpollack ?

ilayaperumalg:
Hi @sunyuhan1998 Sorry about the delay in getting to this. Having the `returnDirect` as MCP schema ToolAnnotations attribute in the MCP JavaSDK is rather invasive as the MCP spec and its implementation shouldn't know anything about Spring AI's behaviour. We need to find a better approach to handle `returnDirect` instead of the current MCP Schema ToolAnnotations. What do you think?

sunyuhan1998:
> Hi @sunyuhan1998 Sorry about the delay in getting to this. Having the `returnDirect` as MCP schema ToolAnnotations attribute in the MCP JavaSDK is rather invasive as the MCP spec and its implementation shouldn't know anything about Spring AI's behaviour. We need to find a better approach to handle `returnDirect` instead of the current MCP Schema ToolAnnotations. What do you think?

You're right. At the time this PR was conceived and submitted, the [mcp-annotations](https://github.com/spring-ai-community/mcp-annotations) project hadn't yet introduced the `@McpTool` annotation. Currently, `@McpTool` is strictly implemented according to the MCP specification and does not include any logic related to `returnDirect`. I suspect that using the `@Tool` annotation on the MCP server side early on was perhaps a workaround or a shortcut? I'm curious about our future plans for balancing the capabilities of `@McpTool` and `@Tool` on the MCP server side. Do we intend to gradually phase out the use of `@Tool` for defining tools in the MCP server going forward?

ilayaperumalg:
@sunyuhan1998 We intend to use `@McpTool` for the tools specific to MCP servers while the Spring AI `@Tool` for non MCP tools. But, currently, for backwards compatibility and the ease of use, we support both the options. 

> Do we intend to gradually phase out the use of @Tool for defining tools in the MCP server going forward?

I agree. This could be something in Spring AI 2.0 cc @tzolov

sunyuhan1998:
> @sunyuhan1998 We intend to use `@McpTool` for the tools specific to MCP servers while the Spring AI `@Tool` for non MCP tools. But, currently, for backwards compatibility and the ease of use, we support both the options.
> 
> > Do we intend to gradually phase out the use of @tool for defining tools in the MCP server going forward?
> 
> I agree. This could be something in Spring AI 2.0 cc @tzolov

Thank you for your explanation. I understand now. If that's the case, does it mean the issue we've been discussing in this PR no longer needs to be followed up on? Should I close this PR?

ilayaperumalg:
@sunyuhan1998 Yes, please. Let's close this PR.

KallivdH:
Thank you for following up!

> Hi @sunyuhan1998 Sorry about the delay in getting to this. Having the `returnDirect` as MCP schema ToolAnnotations attribute in the MCP JavaSDK is rather invasive as the MCP spec and its implementation shouldn't know anything about Spring AI's behaviour. We need to find a better approach to handle `returnDirect` instead of the current MCP Schema ToolAnnotations. What do you think?

@sunyuhan1998 The audience [annotation](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#annotations) look like a right fit for the `returnDirect` functionality. For example, if audience `user` is specified.

PR #3801, opened 2025-07-12
Feature (template): Supports multi character template separators

zjxzjw:
- Allow the use of multiple characters as start and end separators for templates
- Implemented the function of converting custom separators into ST built-in format
- Updated test cases to support new multi character delimiter functionality
- Relevant unit tests have been provided
- This PR resolves Issue # 3800

### DIFF
--- a/spring-ai-template-st/src/main/java/org/springframework/ai/template/st/StTemplateRenderer.java
+++ b/spring-ai-template-st/src/main/java/org/springframework/ai/template/st/StTemplateRenderer.java
@@ -16,9 +16,12 @@
 
 package org.springframework.ai.template.st;
 
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


zjxzjw:
hello 👋 @ilayaperumalg， I have conducted unit testing again and the modifications I made have been successfully approved. Could you please re approve and review them? Thank you

ericbottard:
Hi,

below is my personal opinion on this issue/PR. We'll likely debate this as a team.

I would be in favor of rejecting this PR as it is likely to introduce complications down the line (eg how to distinguish between variables that are now `{foo}` vs. appearances of such tokens in the original text).
I strongly feel that this is not the correct way to address the issue described. Users should embrace the underlying templating engine features, if such features are present. Spring AI trying to hide those will likely result in a roadblock at some point. For the problem at hand, if ST provides a way to escape `{}` so that they are not mistaken for variable placeholders, then users should use that to protect their prompt templates.

Also, we have PRs open for other rendering engines (SpEL & Jinja) which may suit your needs.

Issue #3786, opened 2025-07-11
Add Support for OpenAI based64 format embedding responses


asaikali:
OpenAI supports returning embedding responses as base64 encoded which can save a lot of resources see this excellent blog post [Speed Up OpenAI Embedding By 4x With This Simple Trick!](https://techcommunity.microsoft.com/blog/azuredevcommunityblog/speed-up-openai-embedding-by-4x-with-this-simple-trick/4390081) 

Currently Spring AI assumes that the response from OpenAI is coming back as an array of floats see https://github.com/spring-projects/spring-ai/blob/6a1f982203a1de3f1021a7b2279d366cf45afa93/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiEmbeddingModel.java#L181 

The embedding class constructor assumes a `float[]` https://github.com/spring-projects/spring-ai/blob/6a1f982203a1de3f1021a7b2279d366cf45afa93/spring-ai-model/src/main/java/org/springframework/ai/embedding/Embedding.java#L40

You can find info about how the offical python SDK decodes the response at https://github.com/openai/openai-python/blob/db5c35049accb05f5fb03791ef9c12547fd309a7/src/openai/resources/embeddings.py#L204 

The offical python SDK uses base64 encoding as of version 1.62 and other sdks are turning on base64 as the default. 

Please add support for base64 embeddings to spring AI.

sunyuhan1998:
Hi @ilayaperumalg ,  Could you assign this issue to me? Considering that we now support the `encodingFormat` parameter, I'd like to implement what's described in this issue. We don't need many changes, and I want to submit a PR for it. 

The main idea is to add a custom `JsonDeserializer` for the `embedding` field in `org.springframework.ai.openai.api.OpenAiApi.Embedding`. If the returned embedding is in base64 format, we can convert it to a `float[]` within the `JsonDeserializer`. 

I believe this approach can fulfill the current requirement with minimal impact on the existing code.

ilayaperumalg:
@sunyuhan1998 Thank you for your interest in contributing. Looking forward to your PR!

netboy33:
> Hi [@ilayaperumalg](https://github.com/ilayaperumalg) , Could you assign this issue to me? Considering that we now support the `encodingFormat` parameter, I'd like to implement what's described in this issue. We don't need many changes, and I want to submit a PR for it.
> 
[...]

PR #3804, opened 2025-07-13
GH-3786: Provide `encodingFormat=base64` support for the OpenAI embedding model

sunyuhan1998:
As mentioned in the issue #3786 , when using embedding models, we need to support the `encodingFormat=base64` parameter. When this format is used, the model returns the `embedding` field as a base64-encoded string representing a float array. This PR adds support for handling that format.

**Changes in this PR:**

1. Introduced a custom JSON deserializer, `OpenAiEmbeddingDeserializer`, for the `embedding` field in `org.springframework.ai.openai.api.OpenAiApi.Embedding`. This deserializer automatically converts the base64-encoded string back into a float array when parsing the API response.
2. Added corresponding unit tests.

Fixes: #3786


sunyuhan1998:
Hi @ilayaperumalg can you help review this PR ? Thank you !

ilayaperumalg:
@sunyuhan1998 Thank you for the PR! Rebased, squashed and merged as 9e43b7e10

Issue #494, opened 2024-03-21
native compilation under java 22 fails for an example that works on java 21


joshlong:
it's not critical or anything but we should probably understand it..

code: 

```java
package com.example.service;

import org.springframework.ai.chat.ChatClient;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.context.annotation.Bean;
import org.springframework.web.servlet.function.RouterFunction;
import org.springframework.web.servlet.function.ServerResponse;

import java.util.Map;

import static org.springframework.web.servlet.function.RouterFunctions.route;
import static org.springframework.web.servlet.function.ServerResponse.ok;

@SpringBootApplication
public class ServiceApplication {

    public static void main(String[] args) {
        SpringApplication.run(ServiceApplication.class, args);
    }

    @Bean
    RouterFunction<ServerResponse> storyRoutes(ChatClient ai ) {
        var prompt = """
                tell me a joke
                """;
        return route()
            .GET("/story", r -> ok().body(Map.of("story", ai.call(prompt))))
            .build();
    }
}
```

build:

```
plugins {
	id 'java'
	id 'org.springframework.boot' version '3.2.4'
	id 'io.spring.dependency-management' version '1.1.4'
	id 'org.graalvm.buildtools.native' version '0.9.28'
}

group = 'com.example'
version = '0.0.1-SNAPSHOT'

java {
	sourceCompatibility = '21'
}

repositories {
	mavenCentral()
	maven { url 'https://repo.spring.io/milestone' }
}

ext {
	set('springAiVersion', "0.8.1")
}

dependencies {
	implementation 'org.springframework.boot:spring-boot-starter-web'
	implementation 'org.springframework.ai:spring-ai-vertex-ai-gemini-spring-boot-starter'
	testImplementation 'org.springframework.boot:spring-boot-starter-test'
}

dependencyManagement {
	imports {
		mavenBom "org.springframework.ai:spring-ai-bom:${springAiVersion}"
	}
}

tasks.named('test') {
	useJUnitPlatform()
}
```

error

```

➜  service git:(main) ./gradlew nativeCompile  | pbcopy 
Mar 21, 2024 6:20:20 PM com.google.api.gax.nativeimage.NativeImageUtils registerClassForReflection
WARNING: Failed to find io.grpc.netty.shaded.io.netty.channel.ProtocolNegotiators on the classpath for reflection.
[...]

PR #3833, opened 2025-07-17
Fix GraalVM native compilation issue with Java 22

academey:
## Summary
- Fixes native compilation failure when using Java 22 with GraalVM
- Adds proper RuntimeHints for SLF4J service providers
- Ensures compatibility with both Java 21 and Java 22

## Details

This PR resolves issue #494 where native compilation fails under Java 22 but works on Java 21. The root cause was that SLF4J service providers (`NOP_FallbackServiceProvider` and `SubstituteServiceProvider`) were being initialized at runtime instead of build time in GraalVM native images.

### Changes made:
1. Updated `SpringAiCoreRuntimeHints.java` to register SLF4J classes for reflection and build-time initialization
2. Added comprehensive tests in `SpringAiCoreRuntimeHintsTests.java`

### Error fixed:
```
Error: An object of type 'org.slf4j.helpers.NOP_FallbackServiceProvider' was found in the image heap...
Error: An object of type 'org.slf4j.helpers.SubstituteServiceProvider' was found in the image heap...
```

## Test Plan
- [x] Added unit tests for all registered types
- [x] Verified SLF4J types are properly registered
- [ ] Test native compilation with Java 21
- [ ] Test native compilation with Java 22
- [ ] Ensure no regression in regular JVM mode

Fixes spring-projects#494


spring-ai-model/src/test/java/org/springframework/ai/aot/SpringAiCoreRuntimeHintsTests.java
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright 2023-2024 the original author or authors.
```
sunyuhan1998: 2025

dingcodingco: I fixed it thanks

ilayaperumalg:
@joshlong Could you help with the review of the changes? Thanks

joshlong:
hi - thanks for submitting ths.  unfortunately im not sure how to break the code using java 23, 24, and 25 and modern versions of spring ai. the bug against which the original issue was filed was from march of 2024. 

would you mind sending an example of something that breaks and which then works if you add the hints you show in the PR?

joshlong:
if i go to the spring initializr, choose java 25, maven, gemini, and web, then add 


```java
package com.example.pr;

import org.springframework.boot.ApplicationRunner;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;

@SpringBootApplication
public class PrApplication {

        SpringApplication.run(PrApplication.class, args);
    }
    @Bean
    ApplicationRunner runner(ChatClient.Builder builder) {
        return args -> {
            var res = builder.build().prompt("tell me a joke").call().content();
            System.out.println(res);
        };
    }

}

```
 
this seems to work just fine. not sure what error im looking for

prashantguleria:
Same problem with Spring boot 4.0.0-M3

academey:
Hi @joshlong (it's really really really really big honor to me for mentioning you)
You're right. I investigated and confirmed that `com.google.api:gax:2.65.0` (currently used via`com.google.cloud:libraries-bom:26.60.0`) no longer contains `org.slf4j.LoggerFactory` in its native-image configuration. The upstream gax library has already fixed this issue.

Regarding the Spring Boot 4.0.0-M3 comment @prashantguleria, Could you share the specific error you encountered? This would help determine ifthere's a different issue, since my tests show the problem is already resolved in the current dependency versions. Unless we can reproduce the issue with the current dependencies, this PR appears unnecessary.

ericbottard:
Reproduced using an adaptation of Josh's pom.xml in the linked PR, + Graal 23. Fixed by building against the PR changes.
If this turned out to be a false positive, we can always reconsider

Merged to main as 4969dbc39, thanks for your contribution @academey

PR #3874, opened 2025-07-22
GH-3873 Add an `api-version` option for Azure OpenAI

### LINKED ISSUE #3873: API version not supported

**Expected Behavior**

Should be able to overwrite the api version

**Current Behavior**

By deafult going as gpt-4o/chat/completions?api-version=2025-01-01-preview,
even spring.ai.azure.openai.chat.options.api-version=2024-11-21
did not help

**Context**

sunyuhan1998:
As mentioned in the issue, almost all interfaces of `azure-openai` support the `api-version` option. This is not a parameter within the request body, but rather passed as a query parameter in the URL (refer to the official documentation: https://learn.microsoft.com/en-us/azure/ai-foundry/openai/reference-preview-latest#create-chat-completion). Currently, Spring AI does not support this option. This PR adds that support.

What this PR does:

1. Adds an `apiVersion` configuration option to `AzureOpenAiChatOptions`, `AzureOpenAiAudioTranscriptionOptions`, and `AzureOpenAiEmbeddingOptions`. The parameter is then passed through to the corresponding models and ultimately provided to the Azure SDK. The transmission is achieved via the SDK's `RequestOptions`, which supports appending query parameters to the request URL.

2. I initially intended to add this option to `AzureOpenAiImageOptions` as well. However, the Azure SDK does not expose any method accepting `RequestOptions` as public for image-related operations(`com.azure.ai.openai.OpenAIClient#getImageGenerationsWithResponse(java.lang.String, com.azure.core.util.BinaryData, com.azure.core.http.rest.RequestOptions)`). I believe this is a bug, as it is inconsistent with the behavior of other interfaces. I have submitted a PR to Azure to attempt to fix it: 
https://github.com/Azure/azure-sdk-for-java/pull/46102
Once that PR is accepted, I will follow up with another PR to Spring AI to add the `apiVersion` option to `AzureOpenAiImageOptions`.

Fixes: #3873

### DIFF
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiAudioTranscriptionModel.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiAudioTranscriptionModel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


sdeleuze:
Azure OpenAI model is not supported anymore, please use OpenAI model instead.

Issue #3884, opened 2025-07-22
When the Tool parameter is an empty string, a type conversion error occurs, reporting java.lang.NumberFormatException


lambochen:
Please do a quick search on GitHub issues first, there might be already a duplicate issue for the one you are about to create.
If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

**Bug description**
A clear and concise description of what the bug is about.

A type conversion error occurs when the Tool parameter is an empty string, throwing java.lang.NumberFormatException.

- The parameter type of the tool is Long.
- The Tool parameter responded by the LLM is an empty string.

```
java.lang.NumberFormatException
	at java.base/java.math.BigDecimal.<init>([BigDecimal.java:692](http://bigdecimal.java:692/))
	at java.base/java.math.BigDecimal.<init>([BigDecimal.java:471](http://bigdecimal.java:471/))
	at java.base/java.math.BigDecimal.<init>([BigDecimal.java:900](http://bigdecimal.java:900/))
	at [org.springframework.ai](http://org.springframework.ai/).util.json.JsonParser.toTypedObject([JsonParser.java:139](http://jsonparser.java:139/))
	at [org.springframework.ai](http://org.springframework.ai/).tool.method.MethodToolCallback.buildTypedArgument([MethodToolCallback.java:150](http://methodtoolcallback.java:150/))
	at [org.springframework.ai](http://org.springframework.ai/).tool.method.MethodToolCallback.lambda$buildMethodArguments$1([MethodToolCallback.java:139](http://methodtoolcallback.java:139/))
	at java.base/[java.util.stream](http://java.util.stream/).ReferencePipeline$3$1.accept([ReferencePipeline.java:197](http://referencepipeline.java:197/))
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining([Spliterators.java:992](http://spliterators.java:992/))
	at java.base/[java.util.stream](http://java.util.stream/).AbstractPipeline.copyInto([AbstractPipeline.java:509](http://abstractpipeline.java:509/))
	at java.base/[java.util.stream](http://java.util.stream/).AbstractPipeline.wrapAndCopyInto([AbstractPipeline.java:499](http://abstractpipeline.java:499/))
	at java.base/[java.util.stream](http://java.util.stream/).AbstractPipeline.evaluate([AbstractPipeline.java:575](http://abstractpipeline.java:575/))
[...]

PR #3885, opened 2025-07-22
bugfix: When the Tool parameter is an empty string, a type conversion error occurs

lambochen:
ref issue: https://github.com/spring-projects/spring-ai/issues/3884

Thank you upfront!

### DIFF
--- a/spring-ai-model/src/test/java/org/springframework/ai/util/json/JsonParserTests.java
+++ b/spring-ai-model/src/test/java/org/springframework/ai/util/json/JsonParserTests.java
@@ -262,6 +262,12 @@ class JsonParserTests {
 		assertThat(input).isEqualTo(result);
 	}
 
+	@Test
+	void emptyString() {
+		var value = JsonParser.toTypedObject("", Long.class);
+		assertThat(value).isEqualTo(null);
+	}
+
 	record TestRecord(String name, Integer age) {
 	}
 


spring-ai-model/src/main/java/org/springframework/ai/util/json/JsonParser.java
```diff
@@ -141,7 +143,11 @@ public static Object toTypedObject(Object value, Class<?> type) {
 		if (javaType == String.class) {
 			return value.toString();
 		}
-		else if (javaType == Byte.class) {
+		if (value.toString().isEmpty()) {
+			return null;
```
sunyuhan1998, line 147: The `package-info` for the package containing `JsonParser` has already been declared as `@NonNullApi`, so I think we should never return null under any circumstances. 

I believe the current issue is essentially caused by a type mismatch between the parameters provided by the model when calling the tool and the actual parameter types expected by the tool. Therefore, I suggest we directly throw a clear exception here, including in the exception message the parameter types expected by the tool, as well as the parameter types (and values) actually provided by the model. 

This would also help users improve their prompts and tool descriptions.

sunyuhan1998 (changes requested):

spring-ai-model/src/main/java/org/springframework/ai/util/json/JsonParser.java
hongling-dev: Hi, @sunyuhan1998 
Thank you for your comment.

> throw a clear exception

Regarding this point, I have a different opinion.
From the perspective of this package, it indeed should not return null, as @NonNullApi has been declared. Excuse me for overlooking this when coding; I will make the adjustment.
But from the perspective of Tool Calling, parameter conversion can allow null values because the tool parameter may be optional.

In summary, I think solving this issue might be better addressed by the caller of JsonParser, i.e., the Tool Calling parameter parsing layer.

sunyuhan1998: > parameter conversion can allow null values because the tool parameter may be optional.

I agree with your point that tool parameters should indeed allow null values. 

I think perhaps we are currently facing two distinct issues: 

1. Scenarios where we should allow tool parameters to be null. 
3. Scenarios where the parameters provided by the model do not match the required parameter types of the tool. 

I’m trying to express mainly relates to the second scenario. In your example, the tool expects a parameter of type `Long`, but the model outputs an empty string. If we extend this discussion further, suppose the tool expects a `Boolean` parameter, but the model provides the number `5`—how should we handle such cases? Therefore, I believe it's necessary to provide users with clear feedback when such mismatches occur.

hongling-dev: I agree with your point, provide users with clear feedback when there is a type mismatch.

sobychacko:
Thanks for the PR!

Closing as stale. This logic moved to a different class and the original is now deprecated.

Also unresolved from review: returning null isn't allowed in this package, and the suggestion was to throw a clear error instead. Issue #3884 stays open, and #6018 takes that approach, though it needs the same relocation.

Issue #3790, opened 2025-07-11
Add ChatClient getChatModel() method


timothyBrake:
**Expected Behavior**

When injecting the ChatClient in any service I want to log the actual underlying model being used. Because we have multiple ChatClients of different underlying models, we want to log the actual model being used in the service that uses the ChatClient. E.g.

```
public class QcrIntentAgent {

    @Autowired
    private ChatClient chatClient;


    public void doSomething(){
        log.debug("LLM being used: {}", chatClient.getChatModel().getModel());
       // E.g.: 
       // LLM being used: gpt-4.1-nano 
       // LLM being used: gpt-4o-mini 
       // LLM being used: mistral  
       //. LLM being used: llama3  
       chatClient.prompt(...);
    }

```

**Current Behavior**

There is no clean way of getting this info besides injecting the ChatModel itself. Resulting in either injecting both ChatModel and ChatClient OR injecting just the model and creating a new ChatClient here.

```
    @Autowired
    private ChatClient chatClient;

    @Autowired
    private ChatModel chatModel;

     public void doSomething(){
        log.debug("LLM being used: {}", chatModel.getDefaultOptions().getModel());
         chatClient.prompt(...);
     }

// OR


   @Autowired
    private ChatModel chatModel;

    public void doSomething(){
        log.debug("LLM being used: {}", chatModel.getDefaultOptions().getModel());
        ChatClient.builder(chatModel).build().prompt(...);
     }


 ```

**Context**

Alternatives are see Current Behavior.
Creating a utility class 

 ```
@UtilityClass
public class ChatClientUtils {

    /**
     * Returns the model configured for the ChatClient.
     * This workaround is needed because Spring AI 1.0.0 does not expose the model via public API.
     * WARNING: This depends on the internal structure of DefaultChatClient.
     */
    public static String getModel(ChatClient chatClient) {
        if (chatClient instanceof DefaultChatClient defaultClient) {
            DefaultChatClient.DefaultChatClientRequestSpec requestSpec = (DefaultChatClient.DefaultChatClientRequestSpec) defaultClient.prompt("Hello");
            return requestSpec.getChatOptions().getModel();
        } else {
            throw new IllegalArgumentException("Unsupported ChatClient implementation: " + chatClient.getClass().getName());
        }
    }
}
 ```

zjxzjw:
PR has been submitted. Please review and approve it. Thank you

YunKuiLu:
You can try the following way to get the model name: 
[...]

PR #3889, opened 2025-07-23
Adds support for retrieving the model name directly from DefaultChatClient

zjxzjw:
✨ PR: Expose Model Name from ChatClient
﻿
📝 Summary
﻿
This PR adds support for retrieving the model name directly from `DefaultChatClient`.

🚀 Motivation
﻿
In multi-model environments, we often need to log or observe which underlying model is being used (e.g., `gpt-4o`, `llama3`, etc.).  
Currently, this requires injecting both `ChatClient` and `ChatModel`, or rebuilding the client manually.  
This change simplifies the process and improves observability.
﻿
🔧 Changes
﻿
- 🧠 Stored the `ChatModel` inside `DefaultChatClient` for later access  
- 🏗️ Modified `DefaultChatClientBuilder` to pass the `ChatModel` when building  
- ✅ Added unit test to validate the new functionality  

This PR resolves issue #3790

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
@@ -20,6 +20,7 @@ import java.net.URL;
 import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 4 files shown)


ilayaperumalg:
It would be better if we can improve the logging of ChatModel related metadata instead of exposing the ChatModel object from the ChatClient. Also, by design, we don't want to expose the `ChatModel` which in fact, is a REST client and needs to be immutable and well encapsulated.

Issue #3896, opened 2025-07-23
Similarity searches with the `MariaDBVectorStore` do not provide a score


Til7701:
**Bug description**
The `MariaDBVectorStore` does not return the score when doing a similarity search. Instead it adds the distance to the metadata of the document.

**Environment**
Spring AI 1.0.0

**Steps to reproduce**
1. Embed documents and save them in the `MariaDBVectorStore`
2. Do a similarity search

**Expected behavior**
Get the score via the getter of the document.

**Actual behaviour**
The score is `null` and the distance is in the metadata.

**Minimal Complete Reproducible example**
I did not create an example application, but here is the mapper method for the `MaraDBVectorStore`.

```java
Override
public Document mapRow(ResultSet rs, int rowNum) throws SQLException {
	String id = rs.getString(1);
	String content = rs.getString(2);
	Map<String, Object> metadata = toMap(rs.getString(3));
	float distance = rs.getFloat(4);

	metadata.put("distance", distance);

	return new Document(id, content, metadata);
}
```

In comparison, the `PgVectorStore`, which also uses the `JdbcTemplate` with a custom `DocumentRowMapper` does this differently:

```java
@Override
public Document mapRow(ResultSet rs, int rowNum) throws SQLException {
	String id = rs.getString(COLUMN_ID);
	String content = rs.getString(COLUMN_CONTENT);
	PGobject pgMetadata = rs.getObject(COLUMN_METADATA, PGobject.class);
	Float distance = rs.getFloat(COLUMN_DISTANCE);

	Map<String, Object> metadata = toMap(pgMetadata);
	metadata.put(DocumentMetadata.DISTANCE.value(), distance);

	// @formatter:off
	return Document.builder()
		.id(id)
		.text(content)
		.metadata(metadata)
		.score(1.0 - distance)
		.build(); // @formatter:on
}
```

moupriyaroy25:
If it's a open bug, can I take this up n try?

Til7701:
I already opened a PR for this. If you are interested, you could test, if the fix works correctly.

PR #3898, opened 2025-07-23
GH-3896 Similarity searches with the MariaDBVectorStore do not provide a score

Til7701:
Fixes GH-3896 (https://github.com/spring-projects/spring-ai/issues/3896)

The MariaDBVectorStore no longer puts the distance in the metadata when doing a similarity search. Instead, the score is calculated from the distance and set via the builder.


sobychacko:
@Til7701 The PR is merged. I restored providing the distance as part of the metadata for backward-compatibility reasons while keeping the score changes you introduced. Also, back-ported to `1.0.x`.

PR #3900, opened 2025-07-23
Add `KotlinModule` support for schema generation in `BeanOutputConverter`

dmitrysulman:
It is stated in the **Integrating with BeanOutputConverter Utilities** section of the OpenAI ChatModel [documentation](https://docs.spring.io/spring-ai/reference/api/chat/openai-chat.html#_integrating_with_beanoutputconverter_utilities) that

> Kotlin reflection is used to infer which property are required or not based on the nullability of types and default values of parameters, so for most use case `@get:JsonProperty(required = true)` is not needed.

However, this is not true for `BeanOutputConverter`, because `KotlinModule` that infers the required Kotlin properties is used only in the `ModelOptionsUtils` class.

This PR fixes the issue by adding Kotlin class support to the `BeanOutputConverter`.


dmitrysulman:
Hi @ilayaperumalg, kindly asking if there’s a chance to review and approve this PR when you have time. Thank you!

PR #3901, opened 2025-07-23
Add MCP Client tool predicate for filtering the MCP tools

ilayaperumalg:
- Introduce MCP Sync/Async client BiPredicate interface as a tool filter for the MCP Sync/Async ToolCallbackProvider to use when filtering the MCP tools
 - Update MCP ToolCallbackAutoConfiguration to use these BiPredicate beans when defined (default is to allow all)
 - Add test verifying the tool filter configuration on both sync and async toolcallback provider auto-configuration
 - Update the unit tests for the MCP toolcallback provider

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/main/java/org/springframework/ai/mcp/client/common/autoconfigure/McpToolCallbackAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/main/java/org/springframework/ai/mcp/client/common/autoconfigure/McpToolCallbackAutoConfiguration.java
@@ -17,13 +17,18 @@
 package org.springframework.ai.mcp.client.common.autoconfigure;
 
 import java.util.List;
+import java.util.function.BiPredicate;
 
 import io.modelcontextprotocol.client.McpAsyncClient;
 import io.modelcontextprotocol.client.McpSyncClient;
+import io.modelcontextprotocol.spec.McpSchema;
 
 import org.springframework.ai.mcp.AsyncMcpToolCallbackProvider;
+import org.springframework.ai.mcp.McpAsyncClientBiPredicate;
+import org.springframework.ai.mcp.McpSyncClientBiPredicate;
 import org.springframework.ai.mcp.SyncMcpToolCallbackProvider;
 import org.springframework.ai.mcp.client.common.autoconfigure.properties.McpClientCommonProperties;
+import org.springframework.ai.tool.annotation.Tool;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.AllNestedConditions;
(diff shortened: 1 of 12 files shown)


auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/main/java/org/springframework/ai/mcp/client/common/autoconfigure/McpToolCallbackAutoConfiguration.java
```diff
@@ -51,16 +56,21 @@ public class McpToolCallbackAutoConfiguration {
 	@Bean
 	@ConditionalOnProperty(prefix = McpClientCommonProperties.CONFIG_PREFIX, name = "type", havingValue = "SYNC",
 			matchIfMissing = true)
-	public SyncMcpToolCallbackProvider mcpToolCallbacks(ObjectProvider<List<McpSyncClient>> syncMcpClients) {
+	public SyncMcpToolCallbackProvider mcpToolCallbacks(ObjectProvider<McpSyncClientBiPredicate> syncClientsToolFilter,
```
sunyuhan1998: The documentation of the `syncClientsToolFilter` parameter needs to be supplemented in the method notes here, the rest looks fine to me, thanks for your excellent work!

sunyuhan1998 (changes requested):

auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/main/java/org/springframework/ai/mcp/client/common/autoconfigure/McpToolCallbackAutoConfiguration.java
ilayaperumalg: @sunyuhan1998 Thanks for the review!

ilayaperumalg: I wonder if it would be better to have list of filters instead of a single filter as one can provide multiple filters.

sunyuhan1998: > I wonder if it would be better to have list of filters instead of a single filter as one can provide multiple filters.

If we are to accept multiple filters here, does that mean the final filtering result is the intersection of all the filters' results?

ilayaperumalg: Yeah, it would be an intersection of all the filters. But, we can only have tool specific filtering, not a client specific ones. Given the current implementation returns the ToolCallbacks for all the clients, for now, it wouldn't be possible to have any client specific filtering logic. We need a separate story to track this.

sunyuhan1998: > we can only have tool specific filtering, not a client specific ones. Given the current implementation returns the ToolCallbacks for all the clients, for now, it wouldn't be possible to have any client specific filtering logic. We need a separate story to track this.

I agree that we should track separately the issue of how to filter different tools for different clients.

Regarding the current filter design, I suggest that keeping a single filter should be sufficient, as this would keep our interface clear and simple. If users truly have the need for combined filtering, they can fully implement such composite filtering logic within `syncClientsToolFilter`. Even nesting multiple `Predicate` instances inside `syncClientsToolFilter` would be perfectly acceptable. — This is just my suggestion.

ilayaperumalg: Great feedback! Thank you. I like the idea of using nested Predicate instances if there is a need for multiple filters.

ilayaperumalg:
Updated the PR to introduce MCP client metadata for filtering.

mcp/common/src/main/java/org/springframework/ai/mcp/AsyncMcpToolCallbackProvider.java
```diff
@@ -147,7 +146,8 @@ public ToolCallback[] getToolCallbacks() {
 			ToolCallback[] toolCallbacks = mcpClient.listTools()
 				.map(response -> response.tools()
 					.stream()
-					.filter(tool -> this.toolFilter.test(mcpClient, tool))
+					.filter(tool -> this.toolFilter.test(new McpClientMetadata(mcpClient.getClientCapabilities(),
+							mcpClient.getClientInfo(), mcpClient.initialize().block()), tool))
```
ilayaperumalg: The use of client.initialize() will be replaced once we have access to the client.getCurrentInitializationResult() from https://github.com/modelcontextprotocol/java-sdk/pull/424

mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallbackProvider.java
```diff
@@ -133,7 +131,8 @@ public ToolCallback[] getToolCallbacks() {
 			.flatMap(mcpClient -> mcpClient.listTools()
 				.tools()
 				.stream()
-				.filter(tool -> this.toolFilter.test(mcpClient, tool))
+				.filter(tool -> this.toolFilter.test(new McpClientMetadata(mcpClient.getClientCapabilities(),
+						mcpClient.getClientInfo(), mcpClient.initialize()), tool))
```
ilayaperumalg: The use of client.initialize() will be replaced once we have access to the client.getCurrentInitializationResult() from https://github.com/modelcontextprotocol/java-sdk/pull/424

tzolov:
Rebase, refactored and merged at b2b0757006486df9406894f6b89e0a3c984cdc70

PR #3906, opened 2025-07-24
chore: made embedding options builder methods consistent

garethjevans:
* Deprecated methods that start with set and with.
* Added replacements


ilayaperumalg:
@garethjevans Thanks for the PR!

PR #3927, opened 2025-07-28
Update BeanOutputConverter.java

chang6666:
Resolved the exceptional issue of converting the large model deployed locally with Ollama into a Java Bean.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
@@ -159,6 +159,7 @@ public class BeanOutputConverter<T> implements StructuredOutputConverter<T> {
 	@Override
 	public T convert(@NonNull String text) {
 		try {
+			text = text.replaceAll("<think>[\\s\\s]*?</think>","")
 			// Remove leading and trailing whitespace
 			text = text.trim();
 


spring-ai-model/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
```diff
@@ -159,6 +159,7 @@ private void generateSchema() {
 	@Override
 	public T convert(@NonNull String text) {
 		try {
+			text = text.replaceAll("<think>[\\s\\s]*?</think>","")
```
markpollack, line 162: I think we are going to have to opt into this 'exclusion' in the output converter. I'd like to make this a pluggable stragtegy, who knows how many ways the thinking response will be done.  is this from anthropic or deepseek?.  we have to overhaul the whole code base to handle 'thinking' responses in the assistant messgae, but hacks like this are useful.

also missing semicolon

YunKuiLu: Should it be `\s` and `\S` here? 
```suggestion
			text = text.replaceAll("<think>[\\s\\S]*?</think>","");
```
Although this regex is simple, it's better to have unit tests.

nicolaskrier: I have implemented something similar for **qwen3:14b** model running locally on my Ollama. The solution is a bit more complex but it works properly until now.

```
@Override
public T convert(@NonNull String text) {
    var thinkingPattern = Pattern.compile("^<think>[\\s\\d\\p{L}\\p{P}]*</think>", Pattern.MULTILINE);
    var thinkingMatcher = thinkingPattern.matcher(text);
    var cleanedText = thinkingMatcher.replaceFirst("");

    return super.convert(cleanedText);
}
```

nicolaskrier:
Hi @ilayaperumalg, I believe this PR could be closed since it has been superseded by #4667.

ilayaperumalg:
@nicolaskrier Thanks for following it up! sure, closing this one as resolved via #4667

PR #3969, opened 2025-08-01
Update Ollama API to latest version

ThomasVitale:
* Add missing fields to Ollama API classes
* Deprecate Ollama options that are no longer supported
* Update integration tests using Ollama 0.10.1 (latest version) and more performant models.


sunyuhan1998:
Thank you for your work, it looks great! Additionally, I noticed that you've added support for "think" in `OllamaApi`. I think if this PR is accepted, my previous PR #3386 will no longer be needed.

dnwwdwd:
hello.I want to ask if this change was applied in the latest 1.0.1 version, as it seems I didn't see this change.

ThomasVitale:
@sunyuhan1998 thanks for your comment. This PR only updates the Ollama API (including the new `think` field), but it doesn't expose it yet. I see your PR provides a way to use it from the `OllamaChatModel`, which I think it'd still be beneficial to have.

@dnwwdwd this PR has not been merged yet, so it's not included in the latest 1.0.1 version.

sobychacko:
@ThomasVitale Do you think we should back-port this PR into `1.0.x`? Thanks.

ThomasVitale:
@sobychacko That would be nice to have already in 1.0.x, if it's possible. It would give us the possibility to remove the deprecated Ollama options in 1.1.x.

This PR is backward compatible as it only adds new fields in OllamaApi and deprecates options and fields that Ollama doesn't support anymore. Builders and custom constructors help ensuring the backward compatibility even if some records got new fields.

sobychacko:
@ThomasVitale Merged upstream and backported to `1.0.x` (https://github.com/spring-projects/spring-ai/issues/4037).

ThomasVitale:
@sobychacko thank you!

PR #3993, opened 2025-08-04
Fix bug in example code in documentation

huidongyin:
This PR fixes a minor bug in the example code in the documentation for ToolCallback. 
The incorrect method signature has been corrected to align with the actual interface requirements.


jiafu1115:
LGTM!

PR #4003, opened 2025-08-05
fix: When parameter "value" is json string, JsonParser should convert json string to Object directly.

bourne7:
when parameter "value" is json string:
```json
{
    "name": "foo"
}
```

current code will convert it to json string with escaped quotes:

```
{\\"name\\":\\"foo\\"}
```

following code will fail to parse it from escaped json string to object.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/util/json/JsonParser.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/util/json/JsonParser.java
@@ -168,8 +168,22 @@ public final class JsonParser {
 			return Enum.valueOf((Class<Enum>) javaType, value.toString());
 		}
 
-		String json = JsonParser.toJson(value);
-		return JsonParser.fromJson(json, javaType);
+		Object result = null;
+		if (value instanceof String jsonString) {
+			try {
+				result = JsonParser.fromJson(jsonString, javaType);
+			}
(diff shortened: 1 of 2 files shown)


sobychacko:
@bourne7 Could you add a unit test to verify these changes? in `JsonParserTests`.

bourne7:
sure

bourne7:
sy, i find that this problem has been fixed in this commit:

18ae16e6 Ahmed Maruf <81756975+ohmaruf@users.noreply.github.com> on 2025-05-27 at 16:56
fix(json): prevent double-serialization of already valid JSON strings in JsonParser.toJson

bourne7:
Additional improvements:

modification on 05-28 added isValidJson method, it actually attempts deserialization to determine if the string is JSON type. In fact, this could be optimized - change it to immediately return the object if successful, avoiding double deserialization.

In most cases, Model returns a JSON string, so this optimization can eliminate one deserialization step in the majority of scenarios.

sobychacko:
We still need a unit test to validate the changes.

bourne7:
@sobychacko unittest added ;)

YunKuiLu:
Can you add a unit test like this?
```java
String jsonString = "\"Hello world!\"";
var value = JsonParser.toTypedObject(jsonString, String.class);
```

Issue #2857, opened 2025-04-23
Incorrect Error handling when calling MCP tools


Ga-Ol-St:
Codebase: M7
**Issue Description:**
When we execute MCP Tools it's normal that tool can report errors, for example:
MCP Tool is handling renaming files, we fire this tool and trying to rename a file which don't exist, Tool will report an error.

**Current state:**
When LLM execute some MCP tool and MCP tool reports error - this will terminate execution flow with following error:

> java.lang.IllegalStateException: Error calling tool: [TextContent[audience=null, priority=null, text=file not found]]
>         at org.springframework.ai.mcp.SyncMcpToolCallback.call(SyncMcpToolCallback.java:117) ~[spring-ai-mcp-1.0.0-M7.jar:1.0.0-M7]
>         at org.springframework.ai.mcp.SyncMcpToolCallback.call(SyncMcpToolCallback.java:125) ~[spring-ai-mcp-1.0.0-M7.jar:1.0.0-M7]
>         at org.springframework.ai.model.tool.DefaultToolCallingManager.executeToolCall(DefaultToolCallingManager.java:227) ~[spring-ai-model-1.0.0-M7.jar:1.0.0-M7]
>         at org.springframework.ai.model.tool.DefaultToolCallingManager.executeToolCalls(DefaultToolCallingManager.java:139) ~[spring-ai-model-1.0.0-M7.jar:1.0.0-M7]
>         at com.example.demo.MyController.processRequest(MyController.java:234) ~[main/:na]

**Expected Behavior:**
Error should be processed by **ToolExecutionExceptionProcessor** and result should be returned to LLM

**Issue analysis:**
**DefaultToolCallingManager** should catch this kind of errors but in current code state if MCP Tool calls produce an error they will throw **IllegalStateException** but **DefaultToolCallingManager** is expecting **ToolExecutionException** in case of error in Tool execution.

**Reference points:**
https://github.com/spring-projects/spring-ai/blob/81b715b3d2ea06e0216b13bf5c7a79a31b52e4e6/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java#L204-L210

https://github.com/spring-projects/spring-ai/blob/81b715b3d2ea06e0216b13bf5c7a79a31b52e4e6/mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallback.java#L111-L120

[...]

PR #4012, opened 2025-08-06
fix: handle errors when toolExecution fails for functionToolCallback

emaarco:
## FunctionToolCallback does not wrap exceptions in ToolExecutionException

fixes issue [2857](https://github.com/spring-projects/spring-ai/issues/2857)

## Description

The `FunctionToolCallback` does not properly handle exceptions thrown by user-defined tool functions. Unlike other tool callback implementations (such as `SyncMcpToolCallback`), it does not wrap exceptions in `ToolExecutionException`, which prevents proper error handling by the framework's exception processing mechanisms.

## Current Behavior

When a tool function throws an exception, it propagates uncaught through the `FunctionToolCallback.call()` method:

```java
@Override
public String call(String toolInput, @Nullable ToolContext toolContext) {
    Assert.hasText(toolInput, "toolInput cannot be null or empty");
    logger.debug("Starting execution of tool: {}", this.toolDefinition.name());
    I request = JsonParser.fromJson(toolInput, this.toolInputType);
    O response = this.toolFunction.apply(request, toolContext); // Exception not caught
    logger.debug("Successful execution of tool: {}", this.toolDefinition.name());
    return this.toolCallResultConverter.convert(response, null);
}
```

## Expected Behavior

Exceptions should be caught and wrapped in `ToolExecutionException` to maintain consistency with other tool callback implementations and enable proper error processing.

## Proposed Solution

Wrap the tool function execution in a try-catch block, similar to the implementation in [SyncMcpToolCallback](https://github.com/spring-projects/spring-ai/blob/main/mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallback.java):

```java
try {
    response = this.toolFunction.apply(request, toolContext);
} catch (Exception e) {
    logger.error("Error executing tool: {}", this.toolDefinition.name(), e);
    throw new ToolExecutionException(this.toolDefinition, e);
}
```

## Alternatives / Contribution for other options

If this is not an appropriate way to solve this issue please let me know! :) I would be happy to also contribute more general solutions - like making the `ToolExecutionExceptionProcessor` more generic (so that it can handle all kind of exceptions - and not just ToolExecutionExceptions)

## Impact

This change would:
- Enable error processing by the framework's exception processor (`ToolExecutionExceptionProcessor`)
- Improve debugging and error reporting for FunctionToolCallback execution failures

## Example Use Case

```kotlin
@Bean
fun exampleTool(): FunctionToolCallback<Void, List<String>> {
    return FunctionToolCallback
        .builder("ExampleTool", exampleTool::performCall)
        .description("This is an example tool that returns a list of example strings.")
        .build()
}

class ExampleTool {
    fun performCall(): List<String> {
        val response = client.execute("query { exampleField }")
        if (response.hasErrors()) {
            throw RuntimeException("Error fetching example data: $response") // This should be wrapped
        } else {
            val data = response.getData("exampleField")
            return data as List<String>
        }
    }
}
```

Currently, the `RuntimeException` thrown in `performCall()` is not wrapped in a `ToolExecutionException`, making it difficult for error processors to handle tool-specific failures appropriately.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/function/FunctionToolCallback.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/function/FunctionToolCallback.java
@@ -31,6 +31,7 @@ import org.springframework.ai.tool.definition.DefaultToolDefinition;
 import org.springframework.ai.tool.definition.ToolDefinition;
 import org.springframework.ai.tool.execution.DefaultToolCallResultConverter;
 import org.springframework.ai.tool.execution.ToolCallResultConverter;
+import org.springframework.ai.tool.execution.ToolExecutionException;
 import org.springframework.ai.tool.metadata.ToolMetadata;
 import org.springframework.ai.tool.support.ToolUtils;
 import org.springframework.ai.util.json.JsonParser;
@@ -99,7 +100,15 @@ public class FunctionToolCallback<I, O> implements ToolCallback {
 		logger.debug("Starting execution of tool: {}", this.toolDefinition.name());
 
 		I request = JsonParser.fromJson(toolInput, this.toolInputType);
-		O response = this.toolFunction.apply(request, toolContext);
+		O response;
+
+		try {
+			response = this.toolFunction.apply(request, toolContext);
+		}
+		catch (Exception e) {
+			logger.error("Error executing tool: {}", this.toolDefinition.name(), e);
+			throw new ToolExecutionException(this.toolDefinition, e);
+		}
 
 		logger.debug("Successful execution of tool: {}", this.toolDefinition.name());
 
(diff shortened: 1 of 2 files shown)


YunKuiLu:
Related https://github.com/spring-projects/spring-ai/pull/3918

sobychacko:
@emaarco Just merged the other PR - https://github.com/spring-projects/spring-ai/pull/3918. Is this PR still needed since both of them address the same issue? If so, please reopen and add more context. Thanks!

emaarco:
@sobychacko - thank you. I'll have a look, but i think its solved. 
Only aspect would be that it does not behave like "MethodToolCallback". 
MethodToolCallback is only catching `InvocationTargetException`. The fix however is catching all `Exception`(s)

Moreover, i have the feeling that given the fact, that this "try-catch" logic is in multiple tool-implementations (like FunctionToolCallback, SyncMcpToolCallback & MethodToolCallback) one could think about modifying the `ToolExecutionExceptionProcessor`. 

Whats your take on that? (can also open a dedicated issue on that - in case it does not already exist)

PR #4013, opened 2025-08-06
test: Add edge case tests for `VertexAI` `Gemini` runtime hints registration

alxkm:
Adds new test cases to strengthen VertexAiGeminiRuntimeHintsTests:

- **Null ClassLoader handling** - Ensures hint registration works with null ClassLoader (common in AOT)
- **Basic reflection hints validation** - Verifies hints are actually registered 
- **Idempotent registration** - Prevents duplicate hint registration on multiple calls
- **Package validation** - Confirms JSON annotated classes are found from correct package
- **Hint count validation** - Prevents excessive hint bloat that impacts native image build time

These tests catch potential AOT/GraalVM native compilation issues that could cause runtime failures.


sobychacko:
@alxkm Thanks for all these PR's. One minor suggestion - if you are planning to add more PR's increasing the test coverage, try to batch a few tests in a single PR (and mention those in the commit message) instead of sending individual PR's per test. It will be easier for us to review/merge. Thanks!

PR #4068, opened 2025-08-07
Add GPT-5 model support and temperature documentation

markpollack:
## Summary

This PR adds support for testing GPT-5 models and documents their temperature requirements.

### Test Implementation
- Added parameterized test chatCompletionEntityWithNewModels in OpenAiApiIT.java
- Tests both gpt-5 and gpt-5-2025-08-07 models
- Fixed temperature value to 1.0 for GPT-5 compatibility

### Documentation Update
- Added NOTE section in openai-chat.adoc about GPT-5 temperature requirement
- Warns users that GPT-5 models only support temperature=1.0

## Background

GPT-5 models have a strict temperature requirement and will return an error if any value other than 1.0 is used.

## Testing

- Test passes for both GPT-5 models with temperature=1.0
- Documentation renders correctly with the new NOTE section

## Related Issues

This change addresses model compatibility issues with the new GPT-5 models and improves user documentation to prevent configuration errors.

PR #4086, opened 2025-08-08
Feat: add gpt 5 models and verbosity param

apappascs:
Add gpt-5 models and update their tests

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -482,18 +482,37 @@ public class OpenAiApi {
 		GPT_5("gpt-5"),
 
 		/**
-		 * <b>GPT-5 (2025-08-07)</b> is a specific snapshot of the GPT-5 model from August
-		 * 7, 2025, providing enhanced capabilities for complex reasoning and
-		 * problem-solving tasks.
+		 * GPT-5 mini is a faster, more cost-efficient version of GPT-5. It's great for
+		 * well-defined tasks and precise prompts.
 		 * <p>
-		 * Note: GPT-5 models require temperature=1.0 (default value). Custom temperature
-		 * values are not supported and will cause errors.
+		 * Model ID: gpt-5-mini
 		 * <p>
-		 * Model ID: gpt-5-2025-08-07
-		 * <p>
-		 * See: <a href="https://platform.openai.com/docs/models/gpt-5">gpt-5</a>
+		 * See:
+		 * <a href="https://platform.openai.com/docs/models/gpt-5-mini">gpt-5-mini</a>
 		 */
-		GPT_5_2025_08_07("gpt-5-2025-08-07"),
+		GPT_5_MINI("gpt-5-mini"),
+
+		/**
+		 * GPT-5 Nano is the fastest, cheapest version of GPT-5. It's great for
+		 * summarization and classification tasks.
+		 * <p>
+		 * Model ID: gpt-5-nano
+		 * <p>
+		 * See:
+		 * <a href="https://platform.openai.com/docs/models/gpt-5-nano">gpt-5-nano</a>
+		 */
+		GPT_5_NANO("gpt-5-nano"),
+
+		/**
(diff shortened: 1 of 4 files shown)


apappascs:
@sobychacko @markpollack updating the latest changes from https://github.com/spring-projects/spring-ai/commit/1a3ed95b4d62811a5baa9f7a27d94f5ba9ab8cc2

Gpt-5 models: https://platform.openai.com/docs/models
New params: https://cookbook.openai.com/examples/gpt-5/gpt-5_new_params_and_tools
Using Gpt-5: https://platform.openai.com/docs/guides/latest-model

amresh-das-promptpath (approved):

sunyuhan1998 (approved):

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -482,18 +482,39 @@ public enum ChatModel implements ChatModelDescription {
...
 		/**
-		 * <b>GPT-5 (2025-08-07)</b> is a specific snapshot of the GPT-5 model from August
-		 * 7, 2025, providing enhanced capabilities for complex reasoning and
-		 * problem-solving tasks.
+		 * GPT-5 Chat points to the GPT-5 snapshot currently used in ChatGPT. GPT-5 is our
+		 * next-generation, high-intelligence flagship model. It accepts both text and
+		 * image inputs, and produces text outputs.
 		 * <p>
-		 * Note: GPT-5 models require temperature=1.0 (default value). Custom temperature
-		 * values are not supported and will cause errors.
```
sobychacko, line 490: Is this not the case anymore? In that case, we need to update the docs changed in this commit: https://github.com/spring-projects/spring-ai/pull/4068/files

apappascs: thank youi for pointing out this one @sobychacko  . it is the case for all gpt 5 models but not the gpt-5-chat model . https://community.openai.com/t/temperature-in-gpt-5-models/1337133/25  
I have update documentation and added tests as well

sobychacko: Does it mean that the gpt 5 models default to a value of temperature `1.0`? If so, I think we should make that clear in the docs or javadocs. From what I hear from you, the GPT 5 models completely ignore temperature and internally default to a value of `1.0`. Is that correct?

apappascs: Yes, that is correct. Azure mentions also more params that are not supported, but I am am not sure if that info has to be on spring ai documentation as it is not a limitation of the framework but from the LLM model itself.

The following are currently unsupported with reasoning models:

temperature, top_p, presence_penalty, frequency_penalty, logprobs, top_logprobs, logit_bias, max_tokens

https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/reasoning?tabs=gpt-5%2Cpython-secure%2Cpy#not-supported

xak2000: > GPT 5 models completely ignore temperature and internally default to a value of 1.0

I just wanted to point out that it is not technically correct. In my experiments when the temperature is set to any value other than `1.0` the OpenAI API actually throws an error. I.e. it **does not** ignore the `temperature` parameter, instead it expects/forces it to always be `1.0`.

This is important difference because e.g. with the default auto-configuration the `temperature` is `0.7` (see `OpenAiChatProperties`). So, if the model is set to `gpt-5-*` and the `temperature` is not explicitly set to `1.0`, the default `0.7` is used in the API request and the `ChatClient` throws an exception.

I'm not sure how this problem should be solved on the library level. Automatically set the `temperature` to `1.0` when model is set to one of well-know model names? Probably this approach would be acceptable on `OpenAiChatProperties` level, when it creates the `options` object and sets the default `temperature`.. It could set it to `1.0` (or skip setting it if it works too) if the model is one of well-known models that does not support custom `temperature`.

PR #4104, opened 2025-08-10
chore: remove unnecessary null checks

csbiy:
This PR removes unnecessary null checks from the StTemplateRenderer class to simplify the code and avoid redundant validation.

### Changes
- Removed redundant null checks on template variables map keys in the apply method.
- Simplified validation logic by eliminating defensive checks that are unlikely to occur.


sunyuhan1998 (approved):
LGTM

Issue #4109, opened 2025-08-10
Enable passing of requestMetadata to the Bedrock Converse API.


StuartLox:
Currently, SpringAI does not expose a way to pass custom requestMetadata to the Bedrock Converse API.

AWS Bedrock allows including requestMetadata in requests so that additional contextual information can appear in Bedrock audit logs in CloudWatch.

This is particularly useful for tracing requests, where correlation IDs, user session IDs, or other operational tags are required for audit and compliance purposes.

StuartLox:
Going to close this Issue with this PR: #4110 

Users can enable clients to define requestMetadata in the following way:

```java
ChatResponse response = ChatClient.create(this.chatModel).prompt()
		.advisors(new SimpleLoggerAdvisor())
		.system(s -> s.text(this.systemTextResource)
				.param("name", "Bob")
				.param("voice", "pirate"))
		.user(u -> u.text("Tell me about 3 famous pirates from the Golden Age of Piracy and what they did")
                // Set requestParams for Bedrock Logging here:
				.metadata("requestId", "12345")
		)
		.call()
		.chatResponse();
```

PR #4110, opened 2025-08-10
[Enhancement] - Enable passing of requestMetadata to the Bedrock Converse API.

StuartLox:
## Description

Currently, SpringAI does not expose a way to pass custom requestMetadata to the Bedrock Converse API.

AWS Bedrock allows including requestMetadata in requests so that additional contextual information can appear in Bedrock audit logs in CloudWatch.

This is particularly useful for tracing requests, where correlation IDs, user session IDs, or other operational tags are required for audit and compliance purposes.

See RequestMetadata in Converse API here: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html

## Proposed Change

This change leverages the metadata field from the user prompt Object to provide a mechanism to set the requestMetadata parameter in the ConverseAPI Request.

* **Closes**: #4109

### DIFF
--- a/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockChatOptions.java
+++ b/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockChatOptions.java
@@ -130,7 +130,9 @@ public class BedrockChatOptions implements ToolCallingChatOptions {
 		this.maxTokens = maxTokens;
 	}
 
-	public Map<String, String> getRequestParameters() { return this.requestParameters; }
+	public Map<String, String> getRequestParameters() {
+		return this.requestParameters;
+	}
 
 	public void setRequestParameters(Map<String, String> requestParameters) {
 		this.requestParameters = requestParameters;
@@ -262,9 +264,8 @@ public class BedrockChatOptions implements ToolCallingChatOptions {
 	@Override
 	public int hashCode() {
 		return Objects.hash(this.model, this.frequencyPenalty, this.maxTokens, this.presencePenalty,
-				this.requestParameters, this.stopSequences, this.temperature, this.topK, this.topP,
-				this.toolCallbacks, this.toolNames, this.toolContext,
-				this.internalToolExecutionEnabled);
+				this.requestParameters, this.stopSequences, this.temperature, this.topK, this.topP, this.toolCallbacks,
+				this.toolNames, this.toolContext, this.internalToolExecutionEnabled);
 	}
 
 	public static class Builder {
--- a/models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockConverseChatClientIT.java
+++ b/models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockConverseChatClientIT.java
@@ -80,7 +80,6 @@ class BedrockConverseChatClientIT {
 		logger.info("" + response);
 		assertThat(response.getResults()).hasSize(1);
 		assertThat(response.getResults().get(0).getOutput().getText()).contains("Blackbeard");
-		assertThat(response.getResults().get(0).getMetadata().containsKey("requestId"));
 	}
 
 	@Test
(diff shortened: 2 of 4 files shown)


models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockChatOptions.java
```diff
@@ -53,6 +53,9 @@ public class BedrockChatOptions implements ToolCallingChatOptions {
 	@JsonProperty("presencePenalty")
 	private Double presencePenalty;
 
+	@JsonIgnore
+	private Map<String, String> requestParameters = new HashMap<>();
```
sunyuhan1998, line 57: Initializing the `Map` object at the time of declaration here seems potentially redundant, as it will be completely overwritten by parameters in either the set method or the Builder, meaning the initial value is never actually used.

StuartLox: The requestParameters may not always be passed. This can cause a null pointer exception if we don't initialise the value here I believe.

Lets suppose we run one of the integration tests against AWS (validateCallResponseMetadata):

### Fails - When requestParams not initialised:

image

### Passes - When requestParams set:
image

StuartLox:
@sunyuhan1998 just checking do you know how to get a review from the maintainers?

StuartLox:
@sobychacko any feedback on this one?

sunyuhan1998:
> @sunyuhan1998 just checking do you know how to get a review from the maintainers?

I suppose they'll get to it once they see this PR—after all, the team might be busy with other tasks at times.

models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockConverseChatClientIT.java
```diff
@@ -70,7 +70,9 @@ void call() {
 				.system(s -> s.text(this.systemTextResource)
 						.param("name", "Bob")
 						.param("voice", "pirate"))
-				.user("Tell me about 3 famous pirates from the Golden Age of Piracy and what they did")
+				.user(u -> u.text("Tell me about 3 famous pirates from the Golden Age of Piracy and what they did")
+						.param("requestId", "1234")
```
ilayaperumalg: should be `.metadata` to verify the request metadata.

ilayaperumalg: It would be nice to verify the ConverseRequest updated with metadata. Also, Is there a way to verify the response based on the request metadata being sent?

ilayaperumalg:
@StuartLox Thanks for the PR! The changes look good; Please check if you can improve the Integration test coverage. Thanks!

StuartLox:
@ilayaperumalg have incorporated your suggestions, by updating the IT to validate the expected metadata is in the response output.

models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockConverseChatClientIT.java
```diff
@@ -70,14 +70,17 @@ void call() {
...
+						.metadata("requestId", "12345")
+				)
 				.call()
 				.chatResponse();
 		// @formatter:on
 
 		logger.info("" + response);
 		assertThat(response.getResults()).hasSize(1);
 		assertThat(response.getResults().get(0).getOutput().getText()).contains("Blackbeard");
+		assertThat(response.getResults().get(0).getMetadata().containsKey("requestId"));
```
ilayaperumalg: This isn't complete yet. Could you check?

StuartLox:
@ilayaperumalg the problem is that the requestMetadata is not returned in the response, so its not possible to assert any details related to the requestMetadata in the Response unfortunatly:

See the Response details here:
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html#API_runtime_Converse_ResponseElements

ilayaperumalg:
@StuartLox ok, thanks for clarifying. I manually verified that request metadata is being passed as part of Converse client request. Will merge this PR soon.

ilayaperumalg:
@StuartLox Thanks once again for the changes. Rebased, squashed and merged as 45baa39f7

PR #4112, opened 2025-08-11
GH-4089 Fixed the issue where sensitive words in chat memory repeatedly triggered SafeGuardAdvisor interception

### LINKED ISSUE #4089: The default order of SafeGuardAdvisor should be higher than MessageChatMemoryAdvisor
**Bug description**
The default order of SafeGuardAdvisor should be higher than MessageChatMemoryAdvisor, if we configure SafeGuardAdvisor and MessageChatMemoryAdvisor with default order at the same time, and then we trigger the SafeGuardAdvisor with some sensitive words, SafeGuardAdvisor will be repeatedly triggered incorrectly , even if there is no sensitive word, because of the chat memory from MessageChatMemoryAdvisor contains sensitive word

sunyuhan1998:
As described in the issue, the current default order of `SafeGuardAdvisor` is lower than that of `MessageChatMemoryAdvisor`, which causes a problem: when users employ both advisors simultaneously without manually adjusting their `order`, if any historical message contains sensitive content, all subsequent messages will be blocked by `SafeGuardAdvisor`. This behavior is clearly undesirable. 

This PR addresses the issue with the following changes:  

1. Adjusts the default priority of `SafeGuardAdvisor` to be higher than that of `MessageChatMemoryAdvisor`, ensuring that content filtering occurs before chat memory is stored.  
2. Enhances the documentation for `SafeGuardAdvisor` to alert users about this potential issue and recommend appropriate priority configuration based on their specific use cases.
3. Add unit tests for the `SafeGuardAdvisor`.

Fixes: #4089


sobychacko:
@sunyuhan1998 It might be beneficial to verify the changes via a test. What do you think?

sunyuhan1998:
> @sunyuhan1998 It might be beneficial to verify the changes via a test. What do you think?

No problem, I'll add the tests soon.

sunyuhan1998:
Hi @sobychacko , I noticed that `SafeGuardAdvisor` currently lacks unit tests, so I've added a comprehensive test class for it, which also includes the changes from this PR. Could you please review it again when you have a chance? Thanks!

laurentperez:
Hi. Stumbled upon the same issue.

Could you kindly rebase and resolve conflicts ? It has tests now since @dimitarproynov https://github.com/spring-projects/spring-ai/pull/6613/changes

From my point of view your `Advisor.DEFAULT_CHAT_MEMORY_PRECEDENCE_ORDER - 1;` is correct with 2 minor remarks : 

- It will now order before the zero of https://github.com/spring-projects/spring-ai/blob/main/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/SimpleLoggerAdvisor.java#L130 so this will not be logged.

- I'm not sure if the now negative order should be documented as a breaking change or not : people **may** have desired to persist the guarded prompt and its assistant fallback response for observability purposes. I agree with your "This behavior is clearly undesirable", yet, people may have desired to keep it.

PR #4116, opened 2025-08-11
Fix NPE when metadata is null in MilvusVectorStore

### LINKED ISSUE #4115: NullPointerException when metadata is null during distance injection in Document mapping
**Bug description**
When mapping similarity search results to Document, the code injects a distance field into a JsonObject read from the row’s metadata. If rowRecord.get(metadataFieldName) returns null, the subsequent call to metadata.addProperty(...) throws an NPE. The code currently catches only ParamException, so a NullPointerException escapes the try/catch.

Stack trace (excerpt):
``` java
java.lang.NullPointerException: Cannot invoke "com.google.gson.JsonObject.addProperty(String, java.lang.Number)" because "metadata" is null
    at org.springframework.ai.vectorstore.milvus.MilvusVectorStore.lambda$similaritySearch$2(...)
    at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:...)
    at org.springframework.ai.vectorstore...ObservationAwareMilvusVectorStore.lambda$similaritySearch$7(...)
    at ai.shb.shbai.service.DocumentService.searchSimilarDocuments(DocumentService.java:397)
    at ai.shb.shbai.controller.DocumentController.searchDocuments(DocumentController.java:36)

```

**Environment**
Spring AI version: 1.0.0
Java version: 17.0.14
Vector store: MilvusVectorStore
Gson version: 2.13.1
Module / file / class: org.springframework.ai.vectorstore.milvus.MilvusVectorStore

**Steps to reproduce**
1. Ensure a search result row where the metadata field exists but its value is null (common with legacy/custom data).

2. Run similarity search and map rows to Document with code like:
``` java
JsonObject metadata = (JsonObject) rowRecord.get(this.metadataFieldName); // may be null
metadata.addProperty(DocumentMetadata.DISTANCE.value(), 1 - getResultSimilarity(rowRecord)); // NPE here

```

**Expected behavior**
Missing/null metadata should be handled gracefully (use an empty metadata object/map as fallback).
distance is always injected.

Self-contained snippet that reproduces the failure without external deps:
``` java
import com.google.gson.*;
import com.google.gson.reflect.TypeToken;
import java.lang.reflect.Type;
import java.util.Map;

public class MinimalRepo {

    static Map<String, Object> mapRecord(Object rawMetadata, double similarity) {
        JsonObject metadata = (JsonObject) rawMetadata; // may be null
        // This line throws when metadata == null
        metadata.addProperty("distance", 1.0 - similarity);
        Type t = new TypeToken<Map<String, Object>>(){}.getType();
        return new Gson().fromJson(metadata, t);
    }

        System.out.println("Testing with null metadata...");
        try {
            mapRecord(null, 0.8); // -> NullPointerException
        } catch (NullPointerException e) {
            System.out.println("✓ NullPointerException caught: " + e.getMessage());
            e.printStackTrace();
        }
    }
}
```

little-huang:
### Title
Guard against NPE when metadata is null during distance injection

### Description
Adds a null check before injecting the distance property into metadata to prevent a NullPointerException if metadata is missing from the row record. The change is intentionally minimal and does not alter existing behavior beyond avoiding the crash.

### Motivation / Background
Some search results do not carry metadata (or the metadata value is null). Calling JsonObject.addProperty(...) on a null reference throws an NPE and breaks the mapping pipeline.

### Changes
Add a null guard around the distance injection step to ensure we do not call addProperty(...) on a null metadata.

### Behavior
No functional changes other than preventing the NPE when metadata is null.

### Related issues
Fixes #4115

### DIFF
--- a/vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/milvus/MilvusVectorStore.java
+++ b/vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/milvus/MilvusVectorStore.java
@@ -379,8 +379,10 @@ public class MilvusVectorStore extends AbstractObservationVectorStore implements
 				JsonObject metadata = new JsonObject();
 				try {
 					metadata = (JsonObject) rowRecord.get(this.metadataFieldName);
-					// inject the distance into the metadata.
-					metadata.addProperty(DocumentMetadata.DISTANCE.value(), 1 - getResultSimilarity(rowRecord));
+					if (metadata != null) {
+						// inject the distance into the metadata.
+						metadata.addProperty(DocumentMetadata.DISTANCE.value(), 1 - getResultSimilarity(rowRecord));
+					}
 				}
 				catch (ParamException e) {
 					// skip the ParamException if metadata doesn't exist for the custom


sobychacko:
Please address the DCO commit signing. https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring

little-huang:
> Please address the DCO commit signing. https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring

Thanks, @sobychacko  — signed off and force-pushed. DCO is green now.

PR #4131, opened 2025-08-13
Fix: Make metadata mutable in AzureVectorStore (Azure AI Search)

### LINKED ISSUE #4117: AzureVectorStore throws UnsupportedOperationException when performing similarity search due to immutable metadata map
## Bug Description

The `AzureVectorStore` throws an `UnsupportedOperationException` when performing similarity searches. The error occurs because the code attempts to modify an immutable metadata map during result processing.

## Environment
- **Spring AI version**: 1.0.1
- **Java version**: 21

## Steps to Reproduce

1. Configure AzureVectorStore with Azure AI Search backend
2. Index documents using Azure AI Search indexer with skillset (document chunking and embedding generation)
3. Perform a similarity search using Spring AI's VectorStore interface:
   ```java
   List<Document> results = vectorStore.similaritySearch("your query here");

jinlee1703:
## Summary

When running a similarity search against Azure AI Search, `AzureVectorStore` injects the computed distance into the document metadata. For results with no metadata, the previous code used `Map.of()` (immutable), so `metadata.put(...)` could throw an `UnsupportedOperationException`.

This PR ensures the metadata map is always mutable before enrichment.

## Root Cause

```java
Map<String, Object> metadata = (StringUtils.hasText(entry.metadata()))
        ? JSONObject.parseObject(entry.metadata(), new TypeReference<>() {})
        : Map.of(); // ← immutable
metadata.put(DocumentMetadata.DISTANCE.value(), 1.0 - result.getScore()); // throws
```

## What’s changed

- Added `parseMetadataToMutable(@Nullable String)` helper in `AzureVectorStore`
  - Parses JSON into `Map<String,Object>` and always returns a new `LinkedHashMap<>`
  - Falls back to empty `LinkedHashMap` when input is blank or parsing returns `null`
  - Logs a warning if parsing fails but does not fail the query path
- Replaced inline parsing in `doSimilaritySearch(...)` with the helper and then appends the `distance` field

## Behavior

- No API change.
- Similarity results always contain a mutable metadata map, so distance injection no longer fails.
- If metadata JSON is malformed, results are returned with an empty metadata map (plus a warning log), rather than failing the query.

## Tests

- Add `AzureVectorStoreMetadataTests` to verify the helper behavior:
  - blank / null / whitespace → returns an empty, mutable map
  - valid JSON → returns a mutable copy and allows enrichment (e.g., `distance`)

## Module

- `vector-stores/spring-ai-azure-store`

## Related

Fixes #4117

### DIFF
--- a/vector-stores/spring-ai-azure-store/src/test/java/org/springframework/ai/vectorstore/azure/AzureVectorStoreMetadataTests.java
+++ b/vector-stores/spring-ai-azure-store/src/test/java/org/springframework/ai/vectorstore/azure/AzureVectorStoreMetadataTests.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.springframework.ai.vectorstore.azure;
 
 import static org.assertj.core.api.Assertions.assertThat;


vector-stores/spring-ai-azure-store/src/main/java/org/springframework/ai/vectorstore/azure/AzureVectorStore.java
```diff
@@ -325,6 +323,21 @@ public <T> Optional<T> getNativeClient() {
 		return Optional.of(client);
 	}
 
+	static Map<String, Object> parseMetadataToMutable(@Nullable String metadataJson) {
+		if (!StringUtils.hasText(metadataJson)) {
+			return new LinkedHashMap<>();
```
dev-jonghoonpark: Is there a particular reason you used a `LinkedHashMap`?
It seems like a regular `HashMap` would work just as well.

dev-jonghoonpark:
At the top of the class, please add your name along with the `@author` tag.

vector-stores/spring-ai-azure-store/src/main/java/org/springframework/ai/vectorstore/azure/AzureVectorStore.java
jinlee1703: Thanks for the question! I used `LinkedHashMap` to preserve iteration order so that the parsed JSON key order is kept and the injected `distance` appears last. This gives deterministic logging/serialization and makes debugging easier. There’s no functional dependency on ordering though—if the project prefers `HashMap`, I’m happy to switch.

dev-jonghoonpark: If it doesn’t affect the outcome, there may not be a significant difference, but since HashMap is more efficient, it seems better to change it to a HashMap.

jinlee1703: Great point—updated to use `HashMap` since ordering isn’t required. The helper now returns a new `HashMap<>` and tests still pass. PTAL. ✅ (4b44359)

vector-stores/spring-ai-azure-store/src/test/java/org/springframework/ai/vectorstore/azure/AzureVectorStoreMetadataTests.java
```diff
@@ -0,0 +1,39 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit tests for {@link AzureVectorStore#parseMetadataToMutable(String)}.
+ *
+ * @author Your Name
```
dev-jonghoonpark: Please write your name here

dev-jonghoonpark:
Please add copyright to top of `AzureVectorStoreMetadataTests.java`

vector-stores/spring-ai-azure-store/src/test/java/org/springframework/ai/vectorstore/azure/AzureVectorStoreMetadataTests.java
```diff
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2023-2024 the original author or authors.
```
dev-jonghoonpark: chore: 2024 → 2025

PR #4153, opened 2025-08-14
fix: correctly handle merges when 'current' is null

garethjevans:


sobychacko:
@garethjevans Could you provide a bit more detail on this change? Is there a corresponding GH issue? If not, can you be a bit more elaborate on the commit message for the motivations behind this change? Thanks!

PR #4166, opened 2025-08-15
Issue 4130 similaritySearch throwing Exception due to schema name fix

### LINKED ISSUE #4130: similaritySearch throwing Exception due to schema name

**Bug description**
When doing a `similaritySearch` using the `VectorStore` on a database where the schema name contains a `-` i run into an exception.

```
PreparedStatementCallback; bad SQL grammar [SELECT *, embedding <=> ? AS distance FROM demo-1998.vector_store WHERE embedding <=> ? < ?  AND metadata::jsonb @@ '($.entityType == "PROCESS" ...)'::jsonpath  ORDER BY distance LIMIT ? ]
```

I'm pretty sure that it is related to the schema name here. A possible fix could be wrapping the schema name in `"` on SpringAI side.

**Environment**
Dependencies using Spring Boot 3.5.4 and Java 20

```
implementation("org.springframework.ai:spring-ai-starter-mcp-server-webmvc")
implementation("org.springframework.ai:spring-ai-starter-model-azure-openai")
implementation("org.springframework.ai:spring-ai-starter-vector-store-pgvector")
```

**Steps to reproduce**
- Do a `similaritySearch` call on a database schema which has a `-` in its name

**Expected behavior**
No exceptions are thrown

-

wilocu:
#4130

Problem
  When using PgVectorStore with schema names containing hyphens (e.g., demo-1998), similaritySearch() operations would throw SQL grammar exceptions due to
  unquoted PostgreSQL identifiers in generated SQL.

  Solution
  - Added proper PostgreSQL identifier quoting in getFullyQualifiedTableName() to generate "schema"."table" format
  - Updated schema creation DDL to use quoted schema names
  - Enhanced schema validator to accept hyphenated names while maintaining security against SQL injection
  - Updated error messages to show proper quoted syntax in help text

  Changes
  - PgVectorStore.java: Fixed SQL generation to use quoted identifiers
  - PgVectorSchemaValidator.java: Updated validation logic and error messages
  - PgVectorStoreSchemaQuotingTest.java: Added test coverage for hyphenated schema names

  Example

  Before (❌ SQL error):
  SELECT * FROM demo-1998.vector_store  -- Syntax error

  After (✅ Valid PostgreSQL):
  SELECT * FROM "demo-1998"."vector_store"  -- Works correctly

  Testing

  - All existing tests pass
  - New test verifies proper quoting for hyphenated schema names
  - Verified SQL generation produces valid PostgreSQL syntax
  
Signed-off-by: Mattia Pasetto [matpat17@gmail.com](mailto:matpat17@gmail.com)

I am still pretty new to contributions, please double check and let me know of any errors.

### DIFF
--- a/vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/vectorstore/pgvector/PgVectorSchemaValidator.java
+++ b/vector-stores/spring-ai-pgvector-store/src/main/java/org/springframework/ai/vectorstore/pgvector/PgVectorSchemaValidator.java
@@ -49,14 +49,28 @@ public class PgVectorSchemaValidator {
 			return false;
 		}
 
-		// Check if the table or schema has Only alphanumeric characters and underscores
-		// and should be less than 64 characters
-		if (!name.matches("^[a-zA-Z0-9_]{1,64}$")) {
+		// Check basic constraints: non-empty and reasonable length
+		if (name.trim().isEmpty() || name.length() > 64) {
 			return false;
 		}
 
 		// Check to ensure the table or schema name is not purely numeric
 		if (name.matches("^[0-9]+$")) {
+			return false;
+		}
+
+		// Reject names with potentially dangerous characters that could indicate SQL
+		// injection
+		// These would be problematic even when quoted
+		if (name.contains(";") || // Statement separator
+				name.contains("--") || // SQL comment
+				name.contains("/*") || // Block comment start
+				name.contains("*/") || // Block comment end
+				name.contains("\0") || // Null byte
+				name.contains("\n") || // Newline
+				name.contains("\r") || // Carriage return
+				name.toLowerCase().matches(".*\\b(drop|alter|create|insert|update|delete|select)\\b.*")) { // SQL
+																											// keywords
 			return false;
 		}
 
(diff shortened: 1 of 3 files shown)


markpollack:
Thanks! There is certainly a bug to fix.

I am not an expert on these database best practices, so I turned to some investigation on chatgpt.   

Here is the reply.  would like to know your thoughts @wilocu 

This is a real bug, and the core fix (always quoting identifiers) is right. The piece to adjust is the **validator**. 

---

Quoting identifiers in `getFullyQualifiedTableName()` and DDL is the correct fix for the hyphenated-schema bug. A couple of tweaks will align this with PostgreSQL rules and avoid false rejections:

1. Identifier rules

* PostgreSQL limits identifiers to **63 bytes (UTF-8)**, not 64 chars.
* Numeric names, reserved words, and symbols (e.g., `-`, `.`) are **legal when quoted**.
* Double quotes inside names must be escaped as `""`.

2. Validator

* Drop bans on numeric-only, keywords, `;`, `--`, etc. If we always quote+escape, those are safe.
* Keep: non-null, non-empty, ≤63 **bytes**, and no NUL/control chars.
* Existence check: use `SELECT to_regclass(quote_ident($1)||'.'||quote_ident($2)) IS NOT NULL` to respect case/quoting instead of `information_schema`.

3. Quoting helper
   Ensure the FQN builder escapes quotes:

```java
private static String q(String s) { return "\"" + s.replace("\"", "\"\"") + "\""; }
// usage: q(schema) + "." + q(table)
```

4. Tests
   Add cases for `"demo-1998"`, `"123"`, `"select"`, `a"b`, non-ASCII, and 63-byte boundary.

With those adjustments, the quoting fix resolves the bug without over-restricting valid PostgreSQL identifiers.

PR #4179, opened 2025-08-18
refactor: restructure MCP server auto-configuraitons, adding streamable-http support

tzolov:
- Extract MCP server functionality into specialized modules:
  * STDIO/SSE servers (existing functionality)
  * Streamable-HTTP servers with change notifications
  * Stateless servers for simplified deployments
- Add new auto-configuration modules:
  * spring-ai-autoconfigure-mcp-streamable-server-common
  * spring-ai-autoconfigure-mcp-streamable-server-webflux/webmvc
  * spring-ai-autoconfigure-mcp-stateless-server-common
  * spring-ai-autoconfigure-mcp-stateless-server-webflux/webmvc
  * Add server-specific property prefixes for different server types
- Extract ToolCallbackConverter into separate auto-configuration with conditional enablement via tool-callback-converter property
- Enhance transport providers with builder patterns and new features:
  * Keep-alive interval support for connection health
  * Add keep-alive-interval and other transport-specific options
- Add comprehensive integration tests for all server types
- Update documentation to reflect new architecture and server options

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server/src/main/java/org/springframework/ai/mcp/server/autoconfigure/ToolCallbackConverterAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server/src/main/java/org/springframework/ai/mcp/server/autoconfigure/ToolCallbackConverterAutoConfiguration.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2025-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.mcp.server.autoconfigure;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.springframework.ai.mcp.McpToolUtils;
+import org.springframework.ai.tool.ToolCallback;
+import org.springframework.ai.tool.ToolCallbackProvider;
+import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
+import org.springframework.core.log.LogAccessor;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.MimeType;
+
+import io.modelcontextprotocol.server.McpServerFeatures;
+
+/**
+ * @author Christian Tzolov
+ */
+@EnableConfigurationProperties(McpServerProperties.class)
+@Conditional(ToolCallbackConverterCondition.class)
+public class ToolCallbackConverterAutoConfiguration {
+
+	private static final LogAccessor logger = new LogAccessor(ToolCallbackConverterAutoConfiguration.class);
+
+	@Bean
+	@ConditionalOnProperty(prefix = McpServerProperties.CONFIG_PREFIX, name = "type", havingValue = "SYNC",
+			matchIfMissing = true)
+	public List<McpServerFeatures.SyncToolSpecification> syncTools(ObjectProvider<List<ToolCallback>> toolCalls,
+			List<ToolCallback> toolCallbacksList, List<ToolCallbackProvider> toolCallbackProvider,
+			McpServerProperties serverProperties) {
+
+		List<ToolCallback> tools = new ArrayList<>(toolCalls.stream().flatMap(List::stream).toList());
+
+		if (!CollectionUtils.isEmpty(toolCallbacksList)) {
+			tools.addAll(toolCallbacksList);
+		}
+
+		List<ToolCallback> providerToolCallbacks = toolCallbackProvider.stream()
+			.map(pr -> List.of(pr.getToolCallbacks()))
+			.flatMap(List::stream)
+			.filter(fc -> fc instanceof ToolCallback)
+			.map(fc -> (ToolCallback) fc)
+			.toList();
+
+		tools.addAll(providerToolCallbacks);
+
+		return this.toSyncToolSpecifications(tools, serverProperties);
+	}
+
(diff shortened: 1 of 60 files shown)


markpollack:
This looks great.  I did some similar things for the IT testing in the integration tests for spring-ai-examples.  i think there are some reusable patterns we can discuss.

auto-configurations/mcp/spring-ai-autoconfigure-mcp-server/src/main/java/org/springframework/ai/mcp/server/autoconfigure/ToolCallbackConverterAutoConfiguration.java
```diff
@@ -0,0 +1,139 @@
...
+
+		return this.toAsyncToolSpecification(tools, serverProperties);
+	}
+
+	private List<McpServerFeatures.AsyncToolSpecification> toAsyncToolSpecification(List<ToolCallback> tools,
+			McpServerProperties serverProperties) {
+		// De-duplicate tools by their name, keeping the first occurrence of each tool
+		// name
+		return tools.stream() // Key: tool name
+			.collect(Collectors.toMap(tool -> tool.getToolDefinition().name(), tool -> tool, // Value:
```
markpollack, line 94: this logic is duplicated in the sync version, can share a utility method

auto-configurations/mcp/spring-ai-autoconfigure-mcp-server/src/test/java/org/springframework/ai/mcp/server/autoconfigure/SseWebClientAndWebFluxServerIT.java
```diff
@@ -0,0 +1,502 @@
...
+
+	private final ApplicationContextRunner serverContextRunner = new ApplicationContextRunner()
+		.withConfiguration(AutoConfigurations.of(McpServerAutoConfiguration.class,
+				ToolCallbackConverterAutoConfiguration.class, McpWebFluxServerAutoConfiguration.class));
+
+	private final ApplicationContextRunner clientApplicationContext = new ApplicationContextRunner()
+		.withConfiguration(AutoConfigurations.of(McpToolCallbackAutoConfiguration.class,
+				McpClientAutoConfiguration.class, SseWebFluxTransportAutoConfiguration.class));
+
+	static AtomicReference<LoggingMessageNotification> loggingNotificationRef = new AtomicReference<>();
```
markpollack: `static` can introduce side effects, make them instance vars and reset state in @BeforeEach.

tzolov: As those are shared between the test and the spring configuration beens we can't use the beforeeach but we can make them a special TestContext been as part for the client test configuration. 
I will fix it for the Sse and the Streamable its

tzolov:
Thanks @markpollack I've addressed the above recomendations

Fottas:
Hi [tzolov](https://github.com/tzolov),

I noticed that PR #4179 was merged recently to add streamable support for MCP server.
A bit of context: I had submitted PR #4053 over two weeks ago with a similar goal but a different approach, and there hasn’t been much follow-up on it yet. To be honest, I was a bit surprised and confused to see a parallel solution merged so quickly.

Looking at the current implementation, I do have some concerns:

Module proliferation – introducing 10+ modules mainly for configuration variations feels like overhead.

Maintainability – each new mode could add more modules in the future, which might lead to unnecessary growth.

Complexity – this design seems to increase cognitive load for both users and maintainers.

In my PR, I tried a different approach:

Using configuration properties to switch between modes

Keeping everything in a single autoconfigure module with conditional beans

Preserving backward compatibility while reducing module sprawl

I think this configuration-driven approach might be more sustainable in the long run.
Would it make sense to revisit the design direction here? I’d be happy to help align and refactor so that we can keep things both flexible and maintainable.

Thanks!

tzolov:
rebased and merged at 0b96f4cde

Fottas:
Hi @tzolov,
I noticed this creates 6+ autoconfiguration modules and multiple Properties classes with duplicate code. This goes against Spring Boot's "convention over configuration" principle.
In my PR #4063, I used a single starter with configuration properties to switch between modes:
`
spring.ai.mcp.server.type=stateless|streamable|sse
`
This approach would reduce module count significantly while maintaining the same functionality. Would you consider consolidating these modules?
Happy to help with a refactor if you're interested.

tzolov:
Hi @Fottas,

Thank you for reaching out, I missed your PR while on PTO and rushed to provide auto-configurations for the MCP 0.11.x (i've released recently). 
#4179 is our first attempt, but I'm working, already, on a second iteration due to issues with user-facing configurations and code repetition. I'll submit a PR shortly.

I appreciate your points but disagree about module count and complexity. Better separation of concerns actually reduces complexity.

Unlike MCP client side (where one host may have multiple clients for different server types), server side always has a single server type. 
Bundling unrelated dependencies into one jar and activating the right one via runtime properties/conditions increases complexity compared to simply replacing the starter dependency.

We already have 3 separate boot starters: spring-ai-starter-mcp-server (stdio), spring-ai-starter-mcp-server-webmvc (sse/webmvc), spring-ai-starter-mcp-server-webflux (sse/webflux) 
Having these share a single auto-configuration was a design mistake that created a monolithic configuration we're now splitting. Boot starters understand dependencies, not properties.

Reducing modules isn't the goal - reducing overall complexity and simplifying user-facing configuration is.

Let me submit the new PR and we can continue the conversation there. I'll make you co-author due to your #4053 work and improvement suggestions.

P.S. #4053 lacks stateless server support. These are different beasts with their own spec and transport APIs. Spring AI doesn't support HttpServletStreamableServerTransportProvider - use WebMvc instead. Also your PR lacked any tests

tzolov:
@Fottas here is the new PR: https://github.com/spring-projects/spring-ai/pull/4196

Fottas:
Hi @tzolov,
Thank you for the detailed explanation and for addressing the code duplication issues! The shared properties module approach makes a lot of sense given your points about separation of concerns.
Thank you for including me as co-author - I really appreciate that recognition!

tzolov:
Here's a streamlined version:

Hi @Fottas,

After further exploration, I've decided to keep boot starter choices simple: webmvc vs. webflux. As you pointed out, having separate starters for sse, streamable, and stateless would be involving for the end users.

PR #4211 consolidates everything into 3 boot starters:
- `spring-ai-starter-mcp-server` (stdio)
- `spring-ai-starter-mcp-server-webmvc` (sse/webmvc)  
- `spring-ai-starter-mcp-server-webflux` (sse/webflux)

It adds `spring.ai.mcp.server.protocol=(SSE|STREAMABLE|STATELESS)` as you suggested (`spring.ai.mcp.server.type` was already taken).

However I've managed to refactor the auto-configuraiton modules to 3 common, webmvc and webflux (similar to the client) and so resolve the existing issue of bundling mvc and flux in single auto-conf.  

You're co-author on the new PR. Thanks again for your contribution.

PR #4183, opened 2025-08-19
Fix IllegalAccessException in MongoDbAtlasLocalContainerConnectionDetails.getSslBundle()

markpollack:
## Summary

Fixes the `IllegalAccessException` introduced by the StackOverflowError fix in commit 03d475e by replacing complex MethodHandles reflection with simple interface delegation.

## Problem

The original fix used `MethodHandles.unreflectSpecial()` incorrectly for interface default methods, causing:
```
IllegalAccessException: no private access for invokespecial: interface 
org.springframework.boot.autoconfigure.mongo.MongoConnectionDetails
```

## Root Cause

- `unreflectSpecial()` is designed for superclass method calls, not interface default methods
- The lookup context lacked the required private access for invokespecial calls
- Access privilege conflict between interface public method and parent class method

## Solution

Replaced 24 lines of complex reflection code with 3 lines of elegant interface delegation:
```java
@Override
public SslBundle getSslBundle() {
    return MongoConnectionDetails.super.getSslBundle();
}
```

## Benefits

- ✅ Eliminates IllegalAccessException by properly overriding with public access
- ✅ Avoids the original StackOverflowError by delegating to interface default method  
- ✅ Removes all reflection code, making it simpler and more maintainable
- ✅ Works across all Spring Boot versions without version-specific checks
- ✅ Resolves access privilege conflict between interface and parent class

## Test Results

- ✅ `MongoDbAtlasLocalContainerConnectionDetailsFactoryIT.addAndSearch` passes
- ✅ All testcontainer module tests pass
- ✅ No regressions introduced

## Fixes

- CI Failure: https://github.com/spring-projects/spring-ai/actions/runs/16981113833
- Original StackOverflowError issue from commit 03d475e

### DIFF
--- a/spring-ai-spring-boot-testcontainers/src/main/java/org/springframework/ai/testcontainers/service/connection/mongo/MongoDbAtlasLocalContainerConnectionDetailsFactory.java
+++ b/spring-ai-spring-boot-testcontainers/src/main/java/org/springframework/ai/testcontainers/service/connection/mongo/MongoDbAtlasLocalContainerConnectionDetailsFactory.java
@@ -16,9 +16,6 @@
 
 package org.springframework.ai.testcontainers.service.connection.mongo;
 
-import java.lang.invoke.MethodHandles;
-import java.lang.reflect.Method;
-
 import com.mongodb.ConnectionString;
 import org.testcontainers.mongodb.MongoDBAtlasLocalContainer;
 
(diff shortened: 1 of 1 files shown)


markpollack:
this is on top of an old base it seems, i'm creating a new PR.

PR #2803, opened 2025-04-18
Introduce first-class chat memory support

ThomasVitale:
- ChatMemory will become a generic interface to implement different memory management strategies. It’s been moved from the “”spring-ai-client-chat” package to “spring-ai-model” package while retaining the same package, so it’s transparent to users.
- A MessageWindowChatMemory has been introduced to provide support for a chat memory that keeps at most N messages in the memory.
- A MessageWindowProcessingPolicy API has been introduced to customise the processing policy for the message window. A default implementation is provided out-of-the-box.
- A ChatMemoryRepository interface has been introduced to support different storage strategies for the chat memory. It’s meant to be used as part of a ChatMemory implementation. This is different than before, where the storage-specific implementation was directly tied to the ChatMemory. This design is familiar to Spring users since it’s used already in the ecosystem. The goal was to use a programming model similar to Spring Session and Spring Data.
- The JdbcChatMemory has been supersed by JdbcChatMemoryRepository.
- The ChatClient now supports memory as a first-class citizen, superseding the need for an Advisor to manage the chat memory. It also simplifies providing a conversationId. This feature lays the foundation for including the intermediate messages in tool calling in the memory as well.
- All the changes introduced in this PR are backword-compatible.

Usage:

```java
ChatClient chatClient = ChatClient.builder(this.chatModel)
	.defaultMemory(MessageWindowChatMemory.builder().build())
	.build();

String conversationId = "007";

ChatResponse response1 = chatClient.prompt("My name is Bond. James Bond.")
	.conversationId(conversationId)
	.call()
	.chatResponse(); // Hello, James Bond!

ChatResponse response2 = chatClient.prompt("What is my name?")
	.conversationId(conversationId)
	.call()
	.chatResponse(); // Your name is James Bond!
```

PR #4189, opened 2025-08-20
GH-4170 Fixed the issue where `SystemMessage` was not being saved to the chat memory

### LINKED ISSUE #4170: Chat memory advisor causes message ordering problem with default system message prompt
**Bug description**

Using `MessageChatMemoryAdvisor` with a default system prompt causes "conversation roles must alternate user/assistant/user/assistant" error. This appears to be the same issue as reported in #2216 which was marked as fixed in `1.0.0-M8` but can be reproduced in `1.0.1`.

**Environment**

  - Spring AI version: 1.0.1
  - Spring Boot version: 3.5.4
  - Java version: 21.0.2
  - AI Provider: OpenAI (Docker Model Runner with `ai/mistral:latest` model for testing)
  - Dependencies: spring-ai-starter-model-chat-memory, spring-ai-starter-model-openai

**Steps to reproduce**

  1. Create a `ChatClient` with `defaultSystem()` prompt and `MessageChatMemoryAdvisor` configured
  2. Make a chat call with a conversation id
  3. Attempt a second chat call with the same conversation id
  4. The second call fails with error "After the optional system message, conversation roles must alternate user/assistant/user/assistant..."

The error occurs during the second `chatClient.prompt().call().content()` call. Upon debugging the `MessageChatMemoryAdvisor#before`, I can see that messages are collected in the wrong order:

1. `UserMessage`
2. `AssistantMessage`
3. `SystemMessage`
4. `UserMessage`

**Expected behavior**

The `MessageChatMemoryAdvisor` should properly order the system prompt and conversation context to allow for continued dialogue with alternating user/assistant roles without errors.

I would expect messages to be collected in the following order instead which I believe should allow the request to succeed:

1. `SystemMessage`
2. `UserMessage`
3. `AssistantMessage`
4. `UserMessage`

https://github.com/sjohnr/spring-ai-chat-memory-bug

sunyuhan1998:
As discussed in the issue, the current implementation of `MessageChatMemoryAdvisor` does not save the `SystemMessage` into `ChatMemory`, which is clearly a BUG. 

This PR fixes the issue by checking, when adding chat memory, whether the current conversation is the first turn or whether the current `SystemMessage` differs from the previous one. If either condition is met, the `SystemMessage` will be added to the chat memory.

Fixes #4170

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
@@ -19,6 +19,8 @@ package org.springframework.ai.chat.client.advisor;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.springframework.ai.chat.messages.SystemMessage;
+import org.springframework.util.CollectionUtils;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Scheduler;
@@ -91,9 +93,17 @@ public final class MessageChatMemoryAdvisor implements BaseChatMemoryAdvisor {
 			.prompt(chatClientRequest.prompt().mutate().messages(processedMessages).build())
 			.build();
 
-		// 4. Add the new user message to the conversation memory.
+		// 4. Add user message and system message to the conversation memory: if the
+		// current conversation is the first round, or if the `systemMessage` has been
+		// updated, add both to the conversation memory.
+		SystemMessage systemMessage = chatClientRequest.prompt().getSystemMessage();
 		UserMessage userMessage = processedChatClientRequest.prompt().getUserMessage();
-		this.chatMemory.add(conversationId, userMessage);
+		if (CollectionUtils.isEmpty(memoryMessages) || !memoryMessages.contains(systemMessage)) {
+			this.chatMemory.add(conversationId, List.of(systemMessage, userMessage));
+		}
+		else {
+			this.chatMemory.add(conversationId, userMessage);
+		}
 
 		return processedChatClientRequest;
 	}


sjohnr:
@sunyuhan1998 I admit I'm new to this project, but I don't think the solution here is the correct one.

The system message is maintained in the `ChatClient` and is part of the system's configuration, not part of the chat memory. It is automatically added to the prompt of the `ChatClientRequest` and is available before the `MessageChatMemoryAdvisor` runs. The logic in `MessageChatMemoryAdvisor#before` is indeed the culprit, but I don't think adding the system message to chat memory makes sense or will help.

sunyuhan1998:
Hi @sjohnr Thanks for your review! 

I think this issue might stem from the inconsistency in design between `org.springframework.ai.chat.memory.MessageWindowChatMemory` and `org.springframework.ai.chat.client.advisor.MessageChatMemoryAdvisor`. If we consider `SystemMessage` as not part of the conversation memory, then certain aspects of `MessageWindowChatMemory`'s design appear redundant. Conversely, if `SystemMessage` should be included in the memory, then the way `MessageChatMemoryAdvisor` adds messages to memory may be flawed. @markpollack , @ThomasVitale , what are your thoughts or suggestions on this matter?

markpollack:
I'll look into it shortly and get back to folks.  I appreciate all the hard work on the thread and the PR.

ThomasVitale:
@sunyuhan1998 thanks for looking into this issue. The memory advisors are not currently designed to persist the `SystemMessage`s. Changing that would be a breaking change from a behaviour perspective and I would recommend not doing it. 

The reason why `MessageWindowChatMemory` includes logic to handle `SystemMessage`s is because it's a core API that should support more general memory handling. It's up to the "caller" to decide what to persist. That logic is not used by the built-in advisors (which were there before `MessageWindowChatMemory` was introduced), but it can potentially be used by other implementations. So it's worth fixing the issue you mentioned in https://github.com/spring-projects/spring-ai/issues/4170#issuecomment-3203987125 about the message ordering. Thank you for reporting that!

There is still a pending-design idea to introduce memory as a "first-class citizen" in `ChatClient`. As part of that work, it's good to review the requirements and have all messages sent to the memory. That's what I had in this initial draft: https://github.com/spring-projects/spring-ai/pull/2803 It might be a good time to pick this up again, together with the short-term memory/long-term memory/history discussion. @markpollack what do you think?

sunyuhan1998:
@ThomasVitale  Thank you for the detailed explanation. I now have a clear understanding of the original design intent, as well as the current state of `MessageWindowChatMemory` and `MessageChatMemoryAdvisor`. I fully agree that conversation memory should be treated as a "first-class citizen" and designed more systematically and thoroughly. Indeed, discussions around long-term and short-term memory mechanisms have come up repeatedly across multiple issues, and the current conversation memory design has indeed caused confusion for many users. As one of the core capabilities of a framework like Spring AI, effective conversation memory management is crucial. I believe it's now the right time to revisit and refine this component. My sincere appreciation goes to the team for all the hard work so far, and I will continue to follow the progress of this capability closely.

andrii-k6a:
I think there's a good reason not to save system prompts to memory. If Spring AI persisted them, you'd run into versioning issues. Imagine a user starts a conversation with one system prompt, you update it a few days later, and then they come back with a follow-up question. Which version should be used - the original or the updated one? I don't have a strong opinion, but it seems like not saving the system prompt actually has benefits here.

PR #4192, opened 2025-08-20
Update McpToolUtils.java

shishuiwuhen2009:
fix McpToolUtils.prefixedToolName formatted in order to support Chinese characters in toolName


markpollack:
Thanks!!!  had some help from ai!  let me know what you think of the final changes @shishuiwuhen2009 .  it is also backported.

shishuiwuhen2009:
Thank you so much for your detailed modification suggestions! This Unicode property-based approach for Chinese character support is much more standardized and comprehensive than my original hard-coded range method. I’ve learned a lot about robust Unicode handling—especially the nuances of Script vs. Block classification across JDK versions and how to ensure full Han character coverage. It’s really helpful for improving the quality of my code. Great work on pointing out these key details! 👍

PR #4195, opened 2025-08-20
Remove duplicated lines in AzureOpenAiChatOptions

asw12:
These lines were accidentally left as duplicates during the cleanup in c0bc623.

You can see the same logic in slightly above in L275-L277.


markpollack:
merged in c9ad1c7d9427792fe5a6ad12a6e3699de9308005 and backported to 1.0.x

thanks @asw12

PR #4139, opened 2025-08-13
feat(anthropic): add support for prompt caching with fix

Claudio-code:
Thank you for taking time to contribute this pull request!
You might have already read the [contributor guide][1], but as a reminder, please make sure to:

* Add a Signed-off-by line to each commit (`git commit -s`) per the [DCO](https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring#how-to-use-developer-certificate-of-origin)
* Rebase your changes on the latest `main` branch and squash your commits
* Add/Update unit tests as needed
* Run a build and make sure all tests pass prior to submission

For more details, please check the [contributor guide][1].
Thank you upfront!

[1]: https://github.com/spring-projects/spring-ai/blob/main/CONTRIBUTING.adoc

PR #4199, opened 2025-08-21
GH-1403: Implements Anthropic's prompt caching feature to improve tok…

sobychacko:
…en efficiency

- Adds cache control support in AnthropicApi and AnthropicChatModel
- Creates AnthropicCacheType enum with EPHEMERAL cache type
- Extends AbstractMessage and UserMessage to support cache parameters
- Updates Usage tracking to include cache-related token metrics (cacheCreationInputTokens, cacheReadInputTokens)
- Adds integration test to verify prompt caching functionality and token usage patterns
- Updates ContentBlock to support CacheControl parameter for low-level API usage
- Adds comprehensive reference documentation with usage examples and best practices

This implementation follows Anthropic's prompt caching API which allows for more efficient token usage by caching frequently used prompts.

Original implementation provided by @Claudio-code (Claudio Silva Junior) See https://github.com/spring-projects/spring-ai/pull/4139/commits/15e50263e515312c159d4176a2914f760bcce465

Fixes https://github.com/spring-projects/spring-ai/issues/1403

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -483,12 +483,25 @@ public class AnthropicChatModel implements ChatModel {
 
 	ChatCompletionRequest createRequest(Prompt prompt, boolean stream) {
 
+		// Get cache control from options
+		AnthropicChatOptions requestOptions = (AnthropicChatOptions) prompt.getOptions();
+		AnthropicApi.ChatCompletionRequest.CacheControl cacheControl = (requestOptions != null)
+				? requestOptions.getCacheControl() : null;
+
 		List<AnthropicMessage> userMessages = prompt.getInstructions()
 			.stream()
 			.filter(message -> message.getMessageType() != MessageType.SYSTEM)
 			.map(message -> {
 				if (message.getMessageType() == MessageType.USER) {
-					List<ContentBlock> contents = new ArrayList<>(List.of(new ContentBlock(message.getText())));
+					List<ContentBlock> contents = new ArrayList<>();
+
+					// Apply cache control if enabled for user messages
+					if (cacheControl != null) {
+						contents.add(new ContentBlock(message.getText(), cacheControl));
+					}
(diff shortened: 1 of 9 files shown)


markpollack:
Why add this to usermessage/abstract message?  I though a chatoption would be sufficient.    I'm concerned about having functionality in the message hiearchy that isn't universal.

adase11:
According to their request validation, Anthropic's API only supports a maximum of 4 cache blocks. I got this error with more than 4: `"A maximum of 4 blocks with cache_control may be provided. Found 5."` 

Also system messages can be cached as well by changing `System` from a `String` to a `List<ContentBlock>` and then applying the logic for cache control when building up the `system` variable for the `ChatCompletionRequest`. I'm happy to demonstrate what these changes would look like.

adase11:
@sobychacko - I added a pull request to your branch that takes care of the above issue while also making the caching using `AnthropicChatOptions` more robust. 

It allows fine-grained configuration of Anthropic Prompt Caching on outgoing chat requests.

I added:

- Configure how many prompt segments (“cache blocks”) to cache. (Anthropic only allows max of 4 right now, user could choose less if they desire)
- Control which message types are eligible for caching. (Allow for selecting based on use-case - for example my use case required SYSTEM and TOOL but not USER or ASSISTANT )
- Choose cache type per message type (e.g., EPHEMERAL, EPHEMERAL_1H). This is supported by Anthropic now, also came in handy for my use case
- Set global minimum length for caching and override it per message type. Allowing users to optimize the messages that are attempted to be cached - allowing them to attempt to optimize their usage of the max 4 cache blocks.

sobychacko:
@adase11 Thanks for the feedback. We are addressing some of this feedback on our end with @markpollack. We will update it via a separate PR. Thanks again for looking into it.

adase11:
Appreciate it @sobychacko if there's anything I can help with I'm happy to. We're eagerly awaiting being able to leverage this.

sobychacko:
@adase11 I see that we are chatting about this issue on a few different places. I am closing this PR because of the other issue that is superseding this one. https://github.com/spring-projects/spring-ai/pull/4300.

adase11:
Sounds good

PR #4204, opened 2025-08-21
Update SyncMcpToolCallback.java

shishuiwuhen2009:
Add getOriginalToolName method for AsyncMcpToolCallback in order to get easy access to the 'original' toolname before the prefix was added


ilayaperumalg:
@shishuiwuhen2009 Thanks for the PR! Could you add the same method in AsyncMcpToolCallback as well?

shishuiwuhen2009:
> @shishuiwuhen2009 Thanks for the PR! Could you add the same method in AsyncMcpToolCallback as well?

done

ilayaperumalg:
@shishuiwuhen2009 Thanks for the update. Rebased, squashed and merged as 66780351a

PR #4233, opened 2025-08-25
Add 'think' option to control reasoning models' thinking behavior

mvawork:
This update introduces the 'think' option, allowing users to dictate whether the model deliberates before responding.

Key modifications:
- Incorporated the `think` field into the OllamaOptions.
- Developed the necessary logic to process this new setting.

This addition enriches the OllamaChatModel, offering greater command over the model's operations, which will be particularly beneficial for sophisticated applications.


sunyuhan1998:
Thank you for your contribution! However, it appears this is a duplicate PR, as this improvement is already being worked on in https://github.com/spring-projects/spring-ai/pull/3386.

mvawork:
I'll wait for the feature to be implemented. For now, I'll use my workaround.

PR #4238, opened 2025-08-25
Add support for OpenAI service_tier in OpenAiChatOptions

### LINKED ISSUE #4235: Add support for OpenAI service_tier in OpenAiChatOptions

**Expected Behavior**

Users should be able to specify the service_tier parameter when making calls to OpenAI through Spring AI's OpenAiClient. This parameter should be configurable via OpenAiChatOptions.

Example usage:

```
OpenAiChatOptions options = OpenAiChatOptions.builder()
    .serviceTier("priority") // or "default", "flex", "auto"
    .build();
```

**Current Behavior**

The service_tier field already exists in the OpenAiApi class but is not exposed through OpenAiChatOptions. Currently, there's no way for users to set this parameter when making API calls to OpenAI through Spring AI.

**Context**

- Impact: Without this parameter, we cannot explicitly control the service tier for OpenAI requests, which affects:
   - Cost management (different tiers have different pricing)
   - Performance characteristics (priority vs default processing)
   - Dedicated capacity usage (for enterprise deployments)

- Use Case: We need to specify service tiers for:
   - Production workloads requiring guaranteed capacity ("priority")
   - Cost-sensitive applications ("default")
   - Dynamic scaling scenarios ("flex")
   - Project-level configuration adherence ("auto")

rafaelrddc:
Description
This MR adds support for the OpenAI service_tier parameter in Spring AI's OpenAiChatOptions, allowing users to specify the processing type used for serving requests to OpenAI.

Changes
- Added serviceTier field to OpenAiChatOptions class
- Updated builder pattern to include serviceTier() method
- Added appropriate getter/setter methods
- Updated documentation to reflect the new option

Resolves #4235

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatOptions.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatOptions.java
@@ -234,6 +234,11 @@ public class OpenAiChatOptions implements ToolCallingChatOptions {
 	private @JsonProperty("web_search_options") WebSearchOptions webSearchOptions;
 
 	/**
+	 * Specifies the <a href="https://platform.openai.com/docs/api-reference/responses/create#responses_create-service_tier">processing type</a> used for serving the request.
+	 */
+	private @JsonProperty("service_tier") String serviceTier;
+
+	/**
 	 * Collection of {@link ToolCallback}s to be used for tool calling in the chat completion requests.
 	 */
 	@JsonIgnore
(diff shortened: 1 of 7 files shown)


sobychacko:
@rafaelrddc Can we use an enum for service_tier if there are advertised values for it? That way, this will be more type safe. Additionally, could you please add a few tests that validate the feature in the chat response? Thanks!

rafaelrddc:
@sobychacko I have just added tests for the implementation. Although the `service_tier` field already has some predefined values, I believe it is more appropriate to keep it open, similar to how it's done for the `model` field. This approach provides greater flexibility and prevents integration breaks if OpenAI changes or adds new values in the future.

To facilitate use and validation, I created an enum with the currently supported values, allowing the user to choose to use it when they want more safety in the values entered.

sobychacko:
Thanks for the updates. Merged via https://github.com/spring-projects/spring-ai/commit/ad2e1bcda55e7be510eb3a98106217ec59056da8.

PR #4241, opened 2025-08-26
Add Elasticsearch Service Connection support

l-trotta:
Adding support for elasticsearch docker compose and testcontainer. 

(Disclaimer, I work for elastic)


eddumelendez:
Hi @l-trotta, spring-boot already offers support Service Connection for Elasticsearch via Docker Compose and Testcontainers. A Spring AI application with the right dependencies should be enough. See https://docs.spring.io/spring-boot/reference/features/dev-services.html#features.dev-services.docker-compose.service-connections

l-trotta:
hey @eddumelendez! I'm aware, but seeing Mongo implementations in both made me think adding it in here too would make it easier for spring ai users. Also the [current documentation](https://docs.spring.io/spring-ai/reference/api/docker-compose.html) makes it seem like the services mentioned in the list are the only supported connections. If it makes more sense I can close this and then update the documentation to show that any spring boot Service Connection can be used?

eddumelendez:
MongoDB Atlas is different from MongoDb implementations. I see Spring Boot 4 added support for MongoDB atlas, so spring ai should drop it when upgrade to the next major version. About the docs, you are right, docs should mention than some rely on existing service connection support provided already by spring boot.

l-trotta:
thanks for the feedback, I'll open another PR to update the documentation then.

Issue #513, opened 2024-03-26
The SDK of the vector database is very inflexible, unable to modify some database fields.


Warrior0x1:
In the current version, taking the Milvus database as an example, I cannot specify the field names of the database collection. If I have a previously used collection, and if the fields do not match the current fixed fields, I cannot perform a query. I strongly suggest that custom field names should be allowed!!!

        public static final String DOC_ID_FIELD_NAME = "doc_id";

	public static final String CONTENT_FIELD_NAME = "content";

	public static final String METADATA_FIELD_NAME = "metadata";

	public static final String EMBEDDING_FIELD_NAME = "embedding";

Warrior0x1:
The cost of rebuilding a vector database that has already been put into production in order to use Spring AI is enormous.

iAMSagar44:
I faced a similar issue with Azure AI Search, where I had imported some data into the vector store using Azure's inbuilt solution for importing and chunking data into Azure AI Search. The import process resulted in different field names compared to the vector store implementations in Spring AI. And when I was trying to do a vector search using Spring AI, it failed. 
The only way that I could think of resolving the issue was to create your own implementation of the Vector Store and model the fields in the code with the same name as in Azure AI Search.

markpollack:
Thanks for reporting this, sorry for the delay.   @iAMSagar44 wrt to the comment "The import process resulted in different field names compared to the vector store implementations in Spring AI."   Were the field names different or were the different fields.  I agree, we need to open up the names that are used.  A simple way would be to create setters for all the current field names, what do you think?

AishwaryaMalagannavar:
> Thanks for reporting this, sorry for the delay. [@iAMSagar44](https://github.com/iAMSagar44) wrt to the comment "The import process resulted in different field names compared to the vector store implementations in Spring AI." Were the field names different or were the different fields. I agree, we need to open up the names that are used. A simple way would be to create setters for all the current field names, what do you think?

@markpollack, This issue persists in the AzureVectorStore and does not appear to have been resolved in the 1.0.0 release.

PR #4246, opened 2025-08-27
feat(mcp-client): add support for server-provided returnDirect, with …

eeaters:
## Background
When migrating local Tools to a standalone service, the returnDirect property stopped working because ToolMetadata in the MCP client currently ignores client-side attributes. Although the MCP SDK schema already defines this field, its logic has not yet been implemented. I previously submitted [PR #513](https://github.com/modelcontextprotocol/java-sdk/pull/513)
 to address this gap, and this change builds on top of that work.

## Change Summary
This update allows the client to apply a global configuration for a specific server. The final returnDirect value is computed by merging client and server settings, requiring both sides to be true for the property to take effect.

Because client-side attributes are spread across multiple configuration sources, I introduced a composite class to unify the logic. This is an initial approach; I haven't yet found a cleaner solution. If there are better implementation ideas, I would greatly appreciate suggestions.

This PR addresses and fixes the issues described in [ISSUE #3481](https://github.com/spring-projects/spring-ai/issues/3481)

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/main/java/org/springframework/ai/mcp/client/common/autoconfigure/McpClientAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/main/java/org/springframework/ai/mcp/client/common/autoconfigure/McpClientAutoConfiguration.java
@@ -36,6 +36,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Import;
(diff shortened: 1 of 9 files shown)


sunyuhan1998:
Thank you for your contribution! I was wondering if this PR addresses the same issue as the previous one: https://github.com/spring-projects/spring-ai/pull/3787?

eeaters:
> Thank you for your contribution! I was wondering if this PR addresses the same issue as the previous one: #3787?

Yes, I saw this PR. However, I think it's somewhat risky to have this parameter entirely controlled by the server side. Therefore, I tried to extend the MCP SDK to also give the client-side the ability to control this parameter simultaneously.

sunyuhan1998:
> > Thank you for your contribution! I was wondering if this PR addresses the same issue as the previous one: #3787?
> 
> Yes, I saw this PR. However, I think it's somewhat risky to have this parameter entirely controlled by the server side. Therefore, I tried to extend the MCP SDK to also give the client-side the ability to control this parameter simultaneously.

I understand now, thank you for the explanation, it looks great!

PR #4252, opened 2025-08-27
GH-4251 Add the capability to parse required variables from templates to `TemplateRenderer`

### LINKED ISSUE #4251: `TemplateRenderer` should support retrieving required variables from the template.
**Expected Behavior**

Currently, `TemplateRenderer` only supports rendering templates. It would be highly beneficial if it could also support parsing the required variables from a given template, as this would greatly assist us in proactively validating templates. For example, in `org.springframework.ai.rag.util.PromptAssert#templateHasRequiredPlaceholders`, the current validation relies solely on `contains`, which is clearly prone to oversights.

**Current Behavior**

Currently, `TemplateRenderer` only supports rendering templates.

sunyuhan1998:
As mentioned in the issue, `TemplateRenderer` currently only supports rendering templates. The ability to also parse and extract the required variables from a given template would be highly beneficial, especially for proactively validating template correctness. This PR implements that capability. 

Specifically, it: 

1. Adds a `getRequiredVariables` method to `TemplateRenderer` to extract the variable names required by the template; 

3. Includes corresponding unit tests to ensure correctness and stability.

Fixes #4251

### DIFF
--- a/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/chat/OpenAiCompatibleChatModelIT.java
+++ b/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/chat/OpenAiCompatibleChatModelIT.java
@@ -68,7 +68,7 @@ public class OpenAiCompatibleChatModelIT {
 				.openAiApi(OpenAiApi.builder()
 					.baseUrl("https://api.groq.com/openai")
 					.apiKey(System.getenv("GROQ_API_KEY"))
-					.build()) 
+					.build())
 				.defaultOptions(forModelName("llama3-8b-8192"))
 				.build());
 		}
--- a/spring-ai-commons/src/main/java/org/springframework/ai/template/NoOpTemplateRenderer.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/template/NoOpTemplateRenderer.java
@@ -17,6 +17,7 @@
 package org.springframework.ai.template;
 
 import java.util.Map;
+import java.util.Set;
 
 import org.springframework.util.Assert;
 
@@ -34,6 +35,11 @@ public class NoOpTemplateRenderer implements TemplateRenderer {
 		Assert.notNull(variables, "variables cannot be null");
 		Assert.noNullElements(variables.keySet(), "variables keys cannot be null");
 		return template;
+	}
+
+	@Override
+	public Set<String> getRequiredVariables(String template) {
+		return Set.of();
 	}
 
 }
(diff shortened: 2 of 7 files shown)


ericbottard:
Hi @sunyuhan1998,
thanks for your contribution!

Could you please rebase against main and also add the following logic: not all implementations may support retrieving variable names. As a consequence, could you make the new method throw NotSupportedException and make it `default`

sunyuhan1998:
> Could you please rebase against main and also add the following logic: not all implementations may support retrieving variable names. As a consequence, could you make the new method throw NotSupportedException and make it `default`

Thank you very much for your review. I agree with your comments. I have merged the code from the main branch and adjusted the `getRequiredVariables` method as you suggested. However, I couldn't seem to find the `NotSupportedException` class in the project, so I used `OperationNotSupportedException` instead. Do you think this is acceptable? Could you please review it once again? Thank you so much!

ericbottard:
> > Could you please rebase against main and also add the following logic: not all implementations may support retrieving variable names. As a consequence, could you make the new method throw NotSupportedException and make it `default`
> 
> Thank you very much for your review. I agree with your comments. I have merged the code from the main branch and adjusted the `getRequiredVariables` method as you suggested. However, I couldn't seem to find the `NotSupportedException` class in the project, so I used `OperationNotSupportedException` instead. Do you think this is acceptable? Could you please review it once again? Thank you so much!

That's the one I meant, sorry.

However, the more I look at this, the more I think this may be a rabbit hole, especially for the case you want to support: what about nested properties? what about method invocations on objects? what about x... ? 
Each templating library (granted we only have one at the moment, but you opened a PR to submit one with jinja and I saw one using SpEL) may support different and complex semantics of things they can do with the model. The caller of `TemplateRenderer` is not capable of deciding whether the implementation will fail or not, in the general case.

sunyuhan1998:
> However, the more I look at this, the more I think this may be a rabbit hole, especially for the case you want to support: what about nested properties? what about method invocations on objects? what about x... ? Each templating library (granted we only have one at the moment, but you opened a PR to submit one with jinja and I saw one using SpEL) may support different and complex semantics of things they can do with the model. The caller of `TemplateRenderer` is not capable of deciding whether the implementation will fail or not, in the general case.

I admit that what you said makes sense. In fact, the reason I submitted this PR (along with the related issue) is that I noticed `org.springframework.ai.rag.util.PromptAssert#templateHasRequiredPlaceholders`. This method attempts to assert that the template contains all specified placeholders, but it relies solely on `contains`, which is clearly unreliable. So I thought, it would be great if we could support this capability directly within `TemplateRenderer`. I'm not sure if there might be a better solution though.

PR #4255, opened 2025-08-27
Parallel Tool Execution

### LINKED ISSUE #4254: Parallel Tool Execution

**Expected Behavior**

The DefaultToolCallingManager should process tool calls in parallel rather than sequentially when multiple independent tool calls are present in an AssistantMessage. This would significantly improve performance for scenarios requiring multiple concurrent AI tool interactions.

**Current Behavior**

The DefaultToolCallingManager processes tool calls sequentially using a simple for-loop:

````
for (AssistantMessage.ToolCall toolCall : assistantMessage.getToolCalls()) {
    // Sequential processing
}
````

This creates performance bottlenecks when dealing with multiple independent tool calls that could be executed concurrently.

**Context**

Proposed Solution
- Add parallel execution capability to DefaultToolCallingManager
- Ensure proper error handling and response ordering

### PR DESCRIPTION
**Pull request by rafaelrddc**
Description
This MR adds parallel execution capability to DefaultToolCallingManager to process multiple AI tool calls concurrently instead of sequentially, significantly improving performance for applications requiring multiple independent tool interactions.

Resolves #4254

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
@@ -20,7 +20,10 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Queue;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentLinkedDeque;
 
 import io.micrometer.observation.ObservationRegistry;
 import org.slf4j.Logger;
@@ -44,6 +47,10 @@ import org.springframework.ai.tool.observation.ToolCallingObservationConvention;
 import org.springframework.ai.tool.observation.ToolCallingObservationDocumentation;
 import org.springframework.ai.tool.resolution.DelegatingToolCallbackResolver;
 import org.springframework.ai.tool.resolution.ToolCallbackResolver;
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.core.task.support.ContextPropagatingTaskDecorator;
+import org.springframework.lang.Nullable;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 
@@ -71,6 +78,8 @@ public final class DefaultToolCallingManager implements ToolCallingManager {
 	private static final ToolExecutionExceptionProcessor DEFAULT_TOOL_EXECUTION_EXCEPTION_PROCESSOR
 			= DefaultToolExecutionExceptionProcessor.builder().build();
 
+	private static final TaskExecutor DEFAULT_TASK_EXECUTOR = buildDefaultTaskExecutor();
+
 	// @formatter:on
 
 	private final ObservationRegistry observationRegistry;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java [hunk 1]
```diff
@@ -245,6 +252,16 @@ public void setObservationConvention(ToolCallingObservationConvention observatio
...
+	private static TaskExecutor buildDefaultTaskExecutor() {
+		ThreadPoolTaskExecutor taskExecutor = new ThreadPoolTaskExecutor();
+		taskExecutor.setThreadNamePrefix("ai-toll-calling-");
+		taskExecutor.setCorePoolSize(4);
+		taskExecutor.setMaxPoolSize(16);
+		taskExecutor.setTaskDecorator(new ContextPropagatingTaskDecorator());
+		taskExecutor.initialize();
+		return taskExecutor;
+	}
+
```
sunyuhan1998: I think, no matter what, we shouldn't hardcode thread pool configurations as "magic values" here, even if it's just a default thread pool. Perhaps we should introduce corresponding configuration options and a dedicated configuration class to manage these parameters.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
```diff
@@ -71,6 +78,8 @@ public final class DefaultToolCallingManager implements ToolCallingManager {
 	private static final ToolExecutionExceptionProcessor DEFAULT_TOOL_EXECUTION_EXCEPTION_PROCESSOR
 			= DefaultToolExecutionExceptionProcessor.builder().build();
 
+	private static final TaskExecutor DEFAULT_TASK_EXECUTOR = buildDefaultTaskExecutor();
```
sunyuhan1998: Initializing the static thread pool by calling `buildDefaultTaskExecutor()` here causes the method to be executed regardless of whether the user provides a thread pool in the constructor, which seems inappropriate.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
```diff
@@ -173,64 +185,59 @@ private static List<Message> buildConversationHistoryBeforeToolExecution(Prompt
...
+							toolResult = this.toolExecutionExceptionProcessor.process(ex);
+						}
+						observationContext.setToolCallResult(toolResult);
+						return toolResult;
+					});
+
+				return new ToolResponseMessage.ToolResponse(toolCall.id(), toolName,
+						toolCallResult != null ? toolCallResult : "");
+			}, this.taskExecutor))
+			.map(CompletableFuture::join)
```
sunyuhan1998: Is this intended to wait for all tools to finish execution in startup order? I'm not sure if this approach makes sense—would using `CompletableFuture.allOf()` be more appropriate?

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
sunyuhan1998: By the way, the spelling of the thread name also appears to be incorrect: "toll" should be "tool".

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
rafaelrddc: I just deployed a change tweaking this behavior. The system will now create a default instance only when no custom configuration is supplied.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
rafaelrddc: I changed it to use allOf, but we still need to use join to get the responses. However, it will return immediately.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
rafaelrddc: I can add it to the ToolCallingProperties class, which will allow us to customize the default values using properties. What do you think?

Exemple
````java
@ConfigurationProperties(ToolCallingProperties.CONFIG_PREFIX)
public class ToolCallingProperties {

	public static final String CONFIG_PREFIX = "spring.ai.tools";

	private final Observations observations = new Observations();

	private final TaskExecutorProperties taskExecutor = new TaskExecutorProperties();
    
        public static class TaskExecutorProperties {
		/**
		 * Whether to enable custom task executor configuration for tool calls.
		 */
		private boolean enabled = false;

		/**
		 * Core number of threads in the pool.
		 */
		private int corePoolSize = Runtime.getRuntime().availableProcessors();

		/**
		 * Maximum number of threads in the pool.
		 */
		private int maxPoolSize = Runtime.getRuntime().availableProcessors() * 2;

		/**
		 * Capacity of the queue for holding tasks before they are executed.
		 */
		private int queueCapacity = 100;

		/**
		 * Prefix for thread names in the pool.
		 */
		private String threadNamePrefix = "tool-call-exec-";
	}
}
````

ssamir0602:
This would be a solid addition, do we need more eyes on this?

PR #4256, opened 2025-08-28
GH-4232 Introduce the `ValidationMode` property to `RewriteQueryTransformer`

### LINKED ISSUE #4232: `RewriteQueryTransformer` should allow custom prompts without {target} placeholder
**Expected Behavior**

The `RewriteQueryTransformer` should allow custom prompt templates without requiring the `{target}` placeholder when the custom prompt doesn't use it.

```java
// This should work with a custom prompt that doesn't need {target}
RewriteQueryTransformer.builder()
    .chatClientBuilder(chatClientBuilder)
    .promptTemplate(customPromptTemplate) // Custom prompt without {target}
    .build(); // Should not throw IllegalArgumentException
```

**Current Behavior**

`RewriteQueryTransformer` currently enforces that all prompt templates must include the `{target}` placeholder, even when using completely custom prompts that don't need this placeholder.

```java
// This fails with IllegalArgumentException even though custom prompt doesn't use {target}
RewriteQueryTransformer.builder()
    .chatClientBuilder(chatClientBuilder)
    .promptTemplate(customPromptTemplate) // Custom prompt without {target}
    .build(); // Throws: "The following placeholders must be present in the prompt template: target"
```

**Error Log:**
```
java.lang.IllegalArgumentException: The following placeholders must be present in the prompt template: target
	at org.springframework.ai.rag.util.PromptAssert.templateHasRequiredPlaceholders(PromptAssert.java:55)
	at org.springframework.ai.rag.preretrieval.query.transformation.RewriteQueryTransformer.<init>(RewriteQueryTransformer.java:71)
	at org.springframework.ai.rag.preretrieval.query.transformation.RewriteQueryTransformer$Builder.build(RewriteQueryTransformer.java:128)
```

**Context**

**How has this issue affected you?**
I'm implementing a custom query transformation system that performs specialized processing beyond the default query rewriting. My custom prompt template doesn't need the `{target}` placeholder but the framework forces me to include it, causing runtime exceptions.

**What are you trying to accomplish?**
Create a flexible query transformation system that can handle custom business logic without being constrained by framework assumptions about required placeholders.

**What other alternatives have you considered?**
1. Using a different transformer class (but `RewriteQueryTransformer` is the most appropriate for this use case)
2. Implementing a custom transformer from scratch (overkill for this simple requirement)

**Are you aware of any workarounds?**
Yes, I can add a dummy `{target}` placeholder to my custom prompt template, but this is a hack that makes the code less clean and the prompt less readable. The placeholder validation should be conditional based on what's actually used in the custom prompt.

**Example Use Case:**
```java
// Custom prompt for specialized query transformation
private val customPromptTemplate = PromptTemplate("""
    You are a query processor for a specialized analysis system.
    
    TASK: Process the user query to extract specific information.
    
    INPUT: {query}
    
    OUTPUT FORMAT (JSON):
    {
      "rewritten_query": "processed query",
      "custom_field": "custom value"
    }
    
    Query: {query}
    """)
```

This prompt doesn't need `{target}` but the framework currently requires it, causing the application to crash at runtime.

sunyuhan1998:
As described in the issue, the current `RewriteQueryTransformer` allows users to customize templates via the constructor, but enforces strict validation of placeholders in the template—this can be overly restrictive in certain scenarios. This PR introduces a new `ValidationMode` property to `RewriteQueryTransformer`, supporting three modes: `THROW`, `WARN`, and `NONE`. The default mode is `THROW`, which preserves the original behavior. If set to `WARN`, the system will perform validation as usual, but when a required variable is missing from the template, it will only log a warning instead of throwing an exception. If set to `NONE`, validation is completely skipped.

Changes included in this PR:
1. Added the `ValidationMode` property to `RewriteQueryTransformer` to enable flexible control over placeholder validation;
2. Implemented corresponding unit tests to ensure correctness across all validation modes.

Fixes #4232

### DIFF
--- a/spring-ai-rag/src/test/java/org/springframework/ai/rag/preretrieval/query/transformation/RewriteQueryTransformerTests.java
+++ b/spring-ai-rag/src/test/java/org/springframework/ai/rag/preretrieval/query/transformation/RewriteQueryTransformerTests.java
@@ -21,6 +21,7 @@ import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.read.ListAppender;
 import org.junit.jupiter.api.Test;
 import org.slf4j.LoggerFactory;
+
 import org.springframework.ai.chat.client.ChatClient;
 import org.springframework.ai.chat.prompt.PromptTemplate;
 


sunyuhan1998:
Hi @markpollack @sobychacko , What do you think about this? Could you please help review this PR? Thank you.

ThomasVitale:
@sunyuhan1998 thanks for raising this issue! The validation that is currently done in the `RewriteQueryTransformer` (but also in other RAG components) is very permissive and high-level. I wonder If we could remove it altogether and come up with a different way to validate the prompts.

The current logic only checks for the presence of words in a prompt template that match the variables expected by the component. But it doesn't really check if they are available as placeholders. Since the introduction of the modular RAG components, we introduced the `TemplateRenderer` API and extended the `PromptTemplate` API. I'm wondering if there's something we could do there to solve this problem so that the same solution can be adopted by any type of component that offers prompt template customization. Something more robust than the current `PromptAssert` that validates the actual placeholders in the template and not just a String occurrence. What do you think?

Besides that, I'm curious about the use cases where it should be ok not to validate the prompt placeholders. Could you elaborate on that? For example, in this case, the prompt should include a `query` placeholder to ask for a rewrite. If the placeholder is not there, wouldn't that mean the operation won't work? For special cases or additional flexibility, it might be better to implement a custom `QueryTransformer` directly and have full flexibility.

ThomasVitale:
Just noticed the original issue. If `target` is problematic, I would suggest removing the validation for the `target` placeholder directly, so to still guarantee the core functionality of the component. Or maybe even update the default prompt to remove the `target` part altogether and have a more generic prompt. If more tailored prompts are needed for specific scenarios, a custom `PromptTemplate` can be provided. It would be a simpler solution. Thoughts?

sunyuhan1998:
> Just noticed the original issue. If `target` is problematic, I would suggest removing the validation for the `target` placeholder directly, so to still guarantee the core functionality of the component. Or maybe even update the default prompt to remove the `target` part altogether and have a more generic prompt. If more tailored prompts are needed for specific scenarios, a custom `PromptTemplate` can be provided. It would be a simpler solution. Thoughts?

Actually, my initial idea was: it would be ideal if we could ensure that all variables corresponding to the placeholders in the prompt have been provided by the user. However, currently `TemplateRenderer` doesn't have the capability to extract all placeholders from a prompt. That's why I submitted this PR (https://github.com/spring-projects/spring-ai/pull/4252). If this PR gets merged, the issue we're discussing in the current PR would be naturally resolved. What do you think?

PR #4257, opened 2025-08-28
GH-4205 Optimized the `getClassOrNull` method in `ToolCallingAutoConfiguration`

### LINKED ISSUE #4205: 升级1.0.01版本启动报ClassNotFoundException: org.springframework.security.oauth2.client.ClientAuthorizationException

**Bug description**
A clear and concise description of what the bug is about.

**Environment**
Please provide as many details as possible: Spring AI version, Java version, which vector store you use if any, etc

**Steps to reproduce**
Steps to reproduce the issue.

**Expected behavior**

### LINKED ISSUE #4249: ClassNotFoundException when enabling DEBUG logging for org.springframework.ai
Bug Report: ClassNotFoundException when enabling DEBUG logging for org.springframework.ai
Spring AI version: 1.0.1
Spring Boot version: 3.5.5
Java version: 21

Description:
When setting the logging level to DEBUG for org.springframework.ai, the application fails to start due to a ClassNotFoundException.
However, using INFO level works without issues.
This behavior makes logging level affect runtime stability, which should not happen.

Steps to Reproduce:
1. Create a Spring Boot project with Spring AI 1.0.1.
2. Do not add the spring-security-oauth2-client dependency.
3.Configure logging in application.properties:
    logging.level.org.springframework.ai=DEBUG
4. Run the application.

Actual Behavior:
Application fails with:
java.lang.ClassNotFoundException: org.springframework.security.oauth2.client.ClientAuthorizationException
    at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
    ...
    at org.springframework.ai.model.tool.autoconfigure.ToolCallingAutoConfiguration.toolExecutionExceptionProcessor(ToolCallingAutoConfiguration.java:87)
    ...

Expected Behavior:
Enabling DEBUG logging should not trigger ClassNotFoundException.
If spring-security-oauth2-client is not present, Spring AI should gracefully skip OAuth2-related configuration.
Logging level should never change the runtime behavior of auto-configuration.

sunyuhan1998:
The original intent of the `ToolCallingAutoConfiguration#getClassOrNull` method is to attempt loading a specified class, returning `null` if the class cannot be found. However, the current implementation prints the full stack trace when a `ClassNotFoundException` is thrown, which may mislead users into thinking there is a real problem (while in fact, this is an expected and normal behavior). This PR improves the exception handling in the method, so that when a `ClassNotFoundException` occurs, only the class name is logged instead of the entire stack trace, thereby avoiding confusion and improving log clarity.

Fixes #4205
Fixes #4249

### DIFF
--- a/auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
+++ b/auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
@@ -126,10 +126,19 @@ public class ToolCallingAutoConfiguration {
 
 	private static Class<? extends RuntimeException> getClassOrNull(String className) {
 		try {
-			return (Class<? extends RuntimeException>) ClassUtils.forName(className, null);
+			Class<?> clazz = ClassUtils.forName(className, null);
+			if (RuntimeException.class.isAssignableFrom(clazz)) {
+				return (Class<? extends RuntimeException>) clazz;
+			}
+			else {
+				logger.debug("Class {} is not a subclass of RuntimeException", className);
+			}
 		}
 		catch (ClassNotFoundException e) {
-			logger.debug("Cannot load class", e);
+			logger.debug("Cannot load class: {}", className);
+		}
+		catch (Exception e) {
+			logger.debug("Error loading class: {}", className, e);
 		}
 		return null;
 	}


sobychacko:
Merged via https://github.com/spring-projects/spring-ai/commit/befaf87c775d91fea254d2a3974c4ecbd4e79232.

auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
```diff
@@ -126,10 +126,19 @@ ToolCallingContentObservationFilter toolCallingContentObservationFilter() {
...
+			if (RuntimeException.class.isAssignableFrom(clazz)) {
+				return (Class<? extends RuntimeException>) clazz;
+			}
+			else {
+				logger.debug("Class {} is not a subclass of RuntimeException", className);
+			}
 		}
 		catch (ClassNotFoundException e) {
-			logger.debug("Cannot load class", e);
+			logger.debug("Cannot load class: {}", className);
```
YunKuiLu, line 138: This log can still confuse users. Maybe it's better to catch this exception at line 88 and explain why this class is needed.

PR #4260, opened 2025-08-28
Polishing

quaff:
1. Fix typo
2. Remove redundant `matchIfMissing = false` since it's default value


ilayaperumalg:
The typo is fixed via https://github.com/spring-projects/spring-ai/pull/4283.

Issue #4250, opened 2025-08-27
conversion from json to POJO error when use tool call in ai stream


wykeli:
Bug Report: com.fasterxml.jackson.core.JsonParseException: Unexpected close marker ']': expected '}' (for Object starting at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 235])
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 289]
Spring AI version: 1.0.1
Spring Boot version: 3.5.5
Java version: 17

Description:
the json content is 
`
{"jobParam": {"clusterType": "xx", "noAddClusterAgain": false, "noQuitCluster": false, "receiveAccount": "xx", "noReturnMachine": true}, "borrowNodeCondition": {"nodeFromLabel": "buffer", "notBorrow": false, "weCubeKeys": [{"field": "idc", "relation": "eq", "value": "x"]}, "creator": "XXX", "rate": 6, "sourceCondition": {"weCubeKeys": [{"field": "idc", "relation": "in", "values": ["s", "a"]}, {"field": "cluster_alias", "relation": "eq", "value": "abc"}, {"field": "node_ip", "relation": "eq", "value": "127.0.0.1"]}, "taskName": ""}, "conversionId": "1"}
`
that is not an normal json style
<img width="2267" height="921" alt="Image" src="https://github.com/user-attachments/assets/679afe4d-6dc2-46e7-9476-ad50abbf1d71" />

------------------------------------------------------------------------------------------------------------------------

what the code when occur in 
`
chatClient.prompt().toolCallbacks(ToolCallbacks.from(XXX)).user(xxx).advisors(spec -> spec.param(ChatMemory.CONVERSATION_ID, XXX)).stream().content()
`
the tool code is 
`
@Tool(description = "发起k8s节点迁移任务")
    public String startMgrTask(
            @ToolParam(description = "迁移任务参数") AiMgrJobParam jobParam, @ToolParam(description = "当前会话id") String conversionId) {
}
`
I guess this is a bug when using tool call in ai stream ,beacus no exception orrur when use the sync way

`chatClient.prompt().system(SYSTEM_PROMPT).toolCallbacks(ToolCallbacks.from()).user().advisors(spec -> spec.param(ChatMemory.CONVERSATION_ID, conversionKey)).call().content()`

kmw10693:
[...]

PR #4274, opened 2025-08-31
Fix: handle JSON parse error in streaming tool calls

kmw10693:
## Overview
This PR fixes a JSON parsing error that occurs when handling **streaming tool calls** in Spring AI.

 Related issue: [spring-projects/spring-ai#4250](https://github.com/spring-projects/spring-ai/issues/4250?utm_source=chatgpt.com)

---

## Why
- In streaming mode, tool call JSON fragments can arrive incomplete.  
- This results in Jackson throwing errors such as: `com.fasterxml.jackson.core.JsonParseException: Unexpected close marker ']': expected '}'`
- The issue only occurs in streaming (`stream()`) calls.  
- Synchronous calls (`call()`) are not affected.  
- This prevents users from reliably using tool callbacks with streaming responses.  

---

## What Changed
- Added a new `ToolExecutionEligibilityPredicate` implementation: **`ToolOnFinishPredicate`**.  
- Ensures tool execution only runs after the assistant signals completion (`finishReason = "tool_calls"` or `"stop"`).  
- Prevents premature execution and malformed JSON during streaming tool calls.  

---

## How to Use
Example usage:

```java
ChatClient client = ChatClient.create(openAiChatModel);

String result = client.prompt()
  .user("Migrate node cluster xx")
  .toolCallbacks(ToolCallbacks.from(new MyTool()))
  .options(OpenAiChatOptions.builder()
      .toolExecutionEligibilityPredicate(new ToolOnFinishPredicate())
      .build())
  .stream()
  .content();

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolOnFinishPredicate.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolOnFinishPredicate.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.model.tool;
+
+import java.util.List;
+import java.util.function.BiPredicate;
+
+import org.jetbrains.annotations.NotNull;
+
+import org.springframework.ai.chat.model.ChatResponse;
+import org.springframework.ai.chat.model.Generation;
+import org.springframework.ai.chat.prompt.ChatOptions;
+
+/**
+ * Executes tools only when the assistant signals completion (finishReason = "tool_calls"
+ * or "stop").
+ */
+public final class ToolOnFinishPredicate implements ToolExecutionEligibilityPredicate {
+
+	@Override
+	public boolean isToolExecutionRequired(ChatOptions opts, ChatResponse resp) {
+		List<Generation> gens = resp.getResults();
+		if (gens.isEmpty()) {
+			return false;
+		}
+
+		Generation gen = gens.get(0);
+		boolean hasToolCalls = !gen.getOutput().getToolCalls().isEmpty();
+		String finish = String.valueOf(gen.getMetadata().get("finishReason"));
+
+		return hasToolCalls && ("tool_calls".equalsIgnoreCase(finish) || "stop".equalsIgnoreCase(finish));
+	}
+
+	@Override
+	public boolean test(ChatOptions chatOptions, ChatResponse chatResponse) {
+		return false;
+	}
+
+	@NotNull
+	@Override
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolOnFinishPredicate.java [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.model.tool;
+
+import java.util.List;
+import java.util.function.BiPredicate;
+
+import org.jetbrains.annotations.NotNull;
```
markpollack: can't use this annotation for not null

markpollack:
The PR looks incomplete.  I've commented on the issue thread for further clarification. #4274

spring-ai-model/src/main/java/org/springframework/ai/model/tool/ToolOnFinishPredicate.java
kmw10693: @markpollack I’ve removed ToolOnFinishPredicate.java and applied the fix directly in OpenAiChatModel.
when the LLM returns malformed or invalid JSON during streaming, the code explicitly throws an exception.

kmw10693:
> The PR looks incomplete. I've commented on the issue thread for further clarification. #4274

@markpollack 
On the issue, the problematic LLM was Qwen3-8B. they found that the chunk returned by the LLM was invalid. 

However, the OpenAIChatModel was not properly detected when the LLM returned an invalid response.
I’ve updated OpenAiChatModel so that if the LLM sends an invalid or malformed JSON response (e.g., broken chunk). the code now explicitly throws an exception. I also added the OpenAiChatModelStreamingTest

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -492,14 +493,20 @@ private ChatResponseMetadata from(ChatResponseMetadata chatResponseMetadata, Usa
...
+			List<Choice> choices = chunk.choices()
+				.stream()
+				.map(chunkChoice -> new Choice(chunkChoice.finishReason(), chunkChoice.index(), chunkChoice.delta(),
+						chunkChoice.logprobs()))
+				.toList();
+
+			return new OpenAiApi.ChatCompletion(chunk.id(), choices, chunk.created(), chunk.model(),
+					chunk.serviceTier(), chunk.systemFingerprint(), "chat.completion", chunk.usage());
+		}
+		catch (Exception e) {
```
markpollack, line 506: I don't see what is going to throw an exception in either one of these calls.  records are being created, what json marshalling is going on?

models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelStreamingTest.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+public class OpenAiChatModelStreamingTest {
+
+	@Test
+	void shouldThrowExceptionOnInvalidJsonChunk() {
+		OpenAiApi mockApi = Mockito.mock(OpenAiApi.class);
+
+		OpenAiApi.ChatCompletionChunk invalidChunk = new OpenAiApi.ChatCompletionChunk("invalid-id", null,
```
ilayaperumalg, line 36: Could you explain what qualifies this as an invalid chunk here? I don't understand how this could fail with JSON parser error.

ilayaperumalg:
@kmw10693 Closing this PR for now. Please feel free to re-open and provide response to the questions above. Thanks for your contribution!

Issue #4302, opened 2025-09-04
zhipu embedding model support batch processing


eeaters:
Zhipu's embedding model now supports batch processing. The current code doesn't, and it's throwing a warning. We need to add support for batch embedding requests.

website: https://docs.bigmodel.cn/cn/guide/models/embedding/embedding-3

sdeleuze:
We are stopping to maintain ZhiPuAI as part of the core Spring AI project as we have no proper way to test it, and have to focus on a smaller set of models. We have not been able to find a maintainer for a support outside of this repository, but if somebody want to volunteer, please reach us on https://github.com/spring-projects/spring-ai/discussions.

PR #4308, opened 2025-09-04
optimize ZhiPu Embedding to support batch embedding.

eeaters:
- support batch embedding
- Make test adjustments based on the official demo

Fixes: [#4232](https://github.com/spring-projects/spring-ai/issues/4302)

### DIFF
--- a/models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/ZhiPuAiEmbeddingModel.java
+++ b/models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/ZhiPuAiEmbeddingModel.java
@@ -16,15 +16,15 @@
 
 package org.springframework.ai.zhipuai;
 
-import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import io.micrometer.observation.ObservationRegistry;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.springframework.ai.chat.metadata.DefaultUsage;
+import org.springframework.ai.chat.metadata.EmptyUsage;
+import org.springframework.ai.chat.metadata.Usage;
 import org.springframework.ai.document.Document;
 import org.springframework.ai.document.MetadataMode;
 import org.springframework.ai.embedding.AbstractEmbeddingModel;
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
@mxsl-gr Could you help with the review of this PR? Thanks

mxsl-gr:
> @mxsl-gr Could you help with the review of this PR? Thanks

No problem, I’ve been busy with other tasks. I’ll review it later today.

models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/ZhiPuAiEmbeddingModel.java
```diff
@@ -150,12 +152,9 @@ public float[] embed(Document document) {
 	@Override
 	public EmbeddingResponse call(EmbeddingRequest request) {
 		Assert.notEmpty(request.getInstructions(), "At least one text is required!");
-		if (request.getInstructions().size() != 1) {
-			logger.warn(
-					"ZhiPu Embedding does not support batch embedding. Will make multiple API calls to embed(Document)");
-		}
 
 		EmbeddingRequest embeddingRequest = buildEmbeddingRequest(request);
+		var zhipuEmbeddingRequest = zhipuEmbeddingRequest(embeddingRequest);
```
mxsl-gr: this seems unnecessary. It might be cleaner to handle it in `buildEmbeddingRequest` instead of wrapping it with another request.

mxsl-gr (changes requested):

models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/ZhiPuAiEmbeddingModel.java
eeaters: > this seems unnecessary. It might be cleaner to handle it in `buildEmbeddingRequest` instead of wrapping it with another request.

Influenced by the EmbeddingModelObservationContext, its context requirement mandates the use of `org.springframework.ai.embedding.EmbeddingRequest`;
However, the ZhiPu API uses ZhiPuAiApi.EmbeddingRequest, so they cannot be directly integrated into a single method.

so, I try to adjusting the order of the code to improve its readability.

mxsl-gr: you‘re right, I really forgot about that. I think that works too.

mxsl-gr (approved):

ilayaperumalg:
@eeaters Thanks for the PR! @mxsl-gr Thanks for the review. Rebased and merged as b7f231ac4

PR #4314, opened 2025-09-04
Bedrock converse chat model to merge toolcalling chat options

ilayaperumalg:
- When building the chat client request, if the ToolCallingChatOptions is passed via prompt, merge the options into Bedrock chat options. Previously, this was ignored.
 - Add test to verify function calling when ToolCallingChatOptions is passed


tzolov:
LGTM, thanks @ilayaperumalg

Issue #4331, opened 2025-09-07
JTokkitTokenCountEstimator : Add Base64 support for more accurate binary data token estimation


Hyeri1ee:
**Expected Behavior**

The JTokkitTokenCountEstimator should be improved to more accurately handle binary data (byte[]) within multimodal prompts. Currently, it simply adds the byte array's length to the token count, but most large language models (LLMs) process binary data by tokenizing a Base64-encoded string representation of it. Therefore, the estimation logic should reflect this process.

**Current Behavior**

Currently, the JTokkitTokenCountEstimator processes binary data (byte[]) in multimodal prompts by adding binaryData.length directly to the token count. As noted in the code's own comment, `// This is likely incorrect.`, this approach is inaccurate.

Most LLMs tokenize binary data after it has been encoded into a Base64 string. Base64 encoding increases the original data's size by approximately 33%, so the current logic significantly underestimates the actual number of tokens consumed by the model. This leads to errors in estimating API usage and costs.

**Proposed Code Example**

```java
// After the proposed improvement
else if (media.getData() instanceof byte[] binaryData) {
    // Convert the byte array to a Base64 string for token estimation
    String base64Data = Base64.getEncoder().encodeToString(binaryData);
    tokenCount += this.estimate(base64Data);
}
```

**Context**

Problem: The current JTokkitTokenCountEstimator underestimates the token count for binary data like image files, resulting in a large discrepancy between estimated and actual API costs.

Objective: The goal is to enhance the JTokkitTokenCountEstimator to provide a more accurate token estimate for binary data, which would help developers manage API costs more reliably.

ilayaperumalg:
Closing this as resolved via #4332

PR #4332, opened 2025-09-07
GH-4331 improve token estimation for binary data

Hyeri1ee:
### Summary
This PR improves the JTokkitTokenCountEstimator by converting binary data to a Base64 string for more accurate token estimation. This change aligns the logic with how most multimodal LLMs actually process this data type, leading to better cost predictions.

### Changes
- Enhancement: The estimate() method now handles byte[] data by converting it to a Base64 string before token counting. **This directly addresses the previous inaccurate behavior of adding the raw byte array length, as noted by the original code comment** `// This is likely incorrect.`

- Validation: Existing tests in TokenCountBatchingStrategyTests have been verified to ensure no regressions.

### Key Features
- Accurate Estimation: Provides a more realistic token count for binary data.

- LLM Alignment: The new approach mirrors the Base64 encoding used by most multimodal models.

- Backward Compatibility: This is an internal change and does not affect the public API.

Closes: #4331

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/tokenizer/JTokkitTokenCountEstimator.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/tokenizer/JTokkitTokenCountEstimator.java
@@ -24,6 +24,8 @@ import org.springframework.ai.content.Media;
 import org.springframework.ai.content.MediaContent;
 import org.springframework.util.CollectionUtils;
 
+import java.util.Base64;
+
 /**
  * Estimates the number of tokens in a given text or message using the JTokkit encoding
  * library.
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
@Hyeri1ee Thanks for the PR improving the way token usage is handled for the binary data.

Hyeri1ee:
> @Hyeri1ee Thanks for the PR improving the way token usage is handled for the binary data.

glad it helped! : )

ilayaperumalg:
Rebased and merged as 677730fd3976f9ce5ee5095a49acfac78efbfb72

ilayaperumalg:
Cherry-picked and pushed 1.0.x as c60d9425c

PR #4338, opened 2025-09-08
feat(mcp): enhance tool naming with server title and prefix shortening

tzolov:
- Add title parameter to McpSchema.Implementation constructor to distinguish between client name and server name
- Update prefixedToolName() to include optional title parameter and shorten long prefixes
- Implement shorten() method to abbreviate prefixes by taking first letter of each word
- Ensure generated tool names stay within 64-character limit
- Update all related tests to use new constructor signature and verify shortened names

This change improves tool name uniqueness when multiple MCP servers provide tools with the same name, while keeping names concise through intelligent prefix shortening.


markpollack:
This is definietly as step in the right direction, but i think there can be some tweaks and changes in a separate follow up issue for 1.1 M2


We should only make changes to the name if we have a name collision in the name of tools that are being sent to the AI model.  This way it is easier to reason about and we can be sure there are no duplicate named by checking the list of names vs. relying on a heuristic that needs to change all mcp tool names with a prefix and still has some corner case (e.g. if not using autoconfig the mcp title needs to be set to make the name unique for the case of the same tool name with different servers.  This will also likely help the AI model not to change the tool name it wants to invoke (which itself is a funny situation!)

ilayaperumalg:
This was merged via [f5e8349](https://github.com/spring-projects/spring-ai/commit/f5e83491efd7eaaee98057a0e9619cf25def1e44)

Issue #4325, opened 2025-09-05
Make Anthropic prompt caching message-type aware (TTLs, eligibility, min-size) and optimize cache-block usage


adase11:
Follow up on discussion from #4300 

⸻

## Context

First off, thanks for landing prompt caching support—this is a big unlock. The recent work wires cache_control into Spring AI’s Anthropic integration via AnthropicChatOptions, including a single cacheTtl and a “strategy” knob, plus usage tracking fields for cache reads/writes. That’s a solid baseline.  ￼

In real applications, we can squeeze substantially more value (and avoid pitfalls) by making caching aware of message types and by optimizing how we “spend” Anthropic’s limited cache breakpoints.

For reference, I prototyped this in a small PR (https://github.com/sobychacko/spring-ai/pull/2) against @sobychacko 's branch which added fine-grained controls (per-message-type TTLs, eligibility, and minimum sizes) in addition to some of the other changes already incorporated into #4300 .

## Why now

Anthropic’s behavior has a few important constraints/opportunities:
	•	You can define up to 4 cache breakpoints; automatic prefix checking looks back ~20 content blocks. Using multiple, intentional breakpoints drives higher hit rates.  ￼
	•	Default TTL is 5 minutes, and 1 hour is also available. Mixing TTLs is allowed but longer TTLs must appear before shorter TTLs—the order affects both correctness and billing.  ￼[source](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching?utm_source=chatgpt.com#mixing-different-ttls)
    •	Minimum cacheable length is 1024 tokens (most models) or 2048 for Haiku; short content won’t be cached even if marked. A library-level guard keeps us from wasting breakpoints on tiny content.  ￼

> You can use both 1-hour and 5-minute cache controls in the same request, but with an important constraint: Cache entries with longer TTL must appear before shorter TTLs (i.e., a 1-hour cache entry must appear before any 5-minute cache entries).

   
Given those constraints, two knobs consistently drive better cache utilization in production:
	1.	Per-message-type TTL (e.g., SYSTEM & TOOLS at 1h; USER & ASSISTANT at 5m)
	2.	Minimum content size thresholds (global + per-message-type) so we don’t “spend” scarce breakpoints on small or low-value segments

In my testing, those two changes notably improved “tokens cached as a % of total input” while staying within Anthropic’s rules. 

⸻

### Message-type-aware caching policy

[...]

PR #4342, opened 2025-09-09
gh-4325: Enhance cache management for Anthropic API by introducing per-message TTL and configurable content block usage optimization.

adase11:
# Summary: 
Introduces a structured prompt caching API for Anthropic via AnthropicCacheOptions, applies cache_control deterministically, adds per-message TTLs (5m/1h), content-length eligibility, and clarifies tool caching behavior. Updates docs and adds tests to validate wire format, headers, and limits.

# API

   - AnthropicChatOptions: adds cacheOptions(AnthropicCacheOptions); default remains no caching.
  - AnthropicCacheOptions: strategy (NONE, SYSTEM_ONLY, SYSTEM_AND_TOOLS, CONVERSATION_HISTORY); per-message TTLs via AnthropicCacheTtl (FIVE_MINUTES, ONE_HOUR);
messageTypeMinContentLength; contentLengthFunction for custom token estimates.
  - AnthropicChatModel: applies cache_control only when eligible; caches the last tool definition; never caches the latest user question; uses array system format when caching;
auto-sets Anthropic beta header for 1h TTL.
  - AnthropicApi: models cache_control on content blocks and tools; exposes Usage cacheCreationInputTokens and cacheReadInputTokens.
  - Utilities: CacheEligibilityResolver (with CacheBreakpointTracker) enforces Anthropic’s 4-breakpoint limit.
  

# Tests

  - Unit: AnthropicCacheOptionsTests; CacheEligibilityResolverTests; AnthropicPromptCachingMockTest (wire format, TTL beta header, 4-breakpoint limit).
  - IT: AnthropicPromptCachingIT (guarded by ANTHROPIC_API_KEY; validates real usage fields when available).

# Documentation

  - Updates spring-ai-docs Anthropic page to use cacheOptions with strategy examples.
  - Adds per-message TTL examples (1h) and notes automatic beta header.
  - Adds eligibility guidance (min lengths, custom `contentLengthFunction`).
  - Clarifies tool caching (last tool definition) and that latest user message is not cached in conversation history.

# Compatibility:
  - Default behavior unchanged (caching disabled unless cacheOptions provided).
  - Prior doc examples using cacheStrategy/cacheTtl are replaced with cacheOptions; docs updated accordingly.

Closes: #4325

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -42,7 +42,9 @@ import org.springframework.ai.anthropic.api.AnthropicApi.ContentBlock;
 import org.springframework.ai.anthropic.api.AnthropicApi.ContentBlock.Source;
 import org.springframework.ai.anthropic.api.AnthropicApi.ContentBlock.Type;
 import org.springframework.ai.anthropic.api.AnthropicApi.Role;
-import org.springframework.ai.anthropic.api.AnthropicCacheStrategy;
+import org.springframework.ai.anthropic.api.AnthropicCacheOptions;
+import org.springframework.ai.anthropic.api.AnthropicCacheTtl;
+import org.springframework.ai.anthropic.api.utils.CacheEligibilityResolver;
 import org.springframework.ai.chat.messages.AssistantMessage;
 import org.springframework.ai.chat.messages.Message;
 import org.springframework.ai.chat.messages.MessageType;
@@ -93,6 +95,7 @@ import org.springframework.util.StringUtils;
  * @author Alexandros Pappas
  * @author Jonghoon Park
  * @author Soby Chacko
+ * @author Austin Dase
  * @since 1.0.0
  */
 public class AnthropicChatModel implements ChatModel {
@@ -463,11 +466,9 @@ public class AnthropicChatModel implements ChatModel {
 			requestOptions.setToolContext(ToolCallingChatOptions.mergeToolContext(runtimeOptions.getToolContext(),
 					this.defaultOptions.getToolContext()));
 
-			// Merge cache strategy and TTL (also @JsonIgnore fields)
-			requestOptions.setCacheStrategy(runtimeOptions.getCacheStrategy() != null
-					? runtimeOptions.getCacheStrategy() : this.defaultOptions.getCacheStrategy());
-			requestOptions.setCacheTtl(runtimeOptions.getCacheTtl() != null ? runtimeOptions.getCacheTtl()
-					: this.defaultOptions.getCacheTtl());
+			// Merge cache options that are Json-ignored
+			requestOptions.setCacheOptions(runtimeOptions.getCacheOptions() != null ? runtimeOptions.getCacheOptions()
+					: this.defaultOptions.getCacheOptions());
 		}
 		else {
 			requestOptions.setHttpHeaders(this.defaultOptions.getHttpHeaders());
@@ -498,41 +499,23 @@ public class AnthropicChatModel implements ChatModel {
 		AnthropicChatOptions requestOptions = null;
 		if (prompt.getOptions() instanceof AnthropicChatOptions) {
 			requestOptions = (AnthropicChatOptions) prompt.getOptions();
-			logger.debug("DEBUGINFO: Found AnthropicChatOptions - cacheStrategy: {}, cacheTtl: {}",
-					requestOptions.getCacheStrategy(), requestOptions.getCacheTtl());
+			logger.debug("DEBUGINFO: Found AnthropicChatOptions - cacheOptions {}", requestOptions.getCacheOptions());
 		}
 		else {
 			logger.debug("DEBUGINFO: Options is NOT AnthropicChatOptions, it's: {}",
 					prompt.getOptions() != null ? prompt.getOptions().getClass().getName() : "null");
 		}
 
-		AnthropicCacheStrategy strategy = requestOptions != null ? requestOptions.getCacheStrategy()
-				: AnthropicCacheStrategy.NONE;
-		String cacheTtl = requestOptions != null ? requestOptions.getCacheTtl() : "5m";
+		AnthropicCacheOptions cacheOptions = requestOptions != null ? requestOptions.getCacheOptions()
+				: AnthropicCacheOptions.DISABLED;
 
-		logger.debug("Cache strategy: {}, TTL: {}", strategy, cacheTtl);
-
-		// Track how many breakpoints we've used (max 4)
-		CacheBreakpointTracker breakpointsUsed = new CacheBreakpointTracker();
-		ChatCompletionRequest.CacheControl cacheControl = null;
-
-		if (strategy != AnthropicCacheStrategy.NONE) {
-			// Create cache control with TTL if specified, otherwise use default 5m
-			if (cacheTtl != null && !cacheTtl.equals("5m")) {
-				cacheControl = new ChatCompletionRequest.CacheControl("ephemeral", cacheTtl);
-				logger.debug("Created cache control with TTL: type={}, ttl={}", "ephemeral", cacheTtl);
-			}
-			else {
-				cacheControl = new ChatCompletionRequest.CacheControl("ephemeral");
-				logger.debug("Created cache control with default TTL: type={}, ttl={}", "ephemeral", "5m");
-			}
-		}
-
-		// Build messages WITHOUT blanket cache control - strategic placement only
-		List<AnthropicMessage> userMessages = buildMessages(prompt, strategy, cacheControl, breakpointsUsed);
+		CacheEligibilityResolver cacheEligibilityResolver = CacheEligibilityResolver.from(cacheOptions);
 
 		// Process system - as array if caching, string otherwise
-		Object systemContent = buildSystemContent(prompt, strategy, cacheControl, breakpointsUsed);
+		Object systemContent = buildSystemContent(prompt, cacheEligibilityResolver);
+
+		// Build messages WITHOUT blanket cache control - strategic placement only
+		List<AnthropicMessage> userMessages = buildMessages(prompt, cacheEligibilityResolver);
 
 		// Build base request
 		ChatCompletionRequest request = new ChatCompletionRequest(this.defaultOptions.getModel(), userMessages,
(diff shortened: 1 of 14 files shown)


adase11:
I'm aware of https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#mixing-different-ttls where

> You can use both 1-hour and 5-minute cache controls in the same request, but with an important constraint: Cache entries with longer TTL must appear before shorter TTLs (i.e., a 1-hour cache entry must appear before any 5-minute cache entries).

I din't make any attempt to enforce that as, in my opinion, it's up to the user to configure that properly. (And also not worth the complexity). I probably could have mentioned that in the documentation though.

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -634,30 +619,33 @@ private List<AnthropicMessage> buildMessages(Prompt prompt, AnthropicCacheStrate
 							var source = getSourceByMedia(media);
 							return new ContentBlock(contentBlockType, source);
```
adase11, line 630: Technically these can be cached too - maybe as a next step including these content blocks. https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#what-can-be-cached

sobychacko:
Thanks for the PR! We will start reviewing it soon.

sobychacko:
@adase11 Can you add your name as an author to the classes you changed?

I have a couple of questions regarding the `messageTypeMinContentLengths` in `AnthropicCacheOptions`.  Why are we defaulting to the min content length of 1? Aren't the model requirements much higher for minimum tokens? 1024 and 2048? Also, content length function as it stands, does not map 1-to-1 with the number of tokens. I know the function is flexible so that more sophisticated token count mechanisms can be injected, but still I wonder if using the basic content length is the way to go there? But more importantly, my concern is with the global min requirements in the map. Since we default to `1` and since its unlikely users may override that, we may end up with a situation where we try to cache contents with shorter length than stipulated by the models. You have a check for `length < this.messageTypeMinContentLengths.get(messageType)` in `CacheEligibilityResolver` in which you return null if true. Imagine a content with 200 length. That check will fail because the content is longer than the minimum default of 1, which then ends up being cached and thus wasting the break point, right? Please correct me if I am wrong. 

Another general question, how does token count requirements work in the case of tool segments?

Thanks!

adase11:
@sobychacko Thanks! I'll add my name as the author to the classes I changed. With regard to your questions:

> Why are we defaulting to the min content length of 1? Aren't the model requirements much higher for minimum tokens? 1024 and 2048? Also, content length function as it stands, does not map 1-to-1 with the number of tokens.

I didn't want to be out-of-the-gate prescriptive to users about a minimum, so rather than guessing at a default that will continue to be valid for Anthropic's API spec I wanted to leave that up to the user to decide. And plus, like you said `String` text length doesn't map one-to-one with token length, so rather than guess at the right minimum string content I thought it was easier to just allow everything by default and then let users tweak their configurations to fit their specific use case. 

> Also, content length function as it stands, does not map 1-to-1 with the number of tokens. I know the function is flexible so that more sophisticated token count mechanisms can be injected, but still I wonder if using the basic content length is the way to go there?

Definitely understand what you're saying, I wanted to take the least intrusive / overhead approach as possible by default. My anticipation is that for a vast majority of the time a rough approximation will be sufficient. And then for the times when its not - if that means introducing a more sophisticated token approximation function that probably means much more overhead (the most robust way would be to actually introduce a tokenizer) than I would want to add to users without their explicit opt-in.

> Since we default to 1 and since its unlikely users may override that, we may end up with a situation where we try to cache contents with shorter length than stipulated by the models. You have a check for length < this.messageTypeMinContentLengths.get(messageType) in CacheEligibilityResolver in which you return null if true. Imagine a content with 200 length. That check will fail because the content is longer than the minimum default of 1, which then ends up being cached and thus wasting the break point, right? Please correct me if I am wrong.

That's correct, and I believe it reflects the current behavior of the prompt caching (where users can optimize a bit using the different strategies). The goal of my PR is partly to address this behavior in order to give users more control over what is attempted to be cached (optimizing their allotted cache blocks). However, my intent is to let users opt into this optimization rather than enforcing it by default, since user contexts vary widely. Any default I set risks being a poor fit—or worse, masking inefficiencies until usage scales (e.g., small initial workloads seem fine under defaults, but performance degrades as volume grows).

I chose 1 instead of 0 because the Anthropic API doesn’t allow caching of empty text blocks ([docs](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#what-cannot-be-cached)).

> Another general question, how does token count requirements work in the case of tool segments? 

For Tool definitions I took the easy way out, I think I should have documented better but I left the comment in there that addresses it `Tool definition messages are always considered for caching if the strategy includes system messages.`.

Let me know what you think and I'm happy to make any changes. I don't feel exceptionally strongly about the default min to `1` vs something like `1000` so i'm happy to change that and adjust documentation if you like.

adase11:
@sobychacko - I went ahead and added the author tag as well as updated the documentation to talk about how the tool definitions are handled.

sobychacko:
@adase11 Sounds good to me. I will let @markpollack take a look before we can proceed with the PR.

adase11:
Thanks!

spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/anthropic-chat.adoc
```diff
@@ -318,14 +326,45 @@ ChatResponse response = chatModel.call(
         List.of(new SystemMessage(largeSystemPrompt)),
         AnthropicChatOptions.builder()
             .model("claude-sonnet-4")
-            .cacheStrategy(AnthropicCacheStrategy.SYSTEM_ONLY)
-            .cacheTtl("1h")  // 1-hour cache lifetime
+            .cacheOptions(AnthropicCacheOptions.builder()
+                .strategy(AnthropicCacheStrategy.SYSTEM_ONLY)
+                .messageTypeTtls(MessageType.SYSTEM, AnthropicCacheTtl.ONE_HOUR)
```
markpollack: i think just 'ttl' as a builder name is cleaner

adase11: sounds good

adase11: Done

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -601,30 +599,18 @@ private List<AnthropicMessage> buildMessages(Prompt prompt, AnthropicCacheStrate
...
+				String content = message.getText();
+				boolean isLastUserMessage = lastUserIndex == i;
+				ContentBlock contentBlock = new ContentBlock(content);
+				if (isLastUserMessage) {
+					// Never cache the latest user message
+					contentBlocks.add(contentBlock);
 				}
 				else {
-					contents.add(new ContentBlock(message.getText()));
+					contentBlocks.add(cacheAwareContentBlock(contentBlock, messageType, cacheEligibilityResolver));
```
sobychacko: @adase11 Doesn't this make all the user messages in the request get cached? The first if conditional checks if it's the last user message and then skips adding the cache, and in the else, you just add it to the cache as long as it's a user message (based on the outer if condition). Isn't that wrong though? I think we only need to add the cache control to the last user message, right?

adase11: You are correct, thanks

sobychacko: Do you mean it is not correct right now as it stands in this PR?

adase11: Yes, it's currently incorrect, I should have only attempted to cache the last user message. And I can check that content size based on the last 20 user messages https://docs.claude.com/en/docs/build-with-claude/prompt-caching#continuing-a-multi-turn-conversation

sobychacko: Ok. Are you going to update the PR?

adase11: Yep I'll make the change, looks like previously it was only the second to last user message - do you know if there was a particular reason for that?

sobychacko: I think the reasoning was that if there was only one user message, then we won't cache it. I was also going to check with @markpollack on why that was the case. But i think as long as it is the last user message, then we can go ahead and add the cache control to it.

adase11: Thanks, @sobychacko appreciate it

adase11: Yeah I see 


> Cache the entire conversation history up to (but not including) the current user question. This is ideal for multi-turn conversations where you want to reuse the conversation context while asking new questions.

on the documentation on `AnthropicCacheStrategy.CONVERSATION_HISTORY` - I think from the Anthropic docs that we could be caching including the last user message. I'll make an update to keep with the old logic - only apply to the next to last user message - and then if we're in agreement that actually the last one is ok to make eligible for caching that's an easy change to make.

adase11: @sobychacko - updated, and added some ITs in `AnthropicPromptCachingIT` for better coverage. If we do decide that all user messages are eligible including the last one that would make things a little simpler. But for now I left that logic the same and just considered the eligible (according to Anthropic) content when lookin at the content size (i.e. the prior 20 content blocks https://docs.claude.com/en/docs/build-with-claude/prompt-caching#when-to-use-multiple-breakpoints).

adase11: I could have gotten more sophisticated and attempted to add additional breakpoints if there are > 20 messages in the conversation history. I chose not to consider that for the moment in order to keep things simple but if you prefer me to do that I can.

sobychacko: Let's review your changes with Mark first and we can proceed from there. Thanks!

adase11: sounds good

markpollack:
thanks!  it is merged and will be in 1.1 M2.  I do want to experiement a bit more to fine tune.

merged in 1d5ab9b67558790ba914eb3ecfefd47581631956

sobychacko:
@adase11 Quick follow up question. Have you encountered any issues with the 20 block limitations that Anthropic places on automatic prefix scanning for cache hits? I am curious to find out how you were able to address some edge case use cases I am thinking of. Here is a hypothetical scenario: Imagine you have a conversation history with ~1000 turns as content blocks (assuming that this request fits within Anthropic's context window). Now, if we put only one `cache_control` at the end, then that would definitely be a problem since it can only go back to block ~980 because of the 20 block limitation in auto scanning. This means that blocks from 1-980 will be recomputed for cache and incur new cache writes, cost etc. Even we we put 4 `cache_control` entries throughout the messages, still a majority of the messages will be missed (again due to the only 20 blocks going back situation). One solution, I found in my reading of this topic, is that we could summarize multiple initial block ranges and re-add them. For example blocks 1-200 can be summarized and put them into 20 blocks and then place a `cache_control` after that. It is rarely that those initial blocks will be changed, so it might be safe to summarize them. Did you run into any such situations with larger messages? Basically, I am trying to understand how the 20-block limitation plays any role in the real-world examples. Thanks!

adase11:
> @adase11 Quick follow up question. Have you encountered any issues with the 20 block limitations that Anthropic places on automatic prefix scanning for cache hits? I am curious to find out how you were able to address some edge case use cases I am thinking of. Here is a hypothetical scenario: Imagine you have a conversation history with ~1000 turns as content blocks (assuming that this request fits within Anthropic's context window). Now, if we put only one `cache_control` at the end, then that would definitely be a problem since it can only go back to block ~980 because of the 20 block limitation in auto scanning. This means that blocks from 1-980 will be recomputed for cache and incur new cache writes, cost etc. Even we we put 4 `cache_control` entries throughout the messages, still a majority of the messages will be missed (again due to the only 20 blocks going back situation). One solution, I found in my reading of this topic, is that we could summarize multiple initial block ranges and re-add them. For example blocks 1-200 can be summarized and put them into 20 blocks and then place a `cache_control` after that. It is rarely that those initial blocks will be changed, so it might be safe to summarize them. Did you run into any such situations with larger messages? Basically, I am trying to understand how the 20-block limitation plays any role in the real-world examples. Thanks! 

Apologies for the delay in getting back to you here @sobychacko  - I have not forgotten about your question.

adase11:
@sobychacko In my experience the choice on how to handle older messages in the conversation history - as the history gets close to the context window limit or maybe as it overflows the cache block limit as well - typically comes down to a business/product decision. I've seen typically either the summarization approach (for example claude code or codex will do this for you if you get into a big conversation) or just dropping old messages as well. But these decisions have impact on the actual product behaves and so it's something that I think that the engineering team can't decide what to do without getting buy in from product stakeholders and probably trying out various approaches (some approaches are purely product driven like coaching the user to just start a brand new conversation).  

One clarification to the situation that you outlined though, if I read it right I believe in that situation blocks 1-980 would just be ignored and 981 - 1000 would be eligible for caching (either read or write). 

You're correct that as the conversation grows there's less "room" for caching available, that's part of the reason I wanted to make sure we could apply something like the cache eligibility function - in order to maximize our cache hit rate. Working with this implementation of caching in production we've seen great success for the usage we have (users having conversations that require large system messages and lost of tool call requests and responses that produce lots of data) our cache hit rate maintains above 60% most of the time which has dramatically improved our ability to stay within Anthropic's TPM limits and also improves effectiveness.

PR #4350, opened 2025-09-09
feat: Make advisors observations configurable

ThomasVitale:
* Add possibility to provide a custom AdvisorObservationConvention to the ChatClient for customizing the conventions for advisor spans and metrics.
* Add ChatClientResponse to ChatClientObservationContext for achiving full visibility into both request and response.

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
@@ -70,7 +70,7 @@ public interface ChatClient {
 	 * @deprecated in favor of
 	 * {@link #create(ChatModel, ObservationRegistry, ChatClientObservationConvention, AdvisorObservationConvention)}.
 	 */
-	@Deprecated
+	@Deprecated(since = "1.1.0", forRemoval = true)
 	static ChatClient create(ChatModel chatModel, ObservationRegistry observationRegistry,
 			@Nullable ChatClientObservationConvention chatClientObservationConvention) {
 		return create(chatModel, observationRegistry, chatClientObservationConvention, null);
@@ -93,7 +93,7 @@ public interface ChatClient {
 	 * @deprecated in favor of
 	 * {@link #builder(ChatModel, ObservationRegistry, ChatClientObservationConvention, AdvisorObservationConvention)}.
 	 */
-	@Deprecated
+	@Deprecated(since = "1.1.0", forRemoval = true)
 	static Builder builder(ChatModel chatModel, ObservationRegistry observationRegistry,
 			@Nullable ChatClientObservationConvention chatClientObservationConvention) {
 		return builder(chatModel, observationRegistry, chatClientObservationConvention, null);
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
@@ -662,7 +662,7 @@ public class DefaultChatClient implements ChatClient {
 		/**
 		 * @deprecated in favor of the other constructor.
 		 */
-		@Deprecated
+		@Deprecated(since = "1.1.0", forRemoval = true)
 		public DefaultChatClientRequestSpec(ChatModel chatModel, @Nullable String userText,
 				Map<String, Object> userParams, Map<String, Object> userMetadata, @Nullable String systemText,
 				Map<String, Object> systemParams, Map<String, Object> systemMetadata, List<ToolCallback> toolCallbacks,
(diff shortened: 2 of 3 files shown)


markpollack:
@jonatan-ivanov could you review this?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
```diff
@@ -65,22 +66,46 @@ static ChatClient create(ChatModel chatModel, ObservationRegistry observationReg
 		return create(chatModel, observationRegistry, null);
 	}
 
+	/**
+	 * @deprecated in favor of
+	 * {@link #create(ChatModel, ObservationRegistry, ChatClientObservationConvention, AdvisorObservationConvention)}.
+	 */
+	@Deprecated
```
jonatan-ivanov: Should we mark this for deletion in 2.0 as well as the other deprecated components in this PR?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
```diff
@@ -78,6 +82,15 @@ public boolean isStream() {
...
 
+	@Nullable
+	public ChatClientResponse getChatClientResponse() {
+		return this.chatClientResponse;
+	}
+
+	public void setChatClientResponse(@Nullable ChatClientResponse chatClientResponse) {
+		this.chatClientResponse = chatClientResponse;
+	}
+
```
jonatan-ivanov, line 93: This will not work in the streaming (Reactor) scenario.
Maybe we should fix that first in a separate PR (along with the logging use-case), I have a PoC for that: https://github.com/spring-projects/spring-ai/compare/main...jonatan-ivanov:spring-ai:promt-on-span

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
ThomasVitale: That'd be fine for me. Should I add the markedForDeletion flag to the annotation?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
ThomasVitale: Both the ChatClient and Advisor observation contexts don't get the response populated at the moment, for the reactive scenarios.

I didn't to that here since I don't feel comfortable making changes via the reactor context propagation.

So it would be great if it could be added in a separate PR, thank you!

ThomasVitale: The ChatClientResponse was already part of the AdvisorObservationContext, but it was never populated. In this PR, I made a change to populate it in non-streaming scenarios.

And for consistency, I've done the same for ChatClientObservationContext.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
jonatan-ivanov: I think so, @markpollack?

markpollack: sure, we can do that.

ThomasVitale: Done

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
jonatan-ivanov: What do you think about this? https://github.com/spring-projects/spring-ai/pull/4417
The interesting part is small, see: `DefaultChatClient` and `ChatClientObservationContext`. This PR could depend on this and we can also re-use this in the prompt/completion on spans scenario.

ThomasVitale: Thanks for sharing! I checked the PR and looks good to me. The only comment I left is about the `ChatClientObservationContext`: https://github.com/spring-projects/spring-ai/pull/4417#discussion_r2372825116

ThomasVitale: If you agree with that comment, I would suggest having that PR build on this one, adding support for the reactive scenario and completion logs.

jonatan-ivanov:
It seems neither @sobychacko nor I can push to this PR (is edit by maintainers off?) so I rebased it on `main` and added reactive/stream support here: https://github.com/spring-projects/spring-ai/pull/4769

jonatan-ivanov:
Superseded by https://github.com/spring-projects/spring-ai/pull/4769

PR #4351, opened 2025-09-09
feat: Add builder for AssistantMessage

ThomasVitale:
In alignment with SystemMessage and UserMessage, introduced a builder for AssistantMessage to simplify instantiation considering the many arguments currently needed by the constructors.

Introduced missing tests for AssistantMessage.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AssistantMessage.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AssistantMessage.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2024 the original author or authors.
+ * Copyright 2023-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


sunyuhan1998 (approved):

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AssistantMessage.java
```diff
@@ -45,14 +46,26 @@ public AssistantMessage(String content) {
 		this(content, Map.of());
```
ilayaperumalg, line 46: @ThomasVitale What do you think about deprecating this public usage as well - in favour of the builder? This way, new instance of AssistantMessage can only be created via builder only.

ilayaperumalg:
@ThomasVitale Thanks for the PR!

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AssistantMessage.java
ThomasVitale: That would be fine by me. I know that for `UserMessage` and `SystemMessage` we opted for keeping the 1-argument constructor for convenience, and forcing the `Builder` for any other use cases where more than one argument needs to be provided. What do you think?

ilayaperumalg: I see. That's a good point. I like the convenience aspect as well. We can keep the single argument constructor for AssistantMessage as well.

ilayaperumalg:
Rebased and merged as f4a1c96bd

ThomasVitale:
@ilayaperumalg thank you!

PR #4364, opened 2025-09-10
Use Ollama system role for system message

nicolaskrier:
- Use Ollama's system role for system messages (instead of user role).
- Add unit test to validate Spring AI to Ollama message conversion for all message types.


nicolaskrier:
Hi @ilayaperumalg, could you check this PR please? It is relative to [this commit](https://github.com/spring-projects/spring-ai/commit/8968df7a77bd67ea05112e831f7c088e4ddf7e1f).

sunyuhan1998:
You're right, that was my oversight in that PR. Thank you very much for the correction!

ilayaperumalg:
@nicolaskrier Thanks for the quick fix and sorry about missing that during the merge. Will get to this shortly.

PR #4366, opened 2025-09-11
Change hsqldb scope from runtime to test in spring-ai-autoconfigure-model-chat-memory-repository-jdbc

making:
`spring-ai-autoconfigure-model-chat-memory-repository-jdbc` has a runtime-scope dependency on hsqldb, so when I use `spring-ai-starter-model-chat-memory-repository-jdbc` it pulls in hsqldb, which is not normally needed.

This pull request changes hsqldb scope from runtime to test.


ilayaperumalg:
@making Thanks for the PR fixing this.

PR #4369, opened 2025-09-11
refactor: Improve MCP tool name prefix generation with automatic duplicate handling

tzolov:
- Introduce DefaultMcpToolNamePrefixGenerator to ensure unique tool names across MCP connections
- Replace default prefix generation strategy from client/title-based to duplicate detection
- Automatically add counter prefixes (dp_1_, dp_2_) when duplicate tool names are detected
- Track existing connections and tool names to maintain idempotency
- Update tests to reflect new duplicate handling behavior (keep existing tool on duplicate)
- Update documentation to describe the new default generator behavior
- Change default from McpToolNamePrefixGenerator.defaultGenerator() to new DefaultMcpToolNamePrefixGenerator

BREAKING CHANGE: The default tool name generation strategy has changed. Tools now use their original names by default, with automatic prefixing only when duplicates are detected across different connections.

### DIFF
--- a/mcp/common/src/main/java/org/springframework/ai/mcp/DefaultMcpToolNamePrefixGenerator.java
+++ b/mcp/common/src/main/java/org/springframework/ai/mcp/DefaultMcpToolNamePrefixGenerator.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2025-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.mcp;
+
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import io.modelcontextprotocol.spec.McpSchema;
+import io.modelcontextprotocol.spec.McpSchema.Implementation;
+import io.modelcontextprotocol.spec.McpSchema.Tool;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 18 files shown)


mcp/common/src/main/java/org/springframework/ai/mcp/McpToolNamePrefixGenerator.java
```diff
@@ -45,10 +45,11 @@ public interface McpToolNamePrefixGenerator {
 	 * name.
 	 * @return the default prefix generator
 	 */
-	static McpToolNamePrefixGenerator defaultGenerator() {
-		return (mcpConnectionIfo, tool) -> McpToolUtils.prefixedToolName(mcpConnectionIfo.clientInfo().name(),
-				mcpConnectionIfo.clientInfo().title(), tool.name());
-	}
+	// static McpToolNamePrefixGenerator defaultGenerator() {
```
ilayaperumalg, line 48: will remove when merging.

mcp/common/src/main/java/org/springframework/ai/mcp/DefaultMcpToolNamePrefixGenerator.java
```diff
@@ -0,0 +1,81 @@
...
+	@Override
+	public String prefixedToolName(McpConnectionInfo mcpConnectionInfo, McpSchema.Tool tool) {
+
+		String uniqueToolName = McpToolUtils.format(tool.name());
+
+		if (this.existingConnections
+			.add(new ConnectionId(mcpConnectionInfo.clientInfo(), (mcpConnectionInfo.initializeResult() != null)
+					? mcpConnectionInfo.initializeResult().serverInfo() : null, tool))) {
+			if (!this.allUsedToolNames.add(uniqueToolName)) {
+				uniqueToolName = "dp_" + this.counter.getAndIncrement() + "_" + uniqueToolName;
```
ilayaperumalg, line 69: While one can provide a custom prefix generator, It would be good if we can make the prefix "dp_" customisable.

ilayaperumalg: perhaps "sa" to denote "Spring AI" as a prefix here instead of "dp_"?

ilayaperumalg: After discussion, changing this to use "alt_" when merging.

ilayaperumalg:
Rebased, updated the prefix for the duplicate tool name to use "alt_", merged as 31db86d56

PR #4376, opened 2025-09-12
perf: use AssistantMessage#builder()

leeyazhou:
perf: use AssistantMessage#builder()


ilayaperumalg:
@leeyazhou Thanks for the PR! Rebased and merged as 1decd348f

PR #4378, opened 2025-09-12
Polish McpServerAutoConfiguration

quaff:
1. Rename method parameter `rootsChangeConsumers` to align `mcpSyncServer` with `mcpAsyncServer`.
2. Simplify lambda.

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerAutoConfiguration.java
@@ -112,7 +112,7 @@ public class McpServerAutoConfiguration {
 			ObjectProvider<List<SyncResourceTemplateSpecification>> resourceTemplates,
 			ObjectProvider<List<SyncPromptSpecification>> prompts,
 			ObjectProvider<List<SyncCompletionSpecification>> completions,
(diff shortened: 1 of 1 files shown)


auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerAutoConfiguration.java
```diff
@@ -179,10 +179,8 @@ public McpSyncServer mcpSyncServer(McpServerTransportProviderBase transportProvi
 			}
 		}
 
-		rootsChangeConsumers.ifAvailable(consumer -> {
-			BiConsumer<McpSyncServerExchange, List<McpSchema.Root>> syncConsumer = (exchange, roots) -> consumer
-				.accept(exchange, roots);
-			serverBuilder.rootsChangeHandler(syncConsumer);
+		rootsChangeConsumer.ifAvailable(consumer -> {
+			serverBuilder.rootsChangeHandler(consumer);
```
ilayaperumalg, line 202: @quaff Thanks for the PR. I believe this is safer to do as the consumer type would always be BiConsumer<McpSyncServerExchange, List<McpSchema.Root>>. But, I want to double check with why it was done this way at first place. @tzolov what's your thoughts on this?

quaff: I think it's copied from `mcpAsyncServer` method which cannot use it directly.

Issue #4390, opened 2025-09-15
Make ToolCallResultConverter a spring bean


SagenKoder:
I did not find any request for this, if its duplicate, let me know.

**Expected Behavior**

We rely on our very custom ObjectMapper in our big monolith. We expect to simply create a new bean that implements ToolCallResultConverter for it to override the default implementation. Today this can be overridden per @Tool annotation but not globally.

Alternative:
Having the default implementation as a part of a builder would also solve the same problem for us. Not sure what builder it should be added to in this case. This could allow to have different implementations per ChatClient for example.

**Current Behavior**

Currently its very hardcoded in `org.springframework.ai.tool.support.ToolUtils#getToolCallResultConverter`. If an @Tool annotation is present it will use its tool result converter, and if not present it will fallback to a hardcoded `org.springframework.ai.tool.execution.DefaultToolCallResultConverter`.

**Context**

We have a huge monolith with very custom frameworks. A big part of this is our API framework where we do auth filtering in the api serialization through Jackson filter providers.

To get started with Spring AI quickly and safely we will reuse the same framework to avoid leaking unauthorized information to the LLM or end user. Currently we get around this by having tests to ensure that all @Tool annotations setup our own ToolCallResultConverter implementation.

MartinHaeusler:
I just ran into this as well. I was wondering why I got a JSON error during the tool result processing and I discovered that Spring AI is *not* using the global Jackson instance, but creates a new one. Not being able to set this globally is a big issue, I don't want to explicitly state the tool response converter in each `@Tool` annotation.

Indaplace:
The `ToolCallResultConverter` may be injectable with the same mechanism as `ConstraintValidator` with [LocalValidatorFactoryBean](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/validation/beanvalidation/LocalValidatorFactoryBean.html).

PR #4404, opened 2025-09-16
Make default MethodToolCallbackProvider configurable

SagenKoder:
This PR is an attempt at fixing #4390

1. MethodToolCallbackProvider.Builder (spring-ai-model)

- Added toolCallResultConverter field
- Added toolCallResultConverter(ToolCallResultConverter) method
- Updated constructor to accept converter parameter
- Modified getToolCallbacks() to use custom converter when provided, fallback to ToolUtils.getToolCallResultConverter() otherwise

2. ToolCallbacks.from() (spring-ai-model)

- Added overloaded method from(ToolCallResultConverter, Object...)
- Original from(Object...) now delegates to new method with null converter

3. ChatClient.Builder Interface (spring-ai-client-chat)

- Added defaultToolCallResultConverter(ToolCallResultConverter) method

4. DefaultChatClientBuilder (spring-ai-client-chat)

- Implemented defaultToolCallResultConverter() method

5. DefaultChatClient (spring-ai-client-chat)

- Added toolCallResultConverter field to DefaultChatClientRequestSpec
- Updated constructors to accept converter parameter
- Modified tools() method to pass converter to ToolCallbacks.from()
- Added toolCallResultConverter() method to request spec
- Updated mutate() to preserve converter

6. Test Coverage (spring-ai-model)

- Added test whenCustomToolCallResultConverterIsProvidedThenUsesIt()
- Added ToolWithAnnotations test class

Usage

    // Option 1: Global via ChatClient
    ChatClient.builder(chatModel)
        .defaultToolCallResultConverter(customConverter)
        .build();

    // Option 2: Per provider
    MethodToolCallbackProvider.builder()
        .toolObjects(tools)
        .toolCallResultConverter(customConverter)
        .build();
        
Feel free to give me feedback on this change.

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
@@ -307,6 +307,22 @@ public interface ChatClient {
 
 		Builder defaultToolContext(Map<String, Object> toolContext);
 
+		/**
+		 * Sets a default
+		 * {@link org.springframework.ai.tool.execution.ToolCallResultConverter} to be
+		 * used for all tools created via the {@code tools()} method.
+		 * <p>
+		 * This converter will be used to serialize tool results before sending them back
+		 * to the AI model. Tools created with {@code toolCallbacks()} must have their
+		 * converter set at callback creation time.
+		 * </p>
+		 * @param converter the converter to use for tool results, or null to use the
+		 * default
+		 * @return this builder
+		 * @since 1.0.0
+		 */
+		Builder defaultToolCallResultConverter(org.springframework.ai.tool.execution.ToolCallResultConverter converter);
+
 		Builder clone();
 
 		ChatClient build();
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
@@ -57,6 +57,7 @@ import org.springframework.ai.template.TemplateRenderer;
 import org.springframework.ai.template.st.StTemplateRenderer;
 import org.springframework.ai.tool.ToolCallback;
 import org.springframework.ai.tool.ToolCallbackProvider;
+import org.springframework.ai.tool.execution.ToolCallResultConverter;
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.core.io.Resource;
 import org.springframework.lang.Nullable;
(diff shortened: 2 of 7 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
```diff
@@ -307,6 +307,22 @@ interface Builder {
...
+		 * This converter will be used to serialize tool results before sending them back
+		 * to the AI model. Tools created with {@code toolCallbacks()} must have their
+		 * converter set at callback creation time.
+		 * </p>
+		 * @param converter the converter to use for tool results, or null to use the
+		 * default
+		 * @return this builder
+		 * @since 1.0.0
+		 */
+		Builder defaultToolCallResultConverter(org.springframework.ai.tool.execution.ToolCallResultConverter converter);
```
ilayaperumalg, line 324: Let's not expose the tool call result converter at the ChatClient level as the tool call result converter is applicable only for method and function tools.

ThomasVitale: @SagenKoder thanks for your PR!

The `ToolCallResultConverter` was designed for the specific `MethodToolCallback` implementation of `ToolCallback`. Also, it was designed to be customisable for each tool via the `@Tool` annotation (hence, it doesn't rely on DI since the tool method can be defined in any Java class, even if it's not a Spring bean). I agree with @ilayaperumalg that we shouldn't raise it to a `ChatClient`-level abstraction.

Considering the specific problem described in https://github.com/spring-projects/spring-ai/issues/4390, I would suggest evaluating whether providing a custom `ToolCallback` implementation would be a feasible solution for your specific case. That would give you the flexibility to process tools re-using your existing logic in your monolithic application.

I acknowledge that the current solution has its limits. But before changing the nature of `ToolCallResultConverter`, we should probably consider the broader set of requirements that involve parsing a tool call result to make sure the new API is sound.

SagenKoder: Hi, thank you for very valuable feedback.

Looking at it now I do agree with your takes. Making ToolCallback customizable would be a more generic solution.

I'll look more into better solutions tomorrow to see if we can find something more flexible and "correct".

ilayaperumalg:
@SagenKoder I am closing this PR for now as customising the ToolCallback would probably the appropriate way to achieve your use case. Please feel free to re-open if you still think, having the configurable option for the MethodToolCallbckProvider would be beneficial. Thanks!

PR #4417, opened 2025-09-18
Log chat client completion data and make the response text available in the Observation Context

jonatan-ivanov:
The main driver of this change to make the response test available in the Chat Client Observation Context in `call()` and `stream()` scenarios.
This also introduces chat client log completion which was missing before mainly because the response text was not available in the `stream()` use-case.

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
@@ -17,12 +17,16 @@
 package org.springframework.ai.chat.client.observation;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 import io.micrometer.observation.Observation;
 
 import org.springframework.ai.chat.client.ChatClientAttributes;
 import org.springframework.ai.chat.client.ChatClientRequest;
+import org.springframework.ai.chat.client.ChatClientResponse;
 import org.springframework.ai.chat.client.advisor.api.Advisor;
+import org.springframework.ai.chat.messages.Message;
+import org.springframework.ai.chat.model.Generation;
 import org.springframework.ai.observation.AiOperationMetadata;
 import org.springframework.ai.observation.conventions.AiOperationType;
 import org.springframework.ai.observation.conventions.AiProvider;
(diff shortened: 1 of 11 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/observation/ChatClientObservationContext.java
```diff
@@ -35,12 +39,16 @@
...
  * @author Thomas Vitale
+ * @author Jonatan Ivanov
  * @since 1.0.0
  */
 public class ChatClientObservationContext extends Observation.Context {
 
 	private final ChatClientRequest request;
 
+	@Nullable
+	private String responseText;
```
ThomasVitale: How about using the `ChatClientResponse` instead of a `String`, like I did in https://github.com/spring-projects/spring-ai/pull/4350/files#diff-23be91ee783d94e4bd9a0031f7728a273bb72f1eb34945b99917932d25533aa2R53? That would be consistent with how the Advisors are instrumented (`ChatClientRequest`/`ChatClientResponse` pair) and would allow to observe useful output besides the response text.

I understand the reactive scenario might only deal with a simple `String`, but then at least we get full observability for non-reactive scenarios. And for reactive scenarios, we can provide utility methods for accepting a String while still storing a `ChatClientResponse` in the context.

jonatan-ivanov: I was thinking about it, the problem with `ChatClientResponse` is that we need to merge/aggregate them somehow in the reactive scenarios and I was not sure if it is possible. Also, if it is, what other issues we can bump into. If someone could give me some guidance, I'm open to try it.
/cc @tzolov

jonatan-ivanov: @chemicL suggested using `ChatClientMessageAggregator` which seems to be solving this, I updated the PR.

ThomasVitale: That's great!

ThomasVitale (approved):

sobychacko:
Merged upstream via https://github.com/spring-projects/spring-ai/commit/13de2192c3fe42a2af507396c0f4f0d3fcaaee67.

PR #4418, opened 2025-09-18
test: Add runtime hints validation for Ollama reflection registration

alxkm:
This PR adds tests to validate that Ollama runtime hints are properly configured.

- Hint Scope Validation: Ensures only necessary hint types are registered (excludes proxy/serialization hints)
- API Type Coverage: Confirms essential API classes like responses and tools are registered
- Native Image Compatibility: Validates hint configuration meets GraalVM compilation requirements

These tests prevent runtime failures in GraalVM native images by ensuring:

- Required classes can be accessed via reflection
- No unnecessary hints that bloat the native image
- Critical functionality like parsing and tool calling works correctly

The tests catch configuration issues early rather than discovering them during native compilation or at runtime.


ilayaperumalg:
@alxkm Thanks for the PR improving the test coverage.

PR #4420, opened 2025-09-19
refactor: Refactored the `initializeContainer` method of `CosmosDBVectorStore`

sunyuhan1998:
The `initializeContainer` method of `CosmosDBVectorStore` has some issues:  
1. Most of the parameters in the method signature are not used, and these parameters don't need to be passed as arguments since they have already been assigned to instance properties in the constructor and can be directly accessed.  
2. The reassignment of parameters within the method is meaningless.  

### This PR makes the following changes:  
1. Removed all parameters from the method signature, as they are already accessible via instance properties.  
2. Eliminated meaningless reassignments of method parameters.


ilayaperumalg:
@sunyuhan1998 Thanks for the PR cleaning up the code. LGTM, merging.

ilayaperumalg:
Rebased and merged as 84efb6a63. Also, cherry-picked into 1.0.x as f4a2aa9ef

PR #4453, opened 2025-09-23
GH-4452 Added Builder pattern support for `DeepSeekAssistantMessage`

### LINKED ISSUE #4452: `DeepSeekAssistantMessage.builder()` returned an unexpected type
**Bug description**
Since we recently added support for the Builder pattern to `AssistantMessage`, this causes classes like `DeepSeekAssistantMessage`—which inherit from `AssistantMessage`—to also inherit the builder of `AssistantMessage`, which is clearly problematic.

**Steps to reproduce**
Try using `DeepSeekAssistantMessage.builder().build()`, and you'll get an `AssistantMessage` instead of a `DeepSeekAssistantMessage`.

**Expected behavior**
`DeepSeekAssistantMessage.builder().build()` should return a `DeepSeekAssistantMessage`.

sunyuhan1998:
As mentioned in the issue, `DeepSeekAssistantMessage` also requires comprehensive Builder pattern support. 

This PR follows the approach proposed in #3529 and includes the following changes:  
1. Adjusted the visibility of some fields and methods in the `AssistantMessage` Builder to support inheritance in the `DeepSeekAssistantMessage` Builder.  
2. Added a Builder for `DeepSeekAssistantMessage`.  
3. Added relevant unit tests.
4. Fixed the bug in the `toString` method

Fixes #4452

### DIFF
--- a/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekAssistantMessage.java
+++ b/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekAssistantMessage.java
@@ -119,35 +119,35 @@ public class DeepSeekAssistantMessage extends AssistantMessage {
 				+ this.prefix + ", metadata=" + this.metadata + "]";
 	}
 
-	public static Builder builder() {
-		return new Builder();
-	}
+	public static final class Builder {
 
-	public static final class Builder extends AssistantMessage.Builder {
+		private String content;
+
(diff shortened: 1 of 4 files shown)


sunyuhan1998:
Hi @ericbottard , I've made some adjustments in the current PR, implementing the solution we finalized in #4457. Could you please help review it again? Thanks so much!

ericbottard:
Merged on main via 7d0aff063589c47380843e0d8bdeac31bb6a2bc5

PR #4457, opened 2025-09-24
GH-4454 Added Builder pattern support for `ZhiPuAiAssistantMessage`

### LINKED ISSUE #4454: `ZhiPuAiAssistantMessage.builder()` returned an unexpected type
This is an issue similar to #4452.

**Bug description**
Since we recently added support for the Builder pattern to `AssistantMessage`, this causes classes like `ZhiPuAiAssistantMessage`—which inherit from `AssistantMessage`—to also inherit the builder of `AssistantMessage`, which is clearly problematic.

**Steps to reproduce**
Try using `ZhiPuAiAssistantMessage.builder().build()`, and you'll get an `AssistantMessage` instead of a `ZhiPuAiAssistantMessage`.

**Expected behavior**
`ZhiPuAiAssistantMessage.builder().build()` should return a `ZhiPuAiAssistantMessage`.

sunyuhan1998:
As mentioned in the issue, `ZhiPuAiAssistantMessage` also requires comprehensive Builder pattern support. 

This PR follows the approach proposed in #3529 and includes the following changes:  
1. Adjusted the visibility of some fields and methods in the `AssistantMessage` Builder to support inheritance in the `ZhiPuAiAssistantMessage` Builder.  
2. Added a Builder for `ZhiPuAiAssistantMessage`.  
3. Added relevant unit tests.

Fixes #4454

### DIFF
--- a/models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/ZhiPuAiAssistantMessage.java
+++ b/models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/ZhiPuAiAssistantMessage.java
@@ -25,6 +25,7 @@ import org.springframework.ai.content.Media;
 
 /**
  * @author YunKui Lu
+ * @author Sun Yuhan
  */
 public class ZhiPuAiAssistantMessage extends AssistantMessage {
 
@@ -76,6 +77,50 @@ public class ZhiPuAiAssistantMessage extends AssistantMessage {
 		return "ZhiPuAiAssistantMessage{" + "media=" + this.media + ", messageType=" + this.messageType + ", metadata="
 				+ this.metadata + ", reasoningContent='" + this.reasoningContent + '\'' + ", textContent='"
 				+ this.textContent + '\'' + '}';
+	}
+
+	public static Builder builder() {
+		return new Builder();
+	}
+
+	public static final class Builder extends AssistantMessage.Builder {
+
+		private String reasoningContent;
+
+		@Override
+		public Builder content(String content) {
+			this.content = content;
+			return this;
+		}
+
+		@Override
+		public Builder properties(Map<String, Object> properties) {
+			this.properties = properties;
+			return this;
+		}
+
+		@Override
+		public Builder toolCalls(List<ToolCall> toolCalls) {
+			this.toolCalls = toolCalls;
+			return this;
+		}
+
+		@Override
+		public Builder media(List<Media> media) {
+			this.media = media;
+			return this;
+		}
+
+		public Builder reasoningContent(String reasoningContent) {
+			this.reasoningContent = reasoningContent;
+			return this;
+		}
+
+		public ZhiPuAiAssistantMessage build() {
+			return new ZhiPuAiAssistantMessage(this.content, this.reasoningContent, this.properties, this.toolCalls,
+					this.media);
+		}
+
 	}
 
 }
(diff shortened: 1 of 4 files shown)


models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/ZhiPuAiAssistantMessage.java
```diff
@@ -78,4 +79,48 @@ public String toString() {
 				+ this.textContent + '\'' + '}';
 	}
 
+	public static Builder builder() {
+		return new Builder();
+	}
+
+	public static final class Builder extends AssistantMessage.Builder {
```
ericbottard: After reviewing this PR, we agreed that Builders should not be made extensible, even for cases like this where there is a class hierarchy of things being constructed.

Could you then, instead of opening up `AssistantMessage.Builder`, create the `ZhiPuAiAssistantMessage.Builder` by copy/pasting all the fields from the superclass and adding the one of interest (`reasoningContent` here).

The same reasoning would apply to https://github.com/spring-projects/spring-ai/pull/4453

sunyuhan1998: > Could you then, instead of opening up `AssistantMessage.Builder`, create the `ZhiPuAiAssistantMessage.Builder` by copy/pasting all the fields from the superclass and adding the one of interest (`reasoningContent` here).
> 
> The same reasoning would apply to #4453

Thank you for your review! I fully understand your suggestion. In fact, I initially considered implementing it the way you proposed, but later ran into some issues. Currently, we obtain a Builder instance in `AssistantMessage` via the method `public static Builder builder()`. If we were to define a similar method in a subclass, we couldn't use the same signature `public static Builder builder()`—because it would exactly match the parent class's method signature, yet return a different (incompatible) Builder type specific to the subclass. 

While one workaround would be to change the method name in the subclass (e.g., `builderXxx()`), doing so would hurt API consistency, feel inelegant, and potentially confuse users. Therefore, my current approach is to let the subclass's Builder inherit from the parent's Builder, while appropriately exposing necessary fields or methods in the parent class to support extension.

A similar issue was previously encountered in #3529. For such scenarios involving inheritance combined with the Builder pattern, do you have any better design suggestions? I’d really appreciate your insights.

ericbottard: You're right about the conflict in method names, even if they are static.

Nevertheless, I think what we want to avoid at all costs is making the Builder extensible.
The `.builder()` methods are just convenience (and you're right in saying having them all exist and simply be named `.builder()` would be optimal) for constructing the builder. Nothing prevents users calling `new ZhiPuAiAssistantMessage.Builder()` (or even `new Builder()` with a static import).

I'll circle the compromise with the team at the first opportunity, but I think this is still the best solution (_i.e._ to clarify: everything in our previous comments, minus the dedicated `.builder()` method)

sunyuhan1998: No problem. Since the `Builder` pattern is heavily used in the current project, I believe properly coordinating the `Builder` classes between parent and child classes is an important matter, and I'm looking forward to the team proposing a better, unified solution. Additionally, I've followed your suggestion to temporarily remove the `.builder()` method and reverted all changes made to the parent class. For now, usage of the `Builder` in the child class has been changed to direct instantiation with `new`. Would this approach be acceptable? Could you please review it again? Once this PR is finalized, I'll apply the same solution to #4453. Thank you very much!

ericbottard: Hi, this seems like the best solution.

Thanks for taking the time to work on this (and #4453).

ericbottard:
Merged on main via 01082b8bf1972efdc82cff101fbca5418aba2552

Issue #3232, opened 2025-05-18
MPC Client fails application startup when server is unavailable - need graceful degradation option


astilt322:
When integrating Spring AI's MPC client in a Spring Boot application, if the MPC server is unavailable during client application startup, the entire application fails to start.

This creates an undesired hard dependency between client and server components, which is problematic in distributed systems. The client application should be able to start successfully even when the MPC server is temporarily unavailable, and should attempt to connect when the client is actually used and the server becomes available.

### Error Details
The following exception occurs during application startup when the MPC server is unavailable:

ava.util.concurrent.CompletionException: java.net.ConnectException
  at java.base/java.util.concurrent.CompletableFuture.encodeRelay(CompletableFuture.java:368) ~[na:na]
  at java.base/java.util.concurrent.CompletableFuture.completeRelay(CompletableFuture.java:377) ~[na:na]
  at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1152) ~[na:na]
  at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510) ~[na:na]
  at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run$$$capture(CompletableFuture.java:1773) ~[na:na]
  at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java) ~[na:na]
  at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[na:na]
  at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[na:na]
  at java.base/java.lang.Thread.run(Thread.java:842) ~[na:na]
Caused by: java.net.ConnectException: null
  at java.net.http/jdk.internal.net.http.common.Utils.toConnectException(Utils.java:1055) ~[java.net.http:na]
  at java.net.http/jdk.internal.net.http.PlainHttpConnection.connectAsync(PlainHttpConnection.java:198) ~[java.net.http:na]
  at java.net.http/jdk.internal.net.http.PlainHttpConnection.checkRetryConnect(PlainHttpConnection.java:230) ~[java.net.http:na]
  at java.net.http/jdk.internal.net.http.PlainHttpConnection.lambda$connectAsync$1(PlainHttpConnection.java:206) ~[java.net.http:na]
[...]

PR #4459, opened 2025-09-24
Defer call of getToolCallbacks() to prevent initializing ToolCallbackProvider eagerly

quaff:
Before this commit, `McpSyncClient` is initialized even if `spring.ai.mcp.client.initialized` is set to `true`.

See GH-3232


imod:
Although this is helpful, it only solves a part of the problem: as soon as the connection gets lost (e.g. reboot of the MCP Server) the connection does not get reestablished again and one needs to reboot the client too.

auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/test/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfigurationTests.java
```diff
@@ -45,12 +46,16 @@
 import org.springframework.util.ReflectionUtils;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.BDDMockito.then;
```
Kehrlann: Prefer `verify` over `then`.

auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/test/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfigurationTests.java
```diff
@@ -212,7 +230,10 @@ static class Config {
 		// ToolCallbacks.from(...) utility method.
 		@Bean
 		public ToolCallbackProvider toolCallbacks() {
-			return MethodToolCallbackProvider.builder().toolObjects(new WeatherService()).build();
+			ToolCallbackProvider toolCallbackProvider = MethodToolCallbackProvider.builder()
+				.toolObjects(new WeatherService())
+				.build();
+			return Mockito.spy(toolCallbackProvider);
```
Kehrlann: Prefer a mock over a spy.

spring-ai-model/src/main/java/org/springframework/ai/tool/resolution/ProviderToolCallbackResolver.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
```
Kehrlann:
```suggestion
 * Copyright 2025-2025 the original author or authors.
```

Kehrlann (commented):
The PR mention MCP specifically, but we're now wrapping ALL `ToolCallbackProvider`s, including `MethodToolCallbackProvider` - but those do not have an initialization issue. Is that desired? 

This is a real issue for bootstrapping MCP, and this addresses the bootstrapping part. There's another issue linked to this: remote MCP tools can change on the server, over time, and the client gets notified. In this case, the cached MCP tools are not updated. I wonder if we should have an MCP-specific abstraction rather than a `ToolCallbackProvider` abstraction.


If we want to delay this MCP-focused abstraction, then this PR is neat 👌

imod:
I was hoping this makes it into 1.1.0-M3 :(

auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/test/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfigurationTests.java
quaff: Existing tests rely on spy.

imod:
hmm, I tried this locally but did not succeed - the mcp tools still get initialised at startup and the server start fails if the mcp server is not available.

This is how I configure my `ChatClient`:

```
    @Bean
    fun chatClient(
        chatClientBuilder: ChatClient.Builder,
        mcpSyncClients: List<McpSyncClient>,
    ): ChatClient =
        chatClientBuilder
            .defaultSystem(systemPromptResource)
            .defaultToolCallbacks(
                SyncMcpToolCallbackProvider
                    .builder()
                    .mcpClients(mcpSyncClients)
                    .build(),
            ).defaultAdvisors(
                MessageChatMemoryAdvisor.builder(MessageWindowChatMemory.builder().build()).build(),
                SimpleLoggerAdvisor(),
            ).build()
```

Kehrlann:
@imod doing `.defaultToolCallbacks(...)` explicitly will always trigger a `tools/list` on MCP clients.

imod:
ok, so I always have to set them directly on the ChatClient when executing the prompt? e.g.

```
        val response =
            chatClient
                .prompt()
                .toolCallbacks(
                    SyncMcpToolCallbackProvider
                        .builder()
                        .mcpClients(mcpSyncClients)
                        .build(),
                ).toolContext(mapOf("progressToken" to "token-" + UUID.randomUUID().toString()))
                .advisors { advisor -> advisor.param(ChatMemory.CONVERSATION_ID, conversationId) }
                .user { u -> u.text(resolvedUserPrompt) }
                .call()
                .content()
```

Kehrlann:
Correct @imod 

You can put your `SyncMcpToolCallbackProvider` in a variable, you don't have to create a new one every time, but you need to call `.toolCallbacks()` on the chat client every time.

imod:
@Kehrlann if the remote MCP server is down during startup, then doing so, would cause an error during startup. 
My expectation is, that I can start my mcp client without failing if the remote mcp server is not available.

Kehrlann:
> @Kehrlann if the remote MCP server is down during startup, then doing so, would cause an error during startup. My expectation is, that I can start my mcp client without failing if the remote mcp server is not available.

@imod I may be wrong, but I think we only call the remote when using `#getToolCallbacks()`. This is a "provider" after all.

imod:
@Kehrlann I testet it multiple times, it fails at startup if the MCP server is down. 

```
@RestController
@RequestMapping("/chat/sse")
class SseChatController(
    val chatClient: ChatClient,
    val mcpPromptLoaderService: McpPromptLoaderService,
    val mcpSyncClients: List<McpSyncClient>,
    val progressTokenRegistry: ProgressTokenRegistry,
    val componentParser: ComponentParser,
) {
    val logger: Logger = LoggerFactory.getLogger(SseChatController::class.java)

    val toolCallbackProvider =
        SyncMcpToolCallbackProvider
            .builder()
            .mcpClients(mcpSyncClients)
            .build()

    @PostMapping("/ask", produces = [MediaType.TEXT_EVENT_STREAM_VALUE])
    fun ask(
        @RequestBody prompt: ChatRequest,
    ): SseEmitter {
        val emitter = SseEmitter(300000L) // 5 minute timeout
        val conversationId = prompt.conversationId ?: UUID.randomUUID().toString()

        try {
            val responseBuilder = StringBuilder()

            chatClient
                .prompt()
                .toolCallbacks(
                    toolCallbackProvider,
                )
                .advisors { advisor -> advisor.param(ChatMemory.CONVERSATION_ID, conversationId) }
                .user { u -> u.text(prompt.message) }
                .stream()
                .content()
                .subscribe(
                ...
```

PR #4475, opened 2025-09-26
fix: Fixed the example code error in multimodality doc

sunyuhan1998:
In the following section of the `Multimodality API` documentation, the constructor method for `UserMessage` used in the example code no longer exists: 

https://github.com/spring-projects/spring-ai/blob/f9071963c34dbdf836c9aeb2dfc7912281b6b083/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/multimodality.adoc?plain=1#L38-L50

The current PR modifies it to use the `Builder` pattern for construction.


ilayaperumalg:
@sunyuhan1998 Thanks for the PR fixing the doc.

Issue #4494, opened 2025-09-28
Cannot exclude certain auto-configurations in a consuming Boot application


Kehrlann:
**Bug description**

There is a special case in MCP tools, where excluding the `ToolCallingAutoConfiguration` would be desirable. In a regular Boot application, this would be achieved through:

```java
@SpringBootApplication(exclude = { ToolCallingAutoConfiguration.class })
public class SampleMcpClientWebfluxApplication {

//...

}
```

However, this does exclude the configuration, because, in my application, I use Anthropic and `AnthropicChatAutoConfiguration`. But that auto-configuration pulls in `ToolCallingAutoConfiguration` ([source](https://github.com/spring-projects/spring-ai/blob/b8dfc090b0af53dcdaa52a9d8bf8f50fe25441a2/auto-configurations/models/spring-ai-autoconfigure-model-anthropic/src/main/java/org/springframework/ai/model/anthropic/autoconfigure/AnthropicChatAutoConfiguration.java#L60-L61)), which cannot be excluded independently anymore.

**Environment**

Spring AI 1.0.x and 1.1.x are affected.

**Expected behavior**

It is possible to work around this specific case, but it requires diving into the auto-configurations and figuring their dependencies.
I think using `@ImportAutoConfiguration` on auto-configuration classes is an anti-pattern, because it breaks the "exclusion" part of the auto-configuration contract.

What should really happen in this case is that:
1. Conditions for `ToolCallingAutoConfiguration` are fulfilled, so it is actiaved.
1. Conditions for `AnthropicChatAutoConfiguration` are fulfilled, so it is activated.
1. Anthropic is marked `@AutoConfiguration(after = {ToolCallingAutoConfiguration.class})`, so the ordering is taking care of for me.

By doing so, it becomes possible for users to disable auto-configurations.

**Problematic classes**

- [x] ZhiPuAiChatAutoConfiguration (#4570)
- [x] MistralAiChatAutoConfiguration (#4536)
- [x] MistralAiOcrAutoConfiguration (#4536)
- [x] MistralAiEmbeddingAutoConfiguration (#4536)
- [x] MistralAiModerationAutoConfiguration (#4536)
- [x] ElevenLabsAutoConfiguration (#4568)
- [x] OpenAiChatAutoConfiguration (#4548)
- [x] OpenAiModerationAutoConfiguration (#4548)
- [x] OpenAiAudioTranscriptionAutoConfiguration (#4548)
[...]

PR #4495, opened 2025-09-28
Fix ollama auto-configuration

Kehrlann:
Related to #4494

---

This PR removes the `@ImportAutoConfiguration` from Ollama's auto-configuration classes.
It also reworks the "autoconfigure after" declarations, so that they are closest to the actual auto-config that uses them. For example, the `RestClientAutoConfiguration` is used by the `OllamaApiAutoconfiguration` through the `RestClient.Builder` injection ; so the dependency is expressed in `OllamaApiAutoConfiguration`.

We also introduce a testing pattern where all the auto-configs that _would_ be present in a Boot app are injected by default in the test contexts of integration tests.


ilayaperumalg:
@Kehrlann Thanks for the PR. LGTM, rebased and merged as b915413b3

PR #4509, opened 2025-09-29
Fix typo in ollama-embeddings.adoc

kirkster96:
I noticed that the table on this page was wrong. 

Here is my fix based on my understanding of the section prior to the table.

I tested this and it seems to be true.


ilayaperumalg:
@kirkster96 Thanks for fixing the documentation.

PR #4524, opened 2025-09-30
Fix JsonSchemaGenerator for method input parameters

ilayaperumalg:
- For method input parameters, if the JSON schema generator adds OpenAPI "format" property, some LLMs (ex: Mistral) fail to process the request. Hence, removing this specific property from the JSON schema
 - This issue was uncovered by MistralAiChatModelIT#chatMemoryWithTools where the method toolcallback is used

Auto-cherry-pick to 1.0.x
Fixes #4524

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java
@@ -140,6 +140,8 @@ public final class JsonSchemaGenerator {
 				required.add(parameterName);
 			}
 			ObjectNode parameterNode = SUBTYPE_SCHEMA_GENERATOR.generateSchema(parameterType);
+			// Remove OpenAPI format as some LLMs (like Mistral) don't handle them.
+			parameterNode.remove("format");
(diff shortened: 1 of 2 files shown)


spring-ai-model/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java
```diff
@@ -140,6 +140,8 @@ public static String generateForMethodInput(Method method, SchemaOption... schem
 				required.add(parameterName);
 			}
 			ObjectNode parameterNode = SUBTYPE_SCHEMA_GENERATOR.generateSchema(parameterType);
+			// Remove OpenAPI format as some LLMs (like Mistral) don't handle them.
+			parameterNode.remove("format");
```
ilayaperumalg, line 144: When the `parameterNode` doesn't contain "format", this operation just returns null and hence we don't need explicit check on the existence of this property.

ilayaperumalg, line 144: It would be good to find out what other cases the OpenAPI format can be used in the context of ChatClient request.

PR #4544, opened 2025-10-03
Use `assertThatExceptionOfType`

ngocnhan-tran1996:


ericbottard:
Thanks for your contribution @ngocnhan-tran1996

PR #4550, opened 2025-10-05
GH-4494: Fix anthropic model auto-configuration

loveysuby:
Related to #4494

---

This PR fixes conflicts in `AnthropicChatAutoConfiguration` by removing `@ImportAutoConfiguration` and keeping only the proper `@AutoConfiguration` setup.
It also adds `BaseAnthropicIT` to encapsulate Anthropic integration tests and stabilize configuration behavior.

I'm also wondering if having a separate Base{LLM_PROVIDER}IT per model provider is necessary, since they mostly replicate dependency setup from each auto-configuration.


ilayaperumalg:
@loveysuby Thanks for the PR! Rebased and merged as 110605f07

PR #4553, opened 2025-10-06
Ensure MCP Tools Autoconfiguration happens before ChatClient Autoconfiguration

chemicL:
Without this change, a `ChatClient` instance would be configured without the MCP capabilities expressed via `mcp-annotations`.

This surfaces when client-side MCP features are added such as **elicitation**, **sampling**, or **logging**.


ilayaperumalg:
LGTM

tzolov:
I've added this IT related to the issue: #4554
While ensuring the auto-confg order is important it won't help ensuring the bean postprocessor order.  
The #4554 will fail if we use @Lazy ChatClient instead of ObjectProvider<ChatClient.Builder> even with this order fix.

PR #4554, opened 2025-10-06
refactor: Migrate MCP test to use autoconfiguration and add ChatClient integration

tzolov:
Extend the E2E Mcp Client/Server annotations test to leverage and test the mcp-annotations autoconfigurations and handling the ChatClient in Sampling


ilayaperumalg:
Nice set of changes to use autoconfiguration in the IT.

ilayaperumalg:
Rebased and merged as 5ab9365ed

PR #3243, opened 2025-05-18
Add missing meAdded builder pattern to AssistantMessage and ToolResponseMessage for clarity and flexibilityssage builder

JM-Lab:
Unlike UserMessage and SystemMessage, AssistantMessage and ToolResponseMessage did not support the builder pattern. This PR applies the builder pattern to these classes for consistent and flexible object creation.
Additionally, the properties field in AssistantMessage has been renamed to metadata for consistency with other message types.

- Refactor affected usages after AssistantMessage and ToolResponseMessage update
- Added builder pattern to AssistantMessage and ToolResponseMessage.
- Renamed the properties field to metadata in AssistantMessage for improved semantic clarity and consistency.

PR #4569, opened 2025-10-08
Add builder to ToolResponseMessage

ericbottard:
Fixes https://github.com/spring-projects/spring-ai/pull/3243


ilayaperumalg:
Rebased and merged as 2ded9cd4b

myifeng:
@ilayaperumalg @ericbottard 

Could you please explain why `List.of()` is used here? Does this mean the content isn't being saved?

https://github.com/spring-projects/spring-ai/pull/4569/files#diff-2d8894a0ac637ab95a83efd64a5f280eaa3d34b3ca7101471c18df07bba2ea87R151

ericbottard:
The behavior has not changed and is explained by the comment above the line you referred to.

One could argue that the default value for the builder could be initialized to `List.of()` though.

myifeng:
> The behavior has not changed and is explained by the comment above the line you referred to.
> 
> 
> 
> One could argue that the default value for the builder could be initialized to `List.of()` though.

How can I save this value?

PR #4572, opened 2025-10-08
Fix CosmosDBChatMemory javadoc

deejay1:
CosmosDB Chat Memory wasn't available in 1.0.0, it was merged for 1.1.0-M3


ilayaperumalg:
@deejay1 Thanks for catching this and the PR! LGTM, merging.

PR #4574, opened 2025-10-08
Avoid string concatenation in logging calls

deejay1:
Avoid string concatenation in logging calls. Logger accepts positional parameters which are evaluated lazily, so this avoids concatenating the strings especially in debug calls


ilayaperumalg:
@deejay1 Thanks for the PR! @sunyuhan1998 thanks for the review.

PR #4578, opened 2025-10-09
Make most of Builders final

ericbottard:
This is a low hanging fruit change to make most of the Builders we have final, according to our recent design choice. A handful of Builders are left non-final because they we involved in a class hierarchy already.


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 0403a2a8c

PR #4583, opened 2025-10-09
Use modern language constructs

deejay1:
As we're using a base of Java 17, we can certainly make use of more modern language constructs like
enhanced switch and pattern matching for instanceof.


ilayaperumalg:
@deejay1 Thanks for the cleanup! Rebased and merged as d3c9de145

PR #4588, opened 2025-10-10
Safely merge empty tool call name case in streaming mode

yxfengforever:
Signed-off-by: Feng Xie<yxloveforever@gmail.com>

# Overview
The PR is to add a private method `safelyMergeAssistantMessageIfEmptyToolCallPresent` processes and merges tool calls in an AssistantMessage, particularly handling cases where tool calls might have empty name or fragmented arguments due to streaming or other processing issues.

### Key Logic of `safelyMergeAssistantMessageIfEmptyToolCallPresent`
- Reverses tool calls to process from last to first
- Accumulates arguments in StringBuilder
- **Critical**: `StringUtils.hasText(toolCall.name())` check **prevents invalid tool call exceptions**
- Only creates ToolCall objects when name has valid text content
- Reverses processed tool calls back to original order

### Safety Mechanism
`StringUtils.hasText(toolCall.name())` serves as a guard to:
- Prevent creating tool calls with null/empty names
- Avoid downstream execution exceptions
- Filter out malformed streaming data

### Result
Returns new AssistantMessage with merged tool calls, preserving original content, media, and metadata.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
@@ -129,9 +129,9 @@ public final class DefaultToolCallingManager implements ToolCallingManager {
 		Assert.notNull(chatResponse, "chatResponse cannot be null");
 
 		Optional<Generation> toolCallGeneration = chatResponse.getResults()
-				.stream()
-				.filter(g -> !CollectionUtils.isEmpty(g.getOutput().getToolCalls()))
-				.findFirst();
+			.stream()
+			.filter(g -> !CollectionUtils.isEmpty(g.getOutput().getToolCalls()))
+			.findFirst();
 
 		if (toolCallGeneration.isEmpty()) {
 			throw new IllegalStateException("No tool call requested by the chat model");
@@ -148,12 +148,13 @@ public final class DefaultToolCallingManager implements ToolCallingManager {
 				assistantMessage, internalToolExecutionResult.toolResponseMessage());
 
 		return ToolExecutionResult.builder()
-				.conversationHistory(conversationHistory)
-				.returnDirect(internalToolExecutionResult.returnDirect())
-				.build();
+			.conversationHistory(conversationHistory)
+			.returnDirect(internalToolExecutionResult.returnDirect())
+			.build();
 	}
 
-	private AssistantMessage safelyMergeAssistantMessageIfEmptyToolCallPresent(Optional<Generation> toolCallGeneration) {
+	private AssistantMessage safelyMergeAssistantMessageIfEmptyToolCallPresent(
+			Optional<Generation> toolCallGeneration) {
 		if (toolCallGeneration.isEmpty()) {
 			throw new IllegalStateException("No tool call requested by the chat model");
 		}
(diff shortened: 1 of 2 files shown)


tzolov:
Hi @yxfengforever , can you elaborate about the issue?  What models/providers are generating tool call chunks without name?

yxfengforever:
**The model** i use is **deepseek**. 
**The issue is:**
when i use toolcall in the stream mode, the toolcall name might be empty. One of the place in the spring AI framework requires to check if toolcall name presents,`StringUtils.hasText(toolCall.name())`. If the name is empty, then it throws exception. 
Actually, i ran a testing with one tool like "weatherReport". Here is the code.
```java
 Flux<ChatResponse> response = chatClient.prompt(ref.prompt)
                        Map.of("qam-user-id", headers.get("qam-user-id"))))
                .stream().chatResponse();
```
When it uses stream mode, then there are 2 tool calls return, of which one name is empty string. The expected response should be one tool call "weatherReport".

ericbottard:
I think [this PR](https://github.com/spring-projects/spring-ai/pull/5937) addresses the same problem and seems to fix the problem at a better place (allowing to keep tool calls streaming, rather than blocking) but I haven't taken a thorough look yet.

sobychacko:
@yxfengforever Thanks for the PR. However, closing this in favor of #5937, which addresses the same root cause at a better layer in `MessageAggregator` where the empty-name fragments are introduced, rather than post-processing them at execution time.

There are also two issues with this implementation worth noting for the record:

Looks like there is an argument ordering issue. The new method https://github.com/spring-projects/spring-ai/blob/9e6141751b5a8289ce179792b8362c948542f1c9/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java#L163-L176 reverses the tool call list, then appends argument fragments while iterating, then reverses again. From what I can understand from the code, this flips the order of the fragments. For a two-chunk case like `{name: "weatherReport", args: "{\"city\":"} `followed by `{name: "", args: "\"Beijing\"}"}`, I think the merged arguments come out as `"Beijing"}{"city":` instead of `{"city":"Beijing"}`.

The test doesn't reach the new code either. The method `executeToolCalls` selects the generation to process using `findFirst()` on the first generation with non-empty tool calls. https://github.com/spring-projects/spring-ai/blob/9e6141751b5a8289ce179792b8362c948542f1c9/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java#L131-L134  In the test, `generation1`, which only holds `toolCall1` with `name = "multiGenTool"`, is placed first in the list https://github.com/spring-projects/spring-ai/blob/9e6141751b5a8289ce179792b8362c948542f1c9/spring-ai-model/src/test/java/org/springframework/ai/model/tool/DefaultToolCallingManagerTest.java#L466-L469, and `generation2`, where `toolCall3` with the empty name lives, is never selected.
  https://github.com/spring-projects/spring-ai/blob/9e6141751b5a8289ce179792b8362c948542f1c9/spring-ai-model/src/test/java/org/springframework/ai/model/tool/DefaultToolCallingManagerTest.java#L452 The assertion passes with the original code unchanged.

  Thanks again for the PR — closing it as #5937 supersedes it and should get this sorted properly.

Please feel free to re-open if you disagree or there is context that I am missing here.

PR #4595, opened 2025-10-10
Fix ConnectionDetails class is exposed outside its defined visibility scope

quaff:


ilayaperumalg:
@quaff Thanks for the PR fixing the scope. LGTM, merging.

ilayaperumalg:
Rebased and merged as b4392c59b

PR #4598, opened 2025-10-10
GH-4586 Fixed some merge exceptions in `DeepSeekStreamFunctionCallingHelper`

### LINKED ISSUE #4586: DeepSeekStreamFunctionCallingHelper#merge throw NoSuchElementException
**Bug description**
When calling DeepSeek using the stream approach and tools are invoked, the `org.springframework.ai.deepseek.api.DeepSeekStreamFunctionCallingHelper#merge` method returns a NoSuchElementException: null exception.

**Analysis**

Image

An exception was thrown on line 98. The previous stream returned a tool call. The current stream returns an array of tool calls, which is an empty array rather than null. So the `current.toolCalls().iterator().next()` will throw an NoSuchElementException.

I think the judgement on line 94 should be `current.toolCalls() != null && !current.toolCalls().isEmpty()`

**Environment**
Spring AI version: 1.0.3
Java Version: 21
DeepSeek v3.1

**Expected behavior**
Return tools successfully. Don't throw exception.

sunyuhan1998:
As mentioned in the issue, there is currently a problem in `DeepSeekStreamFunctionCallingHelper` when merging instances where `toolCalls()` is an empty list rather than null. This PR fixes that issue. 

Additionally, I noticed that the class previously lacked unit tests, so I've added relevant test cases. 

Through these new tests, I discovered an issue in the `DeepSeekStreamFunctionCallingHelper#merge(ChatCompletionMessage, ChatCompletionMessage)` method:

```java
String content = (current.content() != null ? current.content()
        : "" + ((previous.content() != null) ? previous.content() : ""));
```

This line has an **operator precedence** problem. The string concatenation operator `+` has higher precedence than the ternary operator `? :`, so the expression is effectively parsed as:

```java
String content = current.content() != null 
    ? current.content() 
    : ("" + ((previous.content() != null) ? previous.content() : ""));
```

This means that if `current.content()` is not null, it is returned directly, and only when it is null does the string concatenation with `previous.content()` occur. This behavior clearly deviates from the intended merge logic.

The correct behavior should be to concatenate the contents of both `previous` and `current`. This PR fixes the issue, ensuring that message content is properly merged regardless of whether `current.content()` is null, and the fix is validated through newly added unit tests.

Fixes #4586

### DIFF
--- a/models/spring-ai-deepseek/src/test/java/org/springframework/ai/deepseek/api/DeepSeekStreamFunctionCallingHelperTest.java
+++ b/models/spring-ai-deepseek/src/test/java/org/springframework/ai/deepseek/api/DeepSeekStreamFunctionCallingHelperTest.java
@@ -40,7 +40,7 @@ class DeepSeekStreamFunctionCallingHelperTest {
 
 	@BeforeEach
 	void setUp() {
-		helper = new DeepSeekStreamFunctionCallingHelper();
+		this.helper = new DeepSeekStreamFunctionCallingHelper();
 	}
 
 	@Test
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
@sunyuhan1998 Thanks for the PR! Rebased and merged 68d176fe2

PR #4605, opened 2025-10-11
Polishing

quaff:


spring-ai-commons/src/main/java/org/springframework/ai/document/DefaultContentFormatter.java
```diff
@@ -121,13 +120,13 @@ public String format(Document document, MetadataMode metadataMode) {
 	 * @param metadata Document metadata.
 	 * @return Returns the filtered by configured mode metadata.
 	 */
-	protected Map<String, Object> metadataFilter(Map<String, Object> metadata, MetadataMode metadataMode) {
```
quaff, line 124: `protected` modifier is meaningless since the class is final.

spring-ai-commons/src/main/java/org/springframework/ai/document/DefaultContentFormatter.java
```diff
@@ -121,13 +120,13 @@ public String format(Document document, MetadataMode metadataMode) {
...
+	private Map<String, Object> metadataFilter(Map<String, Object> metadata, MetadataMode metadataMode) {
 
 		if (metadataMode == MetadataMode.ALL) {
-			return new HashMap<>(metadata);
```
quaff, line 127: Creating new HashMap is unnecessary here

PR #4611, opened 2025-10-11
Fix Mistral AI OCR javadoc since version

nicolaskrier:
Fix Mistral AI OCR javadoc since version tag of all new classes introduced in [this commit](https://github.com/spring-projects/spring-ai/commit/0c1a089624244f3143f6cbcc45ccf85759139b9d).


ilayaperumalg:
@nicolaskrier Thanks for fixing! @sunyuhan1998 @deejay1 Thanks for the review!

PR #4624, opened 2025-10-14
Fix wrong configuration key

quaff:


spring-ai-docs/src/main/antora/modules/ROOT/pages/api/embeddings/zhipuai-embeddings.adoc
```diff
@@ -228,7 +228,7 @@ Next, create an `ZhiPuAiEmbeddingModel` instance and use it to compute the simil
 
 [source,java]
 ----
-var zhiPuAiApi = new ZhiPuAiApi(System.getenv("ZHIPU_AI_API_KEY"));
+var zhiPuAiApi = new ZhiPuAiApi(System.getenv("ZHIPUAI_AI_API_KEY"));
```
ilayaperumalg: This should align with the above environment property "ZHIPUAI_API_KEY"

quaff: Updated.

ilayaperumalg:
@quaff Thanks for the fix! Rebased and merged as 1780e88fe

PR #4632, opened 2025-10-15
Support chat memory based on Elasticsearch

jiafu1115:
Hi Team：
I’ve opened this PR to introduce chat memory support based on Elasticsearch. Elasticsearch's indexes can be archived by time (for example, by week), which makes it a good fit for part of chat memory use cases. 
Can anybody help to take a look?
Thanks for review!

BTW: I pasted the integration test result here due to the CI only run the unit test.
image
image

### DIFF
--- a/auto-configurations/models/chat/memory/repository/spring-ai-autoconfigure-model-chat-memory-repository-elasticsearch/pom.xml
+++ b/auto-configurations/models/chat/memory/repository/spring-ai-autoconfigure-model-chat-memory-repository-elasticsearch/pom.xml
@@ -1,0 +1,110 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2023-2025 the original author or authors.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


jiafu1115:
Hi, @ilayaperumalg Any chance you could take a look or assign it to the right module owner? Thanks!

jiafu1115:
@ilayaperumalg @tzolov @markpollack Could someone please help take a look when you have a chance? Many thanks!

sobychacko:
Thanks for this @jiafu1115, and sorry for the slow turnaround. This overlaps with #5573, which also adds an Elasticsearch `ChatMemoryRepository` module. After comparing the two, we are going to move forward with #5573 as the base. A few reasons: it builds on the current `Rest5Client` / `Rest5ClientTransport` path (matching how `ElasticsearchVectorStore` already wires the client), it round-trips tool calls, tool responses, and media rather than message text only (in this PR, `TOOL` messages come back with empty `responses(...)` and media/tool-call data is not persisted), and it ships docs, a starter, and a BOM entry. This is not a reflection on the quality of your work here, the two simply cover the same ground and we can only carry one. Thank you for taking the time to put this together, and we would welcome future contributions.

PR #4641, opened 2025-10-15
test: add null parameter validation tests

alxkm:
This PR adds null parameter validation tests for two components to ensure proper error handling at API boundaries.

### Changes

**AzureOpenAiRuntimeHints:**
- Added test verifying that `registerHints()` throws `NullPointerException` when RuntimeHints parameter is null
- Ensures the method validates required parameters before processing

**ResourceCacheService:**
- Added test verifying that `getCachedResource()` throws `IllegalArgumentException` when URI is null
- Validates that error message contains "Location must not be null" for clear debugging
- Prevents `NullPointerException` from bubbling up to callers


ilayaperumalg:
@alxkm Thanks for the PR! and @sunyuhan1998 Thanks for the review! Rebased and merged as c03a5ca00

PR #4644, opened 2025-10-16
Deprecate public constructor of QuestionAnswerAdvisor in favor of builder

quaff:
And remove null check against searched documents since it never be null.


ilayaperumalg:
@quaff Thanks for the PR!

Issue #4643, opened 2025-10-16
Can @ConditionalOnMissingBean be added in McpServerStreamableHttpWebMvcAutoConfiguration ？


zhb0905:
# Issue Description

## Problem Statement

Currently, the `RouterFunction<?> webMvcStreamableServerRouterFunction` bean in `McpServerStreamableHttpWebMvcAutoConfiguration` and `McpServerStreamableHttpWebFluxAutoConfiguration` lacks the `@ConditionalOnMissingBean` annotation. This prevents me from cleanly customizing my own `WebMvcStreamableServerTransportProvider` and `RouterFunction` implementations without encountering bean definition conflicts.

## Current Behavior

When attempting to define a custom `RouterFunction` bean with the same name (`webMvcStreamableServerRouterFunction`), Spring Boot throws the following error:

```
The bean 'webMvcStreamableServerRouterFunction', defined in class path resource [org/springframework/ai/mcp/server/autoconfigure/McpServerStreamableHttpWebMvcAutoConfiguration.class], could not be registered. A bean with that name has already been defined in class path resource [com/alibaba/middleware/ai/PandoraMcpServerStreamableHttpWebMvcAutoConfiguration.class] and overriding is disabled.

Action:
Consider renaming one of the beans or enabling overriding by setting spring.main.allow-bean-definition-overriding=true
```

## Real-world Example

Here's my actual implementation that demonstrates the issue:

```java
@AutoConfiguration
@AutoConfigureBefore(McpServerStreamableHttpWebMvcAutoConfiguration.class)
@ConditionalOnClass({ McpSchema.class })
@Conditional({ 
    McpServerStdioDisabledCondition.class,
    McpServerAutoConfiguration.EnabledStreamableServerCondition.class 
})
@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)
@EnableConfigurationProperties({ 
    PandoraMcpProperties.class, 
    McpServerProperties.class, 
    McpServerStreamableHttpProperties.class 
})
public class PandoraMcpServerStreamableHttpWebMvcAutoConfiguration {

    @Autowired
    private PandoraMcpProperties pandoraMcpProperties;

    @Bean
    @ConditionalOnMissingBean
    public WebMvcStreamableServerTransportProvider webMvcStreamableServerTransportProvider(
            ObjectProvider<ObjectMapper> objectMapperProvider, 
            McpServerStreamableHttpProperties serverProperties) {
        
[...]

PR #4645, opened 2025-10-16
Allow application to provide custom RouterFunction

quaff:
Fix GH-4643


ilayaperumalg:
@quaff Thanks for the PR! LGTM, rebased and merged as b99f26df7

PR #4649, opened 2025-10-16
fix(deepseek): reset tool_choice handling to prevent infinite loop when returnDirect=false (#4617)

### LINKED ISSUE #4617: DeepSeekChatModel forces tool_choice to be specified to trigger loop tool call

**Bug description**
When I use DeepSeekChatModel to force a ToolCall call using tool_choice, if the tool definition has returnDirect=false, the completion of the push back will trigger ToolCall again, causing an infinite loop.

**Environment**
Java Version: 17
Spring AI Version: 1.0.3

**Steps to reproduce**
1. Define a tool function and specify returnDirect=false.
2. Use the DeepSeekChatModel ChatClient to specify ToolChoice as this tool and perform completion.

**Expected behavior**
Call the function only once and output the text normally

**My Suggestion**

- Before

https://github.com/spring-projects/spring-ai/blob/3fc1ed6750d9521a50abe56a537319681d785d6b/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekChatModel.java#L209-L225

- After

Reset tool_choice to "auto" when push back to model

```java
		if (this.toolExecutionEligibilityPredicate.isToolExecutionRequired(prompt.getOptions(), response)) {
			var toolExecutionResult = this.toolCallingManager.executeToolCalls(prompt, response);
			if (toolExecutionResult.returnDirect()) {
				// Return tool execution result directly to the client.
				return ChatResponse.builder()
					.from(response)
					.generations(ToolExecutionResult.buildGenerations(toolExecutionResult))
					.build();
			}
			else {
							if (prompt.getOptions() instanceof DeepSeekChatOptions) {
								DeepSeekChatOptions options = (DeepSeekChatOptions)prompt.getOptions();
								options.setToolChoice(ChatCompletionRequest.ToolChoiceBuilder.AUTO);
							}

				// Send the tool execution result back to the model.
				return this.internalCall(new Prompt(toolExecutionResult.conversationHistory(), prompt.getOptions()),
						response);
			}
		}

		return response;
```

kuntal1461:
Summary
Fixes #4617 — When DeepSeekChatModel is forced to use a tool_choice with returnDirect=false, 
the model repeatedly re-enters tool-call, causing an infinite loop. 
This change resets tool_choice to auto after tool execution so the model can produce normal assistant text.

Changes
- Reset tool_choice to auto after tool execution in both internalCall and internalStream.
- Add DeepSeekChatModelToolChoiceResetTests to verify forced tool_choice behavior.

How to Reproduce
1. Define a tool with returnDirect=false
2. Force tool_choice to that tool
3. Observe repeated tool invocation (without this fix)

Expected Behavior
Tool is invoked once; model returns normal text.

Environment
- Observed on Spring AI 1.0.3
- Fix applied against main (1.1.0-SNAPSHOT)

Tests
mvn -pl models/spring-ai-deepseek -Dtest=org.springframework.ai.deepseek.DeepSeekChatModelToolChoiceResetTests test

DCO Sign-off
Signed-off-by: Kuntal Maity <kuntal.1461@gmail.com>

### DIFF
--- a/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekChatModel.java
+++ b/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekChatModel.java
@@ -198,7 +198,7 @@ public class DeepSeekChatModel implements ChatModel {
 				if (completionEntity != null && completionEntity.getBody() != null) {
 					usage = chatCompletion.usage();
 				}
-				Usage currentChatResponseUsage = usage != null ? getDefaultUsage(usage) : new EmptyUsage();
+				Usage currentChatResponseUsage = toUsageOrEmpty(usage);
 				Usage accumulatedUsage = UsageCalculator.getCumulativeUsage(currentChatResponseUsage,
 						previousChatResponse);
 				ChatResponse chatResponse = new ChatResponse(generations,
@@ -221,9 +221,7 @@ public class DeepSeekChatModel implements ChatModel {
 			}
 			else {
 				// Reset tool choice to AUTO to prevent forcing repeated tool calls.
-				if (prompt.getOptions() instanceof DeepSeekChatOptions options) {
-					options.setToolChoice(ChatCompletionRequest.ToolChoiceBuilder.AUTO);
-				}
+				resetToolChoiceToAuto(prompt);
 				// Send the tool execution result back to the model.
 				return this.internalCall(new Prompt(toolExecutionResult.conversationHistory(), prompt.getOptions()),
 						response);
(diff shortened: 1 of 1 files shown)


models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekChatModel.java
```diff
@@ -76,6 +76,7 @@
  * backed by {@link DeepSeekApi}.
  *
  * @author Geng Rong
+ * @last Updated By : @kuntal1461
```
jiafu1115: keep same style? @author Geng Rong

kuntal1461: @jiafu1115 It is fixed . Please review once

models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekChatModel.java
```diff
@@ -305,6 +313,10 @@ public Flux<ChatResponse> internalStream(Prompt prompt, ChatResponse previousCha
 									.build());
 						}
 						else {
+							// Reset tool choice to AUTO to prevent forcing repeated tool calls.
```
jiafu1115, line 314: duplicated code with 223. you can extract these duplicated line to seperate method.

sobychacko:
Thanks for the PR!

Closing as obsolete. DeepSeek's internal tool execution loop was removed, so the code this changes is gone.

Issue #4617 is worth re-checking against the current tool-calling path, since the behavior may differ there.

Issue #4648, opened 2025-10-16
Reactive tool support


bcersows:
**Expected Behavior**

I would love to be able to pass Reactor context to tool definitions/processing, so stuff like this would work:

```
    @Tool(name = "greeter")
    public Mono<String> greet(
        @ToolParam(description = "...") final String language
    ) {
      return Mono.deferContextual(contextView-><do stuff based on context>);
    }
```

I am having a filter sitting in front of the actual tool execution that does some stuff like authorization based on the received call's meta data, and would like to pass the authorized user details into the tools that require it. As that is no real MCP parameter, I cannot pass it any other way.

So (at least for ASYNC mode) having the `McpToolUtils#toAsyncToolSpecification` not do a `Mono.fromCallable`, but instead have the call chain `toSyncToolSpecification` -> `toSharedSyncToolSpecification` also handle reactive publishers for the `callHandler`s would be great.
Yet another approach would be to map the callbacks into `AsyncMcpToolMethodCallback`.

**Current Behavior**

Due to the wrapping of async tools first into sync and then shared tools, any reactive context added before the handling is lost.

**Context**

Being able to pass arbitrary data between filters and the actual tools would be great. Alternatively, just allow for the user to pass data to the `ToolContext` created in `toSharedSyncToolSpecification` (or enrich it with the whole reactor context as well).

PR #4654, opened 2025-10-17
Added reactive approach to method toolbacks

bcersows:
Added reactive approach to method toolbacks, allowing to backpressure down until the tool definition, and to pass context to them.


bcersows:
Contains an approach for https://github.com/spring-projects/spring-ai/issues/4648.

PR #3823, opened 2025-07-15
feat: add streamable-HTTP transport support for MCP client

tzolov:
- Split MCP client auto-configuration into transport-specific modules to support new streamable-HTTP transport
  - spring-ai-autoconfigure-mcp-client-common (shared functionality)
  - spring-ai-autoconfigure-mcp-client-httpclient (HttpClient-based transports)
  - spring-ai-autoconfigure-mcp-client-webflux (WebFlux-based transports)
- Update starters to use appropriate transport-specific modules
- Add streamable-HTTP configuration properties (spring.ai.mcp.client.streamable)
- Update documentation with streamable-HTTP transport examples
- Update BOM to reflect new modular structure

PR #4665, opened 2025-10-19
docs: remove wrong note about activation order for mcp clients

jiafu1115:
After splitting the MCP clients/packages, the order is no longer used (useless) and has been removed from the implementation. So the current description of the order doesn’t match what actually happens. So propose to remove the description.
FYI:
https://github.com/spring-projects/spring-ai/pull/3823/files#diff-537607bc8950c21f4af68adc6a0eb8d47730de22883e9714e1e3da17bf8e1caaL64


jiafu1115:
tiny change. cc @tzolov for help review. thanks.

ilayaperumalg:
@jiafu1115 Thanks for the PR fixing the javadoc.

Issue #4670, opened 2025-10-19
MCP Client Initialization Timing and Tool Callback Resolution Issues


tzolov:
There are critical timing and resolution issues in the Spring AI MCP client auto-configuration that prevent MCP-annotated beans from being properly registered and cause tool callbacks to be resolved prematurely.

### Issue 1: MCP Client Initialization Timing

MCP clients are created and configured before all Spring singleton beans have been fully initialized. This creates a race condition where beans with MCP annotations (like `@McpLogging`, `@McpSampling`, `@McpToolListChanged`, etc.) may not have been scanned and registered yet when the MCP clients are being configured.

**Observed Behavior**
1. `McpClientAutoConfiguration` creates `List<McpSyncClient>` and `List<McpAsyncClient>` beans immediately
2. These clients are configured using specifications derived from `ClientMcpAnnotatedBeans`
3. However, `ClientMcpAnnotatedBeans` may not contain all annotated beans yet because:
   - Some `@Configuration` classes may initialize late
   - Some beans with MCP annotations may be created after the MCP clients are already configured
   - The annotation scanner completes before all beans are fully instantiated

**Impact**
- Late-initializing beans with MCP annotations are not included in the MCP client configuration
- Specification beans (like `SyncLoggingSpecification`, `AsyncSamplingSpecification`, etc.) are created with an incomplete view of the application context
- MCP annotation customizers (`McpSyncAnnotationCustomizer`, `McpAsyncAnnotationCustomizer`) are applied with missing specifications
- This results in MCP clients that are missing handlers for annotated methods that were meant to be included

### Issue 2: Eager Tool Callback Resolution in ChatClient

When `ToolCallbackProvider` instances are added to a `ChatClient` request spec, they are resolved immediately during the configuration phase by calling `getToolCallbacks()`. This causes problems for providers that depend on resources or beans that may not be fully initialized yet.

**Problematic Flow**
1. User configures: `chatClient.prompt().toolCallbacks(mcpToolCallbackProvider)`
2. During configuration, `DefaultChatClientRequestSpec` immediately calls `provider.getToolCallbacks()`
3. At this point, MCP clients may not exist yet or the client list may be empty
4. Result: Empty or incomplete tool callback list is registered

**Impact**
- `MCP*ToolCallbackProvider` implementations try to access MCP client lists that may still be empty
- Tool callbacks are computed before the providers are ready
- The lazy initialization pattern expected by Spring AI's architecture is broken
[...]

Issue #4618, opened 2025-10-13
MCPAnotated beans found in wrong order


DannySortino:
**Bug description**
When using MCPAnnotations such as @McpProgress the AbstractAnnotatedMethodBeanPostProcessor adds the found related methods of the beans to the internal registry. However it is then possible that the McpClientSpecificationFactoryAutoConfiguration will be found and created by Spring before any @Component beans meaning that any  methods with @McpProgress will fail to be properly created as listeners.

**Environment**
SpringAI 1.1.0-M2

**Steps to reproduce**
Have a bean defined with @Componet and within that have an @McpProgress method. It will then be noticied that the McpClientSpecificationFactoryAutoConfiguration will be created first and not the Component Bean, and will result in no listener being created.

**Expected behavior**
I would expect any bean that has @McpProgress to be picked up and processed, and not only ones found within the root spring applicaiton class, and within @Configuration classes.

ilayaperumalg:
Addressed via #4802

PR #4671, opened 2025-10-19
Lazy mcp client and toolcallback resolution

tzolov:
Fixes critical timing issues where MCP clients were created before all annotated beans were scanned, and tool callbacks were resolved too early during ChatClient configuration.

## Problems Fixed
1. **MCP Client Timing** - Clients created before all singleton beans initialized, missing late-initializing beans with MCP annotations
2. **Eager Tool Resolution** - Tool callback providers resolved during configuration instead of execution
3. **Static Resolver Pollution** - MCP providers incorrectly included in static resolver instead of lazy resolution
4. **List Reference Semantics** - Inconsistent handling prevented reference sharing needed for deferred initialization

## Solution
- **SmartInitializingSingleton Pattern**: New `McpSyncClientInitializer` and `McpAsyncClientInitializer` classes defer client creation until after all singletons are instantiated
- **Lazy Resolution**: ChatClient now stores providers and resolves them at execution time (`call()`/`stream()`)
- **MCP Provider Filtering**: Excluded from `StaticToolCallbackResolver` to enable lazy resolution
- **Reference Sharing**: Added `mcpClientsReference()` methods to properly share list references

## Key Changes
**Configuration:**
- `McpClientAutoConfiguration.java` - Added initializer classes
- `McpToolCallbackAutoConfiguration.java` - Uses reference sharing
- `McpClientSpecificationFactoryAutoConfiguration.java` - Deprecated (disabled by default)

**Tool Resolution:**
- `DefaultChatClient.java` - Lazy provider resolution
- `ToolCallingAutoConfiguration.java` - MCP provider filtering

**Providers:**
- Added `mcpClientsReference()` to `SyncMcpToolCallbackProvider` and `AsyncMcpToolCallbackProvider`

## Breaking Changes
**Deprecated (still functional):**
- `mcpClients(List)` and `mcpClients(varargs)` → Use `mcpClientsReference(List)` instead

**Behavioral (transparent to users):**
- MCP clients created after all singleton beans instantiated
- Tool callbacks resolved at execution time, not configuration time
- MCP providers excluded from static resolver

## Migration
**Most users**: No changes required - fixes are transparent and backward compatible.
**Custom MCP providers**: Update to use `mcpClientsReference()` instead of deprecated `mcpClients()` methods.

Resolves: #4670, #4618

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/main/java/org/springframework/ai/mcp/client/common/autoconfigure/McpToolCallbackAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/main/java/org/springframework/ai/mcp/client/common/autoconfigure/McpToolCallbackAutoConfiguration.java
@@ -60,9 +60,17 @@ public class McpToolCallbackAutoConfiguration {
 	 * <p>
 	 * These callbacks enable integration with Spring AI's tool execution framework,
 	 * allowing MCP tools to be used as part of AI interactions.
+	 *
+	 * <p>
+	 * IMPORTANT: This method receives the same list reference that is populated by
+	 * {@link McpClientAutoConfiguration.McpSyncClientInitializer} in its
+	 * {@code afterSingletonsInstantiated()} method. This ensures that when
+	 * {@code getToolCallbacks()} is called, even if it's called before full
+	 * initialization completes, it will eventually see the populated list.
 	 * @param syncClientsToolFilter list of {@link McpToolFilter}s for the sync client to
 	 * filter the discovered tools
-	 * @param syncMcpClients provider of MCP sync clients
+	 * @param syncMcpClients the MCP sync clients list (same reference as returned by
+	 * mcpSyncClients() bean method)
 	 * @param mcpToolNamePrefixGenerator the tool name prefix generator
 	 * @return list of tool callbacks for MCP integration
 	 */
@@ -70,15 +78,14 @@ public class McpToolCallbackAutoConfiguration {
 	@ConditionalOnProperty(prefix = McpClientCommonProperties.CONFIG_PREFIX, name = "type", havingValue = "SYNC",
 			matchIfMissing = true)
 	public SyncMcpToolCallbackProvider mcpToolCallbacks(ObjectProvider<McpToolFilter> syncClientsToolFilter,
-			ObjectProvider<List<McpSyncClient>> syncMcpClients,
-			ObjectProvider<McpToolNamePrefixGenerator> mcpToolNamePrefixGenerator,
+			List<McpSyncClient> syncMcpClients, ObjectProvider<McpToolNamePrefixGenerator> mcpToolNamePrefixGenerator,
 			ObjectProvider<ToolContextToMcpMetaConverter> toolContextToMcpMetaConverter) {
 
-		List<McpSyncClient> mcpClients = syncMcpClients.stream().flatMap(List::stream).toList();
-
+		// Use mcpClientsReference to share the list reference - it will be populated by
+		// SmartInitializingSingleton
 		return SyncMcpToolCallbackProvider.builder()
-			.mcpClients(mcpClients)
-			.toolFilter(syncClientsToolFilter.getIfUnique((() -> (McpSyncClient, tool) -> true)))
+			.mcpClientsReference(syncMcpClients)
+			.toolFilter(syncClientsToolFilter.getIfUnique((() -> (mcpClient, tool) -> true)))
 			.toolNamePrefixGenerator(
 					mcpToolNamePrefixGenerator.getIfUnique(() -> McpToolNamePrefixGenerator.noPrefix()))
 			.toolContextToMcpMetaConverter(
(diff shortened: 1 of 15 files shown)


auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/main/java/org/springframework/ai/mcp/client/common/autoconfigure/McpToolCallbackAutoConfiguration.java
```diff
@@ -60,45 +60,67 @@ public McpToolNamePrefixGenerator defaultMcpToolNamePrefixGenerator() {
...
+		// Use mcpClientsReference to share the list reference - it will be populated by
+		// SmartInitializingSingleton
 		return AsyncMcpToolCallbackProvider.builder()
-			.toolFilter(asyncClientsToolFilter.getIfUnique(() -> (McpAsyncClient, tool) -> true))
+			.toolFilter(asyncClientsToolFilter.getIfUnique(() -> (mcpClient, tool) -> true))
 			.toolNamePrefixGenerator(toolNamePrefixGenerator.getIfUnique(() -> McpToolNamePrefixGenerator.noPrefix()))
 			.toolContextToMcpMetaConverter(
 					toolContextToMcpMetaConverter.getIfUnique(() -> ToolContextToMcpMetaConverter.defaultConverter()))
-			.mcpClients(mcpClients)
+			.mcpClientsReference(mcpClients)
```
jiafu1115, line 123: if it can use clients instead of mcpClients. after all. the mcpClients is deprecated and mcpClientsReference is too long

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
```diff
@@ -988,18 +994,39 @@ public ChatClientRequestSpec templateRenderer(TemplateRenderer templateRenderer)
...
+		private void resolveToolCallbacksBeforeExecution() {
+			if (!this.toolCallbackProviders.isEmpty()) {
+				// Resolve all providers and add their callbacks
+				for (ToolCallbackProvider provider : this.toolCallbackProviders) {
+					this.toolCallbacks.addAll(List.of(provider.getToolCallbacks()));
+				}
+				// Clear providers list to avoid re-processing on subsequent calls
+				this.toolCallbackProviders.clear();
+			}
+		}
```
Kehrlann: **[issue]** There is a race condition here, where you could be initializing the toolcallbacks twice, in parallel. And, worse, one thread calling `.clear()` while another thread is still iterating, resulting in an exception.

auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/main/java/org/springframework/ai/mcp/client/common/autoconfigure/annotations/McpClientSpecificationFactoryAutoConfiguration.java
```diff
@@ -51,13 +51,33 @@
...
+ *
  * @author Christian Tzolov
  * @author Fu Jian
+ * @deprecated Since 1.1.0, specifications are now created dynamically after all singleton
+ * beans are initialized. This class will be removed in a future release.
  */
+@Deprecated(since = "1.1.0", forRemoval = true)
 @AutoConfiguration(after = McpClientAnnotationScannerAutoConfiguration.class)
 @ConditionalOnClass(McpLogging.class)
 @ConditionalOnProperty(prefix = McpClientAnnotationScannerProperties.CONFIG_PREFIX, name = "enabled",
```
kuntal1461, line 79: Hi @tzolov – flipping the condition to havingValue="false" unintentionally re-enables this (deprecated) auto-configuration whenever a user sets spring.ai.mcp.client.annotation-scanner.enabled=false. In 1.1.0‑M2 that property meant “turn it off”; with this change the same setting now activates it, so existing apps that disabled the scanner will regress. If the goal is to keep it off by default, the condition should probably remain tied to true (with a new default of false) or be removed entirely.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
kuntal1461, line 776: hi @tzolov – DefaultChatClientRequestSpec.mutate() only copies the already-resolved toolCallbacks list into the new builder and drops the still‑pending toolCallbackProviders. If a request spec is configured with toolCallbacks(provider) and then mutated before executing, the new client silently loses the provider and no MCP tools are ever resolved. (Repro: call chatClient.prompt().toolCallbacks(provider).mutate().build().prompt().call() – the provider is never invoked.) Please carry the providers over as well, e.g. add defaultToolCallbacks(this.toolCallbackProviders.toArray(new ToolCallbackProvider[0])) before returning the builder, or otherwise ensure they survive mutation.

ilayaperumalg:
This is addressed via https://github.com/spring-projects/spring-ai/commit/2a2f1551769b7d56a3cec01ca7e092c345da6dce #4802

PR #4674, opened 2025-10-19
test: Enhance PostgresMlEmbeddingOptions test coverage

alxkm:
## Description
This PR enhances test coverage for `PostgresMlEmbeddingOptions` by adding tests for previously untested methods and documenting important builder behavior.

## Changes
- Added 6 new test cases covering setters, interface methods, and builder behavior
- Documents that builder reuses the same instance (not creating new ones on each build())
- Validates edge cases and null handling

## Test Coverage Added

- **Setter functionality** - Validates setters work directly on options instance
- **Interface compliance** - Tests `getModel()` and `getDimensions()` return null as designed  
- **Builder behavior** - Documents that builder returns same instance on multiple `build()` calls
- **Builder mutation** - Shows that modifying builder after build affects previous instances
- **Null handling** - Verifies kwargs setter accepts null values

## Value
The tests reveal that the builder pattern implementation reuses the same options instance rather than creating new instances on each `build()` call. This behavior is now explicitly tested and documented, which is important for developers to understand as it differs from typical builder pattern expectations.

### DIFF
--- a/models/spring-ai-postgresml/src/test/java/org/springframework/ai/postgresml/PostgresMlEmbeddingOptionsTests.java
+++ b/models/spring-ai-postgresml/src/test/java/org/springframework/ai/postgresml/PostgresMlEmbeddingOptionsTests.java
@@ -203,4 +203,67 @@ public class PostgresMlEmbeddingOptionsTests {
 		assertThat(options.getKwargs()).doesNotContainKey("key1");
 	}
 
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
@alxkm Thanks for the PR and @sunyuhan1998 Thanks for the review. Rebased and merged as 838801fc8

PR #4652, opened 2025-10-17
Add Builder for FactCheckingEvaluator

guanxuc:
* Add Builder for FactCheckingEvaluator
* Add FactCheckingEvaluatorTests

PR #4675, opened 2025-10-20
Revise FactCheckingEvaluator

quaff:
1. use required arguments as parameters of `builder()`.
2. deprecate constructors.

See #4652

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/FactCheckingEvaluator.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/FactCheckingEvaluator.java
@@ -62,6 +62,7 @@ import org.springframework.util.Assert;
  * @author Eddú Meléndez
  * @author Mark Pollack
  * @author guan xu
+ * @author Yanming Zhou
  * @see Evaluator
  * @see EvaluationRequest
  * @see EvaluationResponse
@@ -97,7 +98,9 @@ public class FactCheckingEvaluator implements Evaluator {
 	 * the default evaluation prompt suitable for general purpose LLMs.
 	 * @param chatClientBuilder The builder for the ChatClient used to perform the
 	 * evaluation
+	 * @deprecated in favor of {@link #builder(ChatClient.Builder)}
 	 */
+	@Deprecated(forRemoval = true)
 	public FactCheckingEvaluator(ChatClient.Builder chatClientBuilder) {
 		this(chatClientBuilder, null);
 	}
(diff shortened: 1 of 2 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/FactCheckingEvaluator.java
```diff
@@ -97,7 +98,9 @@ public class FactCheckingEvaluator implements Evaluator {
 	 * the default evaluation prompt suitable for general purpose LLMs.
 	 * @param chatClientBuilder The builder for the ChatClient used to perform the
 	 * evaluation
+	 * @deprecated in favor of {@link #builder(ChatClient.Builder)}
 	 */
+	@Deprecated(forRemoval = true)
```
quaff, line 103: Could it be removed instead of deprecated?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/FactCheckingEvaluator.java
```diff
@@ -108,7 +111,9 @@ public FactCheckingEvaluator(ChatClient.Builder chatClientBuilder) {
 	 * @param chatClientBuilder The builder for the ChatClient used to perform the
 	 * evaluation
 	 * @param evaluationPrompt The prompt text to use for evaluation
+	 * @deprecated in favor of {@link #builder(ChatClient.Builder)}
 	 */
+	@Deprecated
 	public FactCheckingEvaluator(ChatClient.Builder chatClientBuilder, @Nullable String evaluationPrompt) {
```
quaff, line 117: Could it be `private` instead of deprecated?

quaff:
@ericbottard Please review.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/FactCheckingEvaluator.java
ilayaperumalg: We keep the deprecations latest for one release (even for milestone) before removing.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/FactCheckingEvaluator.java
ilayaperumalg: Let's keep this as is until we can completely remove it right after 1.1.0-M4.

ilayaperumalg: I realise this is the constructor used by the builder for instantiation. We should mark this private then.

ericbottard:
Merged as e81cd08e90809924d9a424ff592fd2d12f1beb0a

PR #4676, opened 2025-10-20
feat(ollama): support creating-model API (`/api/create`)

### LINKED ISSUE #4663: Add createModel support in OllamaApi
There’s an existing API in the Ollama application to create models dynamically via the endpoint:

POST /api/create

Currently, the Spring AI OllamaApi does not expose this operation.
Adding a createModel() method would be very useful, especially for initializing or customizing models and "template" at application startup.
 
Right now, there’s no way to change a model’s template through Java — this feature would make it possible to.

Here are a couple of reference screenshots from the Ollama Doc:
[Create Model Doc](https://docs.ollama.com/api#parameters-3)

Image

Image

Looking forward to seeing this feature added — it would make Spring AI’s Ollama integration much more flexible and complete.

yaner-here:
Close #4663

Ollama API provides a [`/api/create` route](https://docs.ollama.com/api#create-a-model) to create a new model from another existing model.

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaApi.java
@@ -246,6 +246,17 @@ public final class OllamaApi {
 				.bodyToFlux(ProgressResponse.class);
 	}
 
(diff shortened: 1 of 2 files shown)


yaner-here:
Hi, any progress?

PR #4683, opened 2025-10-21
Document that Spring Boot Actuator is required for observability

quaff:


ilayaperumalg:
@quaff Thanks for the documentation update!

PR #4687, opened 2025-10-21
Stop using array as annotation attribute value where feasible

quaff:


ilayaperumalg:
@quaff Thanks for the code cleanup! Rebased and merged as 77473d6d5

Issue #3368, opened 2025-05-29
Migrate from Azure OpenAI Java SDK to OpenAI Java Library


Reasch:
# Feature Request: Migrate from Azure OpenAI Java SDK to OpenAI Java Library

## Summary
Request to add support for the official OpenAI Java library (`openai-java`) in Spring AI, as the Azure OpenAI Java SDK has been deprecated and is no longer receiving updates.

## Background
The Azure OpenAI Java SDK (previously used by Spring AI) has been officially deprecated and Microsoft is directing users to migrate to the community-maintained OpenAI Java library. This change affects Spring AI users who rely on OpenAI integration through Azure endpoints.

**Current Status:**
- Azure OpenAI Java SDK: ⚠️ **Deprecated** - No longer receiving updates
- Recommended Alternative: [[OpenAI Java Library](https://github.com/openai/openai-java)]- Actively maintained community project

## Motivation
1. **Sustainability**: The current Azure OpenAI SDK will not receive security updates or new features
2. **Future Compatibility**: New OpenAI API features will only be available through the actively maintained library
3. **Community Support**: The OpenAI Java library has active community support and regular updates
4. **API Parity**: Ensures continued compatibility with latest OpenAI API capabilities

## Proposed Solution
Implement support for the OpenAI Java library while maintaining backward compatibility where possible.

### Suggested Implementation Approach:
1. **Add new dependency**: Include `openai-java` library as an optional dependency
2. **Create new configuration classes**: Implement Spring Boot auto-configuration for the new library
3. **Maintain existing interfaces**: Keep current Spring AI interfaces to minimize breaking changes
4. **Provide migration guide**: Document the transition path for existing users
5. **Deprecation timeline**: Establish a timeline for phasing out the old Azure SDK integration

## Benefits
- **Long-term maintainability**: Ensures Spring AI remains up-to-date with OpenAI developments
- **Enhanced features**: Access to latest OpenAI API capabilities
- **Better performance**: Potential improvements from actively maintained library
- **Community alignment**: Aligns with OpenAI ecosystem recommendations

## Compatibility Considerations
- Maintain existing Spring AI API surface to minimize migration effort
- Provide clear migration documentation
- Consider supporting both libraries during a transition period
- Ensure Azure OpenAI endpoints continue to work with the new library

## Implementation Tasks
- [ ] Research OpenAI Java library integration requirements
- [ ] Design new auto-configuration classes
- [ ] Implement new client adapters
[...]

PR #4688, opened 2025-10-21
Add support for OpenAI Java SDK (official Java library from OpenAI)

jdubois:
This follows the discussion on ticket #3368

This is a draft PR as the work is pretty big, I'll start with the simpler implementations (Embeddings) and finish with the more complex one (Chat).

### DIFF
--- a/models/spring-ai-openai-sdk/src/test/java/org/springframework/ai/openaisdk/chat/client/OpenAiSdkChatClientIT.java
+++ b/models/spring-ai-openai-sdk/src/test/java/org/springframework/ai/openaisdk/chat/client/OpenAiSdkChatClientIT.java
@@ -274,7 +274,7 @@ class OpenAiSdkChatClientIT {
 
 		// @formatter:off
 		String response = ChatClient.create(this.chatModel).prompt()
-				.user(u -> u.text("What's the weather like in San Francisco, Tokyo, and Paris?"))
+				.user(u -> u.text("What's the weather like in San Francisco, Tokyo, and Paris in Celsius?"))
 				.toolCallbacks(FunctionToolCallback.builder("getCurrentWeather", new MockWeatherService())
 					.description("Get the weather in location")
 					.inputType(MockWeatherService.Request.class)
@@ -297,7 +297,7 @@ class OpenAiSdkChatClientIT {
 					.description("Get the weather in location")
 					.inputType(MockWeatherService.Request.class)
 					.build())
-				.defaultUser(u -> u.text("What's the weather like in San Francisco, Tokyo, and Paris?"))
+				.defaultUser(u -> u.text("What's the weather like in San Francisco, Tokyo, and Paris in Celsius?"))
 			.build()
 			.prompt().call().content();
 		// @formatter:on
@@ -312,7 +312,7 @@ class OpenAiSdkChatClientIT {
 
 		// @formatter:off
 		Flux<String> response = ChatClient.create(this.chatModel).prompt()
-				.user("What's the weather like in San Francisco, Tokyo, and Paris?")
+				.user("What's the weather like in San Francisco, Tokyo, and Paris in Celsius?")
 				.toolCallbacks(FunctionToolCallback.builder("getCurrentWeather", new MockWeatherService())
 					.description("Get the weather in location")
 					.inputType(MockWeatherService.Request.class)


jdubois:
With today's commit, we now have:

- Support for EmbeddingModel
- Support for ImageModel

And some basic shared configuration, that will be improved next, so we're ready for the more complex stuff (ChatModel).

jdubois:
The setup logic for the OpenAI SDK should be very similar to this class from LangChain4, because it's basically doing the same thingj:

https://github.com/langchain4j/langchain4j/blob/main/langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/InternalOpenAiOfficialHelper.java

I'm very careful before taking code from another project, but I believe this is fine: I'm the author of that other class, so I have the copyright on this code. Also, it's simple logic to connect to Azure/GitHub Models/OpenAI, nothing that is very specific or critical.

However, if this causes problems:
- I can talk to OpenAI to check if that can be included into their code (I don't think they'll agree as they try to have something generic)
- I can create my own library, but having both projects depend on a small library that I would maintain doesn't seem a good idea to me

Happy to discuss this if needed.

jdubois:
Argh, I updated my fork and now my commit history is all wrong!!!! I'll do my best to fix this.

jdubois:
I had to force push my fork, sorry if anyone had this pulled on their side! I'll check if everything is good, otherwise I have a backup on my laptop.

jdubois:
Thanks @tzolov for our discussion -> I've just added a few more tests, I've put this "ready for review" and I'll be happy to discuss this with you

jdubois:
I've fixed some of the Checkstyle issues but there's a ton of them, so this is far from finished. Does it really make sense to follow of all of those rules?

tzolov:
@jdubois,
I've pushed the https://github.com/jdubois/spring-ai/commit/2e706f14d7dfabc078f8a6a629d6d3fb8d937b65 to add auto-configuration and boot starter. Also tried to improve the chat-options a bit.

tzolov:
Also fixed the audio output support: https://github.com/spring-projects/spring-ai/pull/4688/commits/94274426cfdca2438c5d5a0c4a85842f1a168ffd

jdubois:
Thanks a lot @tzolov !

jdubois:
@tzolov 
- I improved the authentication configuration in the Options* files (adding them to the Builder), and wrote documentation in the README.md file for authentication, as it's the non-obvious thing.
You should be able to run tests on GitHub Models when the model is available ("dall-e-3" isn't available so it won't work, for example).
- We got a big renaming last week, from Azure OpenAI to Microsoft Foundry, so I refactored everything to take this into account. This will prevent from having to do it later....

jdubois:
@tzolov concerning the remaining issues in your review:

- For Azure OpenAI (now Microsoft Foundry), it *should* be good. I'm testing on it, but I don't have a normal subscription, so the only risk is in that difference between my subscription and what normal people have. That should be pretty limited, we're using some very big common models here.
- The failing test in `OpenAiOfficialImageModelObservationIT` is all I have left, and to be honest I don't understand much about it.

jdubois:
@jdubois `OpenAiOfficialImageModelObservationIT` is fixed, everything should be green now!

tzolov:
@jdubois I've updated the openai-sdk chat docs: https://github.com/jdubois/spring-ai/blob/openaiofficial/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/openai-sdk-chat.adoc. 
Can you please check if they make sense? 

Usually the README contains only a reference to the official doc page. 

If you want you can generate the docs with `./mvnw -pl spring-ai-docs antora`  and then navigate to spring-ai-docs/target/antora/site/api/chat/openai-sdk-chat.html

In the meantime I will try to add doc pages for the opneai-sdk embedding and image models.

tzolov:
I've pushed the openai-skd-embedding.adoc and openai-sdk-image.adoc as well.

tzolov:
Great stuff. Thanks @jdubois

tzolov:
Rebased, squashed and merged at 15df47dc40c836825d7e4883ded16ed3cd7df090

jdubois:
Awesome, thanks for all your help @tzolov !!!

PR #4689, opened 2025-10-21
Refactor: Extract AbstractObservableChatModel to centralize all chat models' common logic

jiafu1115:
- Created AbstractObservableChatModel base class for observation, retry, and tool calling
- Refactored OpenAiChatModel and DeepSeekChatModel to extend the base class
- Reduced code duplication in subclass
- Introduced template method pattern for doCall() /doStream()/others

When reviewing #4649, I found that the issue isn’t limited to DeepSeek and all other models have similar logic for observation, retry, return directly and tool calling. So the best solution would be to refactor all the chat model code to centralize the common logic. That way, we can avoid having to make the same change in multiple places (10+ models mean 10+ * work load), and it will also make future maintenance easier across all models.
Example for repeat change in history:
image

BTW, since this is a change which could easily conflict with other ongoing work (though it’s low risk due to method extracting only), and I’m not sure about the community’s thoughts on it, I only applied the refactor to two models for now to get feedback. If this approach looks good and gets reviewed, I can go ahead and update the rest of the models in some more separate PRs.

Thanks for reviewing!

### DIFF
--- a/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekChatModel.java
+++ b/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekChatModel.java
@@ -20,14 +20,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
-import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationRegistry;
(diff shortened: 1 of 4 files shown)


jiafu1115:
#4756 is another similar case. I think it’d be better to do the refactor ASAP.
It’s low-risk, high-reward, and the earlier we do it, the better.
Could anyone help take a look when you’re not too busy? Thanks!
cc @ilayaperumalg @ericbottard

PR #4702, opened 2025-10-23
fix(chroma): handle collection not found exception during initialization

zksite:
Fix collection existence check broken by error message change in Chroma API.

The Chroma backend recently changed its error message from:
  "Collection [...] does not exists"
to:
  "Collection [...] does not exist"

This broke the string-based detection in `getCollection()`, causing 404s to be treated as unexpected errors and preventing initialization.

This PR replaces fragile string matching with reliable HTTP 404 status code checking via `HttpClientErrorException.NotFound`.

### DIFF
--- a/vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/chroma/vectorstore/ChromaVectorStore.java
+++ b/vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/chroma/vectorstore/ChromaVectorStore.java
@@ -48,6 +48,7 @@ import org.springframework.beans.factory.InitializingBean;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
+import org.springframework.web.client.HttpClientErrorException;
 
 /**
  * {@link ChromaVectorStore} is a concrete implementation of the {@link VectorStore}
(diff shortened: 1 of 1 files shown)


vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/chroma/vectorstore/ChromaVectorStore.java
```diff
@@ -117,7 +118,16 @@ public static Builder builder(ChromaApi chromaApi, EmbeddingModel embeddingModel
...
 	public void afterPropertiesSet() throws Exception {
 		if (!this.initialized) {
-			var collection = this.chromaApi.getCollection(this.tenantName, this.databaseName, this.collectionName);
+			//if the collection doesn't exist, chromaApi.getCollection will throw an exception
+			ChromaApi.Collection collection = null;
+			try {
+				collection = this.chromaApi.getCollection(this.tenantName, this.databaseName, this.collectionName);
+			}
+			catch (Exception ex) {
+				if (!(ex.getCause() instanceof HttpClientErrorException.NotFound)) {
```
MatthiasMolitor, line 127: Shouldn't this check be part of [ChromaApi:233](https://github.com/spring-projects/spring-ai/blob/main/vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/chroma/vectorstore/ChromaApi.java#L231)?

If I am not mistaken, the `ChromaApi` still contains the exception message check.
That exception message check will also still prevent `ChromaApi#getCollection()`  from returning `null`, when a collection not exists, although returning `null` seems to be the expected behavior.

zksite: Hi, I’ve looked into this further and realized the issue is in Spring AI’s own Chroma client implementation.

The error message check on line 233 currently uses:

```java
String.format("Collection [%s] does not exists", collectionName)
```

However, the actual response from the Chroma server is:

```
Collection [...] does not exist
```

Since the Chroma HTTP API returns "does not exist" ,As a result, the null-return path for missing collections is never triggered.

The fix is simply to update the expected message::

```java
if (String.format("Collection [%s] does not exist", collectionName).equals(msg)) {
    return null;
}
```

I’m happy to submit a quick PR if that would be helpful!

MatthiasMolitor: Changing the expected message would break compatibility with older Chroma versions.

Checking the type of the exception, as you already did in this Pull Request, might be the better approach.

But that is just my opinion. I am waiting for the same fix to be able to update Chroma in my applications :-)

megla-tlanghorst:
For now, using ChromaDB 1.1.1 seems to work at least. This issue seems to occur >= 1.2.

PR #4718, opened 2025-10-24
chore: clean up parent pom.xml

guanxuc:
- Define the `spring-javaformat-checkstyle.version` property in <properties>
- Format the `maven-gpg-plugin` plugin in the `sonatype` profile


ilayaperumalg:
@guanxuc Thanks for the PR! Rebased and merged as f616cd9b2

PR #4721, opened 2025-10-24
Add Spring Framework 7.x compatibility to SpringAiRetryAutoConfiguration#responseErrorHandler

### LINKED ISSUE #4720: ResponseErrorHandler used in SpringAiRetryAutoConfiguration fails with Spring Framework 7.x
Same issue as #3415 but for `SpringAiRetryAutoConfiguration`

Problem
---

Code breaks when users upgrade to Spring 7 because the deprecated handleError(ClientHttpResponse) method gets completely removed.

Current code:

```java
@Override
public void handleError(@NonNull ClientHttpResponse response) throws IOException {
    // our error handling logic
}
```
What changed:

Spring 6.2.x: Has both handleError(response) and handleError(url, method, response)
Spring 7.x: Removes the deprecated handleError(response) method entirely

Solution
----

Implement both methods, but don't use `@Override` on the old one:

```java
@Override
public void handleError(URI url, HttpMethod method, @NonNull ClientHttpResponse response) throws IOException {
    handleError(response);  // delegate to existing logic
}

// No @Override annotation - works in both versions
public void handleError(@NonNull ClientHttpResponse response) throws IOException {
    // existing error handling logic stays here
}
```
How it works:

- When using Spring 6.2.x, we call the old method directly (no overhead)
- When using Spring 7.x, we call the new method, which delegates to our logic
- Same behavior, works with both versions

ielatif:
Fixes #4720

### DIFF
--- a/auto-configurations/common/spring-ai-autoconfigure-retry/src/main/java/org/springframework/ai/retry/autoconfigure/SpringAiRetryAutoConfiguration.java
+++ b/auto-configurations/common/spring-ai-autoconfigure-retry/src/main/java/org/springframework/ai/retry/autoconfigure/SpringAiRetryAutoConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.ai.retry.autoconfigure;
 
 import java.io.IOException;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
@ielatif The changes you propose are already on `main` which supports Spring Framework 7: https://github.com/spring-projects/spring-ai/blob/920e6f45fd532cd2ad1d3b26343079f00859b8ef/auto-configurations/common/spring-ai-autoconfigure-retry/src/main/java/org/springframework/ai/retry/autoconfigure/SpringAiRetryAutoConfiguration.java#L100 Could you double check? Thanks

ielatif:
@ilayaperumalg you are right the changes are already on main.
I submited the PR a month ago, i havent noticed that it was fixed one week ago.
I'll close the PR.
Thanks.

ilayaperumalg:
@ielatif Sorry, the whole set of changes for Spring Framework 7.x & Spring Boot 4.0 compatibility was tracked and merged via a single PR https://github.com/spring-projects/spring-ai/pull/4774. Thanks for following it up!

PR #4724, opened 2025-10-25
Improve hasJacksonAnnotations performance with early return

jiafu1115:
- Remove unnecessary hasAnnotation variable
- Return immediately when Jackson annotation is found
- Improve performance by avoiding unnecessary reflection operations


ericbottard:
Thanks @jiafu1115

Issue #326, opened 2024-02-13
Review ImageOptions and ImageResponse


markpollack:
There are no setters on `ImageOptions`, but yet there are on `ChatOptions`.

The `ImageResponse` can have more helpers to convert base64 to binary.  

Review for ease of use in using the `ImageResponse`

PR #4736, opened 2025-10-27
Review API design of org.springframework.ai.image (#326)

kuntal1461:
## Summary

This PR contributes to issue #326 by improving the API design of org.springframework.ai.image:
Introduces a new ImageResponseFormat enum for strongly typed response formats.
Aligns ImageOptions across providers (OpenAI, Azure OpenAI, Stability, ZhiPuAi).
Updates tests accordingly.
Enhances documentation to reflect the new type-safe format property.
Ensures observation and response metadata consistency.
This aligns image models with improved usability and type safety.

## Testing

- ./mvnw -q -DskipITs -DskipNpm -Dgpg.skip=true -Dmaven.javadoc.skip=true test

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/image/ImageOptionsBuilder.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/image/ImageOptionsBuilder.java
@@ -100,8 +100,8 @@ public final class ImageOptionsBuilder {
 		}
 
 		@Override
-		public ImageResponseFormat getResponseFormat() {
-			return this.responseFormat;
+		public String getResponseFormat() {
+			return (this.responseFormat != null) ? this.responseFormat.getValue() : null;
 		}
 
 		public void setResponseFormat(ImageResponseFormat responseFormat) {
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiImageModel.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiImageModel.java
@@ -181,7 +181,7 @@ public class AzureOpenAiImageModel implements ImageModel {
 			if (runtimeImageOptions.getResponseFormat() != null) {
 				// b64_json or url
 				imageGenerationOptions.setResponseFormat(
-						ImageGenerationResponseFormat.fromString(runtimeImageOptions.getResponseFormat().getValue()));
+						ImageGenerationResponseFormat.fromString(runtimeImageOptions.getResponseFormat()));
 			}
 			if (runtimeImageOptions.getWidth() != null && runtimeImageOptions.getHeight() != null) {
 				imageGenerationOptions.setSize(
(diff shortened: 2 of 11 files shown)


spring-ai-model/src/test/java/org/springframework/ai/image/observation/DefaultImageModelObservationConventionTests.java
```diff
@@ -90,15 +91,15 @@ void shouldHaveHighCardinalityKeyValuesWhenDefined() {
 			.height(1080)
 			.width(1920)
 			.style("sketch")
-			.responseFormat("base64")
+			.responseFormat(ImageResponseFormat.B64_JSON)
```
ericbottard, line 94: This looks like a change of semantics here

spring-ai-model/src/main/java/org/springframework/ai/image/ImageOptionsBuilder.java
```diff
@@ -96,10 +101,10 @@ public void setModel(String model) {
 
 		@Override
 		public String getResponseFormat() {
-			return this.responseFormat;
+			return (this.responseFormat != null) ? this.responseFormat.getValue() : null;
```
ericbottard, line 104: Assuming we carry through with this change (see my overall comment on the benefits of this PR), could we try to push the null check inside the enum (for example having ImageResponseFormat.NULL a value, and setting all options to that)

kuntal1461:
Hi @ericbottard Thanks for calling that out—yes, the base64 → b64_json change was intentional. OpenAI (and Azure’s wrapper) return the literal b64_json token, so the enum would never round‑trip if we left the old string. The enum isn’t asserting that every provider supports every value; each integration only ever sets the members its API understands (e.g. Stability AI uses the PNG/JSON entries, OpenAI uses URL/b64_json). I do share the concern about future formats—if someone ships a new token we’ll need to extend the enum quickly—but the stronger typing gives us an earlier signal when that happens.

OpenAI’s image API docs list the response_format choices explicitly: url or b64_json. You can see it here (scroll to the request parameters table): https://platform.openai.com/docs/api-reference/images/create


Azure OpenAI mirrors OpenAI’s parameters. The CreateImage endpoint documents response_format taking url or b64_json as well: https://learn.microsoft.com/azure/ai-services/openai/reference#image-generation-create.

spring-ai-model/src/main/java/org/springframework/ai/image/ImageOptionsBuilder.java
kuntal1461: Hi @ericbottard I wouldn’t introduce a ImageResponseFormat.NULL sentinel. Today, if the caller never sets a response format we just return null, and Jackson simply omits response_format from the payload—which is exactly what the provider expects. If we replaced that with an enum constant whose getValue() returns "null", we’d end up serializing response_format: "null" unless every caller remembered to special-case the sentinel, which is both error-prone and not part of the provider’s spec. Keeping the null check where we materialize the request is much clearer; if we want to tighten the API surface, a better direction would be to expose an Optional<ImageResponseFormat> rather than distort the enum domain.

Can you confirm if this approach matches what you’d expect?

sobychacko:
This PR branch has diverged quite a lot and the image package has moved on across several commits.
Better to settle the design on #326 first based on what Eric suggested and reviewed above, then provide an implementation against current `main`. Closing the PR for the time being.

PR #4745, opened 2025-10-28
Fix wrong Javadoc

quaff:


models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/api/ZhiPuAiImageApi.java
```diff
@@ -43,7 +43,7 @@ public class ZhiPuAiImageApi {
 
 	/**
 	 * Create a new ZhiPuAI Image api with base URL set to
-	 * {@code https://api.ZhiPuAI.com}.
+	 * {@code https://open.bigmodel.cn/api/paas}.
```
sobychacko, line 46: It looks like we can't access either of these two URL's. how about removing it altogether?

quaff: It's the value of `ZhiPuApiConstants.DEFAULT_BASE_URL`.

ilayaperumalg: If the current URL is no longer valid, shouldn't the default constant value be changed as well then?

ilayaperumalg:
@quaff Thanks for the PR! @mxsl-gr Could you review this please? Thanks!

models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/api/ZhiPuAiImageApi.java
quaff: It's accessible from my country.

ilayaperumalg: > It's accessible from my country.

I understand that. What I meant is, if the URL "https://api.ZhiPuAI.com" is no longer valid, then shouldn't the URL "https://open.bigmodel.cn/api/paas" be set as the default as well?

quaff: I guess `https://api.zhipuai.com/` is wrong at the beginning.

ilayaperumalg: @quaff I misunderstood the changes. I overlooked the ZhiPuApiConstants.DEFAULT_BASE_URL value as I thought it had "https://api.ZhiPuAI.com" at the first place. Sorry about that. Merging.

ilayaperumalg:
Rebased and merged as cb6afde12

PR #4749, opened 2025-10-28
Defer ToolCallbackProvider resolution to execution time

### LINKED ISSUE #4748: ChatClient eagerly retrieves ToolCallbacks during configuration
**Expected Behavior**

ToolCallbackProviders should be resolved lazily when `call()` or `stream()` is invoked, not during client configuration. This allows proper Spring bean lifecycle management and defers expensive operations until they're actually needed.

**Current Behavior**

In `DefaultChatClient.java`, the `toolCallbacks(ToolCallbackProvider...)` method eagerly calls `getToolCallbacks()` during client configuration

**Context**

This issue creates several lifecycle problems:

- ToolCallbackProviders may depend on Spring beans that aren't fully initialized yet during ChatClient configuration
- The current implementation couples ToolCallback resolution to configuration time rather than execution time
- ToolCallbacks are retrieved even if the chat request is never executed

This particularly affects scenarios where:

- ToolCallbackProviders retrieve callbacks from external systems or APIs (e.g. MCP)
- Callbacks depend on request-time context that isn't available during configuration (e.g. security)
- Multiple ChatClient configurations are created but not all are executed

tzolov:
Changed DefaultChatClient to store ToolCallbackProvider instances instead of eagerly resolving them during configuration. Providers are now resolved lazily when call() or stream() is invoked.

- Added toolCallbackProviders field to DefaultChatClientRequestSpec
- Updated DefaultChatClientUtils to resolve providers at execution time
- Added tests verifying lazy evaluation behavior

Resolves #4748

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
@@ -618,6 +618,8 @@ public class DefaultChatClient implements ChatClient {
 
 		private final List<ToolCallback> toolCallbacks = new ArrayList<>();
 
+		private final List<ToolCallbackProvider> toolCallbackProviders = new ArrayList<>();
+
 		private final List<Message> messages = new ArrayList<>();
 
 		private final Map<String, Object> userParams = new HashMap<>();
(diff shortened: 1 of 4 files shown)


ericbottard:
You should add Fixes https://github.com/spring-projects/spring-ai/pull/4459 as well

PR #4751, opened 2025-10-28
Do not inject MCP ToolCallbackProviders into ToolCallbackResolver

Kehrlann:
## Description

- MCP ToolCallbackProviders should not be "resolved" at startup time by the auto-configuration, ie we don't want to call #getToolCallback eagerly
- We also want to break the following dependency cycle:
    - ChatClient
    - -> ToolCallingManager
    - -> ToolCallbackResolver
    - -> ToolCallbackProvider (incl. SyncMcpToolCallbackProvider)
    - -> McpSyncClient
    - -> ClientMcpAnnotatedBeans
    - -> ChatClient (when there is Sampling)
- This PR ensures that the ToolCallbackResolver does not depend on SyncMcpToolCallbackProvider, thus breaking the cycle.
- MCP callback providers can still be passed to the chat client, but only at runtime, not during the configuration phase.

## Breaking change

MCP tools cannot be resolved by name anymore. The following code would not resolve the MCP tool anymore:

```java
chatClient.prompt("Tell me a joke")
    .toolNames("mcp-laugh")
    .call();
```

Instead you'd have to to:

```java
chatClient.prompt("Tell me a joke")
    .toolCallbacks(/* your MCP ToolCallback or ToolCallbackProvider */)
    .call();
```

### DIFF
--- a/auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
+++ b/auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.ai.model.tool.autoconfigure;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import io.micrometer.observation.ObservationRegistry;
@@ -35,7 +36,14 @@ import org.springframework.ai.tool.resolution.DelegatingToolCallbackResolver;
 import org.springframework.ai.tool.resolution.SpringBeanToolCallbackResolver;
 import org.springframework.ai.tool.resolution.StaticToolCallbackResolver;
 import org.springframework.ai.tool.resolution.ToolCallbackResolver;
+import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.beans.factory.support.BeanDefinitionBuilder;
+import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProcessor;
+import org.springframework.beans.factory.support.DefaultListableBeanFactory;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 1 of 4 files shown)


auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
```diff
@@ -77,6 +100,50 @@ ToolCallbackResolver toolCallbackResolver(GenericApplicationContext applicationC
...
+				excludeMcpToolCallbackProviderBeanDefinition);
+	}
+
+	private static boolean isMcpToolCallbackProvider(ResolvableType type) {
+		if (type.getType().getTypeName().equals("org.springframework.ai.mcp.SyncMcpToolCallbackProvider")
+				|| type.getType().getTypeName().equals("org.springframework.ai.mcp.AsyncMcpToolCallbackProvider")) {
+			return true;
+		}
+		var superType = type.getSuperType();
+		return superType != ResolvableType.NONE && isMcpToolCallbackProvider(superType);
```
tzolov, line 144: After we remove the deprecated constructors It won't have (A)SyncMcpToolCallbackProvider inheritance. 
But currently it is possible.

auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
```diff
@@ -56,17 +65,31 @@
...
 	@ConditionalOnMissingBean
 	ToolCallbackResolver toolCallbackResolver(GenericApplicationContext applicationContext,
-			List<ToolCallback> toolCallbacks, List<ToolCallbackProvider> tcbProviders) {
-
+			List<ToolCallback> toolCallbacks,
+			@Qualifier(EXCLUDE_MCP_TOOL_CALLBACK_PROVIDER) List<ToolCallbackProvider> tcbProviders) {
 		List<ToolCallback> allFunctionAndToolCallbacks = new ArrayList<>(toolCallbacks);
-		tcbProviders.stream().map(pr -> List.of(pr.getToolCallbacks())).forEach(allFunctionAndToolCallbacks::addAll);
+		tcbProviders.stream()
+			.filter(pr -> !isMcpToolCallbackProvider(ResolvableType.forInstance(pr)))
```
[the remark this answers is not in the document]
quaff: If filtering is performed here, why registry bean `EXCLUDE_MCP_TOOL_CALLBACK_PROVIDER`?

auto-configurations/models/tool/spring-ai-autoconfigure-model-tool/src/main/java/org/springframework/ai/model/tool/autoconfigure/ToolCallingAutoConfiguration.java
Kehrlann: Good point. We should make these classes final if their constructors are all private.

PR #4752, opened 2025-10-28
Run format apply instead of check.

ericbottard:
I'm tired of running maven only to realize format is broken.

This PR instead runs the format earlier in the lifecycle. I've used this successfully on other Spring projects.

~Now, this has a shortcoming: what if people don't run maven locally before submitting their PR / or before us merging it?
Well, the formatting may differ and become stale, until it gets fixed again with the next well behaved PR.
I still believe this is better than the current frustrating situation.~

If you disagree, we can come up with a more advanced setup where CI runs `check` and developers run `apply` automatically, having the best of both worlds. Let me know. --> DONE

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -367,20 +367,6 @@
 	<build>
 		<plugins>
 			<plugin>
-				<groupId>io.spring.javaformat</groupId>
-				<artifactId>spring-javaformat-maven-plugin</artifactId>
-				<version>${spring-javaformat-maven-plugin.version}</version>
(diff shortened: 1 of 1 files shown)


ericbottard:
@ilayaperumalg Please TAL at this enhanced version.
Also see https://docs.github.com/en/actions/reference/workflows-and-actions/variables

ilayaperumalg:
Rebased and merged as fb523c8fd

PR #4761, opened 2025-10-30
Add test to ensure that ToolCallbackResolver doesn't use MCP ToolCallbackProviders

quaff:
See GH-4751


ilayaperumalg:
@quaff Thanks for the PR!

ilayaperumalg:
Rebased and merged as 68e5324de

PR #4766, opened 2025-10-30
feat: Add getTextContent method to retrieve message text content

### LINKED ISSUE #4735: Getting textContent but method name getText causes serialization failure issue !!!!
**Bug description**
SpringAI provides a Message interface for its conversation records, which has corresponding implementation classes. For example, UserMessage is the question asked by the user, and AssistantMessage is the question answered by the large model. In these implementation classes, the content of his answers is saved in the textContent field, but there is no get method provided. Instead, a getText method is provided, which results in no content in the JSON when serialized.

**Environment**
SpringAI 1.0.0-M6 , JDK17

**Steps to reproduce**

Image

**Expected behavior**
I expect to add the getTextContent() method on top of the original one

Image

Image

I see that 2 pieces of data were indeed stored, but unfortunately, the specific content was not stored.
That's because the textContent property in the implementation class of the org.springframework.ai.chat.messages.Message interface does not provide a get method, only a getText() method, which results in the inability to retrieve the value.
Later on, I wrote a serialization and deserialization tool class (reflection) specifically for this message, which solved the problem. However, it was quite troublesome, so I decided to add a getTextContent() method directly. This would be very convenient to use, and I hope to receive a reply after seeing it. Thank you

liugddx:
Close #4735

Add getTextContent() method to AbstractMessage class to comply with JavaBean naming conventions.
This ensures that the textContent field is properly serialized by standard JSON serialization libraries.

Changes:
- Added getTextContent() getter method in AbstractMessage
- Method is inherited by all Message implementation classes (UserMessage, AssistantMessage, SystemMessage, etc.)
- Maintains backward compatibility with existing getText() method

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AbstractMessage.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AbstractMessage.java
@@ -109,6 +109,15 @@ public abstract class AbstractMessage implements Message {
 	}
 
 	/**
+	 * Get the text content of the message.
+	 * @return the text content of the message
+	 */
+	@Nullable
+	public String getTextContent() {
+		return this.textContent;
+	}
+
+	/**
 	 * Get the metadata of the message.
 	 * @return the metadata of the message
 	 */


liugddx:
PTAL  @ilayaperumalg

Issue #3379, opened 2025-05-29
[EPIC]  Spring Boot 4.0 compatability


markpollack:
The goal is to support Spring Boot 3.4.x, 3.5.x, and Boot 4.0/Spring Framework 7 within the same branch/version of Spring AI

Spring AI depends on Boot only for the auto-configuraiton modules.

Spring AI has a tight dependency on Spring Framework web client classes such as WebClient and RestClient.

An initial pass to upgrade to Boot 4/ SF 7 was promising.  See this [diff](https://github.com/spring-projects/spring-ai/commit/3de5c3fea98b93ce7072149f91024c4bf00c461a#diff-c5f3cd52895f18b25c379596a29057988f2af57a2b043e11237933f1839acf58).

Takewaway

* Create a wrapper class in RetryUtils so that the method signature change in `ResponseErrorHandler` is handled correctly for both a SF 6 and SF 7 code base. 
* the addAll method for headers has changed when using restClient/webClient.  A workaround is to not call the newly changed method, but instead to iterate over the header map and use the 'add' method.
* `HttpHeaders` removed the method containsKey and replaced it with containsHeader.  Instead can switch to use `if (headers.get(headerName) != null)` in the code base as that will be portable between both SF 6 and SF 7
* `ChatClientExtensions.kt` should use the more strict kotlin 2.1 enforced syntax, it  will be compatible with kotlin 1.9
* Boot 3.5 pulls in elasticsearch-java client 8.18 where as boot 3.4.5 pulls in elasticsearch-java client 8.15.  This requires a code change to work with 8.18.  Spring AI should take over the mgmt of this away from Spring Boot and use the 8.18 client always.

ThomasVitale:
Will Spring AI 2.0 have a Spring Boot 4 baseline? It would be great if it was possible to use the new Programmatic Bean Registration APIs introduced in Spring Framework 7 to support auto-configuration of multiple beans of the same type, enabling auto-config for multiple ChatModels and ChatClients.

ilayaperumalg:
Related PR: https://github.com/spring-projects/spring-ai/pull/4817

PR #4771, opened 2025-10-31
GH-3379: Spring Framework 7.x compatibility

bedrin:
This PR adds compatibility with Spring Framework 7 to enable early adoption

- Use only Spring Framework APIs available both in 6.x and 7.x branches
- Add constructors without logic to `*Api` classes in `models` modules to simplify extensibility; effective final fields are marked as final
- Kotlin 2.x support; use kotlin compiler version 2.2.21
- Update MCP SDK to 0.15.0
- Update MCP Annotations to 0.6.0

NOTE: this PR doesn't add support of Spring Boot 4 - it requires a separate and much bigger change and will be carried over in separate PR. This PR doesn't break existing Spring Boot 3.5 / Spring Framework 6 compatibility and can be merged into Spring AI 1.1.x release


bedrin:
Merged `main` into this branch - ready for review. This feature brings compatibility with Spring Framework 7.x while keeping it compatible with existing versions (It does NOT support Sprig Boot 4 though). @ilayaperumalg do you think we can merge it to `1.1.x` instead? The goal is to allow early adoption and early feedback without having to wait for `2.0` release. In my project I'm not using Spring Boot part of Spring AI so we can already benefit from it.

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/AnthropicApi.java
```diff
@@ -217,7 +231,7 @@ public Flux<ChatCompletionResponse> chatCompletionStream(ChatCompletionRequest c
 		return this.webClient.post()
 			.uri(this.completionsPath)
 			.headers(headers -> {
-				headers.addAll(additionalHttpHeader);
+				headers.addAll(HttpHeaders.readOnlyHttpHeaders(additionalHttpHeader));
```
filiphr, line 234: This works however it is less performant as it keeps creating new objects even when `additionalHttpHeader` is empty. I would advise something more like:

```java
additionalHttpHeader.forEach(headers::addAll);
```

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/metadata/support/OpenAiResponseHeaderExtractor.java
```diff
@@ -71,7 +71,7 @@ public static RateLimit extractAiResponseHeaders(ResponseEntity<?> response) {
 
 	private static Duration getHeaderAsDuration(ResponseEntity<?> response, String headerName) {
 		var headers = response.getHeaders();
-		if (headers.containsKey(headerName)) {
+		if (null != headers.getFirst(headerName)) {
```
filiphr, line 74: This can be optimized a bit since at the moment the headers are fetched twice. The `getFirst` is not needed. The entire section could be 

```java
var values = headers.get(headerName);
if (!CollectionUtils.isEmpty(values)) {
    return DurationFormatter.TIME_UNIT.parse(values.get(0));
}
```

or

```java
var value = headers.getFirst(headerName);
if (value != null) {
    return DurationFormatter.TIME_UNIT.parse(value);
}
```

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/metadata/support/OpenAiResponseHeaderExtractor.java
```diff
@@ -82,7 +82,7 @@ private static Duration getHeaderAsDuration(ResponseEntity<?> response, String h
 
 	private static Long getHeaderAsLong(ResponseEntity<?> response, String headerName) {
 		var headers = response.getHeaders();
-		if (headers.containsKey(headerName)) {
+		if (null != headers.getFirst(headerName)) {
```
filiphr, line 85: Same comment as for `getHeaderAsDuration`

pom.xml
```diff
@@ -267,6 +267,7 @@
 		<maven.compiler.source>${java.version}</maven.compiler.source>
 		<maven.compiler.target>${java.version}</maven.compiler.target>
 		<kotlin.compiler.jvmTarget>${java.version}</kotlin.compiler.jvmTarget>
+		<kotlin.compiler.version>2.2.21</kotlin.compiler.version>
```
filiphr, line 270: I am not a Kotlin expert, but what does it mean if you use Kotlin compiler 2.2.21? When I tried to compile the project using Kotlin version 1.9.25 it was complaining because Spring Framework 7 had classes that were compiled with 2.2. It could mean that users would have the same problem if they want to use Spring AI 1.x with Spring 6.

filiphr (commented):
Thanks @ericbottard for pointing me to this PR. It is indeed a bit more complete that mine at #4817. However, I believe that some of the things I did in my PR are a bit more performant. e.g. the use of `HttpHeaders.readOnlyHttpHeaders` is less optimal than the one I am using 
```java
additionalHttpHeader.forEach(headers::addAll);
```

I've left some comments inline. 

Additionally, I believe that this PR will still not make the entire codebase compile against Spring 7 and that's because of the HuggingFace generated `ApiClient` using removed methods in Spring 7. If the [`ApiClient.mustache`](https://github.com/spring-projects/spring-ai/pull/4817/files#diff-111a294df116430873f6cc4f6fc2e14f67369bb65bd706c65b778cf41ac3ca2c) from my PR is added then this would be complete.

Additionally, I also added a GitHub action (https://github.com/spring-projects/spring-ai/pull/4817/files#diff-e76e5134b85a9eb3e0d9ce8e671ce97c8c577ff3c24162c45d71691078201347) which compiles the project against Spring 7, so if you want to make sure that Spring AI 1.x stays compatible with Spring 7 I would advise adding that as well.

pom.xml
bedrin: Great catch!

bedrin:
> Thanks @ericbottard for pointing me to this PR. It is indeed a bit more complete that mine at #4817. However, I believe that some of the things I did in my PR are a bit more performant. e.g. the use of `HttpHeaders.readOnlyHttpHeaders` is less optimal than the one I am using
> 
> ```java
> additionalHttpHeader.forEach(headers::addAll);
> ```
> 
> I've left some comments inline.

It's already done in PR for Spring Boot 4 support:
https://github.com/spring-projects/spring-ai/pull/4774/files#diff-e5cb77ca54425cc7c57b08f1d70da4118eaf2c0377d381b1d184ace96d6f6e89R417

This PR for Spring Framework 7 support might have some tradeoffs to keep everything compatible with both Spring Framework 6 and 7 but I don't think it makes any noticeable difference.

> 
> Additionally, I believe that this PR will still not make the entire codebase compile against Spring 7 and that's because of the HuggingFace generated `ApiClient` using removed methods in Spring 7. If the [`ApiClient.mustache`](https://github.com/spring-projects/spring-ai/pull/4817/files#diff-111a294df116430873f6cc4f6fc2e14f67369bb65bd706c65b778cf41ac3ca2c) from my PR is added then this would be complete.

I did it in separate PR for Spring Boot 4 support but I agree it should be part of Spring 7 support:
https://github.com/spring-projects/spring-ai/pull/4774/files?file-filters%5B%5D=.mustache&show-viewed-files=true#diff-111a294df116430873f6cc4f6fc2e14f67369bb65bd706c65b778cf41ac3ca2c

> 
> Additionally, I also added a GitHub action (https://github.com/spring-projects/spring-ai/pull/4817/files#diff-e76e5134b85a9eb3e0d9ce8e671ce97c8c577ff3c24162c45d71691078201347) which compiles the project against Spring 7, so if you want to make sure that Spring AI 1.x stays compatible with Spring 7 I would advise adding that as well.

Need input from maintainers here if it makes sense to have it or keep it simple before 2.0

tzolov:
Replaced by #4774

Issue #4787, opened 2025-11-03
DefaultToolExecutionExceptionProcessor should include exception when logging


mkjensen:
[DefaultToolExecutionExceptionProcessor](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolExecutionExceptionProcessor.java#L80) logs a message when a tool call fails, but it does not include the exception.

It would be nice to get access to the stack trace by including the exception in the call to the logger.

PR #4788, opened 2025-11-03
GH-4787: Include exception in call to logger when tool execution fails

mkjensen:
Add exception as argument to logger call when tool execution fails.

Issue: https://github.com/spring-projects/spring-ai/issues/4787


ilayaperumalg:
@mkjensen Thanks for the PR and @liugddx thanks for the review. 

LGTM, Rebased and merged as c5ce4bed3

PR #4730, opened 2025-10-26
feat: introduce McpServerObjectMapperFactory for consistent ObjectMapper configuration

liugddx:
Close #4451

Thank you for taking time to contribute this pull request!
You might have already read the [contributor guide][1], but as a reminder, please make sure to:

* Add a Signed-off-by line to each commit (`git commit -s`) per the [DCO](https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring#how-to-use-developer-certificate-of-origin)
* Rebase your changes on the latest `main` branch and squash your commits
* Add/Update unit tests as needed
* Run a build and make sure all tests pass prior to submission

For more details, please check the [contributor guide][1].
Thank you upfront!

[1]: https://github.com/spring-projects/spring-ai/blob/main/CONTRIBUTING.adoc

PR #4791, opened 2025-11-03
Revisit MCP ObjectMapper config

ericbottard:
This is resurrecting PR https://github.com/spring-projects/spring-ai/pull/4730, with the added trick of `@Bean(defaultCandidate=false)`


auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerAutoConfiguration.java
ilayaperumalg, line 78: Should `org.springframework.ai.mcp.server.common.autoconfigure.McpServerObjectMapperAutoConfiguration` be added here to make sure the order is more explicit (though it will be implicitly driven by the dependency bean defined).

ilayaperumalg (approved):

auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerAutoConfiguration.java
ericbottard: Until proven otherwise, this is not necessary, so I'll abstain. Moreover, it would also apply to other AutoConfig that also use a bean defined by the new AutoConfig this PR introduces.
I'll continue doing some research on what is and isn't necessary, but I think if you're ok with the rest of this PR, it can be merged as is.

ilayaperumalg:
Rebased, squashed and merged as 6d9b8e65a

PR #4802, opened 2025-11-04
Add McpClientHandlersRegistry

### LINKED ISSUE #4670: MCP Client Initialization Timing and Tool Callback Resolution Issues

Kehrlann:
Fixes #4670 

This PR introduces two variants (sync and async) of a "registry" of handlers, `ClientMcpSyncHandlersRegistry`. 

These registries have two main responsibilities:
- Discovering client MCP-annotated beans
- Dispatching method calls per MCP client name

This allows to break the dependency between MCP clients and MCP-annotated beans. This was a problem when doing sampling, where there was a cycle MCP Client -> MCP handlers -> ChatClient -> MCP Client.
With this PR, the MCP clients depend on the registry, but the registry does not depend on the MCP-handlers. Instead, it discovers them dynamically and makes handlers available by client name. See #4751 for the initial attempt at fixing the circular dependency.

The flow to capture annotations has two steps:
1. An initial scanning phase, with a `BeanFactoryPostProcessor`. Bean types are scanned for annotations, and the bean names found are recorded for later use. Additionally, `@McpSampling` and `@McpElicitation` annotations are tracked per client name. This allows to bootstrap MCP clients in the `McpClientAutoConfiguration`: by the time the clients are created, we know their capabilities, even if the annotated beans have not been instantiated yet.
2. After all singletons have been instantiated, using `SmartInitializingSingleton`, actual bean instances are retrieved, and from the bean instances and annotated methods, MCP handlers are populated in the registry, stored by client name.

Notes:
- There can only be one sampling and one elicitation annotation per client, but other annotations support multiple handlers.
- The async version chains handlers sequentially, similar to the original Java SDK implementation ([source](https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java#L930-L932))

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/test/java/org/springframework/ai/mcp/client/common/autoconfigure/annotations/McpClientListChangedAnnotationsScanningIT.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-common/src/test/java/org/springframework/ai/mcp/client/common/autoconfigure/annotations/McpClientListChangedAnnotationsScanningIT.java
@@ -16,15 +16,21 @@
 
 package org.springframework.ai.mcp.client.common.autoconfigure.annotations;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import io.modelcontextprotocol.spec.McpSchema;
+import org.junit.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.springaicommunity.mcp.annotation.McpPromptListChanged;
 import org.springaicommunity.mcp.annotation.McpResourceListChanged;
 import org.springaicommunity.mcp.annotation.McpToolListChanged;
+import reactor.core.publisher.Mono;
 
+import org.springframework.ai.mcp.annotation.spring.ClientMcpAsyncHandlersRegistry;
+import org.springframework.ai.mcp.annotation.spring.ClientMcpSyncHandlersRegistry;
 import org.springframework.boot.autoconfigure.AutoConfigurations;
 import org.springframework.boot.test.context.runner.ApplicationContextRunner;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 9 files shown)


mcp/mcp-annotations-spring/src/main/java/org/springframework/ai/mcp/annotation/spring/AbstractClientMcpHandlerRegistry.java
```diff
@@ -0,0 +1,156 @@
...
+		for (var beanName : beanFactory.getBeanDefinitionNames()) {
+			if (!beanFactory.getBeanDefinition(beanName).isSingleton()) {
+				// Only process singleton beans, not scoped beans
+				continue;
+			}
+			var foundAnnotations = scan(AutoProxyUtils.determineTargetClass(beanFactory, beanName));
+			if (!foundAnnotations.isEmpty()) {
+				this.allAnnotatedBeans.add(beanName);
+			}
+			for (var foundAnnotation : foundAnnotations) {
```
tzolov, line 81: I guess this block can be nested in side the `if (!foundAnnotations.isEmpty()) {` above

mcp/mcp-annotations-spring/src/main/java/org/springframework/ai/mcp/annotation/spring/AbstractClientMcpHandlerRegistry.java
```diff
@@ -0,0 +1,156 @@
...
+	protected Map<Class<? extends Annotation>, Set<Object>> getBeansByAnnotationType() {
+		// Use a set in case multiple handlers are registered in the same bean
+		Map<Class<? extends Annotation>, Set<Object>> beansByAnnotation = new HashMap<>();
+		for (var annotation : CLIENT_MCP_ANNOTATIONS) {
+			beansByAnnotation.put(annotation, new HashSet<>());
+		}
+
+		for (var beanName : this.allAnnotatedBeans) {
+			var bean = this.beanFactory.getBean(beanName);
+			var annotations = scan(bean.getClass());
```
tzolov: `this.scan`

mcp/mcp-annotations-spring/src/main/java/org/springframework/ai/mcp/annotation/spring/AbstractClientMcpHandlerRegistry.java
```diff
@@ -0,0 +1,156 @@
...
+	public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException {
+		this.beanFactory = beanFactory;
+		Map<String, List<String>> elicitationClientToAnnotatedBeans = new HashMap<>();
+		Map<String, List<String>> samplingClientToAnnotatedBeans = new HashMap<>();
+		for (var beanName : beanFactory.getBeanDefinitionNames()) {
+			if (!beanFactory.getBeanDefinition(beanName).isSingleton()) {
+				// Only process singleton beans, not scoped beans
+				continue;
+			}
+			var foundAnnotations = scan(AutoProxyUtils.determineTargetClass(beanFactory, beanName));
```
tzolov: this prefix (e.g. this.scan)

mcp/mcp-annotations-spring/src/main/java/org/springframework/ai/mcp/annotation/spring/ClientMcpAsyncHandlersRegistry.java
```diff
@@ -0,0 +1,267 @@
...
+		var consumers = this.resourceListChangedHandlers.get(name);
+		if (consumers == null) {
+			return Mono.empty();
+		}
+		return Flux.fromIterable(consumers).flatMap(c -> c.apply(updatedResources)).then();
+	}
+
+	@Override
+	public void afterSingletonsInstantiated() {
+		var beansByAnnotation = getBeansByAnnotationType();
```
tzolov: this prefix

mcp/mcp-annotations-spring/src/main/java/org/springframework/ai/mcp/annotation/spring/ClientMcpSyncHandlersRegistry.java
```diff
@@ -0,0 +1,282 @@
...
+			return;
+		}
+		for (var consumer : consumers) {
+			consumer.accept(updatedResources);
+		}
+	}
+
+	@Override
+	public void afterSingletonsInstantiated() {
+		var beansByAnnotation = getBeansByAnnotationType();
```
tzolov: this prefix

tzolov (approved):
Looks great! Thanks @Kehrlann 
There are some small formatting issues I'll resolve while merging

mcp/mcp-annotations-spring/src/main/java/org/springframework/ai/mcp/annotation/spring/AbstractClientMcpHandlerRegistry.java
Kehrlann: It's already pretty nested ; I'm reluctant to nest it more.
The code is functionally equivalent anyway.

But that's not a strong opinion.

tzolov:
Rebased, squashed and merged at 2a2f1551769b7d56a3cec01ca7e092c345da6dce

PR #4803, opened 2025-11-05
[minor-fix, google genai] use lazy load for unknown embedding dimensions and cache results

viacheslav-dobrynin:
**Context**
`super.dimensions()` calls the embedding model API to know actual dimensions number even if model exists in KNOWN_EMBEDDING_DIMENSIONS map.
This leads to unnecessary expenses and problems during testing.

**Solutions**
1. Compute dimensions number lazily.
2. Cache results (questionable, can change to `if res != null res else compute()`)

### DIFF
--- a/models/spring-ai-google-genai-embedding/src/main/java/org/springframework/ai/google/genai/text/GoogleGenAiTextEmbeddingModel.java
+++ b/models/spring-ai-google-genai-embedding/src/main/java/org/springframework/ai/google/genai/text/GoogleGenAiTextEmbeddingModel.java
@@ -259,7 +259,7 @@ public class GoogleGenAiTextEmbeddingModel extends AbstractEmbeddingModel {
 
 	@Override
 	public int dimensions() {
-		return KNOWN_EMBEDDING_DIMENSIONS.getOrDefault(this.defaultOptions.getModel(), super.dimensions());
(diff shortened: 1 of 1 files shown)


models/spring-ai-google-genai-embedding/src/main/java/org/springframework/ai/google/genai/text/GoogleGenAiTextEmbeddingModel.java
```diff
@@ -259,7 +259,7 @@ private DefaultUsage getDefaultUsage(Integer totalTokens) {
 
 	@Override
 	public int dimensions() {
-		return KNOWN_EMBEDDING_DIMENSIONS.getOrDefault(this.defaultOptions.getModel(), super.dimensions());
+		return KNOWN_EMBEDDING_DIMENSIONS.computeIfAbsent(this.defaultOptions.getModel(), model -> super.dimensions());
```
ilayaperumalg, line 262: @viacheslav-dobrynin Thanks for catching this and the fix!

ilayaperumalg:
LGTM, rebased and merged as a8b3982ef

models/spring-ai-google-genai-embedding/src/main/java/org/springframework/ai/google/genai/text/GoogleGenAiTextEmbeddingModel.java
viacheslav-dobrynin: Thanks you for the quick merge!

PR #4810, opened 2025-11-05
Optimize MistralAiEmbeddingModel dimensions method

nicolaskrier:
- Calculate and cache values for unknown models only if necessary
- Make known embedding dimensions a mutable map attribute
- Verify the cache mechanism with MistralAiEmbeddingModelTests

PR #4811, opened 2025-11-05
Refactor overridden EmbeddingModel dimensions method

nicolaskrier:
- Provide a way to override known embedding dimensions
- Make known embedding dimensions maps immutable

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/embedding/AbstractEmbeddingModel.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/embedding/AbstractEmbeddingModel.java
@@ -71,6 +71,24 @@ public abstract class AbstractEmbeddingModel implements EmbeddingModel {
 		}
 	}
 
+	/**
+	 * Return the dimension of the requested embedding generative name. Uses the embedding
+	 * model to retrieve its default dimensions if the generative name is unknown.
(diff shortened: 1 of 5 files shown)


spring-ai-model/src/main/java/org/springframework/ai/embedding/AbstractEmbeddingModel.java
```diff
@@ -71,6 +71,24 @@ public static int dimensions(EmbeddingModel embeddingModel, String modelName, St
...
 
+	/**
+	 * Return the dimension of the requested embedding generative name. Uses the embedding
+	 * model to retrieve its default dimensions if the generative name is unknown.
+	 * @param embeddingModel Embedding model client to determine its known embedding
+	 * dimensions and default dimensions.
+	 * @param modelName Embedding generative name to retrieve the dimensions for.
+	 * @return Returns the embedding dimensions for the model name.
+	 */
+	public static int dimensions(AbstractEmbeddingModel embeddingModel, String modelName) {
```
nicolaskrier, line 82: I am not sure it is the best solution to avoid source code duplication in embedding models overriding `dimensions` method.

nicolaskrier:
Blocked by [this PR (#4810)](https://github.com/spring-projects/spring-ai/pull/4810).

PR #4813, opened 2025-11-06
Remove redundant check from `ResponseErrorHandler::handleError`

quaff:
`handleError()` is called only if `hasError()` returns true.


quaff:
Related to #4721

quaff:
Rebased, please review @ilayaperumalg

PR #4774, opened 2025-10-31
GH-3379: Spring Boot 4.x compatibility

bedrin:
This PR migrates Spring AI framework to **Spring Boot 4**.0.0.RC1
Depending on Spring AI release schedule and planning it should probably be targeted to 2.x dedicated branch (or dedicated 1.1.x branch should be created) since it will break compatibility with Spring Boot 3.5

- Updated Spring Boot dependency to 4.0.0.RC1
- Spring Framework 7 API compatibility (see below)
- Migration to Spring Boot 4.x as per https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide
  * Adopt new modular design of starters
  * Use new Jackson 3.x API partially - some classes like `JsonSchemaGenerator` depend on Jackson 2.x for now
  * Support Property Mapper API changes around null handling
  * Use Spring Framework Retry functionality instead of separate Spring Retry project
- Updated Swagger codegen templates used by huggingface model to align with latest Spring Framework APIs
- Update elasticsearch test container to 9.2.0
- Added `FailureDetectingExternalResource` from old versions of `testcontainers` library to support `gemfire-testcontainers`
- Other related dependencies updates:
  * TestContainers to 2.0.1
  * GemFire testcontainers to 2.3.3
  * opensearch-testcontainers to 4.0.0
  * Rest Assured to 5.5.6
  * swagger-codegen-maven-plugin to 3.0.75

Fixes GH-3379 (https://github.com/spring-projects/spring-ai/issues/3379)

Includes changes from https://github.com/spring-projects/spring-ai/pull/4681

Built on-top of https://github.com/spring-projects/spring-ai/pull/4771
- Use only Spring Framework APIs available both in 6.x and 7.x branches
- Add constructors without logic to `*Api` classes in `models` modules to simplify extensibility; effective final fields are marked as final
- Kotlin 2.x support; use kotlin compiler version 2.2.21

Future tasks:
- [x] Raise issue to migrate from Spring Retry to Spring Framework 7 built-in retry functionality: https://github.com/spring-projects/spring-ai/pull/4681
- [ ] Raise issue with `swagger-codegen-maven-plugin` to support Spring Framework 7
- [x] Raise issue with GemFire to support testcontainers 2.x: https://github.com/gemfire/gemfire-testcontainers/pull/7
- [x] Raise issue with jsonschema-generator to support Jackson 3.x: https://github.com/victools/jsonschema-generator/issues/553

PR #4817, opened 2025-11-06
Use methods from HttpHeaders which are available both from Spring 6 and Spring 7

filiphr:
I know that Spring AI 1.x is only compatible with Spring Boot 3 and Spring 6. However, we are trying to migrate to Spring Boot 4 and keep using Spring AI 1.x. Once Spring AI 2.x is released we are going to migrate to it.

However, in the meantime it is not possible due to few things:

* Spring Boot 4 - big restructure of it's modules. This is OK, since we could create our own auto configurations for what we need
* Spring Web `HttpHeaders` - Use of certain methods from `HttpHeaders` which are no longer available in Spring 7.

This PR is trying to use alternative methods which are available both in Spring 6 and Spring 7. This means that things like the `OpenAiAPI`, `AnthropicApi` etc. can still be used on Spring 7.

I hope that you can consider this PR and consider porting it to Spring AI 1.1.

### DIFF
--- a/auto-configurations/common/spring-ai-autoconfigure-retry/src/main/java/org/springframework/ai/retry/autoconfigure/SpringAiRetryAutoConfiguration.java
+++ b/auto-configurations/common/spring-ai-autoconfigure-retry/src/main/java/org/springframework/ai/retry/autoconfigure/SpringAiRetryAutoConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.ai.retry.autoconfigure;
 
 import java.io.IOException;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 
 import org.slf4j.Logger;
@@ -30,6 +31,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
+import org.springframework.http.HttpMethod;
 import org.springframework.http.client.ClientHttpResponse;
 import org.springframework.lang.NonNull;
 import org.springframework.retry.RetryCallback;
@@ -84,6 +86,11 @@ public class SpringAiRetryAutoConfiguration {
 			@Override
 			public boolean hasError(@NonNull ClientHttpResponse response) throws IOException {
 				return response.getStatusCode().isError();
+			}
+
+			@Override
+			public void handleError(URI url, HttpMethod method, ClientHttpResponse response) throws IOException {
+				handleError(response);
 			}
 
 			// On purposes commented out so that the code can compile both with Spring 6


ericbottard:
Hi @filiphr,

thanks for your PR that could be a good first step indeed.

Forcing the use of spring-web 7 though (by adding
```xml
			<dependency>
				<groupId>org.springframework</groupId>
				<artifactId>spring-web</artifactId>
				<version>7.0.0-M8</version>
			</dependency>

```

in the dependencies sections), and after skipping over some other deprecations, I still get a couple of wrong API usages (in hugging-face first, but there could be others).

Could you take a look and try to cover the last extra mile?

filiphr:
Thanks for reviewing @ericbottard, there were indeed some more spring-web 7 compilation problems. I have that fixed locally. I'll push it. I can also push a GitHub workflow that will run a compile with Spring 7.

filiphr:
@ericbottard I've done the additional adjustments so that the entire code base can compile with Spring 7. I also added a GitHub workflow build so that a PR check is run with the compilation.

The "biggest" change is the `ApiClient.mustache`. I put it in quotes, since that file is the most amount of lines, but if you actually compare it with the one from the Swagger Codegen you'll see that the change is one or 2 lines around the header checks.

Let me know what you think

pom.xml
```diff
@@ -270,6 +270,7 @@
 
 		<!-- production dependencies -->
 		<spring-boot.version>3.5.7</spring-boot.version>
+		<spring-framework.version>6.2.12</spring-framework.version>
```
ericbottard, line 273: I assume changes to this file were not meant to be commited?

ericbottard:
@filiphr Thanks for your prompt replies.

I'll bring up the discussion about this PR with the team today and gauge whether this is a smart inclusion for 1.1

pom.xml
filiphr: I added a `spring_7` to the pr-check GitHub Action. It is needed if you would like to have that run always. If that's not something that you are interested then I will remove the changes in the `pom.xml` and remove the GitHub action.

filiphr:
Thanks a lot for considering this @ericbottard. I really hope that at least you will include the changes around the `HttpHeaders` in the `spring-ai-openai` and `spring-ai-anthropic` in 1.1, this would make it way easier for us and would avoid the need to temporarily copy those classes on our side.

ericbottard:
Hi @filiphr,

It is very likely that this will be considered first thing for the upcoming 1.1.1.

Also, https://github.com/spring-projects/spring-ai/pull/4771 has been brought to my attention, which is a bit broader is scope, so we would probably favor this one. Could you have a look at it and make sure the **Spring Framework 7 features only** from that one are similar with yours?

filiphr:
Thanks for pointing me to that PR @ericbottard. I've reviewed that PR and I've left some comments in there. My PR is missing 2 things compared to that PR:

* Add constructors without logic to `*Api` classes in `models` modules to simplify extensibility; effective final fields are marked as final
* Missing method override in `SpringAiRetryAutoConfiguration`

I will add the missing override, whereas for the constructors, I feel like that is a separate thing that you need to decide whether you want exposed or not like that.

tzolov:
Replaced by #4774

filiphr:
@tzolov thanks for merging #4774. However, that one is for Spring Boot 4, which I assume will come in Spring AI 2.0 in February. This PR was for making Spring AI 1.1 forward compatible with Spring Framework 7. I am well aware that the Auto Configurations cannot be made forward compatible, but I believe that the code that does not depend on Spring Boot can be best effort compatible with Spring 7.

If you agree, then I can adjust my branch and target the Spring AI 1.1.x branch.

PR #4820, opened 2025-11-06
Polish BeanOutputConverter and associated tests

nicolaskrier:


spring-ai-model/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
```diff
@@ -243,12 +245,16 @@ public String getJsonSchema() {
...
-			return this.jsonMapper.readValue(this.jsonSchema, Map.class);
+			return this.jsonMapper.readValue(this.jsonSchema, MAP_TYPE_REFERENCE);
 		}
 		catch (JacksonException ex) {
 			logger.error("Could not parse the JSON Schema to a Map object", ex);
 			throw new IllegalStateException(ex);
 		}
 	}
 
+	private static final class MapTypeReference extends TypeReference<Map<String, Object>> {
```
nicolaskrier, line 247: @sdeleuze, I introduced the same class as you in #6135.

Also, I wonder if it would make sense to ask the Spring Boot team whether they could expose this class with a public accessor. It already exists in Spring Boot as `org.springframework.boot.json.JacksonJsonParser.MapTypeReference`.

nicolaskrier:
Closed due to new contribution policy: this PR corresponds to polish-style changes only.

Issue #2494, opened 2025-03-17
JsonParser is using a self instantiated ObjectMapper, allow usage of a custom ObjectMapper instance


brunorabaca:
**Bug description**
When calling a tool, is invoked the JsonParser.OBJECT_MAPPER to extract tool arguments.
But when the json has '\n' characters the follow exception is thrown

`
com.fasterxml.jackson.core.JsonParseException: Illegal unquoted character ((CTRL-CHAR, code 10)): has to be escaped using backslash to be included in string value
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 191]
	at com.fasterxml.jackson.core.JsonParser._constructReadException(JsonParser.java:2660) ~[jackson-core-2.18.2.jar:2.18.2]
	at com.fasterxml.jackson.core.base.ParserBase._throwUnquotedSpace(ParserBase.java:1409) ~[jackson-core-2.18.2.jar:2.18.2]
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._finishString2(ReaderBasedJsonParser.java:2202) ~[jackson-core-2.18.2.jar:2.18.2]
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._finishString(ReaderBasedJsonParser.java:2173) ~[jackson-core-2.18.2.jar:2.18.2]
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.getText(ReaderBasedJsonParser.java:295) ~[jackson-core-2.18.2.jar:2.18.2]
	at com.fasterxml.jackson.databind.deser.std.UntypedObjectDeserializerNR.deserialize(UntypedObjectDeserializerNR.java:82) ~[jackson-databind-2.18.2.jar:2.18.2]
	at com.fasterxml.jackson.databind.deser.std.MapDeserializer._readAndBindStringKeyMap(MapDeserializer.java:623) ~[jackson-databind-2.18.2.jar:2.18.2]
	at com.fasterxml.jackson.databind.deser.std.MapDeserializer.deserialize(MapDeserializer.java:449) ~[jackson-databind-2.18.2.jar:2.18.2]
	at com.fasterxml.jackson.databind.deser.std.MapDeserializer.deserialize(MapDeserializer.java:32) ~[jackson-databind-2.18.2.jar:2.18.2]
[...]

PR #4833, opened 2025-11-09
Allow overriding JsonMapper via named beans

worldtiki:
## Problem

`JsonParser` and `ModelOptionsUtils` build their `JsonMapper` instances in static initializers with fixed Jackson features and no public extension point. Under Spring Boot 4 / Jackson 3, `JsonMapper` is immutable (`configure(...)` was removed), so applications can't enable features like `JsonReadFeature.ALLOW_UNESCAPED_CONTROL_CHARS` on the mappers Spring AI uses internally.

This surfaces in production as a `StreamReadException` (`Illegal unquoted character`) on tool-call argument deserialization — e.g. `MethodToolCallback.extractToolArguments` → `JsonParser.fromJson` — and on `BedrockProxyChatModel.internalCall` → `ModelOptionsUtils.jsonToMap`. Bedrock (and in general any LLM emitting multi-line tool-call arguments) hits this reliably. Related: #2494, #4838, #5178, #4390.

Today the only workarounds are reflection + `sun.misc.Unsafe.putObject` to overwrite the static final fields, or replacing large chunks of Spring AI (`ToolCallingManager`, `BedrockProxyChatModel`). Neither is acceptable for production.

## Proposal

Per prior review feedback: no new Maven module, no `@ConfigurationProperties`, no dedicated auto-configuration class — just add two named `JsonMapper` beans to the existing `ToolCallingAutoConfiguration`, in the same style as `mcpServerObjectMapper`.

Changes:

- `JsonParser` and `ModelOptionsUtils` gain an `AtomicReference<JsonMapper>` around their default mapper, plus `getJsonMapper()` / `setJsonMapper(JsonMapper)` static accessors. Internal callers resolve the mapper through the `AtomicReference`, so overrides take effect on every public method (`fromJson`, `toJson`, `jsonToMap`, `jsonToObject`, `toJsonString`, `objectToMap`, `mapToClass`, …). `ModelOptionsUtils.JSON_MAPPER` is preserved for binary compatibility with existing direct-field consumers.
- `ToolCallingAutoConfiguration` exposes two named beans — `jsonParserJsonMapper` and `modelOptionsJsonMapper` — each with `defaultCandidate = false` and `@ConditionalOnMissingBean(name = …)`, matching today's default Jackson config exactly.
- An `InitializingBean` pulls both beans via `@Qualifier` and wires them through the new setters. This runs whether the beans are the defaults or user-supplied overrides, so a custom bean always reaches the static utilities (which a simple side-effect in the default `@Bean` method can't guarantee).

## Usage

Users who need custom Jackson configuration now declare their own bean with the published name. Typical case — enabling lenient control-char handling for a Bedrock-backed tool-calling app:

```java
@Configuration
class SpringAiJacksonConfig {

    @Bean
    JsonMapper jsonParserJsonMapper() {
        return JsonParser.getJsonMapper()
            .rebuild()
            .enable(JsonReadFeature.ALLOW_UNESCAPED_CONTROL_CHARS)
            .build();
    }

    @Bean
    JsonMapper modelOptionsJsonMapper() {
        return ModelOptionsUtils.getJsonMapper()
            .rebuild()
            .enable(JsonReadFeature.ALLOW_UNESCAPED_CONTROL_CHARS)
            .build();
    }
}
```

Starting from `.getJsonMapper().rebuild()` preserves Spring AI's default modules and feature flags; callers only have to express the delta.

## Tests

- New `JsonMapperAutoConfigurationTests` covers: (1) defaults are exposed as the named beans and wired into both static utilities; (2) a user-supplied `jsonParserJsonMapper` replaces the default and reaches `JsonParser`; (3) same for `modelOptionsJsonMapper` and `ModelOptionsUtils`.
- Full monorepo build: 10,060 / 10,060 passing.

Addresses #2494.


ericbottard:
Hi @worldtiki, 

thanks for your contribution!

While I can appreciate the need to provide means to customize ObjectMappers used in Spring AI, this PR seems too "overengineered" to me. IMO, there is no need for a dedicated maven module and autoconfiguration class for this. A change to existing configuration should hopefully suffice, keeping the approach of naming the OM beans.

Moreover, introducing ConfigurationProperties to drive the configuration of the OM is overkill. It merely reproduces what Jackson already provides and limits the configuration to what you've provided here. Just let users define their own OM bean if they want to, leveraging every Jackson feature they want if needed.

Lastly, I shall mention that complexifying the Jackson story in Spring AI before we migrate to Spring 7 / Boot 4 (which uses Jackson 3) may not be the best timing, so we should be very careful to not make things more complicated than they need be.

worldtiki:
Yeap that's fair @ericbottard.
Let me take a second look and try to simplify this.

PR #4835, opened 2025-11-10
Provide detailed Exception message when token count exceeds max

jxblum:
Closes #4835

As a developer, it would be really helpful to know when the (default) max token count is exceeded, what was the configured max value and what was the token count of the Document by ID that exceeded it.

This sort of diagnostic information should be given in any Exception to avoid having to resort to "debugging" to get it.

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/embedding/MaxTokenCountExceededException.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/embedding/MaxTokenCountExceededException.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.embedding;
+
+import org.springframework.ai.document.Document;
+
+/**
+ * {@link RuntimeException} thrown when the token count of the provided content exceeds
+ * the configured maximum.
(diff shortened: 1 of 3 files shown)


jxblum:
Any updated on this?

sobychacko:
@jxblum Hope you are doing well! Thank you for the PR, and we apologize for the delay in reviewing it.

I have a quick design question: what would a new public exception type add here beyond a richer error? I am a bit cautious about growing the exception hierarchy for a single case, because it becomes part of the long-term public API, since users may start catching it specifically, and we need to preserve it for backward compatibility.
From what I understand, the main outcome we want is better diagnostic information from the exception stack trace without users spend extra time debugging the message. If that is the goal, I think we could keep throwing `IllegalArgumentException` with a detailed message and still get those diagnostics, while remaining compatible with
code that already catches `IllegalArgumentException`. What do you think? If you are open to that approach, we can agree on a stable message format and adjust the tests accordingly. Or did you have a specific use case that requires a new exception type?

Thanks again for the contribution. You might want to rebase the branch on the latest `main`.

jxblum:
Hi @sobychacko - Nice to hear from you! I am doing fine, thx. Hope you are doing well.


Regarding the domain-specific Exception (i.e. `MaxTokenCountExceededException`), there are several reasons why I would opt for this approach.

1. `IllegalArgumentException` is quite generic. If another call within the execution path ever throws an `IllegalArgumentException` for a different reason then the user is left to parse the contents of the Exception message to derive which argument was illegal. In many cases, an `IllegalArgumentException` should be handled by the caller appropriately.

2. Additionally, a custom Exception, namely the "`MaxTokenCountExceededException`, could hold additional metadata, such as the current `tokenCount` and the configured/permitted `maxTokenCount` as queryable properties on the Exception class. Additionally, it could very well have identified the `Document` by ID that violated the token count as another property. Furthermore, it could hold additional context of the running operation.

3. Spring AI domain-specific exceptions (like `MaxTokenCountExceededException`) could wrap provider-specific Exceptions. Arguably, the `TokenCountEstimator` is a provider interface wrapping different implementations, even though there is only a single provider implementation ATM. Token counts are model specific.

4. Finally, like the core Spring Framework's DAO [Exception Hierarchy](https://docs.spring.io/spring-framework/reference/data-access/dao.html#dao-exceptions), where the Exceptions are very descriptive (rather than generic), I like that the Exceptions are meaningful and consistent.

PR #4846, opened 2025-11-11
Handle malformed OpenAI API responses

tzolov:


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -758,9 +759,16 @@ public String getName() {
...
+		/**
+		 * Handles, empty, NULL and unknown values
+		 */
+		@JsonEnumDefaultValue
+		@JsonProperty("")
+		UNKNOWN,
 		/**
 		 * The model hit a natural stop point or a provided stop sequence.
 		 */
+		@JsonEnumDefaultValue
```
ilayaperumalg: This needs to be removed.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
```diff
@@ -758,9 +759,16 @@ public String getName() {
 	 */
 	public enum ChatCompletionFinishReason {
 
+		/**
+		 * Handles, empty, NULL and unknown values
+		 */
+		@JsonEnumDefaultValue
```
ilayaperumalg: Let's not assume any default value at the API level.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -428,7 +428,7 @@ private Generation buildGeneration(Choice choice, Map<String, Object> metadata,
 		List<Media> media = new ArrayList<>();
 		String textContent = choice.message().content();
 		var audioOutput = choice.message().audioOutput();
-		if (audioOutput != null) {
+		if (audioOutput != null && StringUtils.hasText(audioOutput.data())) {
```
liugddx: Currently, decoding of empty data has been avoided, but an IllegalArgumentException may still be thrown due to invalid Base64. It is recommended to use try/catch and log a warning. Additionally, perform null checks on audioParameters and format, providing safe defaults to avoid NPE.

```
if (audioOutput != null && StringUtils.hasText(audioOutput.data())) {
    var audioParams = request.audioParameters();
    String format = (audioParams != null && audioParams.format() != null)
            ? audioParams.format().name().toLowerCase()
            : "mpeg"; // It can be adjusted according to the project's existing default values.
    String mimeType = String.format("audio/%s", format);
    try {
        byte[] audioData = Base64.getDecoder().decode(audioOutput.data());
        Resource resource = new ByteArrayResource(audioData);
    } catch (IllegalArgumentException ex) {
        // It is recommended to log the warn instead of throwing it to avoid affecting the text portion.
        log.warn("Malformed audio Base64 data in OpenAI response; skipping audio payload", ex);
    }
}```

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
tzolov: I've cleaned in those in the recent commits.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
tzolov: Thanks @liugddx, I've covered this in last commit. 
However, I'm now questioning if it's the right approach. In an agentic loop, we should catch errors like this and return them to the LLM for correction—similar to how StructuredOutputValidationAdvisor works. Thoughts?

liugddx: > Thanks @liugddx, I've covered this in last commit. However, I'm now questioning if it's the right approach. In an agentic loop, we should catch errors like this and return them to the LLM for correction—similar to how StructuredOutputValidationAdvisor works. Thoughts?

Thanks, @tzolov. I think your intuition is right: in an agentic loop we shouldn’t silently “fix” malformed responses at the transport/parsing layer. Instead, we should surface a structured error signal that advisors can consume (similar to StructuredOutputValidationAdvisor) and let the loop prompt the model to repair.

tzolov: Ok, I will revert this change and leave it for a later considerations

ilayaperumalg (approved):

ilayaperumalg:
Rebased and merged as c0cc32cfd

PR #4848, opened 2025-11-11
fix: Correct UserMessage example to use builder

Mo-Greene:
The documentation example for UserMessage incorrectly used a private constructor. This commit updates the example to use the public `builder()` method, making the code functional.

---

I've submitted this fix because the example as written used a private constructor and appeared to be non-functional.

However, I did consider that this might have been an **intentional choice**—perhaps to show a simpler, POJO-like constructor pattern for teaching purposes, even if it wasn't runnable.

If that is the case and you'd prefer to keep the original example for pedagogical reasons, please let me know, and I'll be happy to close this pull request.


ilayaperumalg:
@Mo-Greene Thanks for catching and the PR fixing the documentation!

ilayaperumalg:
Rebased and merged as 68adc264c

PR #2363, opened 2025-03-02
GH-2432: Remove unused json-path dependency in spring-ai-openai

barsha-ghosh721:
Thank you for taking time to contribute this pull request!
You might have already read the [contributor guide][1], but as a reminder, please make sure to:

* Sign the [contributor license agreement](https://cla.pivotal.io/sign/spring)
* Rebase your changes on the latest `main` branch and squash your commits
* Add/Update unit tests as needed
* Run a build and make sure all tests pass prior to submission

PR #4876, opened 2025-11-13
Remove unused json-path dependency from spring-ai-anthropic module

kzwang:
Similar to https://github.com/spring-projects/spring-ai/pull/2363, remove unused json-path dependency in spring-ai-anthropic module as well


sobychacko:
The `spring-ai-anthropic` module has been fully rewritten on the official Anthropic Java SDK (available in `2.0.0-M3`) and the old `pom.xml` no longer exists. The `json-path` dependency is not present in the new module. Thanks for the PR!

PR #4903, opened 2025-11-18
Configure ErrorProne + Nullaway

ericbottard:
Configure ErrorProne + NullAway on the codebase.

Fix various ErrorProne issues across the codebase.

This is the first step towards implementing JSpecify null safety gradually.
This first commit focuses solely on setting up ErrorProne. In doing so, it caught a handful
of issues with our tests, that this commit fixes.

There are also many warnings that may be of interest.

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-openai/src/test/java/org/springframework/ai/model/openai/autoconfigure/tool/FunctionCallbackWithPlainFunctionBeanIT.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-openai/src/test/java/org/springframework/ai/model/openai/autoconfigure/tool/FunctionCallbackWithPlainFunctionBeanIT.java
@@ -317,7 +317,7 @@ class FunctionCallbackWithPlainFunctionBeanIT {
 				.collect(Collectors.joining());
 			logger.info("Response: {}", content);
 
-			assertThat(content).isNotEmpty().withFailMessage("Content returned from OpenAI model is empty");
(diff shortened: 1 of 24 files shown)


sdeleuze (commented):
As discussed today, it could make sense at least initially to configure NullAway to only check null-marked code via `-XepOpt:NullAway:OnlyNullMarked` as done in this PR, that will allow to introduce null-safety annotations package by package.

We typically only enable null-safety checks with Error Prone to avoid slowing down the compilation too much and don't confuse contributors with other kind of checks. This kind be done by specifying `-XepDisableAllChecks` (which was already used on Gradle side). This was not configured in the `maven` branch of my `jspecify-nullaway-demo` project, [it is now fixed](https://github.com/sdeleuze/jspecify-nullaway-demo/commit/febd0d2aea9f685140f9b70accdf53c3ed091a19), so maybe Spring AI could do the same, but as you prefer.

NullAway requires a recent version of `javac`, [this is enforced by NullAway in recent versions](https://github.com/uber/NullAway/pull/1317), and due to the forked compiler, it looks like the NullAway error message is not visible when using an unsupported Java version, so I would advise to use the Maven enforcer plugin to ensure Java 25 is used (combined with the existing `--release 17` to keep the baseline). The CI configuration should be updated to use Java 25 as well on `main`. You can maybe take inspiration of [Spring Data build ](https://github.com/spring-projects/spring-data-build).

ericbottard:
Requires https://github.com/spring-projects/spring-ai/pull/4916 to be merged first. I'll rebase this on main when this is done

PR #4904, opened 2025-11-18
Pass when lower case's evaluation response contains 'yes' for FactCheckingEvaluator and RelevancyEvaluator.

### LINKED ISSUE #4902: RelevancyEvaluator problem
**Bug description**
method evaluate of RelevancyEvaluator always returns passing false

Problem is with Git commit caed4b3705126ed8dfd2077c3177c9c7c2b8a963 of @remeio 

[https://github.com/spring-projects/spring-ai/commit/caed4b3705126ed8dfd2077c3177c9c7c2b8a963](commit)

specifically with this condition:

`if ("yes".equalsIgnoreCase(evaluationResponse))`

Because LLM response is "YES\n", due to DEFAULT_PROMPT_TEMPLATE:

"Answer YES, if the response for the query is in line with context information otherwise NO."

And the condition fails because of the line break after "YES" (because the using of "equals" instead of "contains")

**Environment**
Spring AI version 1.1.0

remeio:
Pass when lower case's evaluation response contains 'yes' for FactCheckingEvaluator and RelevancyEvaluator.

Fix #4902

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/FactCheckingEvaluator.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/FactCheckingEvaluator.java
@@ -138,7 +138,7 @@ public class FactCheckingEvaluator implements Evaluator {
 			.call()
 			.content();
 
-		boolean passing = "yes".equalsIgnoreCase(evaluationResponse);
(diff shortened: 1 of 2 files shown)


andruhon:
I'd improve it a little bit to trim as well. I had the evaluator failing, because model decided to answer with line end character `yes\n`

PR #4909, opened 2025-11-19
fix: SimpleVectorStoreFilterExpressionConverter incorrectly treats "in" substring inside metadata key as IN operator

### LINKED ISSUE #4908: SimpleVectorStoreFilterExpressionConverter incorrectly treats "in" substring inside metadata key as IN operator
#### Bug description

SimpleVectorStoreFilterExpressionConverter has a bug when converting filter expressions.
When a metadata key contains the substring "in", the converter mistakenly interprets it as an IN operator because it checks:

#### Expected Behavior

When converting a filter expression where:
- the operator is EQ
- the metadata key contains "in" as a substring 
- the value is a List 

the converter should output a standard SpEL equality expression.

```java
Filter.Expression expr = new Filter.Expression(
    ExpressionType.EQ,
    new Filter.Key("pin code"),
    new Filter.Value(List.of("1234", "5678"))
);
```
The expected SpEL output is:

``` java
#metadata['pin code'] == {'1234','5678'}
```

#### Current Behavior

The current SimpleVectorStoreFilterExpressionConverter incorrectly interprets the key substring "in" as the IN operator because it checks:

```java
context.lastIndexOf("in")
```

As a result, the converter generates an unintended contains() expression

```java
{'1234','5678'}.contains(#metadata['pin)
```

This is invalid and breaks metadata filtering in SimpleVectorStore.

#### Proposed Fix

Modify the doValue() implementation to distinguish real IN operators from accidental key substrings

```java
String ctx = context.toString().trim();
boolean isInClause = ctx.endsWith(" in");
boolean isNotInClause = ctx.endsWith(" not in");

if (!isInClause && !isNotInClause) {
    context.append(formattedList);
} else {
    appendSpELContains(formattedList, context);
}
```

This ensures substring "in" inside a metadata key is not mistaken for the IN operator.

diydriller:
#### Summary
This PR fixes a substring–based bug in SimpleVectorStoreFilterExpressionConverter where metadata keys containing "in" were incorrectly interpreted as IN operators.

The fix ensures that only real IN / NOT IN operators trigger the contains() conversion logic.

#### Changes
Updated doValue() to detect true IN operators using more precise checks:

```java
String ctx = context.toString().trim();
boolean isInClause = ctx.endsWith(" in");
boolean isNotInClause = ctx.endsWith(" not in");

if (!isInClause && !isNotInClause) {
    context.append(formattedList);
} else {
    appendSpELContains(formattedList, context);
}
```

Closes: https://github.com/spring-projects/spring-ai/issues/4908

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/converter/SimpleVectorStoreFilterExpressionConverter.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/converter/SimpleVectorStoreFilterExpressionConverter.java
@@ -85,7 +85,11 @@ public class SimpleVectorStoreFilterExpressionConverter extends AbstractFilterEx
 			}
 			formattedList.append("}");
 
-			if (context.lastIndexOf("in ") == -1) {
+			String ctx = context.toString().trim();
+			boolean isInClause = ctx.endsWith(" in");
+			boolean isNotInClause = ctx.endsWith(" not in");
+
+			if (!isInClause && !isNotInClause) {
 				context.append(formattedList);
 			}
 			else {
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks for the PR!

The file this PR modifies, `SimpleVectorStoreFilterExpressionConverter`, was removed from `main` in ba9220b22 (released in 2.0.0-M4), which replaced the SpEL string conversion with a direct AST evaluator. That change also eliminates the substring issue reported in #4908, so this PR no longer applies to `main`.

The 1.1.x line is no longer maintained on the OSS side, so we won't be retargeting this there either.

Closing as superseded by ba9220b22. See #4908 for details.

PR #4916, opened 2025-11-19
Introduce Maven toolchains.

ericbottard:
This makes sure we use a Java 25+ JDK to compile,
even though the compiled code is targeted at Java 17.


ilayaperumalg:
Rebased, squashed and merged as d26637b33

PR #4936, opened 2025-11-22
Fix extraBody loss during ModelOptionsUtils.merge()

SenreySong:
> ⚠️ Note: This **does not** cause `extraBody` to appear in the final HTTP payload. It only ensures the map is available for downstream flattening.

### Description  

The `extraBody` field is currently **lost during the merge step** in `ModelOptionsUtils.merge()`, which prevents non-standard model parameters (e.g., `enable_thinking`, `repetition_penalty`) from being passed to the underlying model API. This is a critical issue for models like **Qwen3** on Alibaba Cloud's Bailian platform, which rely on such parameters for advanced features ([Qwen3 API Reference](https://bailian.console.aliyun.com/?tab=api#/api/?type=model&url=271256)).

#### Problem  
`ModelOptionsUtils.merge()` only preserves fields annotated with `@JsonProperty`. Since `extraBody` is **intentionally not annotated** (it should be flattened into the top-level request body, not serialized as a nested field), it gets filtered out **before** any flattening logic can run:

```java
// Current problematic behavior in ModelOptionsUtils.merge():
        : acceptedFieldNames;
// → `extraBody` is absent from `requestFieldNames`, so it's discarded
```

As a result, even if a user sets:
```java
OpenAiChatOptions.builder()
    .extraBody(Map.of("enable_thinking", true))
    .build();
```
… the `extraBody` content never reaches the final `ChatCompletionRequest`.

#### Fix  
Explicitly include `"extraBody"` in the allowed field names when merging into classes that declare it (e.g., `OpenAiChatOptions`). This ensures the data survives the merge and can be properly flattened during serialization—**without changing the final JSON structure**.

#### Impact  
- Enables compatibility with models requiring custom parameters (e.g., Qwen3’s `enable_thinking`).
- No breaking changes; existing behavior is fully preserved.
- Minimal and safe change focused solely on data retention during merge.

#### Test Coverage  
Added `ExtraBodyMergeTest` to verify that `extraBody` is retained after merging and correctly propagated to the request object.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -628,14 +628,15 @@ public class OpenAiChatModel implements ChatModel {
 		ChatCompletionRequest request = new ChatCompletionRequest(chatCompletionMessages, stream);
 
 		OpenAiChatOptions requestOptions = (OpenAiChatOptions) prompt.getOptions();
-		request = ModelOptionsUtils.merge(requestOptions, request, ChatCompletionRequest.class);
+		request = ModelOptionsUtils.merge(requestOptions, request, ChatCompletionRequest.class,
+				ChatCompletionRequest.getMergableFieldNames());
 
 		// Add the tool definitions to the request's tools parameter.
 		List<ToolDefinition> toolDefinitions = this.toolCallingManager.resolveToolDefinitions(requestOptions);
 		if (!CollectionUtils.isEmpty(toolDefinitions)) {
 			request = ModelOptionsUtils.merge(
 					OpenAiChatOptions.builder().tools(this.getFunctionTools(toolDefinitions)).build(), request,
-					ChatCompletionRequest.class);
+					ChatCompletionRequest.class, ChatCompletionRequest.getMergableFieldNames());
 		}
 
 		// Remove `streamOptions` from the request if it is not a streaming request
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -16,6 +16,8 @@
 
 package org.springframework.ai.openai.api;
 
+import java.lang.reflect.Field;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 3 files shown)


linarkou:
I guess the same issue on the line 530 of OpenAiChatModel.java - if you define extraBody in defaultOptions (i.e. in `application.yml`), they won't make it into the final `CompletionRequest`.
IMHO, the problem is deeper, `ModelOptionsUtils::merge` logic shouldn't rely on `@JsonProperty` annotation.
There are 2 cases where `ModelOptionsUtils::merge` is used - to merge default options with request options, and then to pass 'merged' options into CompletionRequest.
I think the "merge" logic of this cases should be separated. My suggestion is to introduce new annotation(s) for that specific goals and rely on them for that mappings.

apappascs:
FYI @tzolov

SenreySong:
Hi @linarkou Could you please review this approach and let me know if it’s viable for inclusion?
I’m especially interested in whether the architectural direction—decoupling merge logic from JSON annotations—aligns with the project’s long-term goals.


#### 1. `@MergeableField` Annotation

**Location**: `spring-ai-model/src/main/java/org/springframework/ai/model/MergeableField.java`

```java
@Target(ElementType.FIELD)
@Retention(RetentionPolicy.RUNTIME)
@Documented
public @interface MergeableField {
    boolean deepMerge() default true;  // Support Map deep merging
    int priority() default 0;           // Reserved for future use
}
```

**Purpose**: Explicitly mark fields that should participate in options merging, decoupling from JSON serialization annotations.

#### 2. `ModelOptionsUtils.mergeOptions()` Method

**Location**: `spring-ai-model/src/main/java/org/springframework/ai/model/ModelOptionsUtils.java`

**Signature**:
```java
public static <T> T mergeOptions(Object source, Object target, Class<T> clazz)
```

**Merge Rules**:
1. Fields with `@MergeableField`: Always merged
2. Fields with `@JsonProperty` (no `@JsonIgnore`): Merged by default
3. Fields with `@JsonIgnore`: Only merged if also annotated with `@MergeableField`
4. Map fields with `@MergeableField(deepMerge=true)`: Deep merge key-value pairs
5. Other types: Source's non-null values override target's values

**Key Algorithm**:
```java
// 1. Create new instance
T result = clazz.getDeclaredConstructor().newInstance();

// 2. Scan all fields
for (Field field : clazz.getDeclaredFields()) {
    if (!shouldMergeField(field)) continue;

    Object sourceValue = field.get(source);
    Object targetValue = field.get(target);

    // 3. Merge field values
    Object mergedValue = mergeFieldValue(field, sourceValue, targetValue);
    field.set(result, mergedValue);
}
```

```java
// Before
OpenAiChatOptions requestOptions = ModelOptionsUtils.merge(
    runtimeOptions, this.defaultOptions, OpenAiChatOptions.class);

// After
OpenAiChatOptions requestOptions = ModelOptionsUtils.mergeOptions(
    runtimeOptions, this.defaultOptions, OpenAiChatOptions.class);
```

Wendell99:
Hope this bug gets fixed in the next milestone — I really want to use it!

ilayaperumalg:
@SenreySong Thanks for the PR. Sorry about the delay in response. This issue has been addressed via #4975 and #5321. Please verify the latest 2.0.0-SNAPSHOT or 1.1.5-SNAPSHOT to validate the fixes and share your feedback.

PR #4953, opened 2025-11-25
Fix Advisor NPEs by enforcing non-null context in ChatClientRequest/Response

### LINKED ISSUE #4952: ChatClient advisors NPE when context contains null values
## Bug description
`ChatClientRequest` only forbids null keys, not null values. The advisors (`ChatModelStreamAdvisor`, `ChatModelCallAdvisor`, `SafeGuardAdvisor`) call `Map.copyOf(request.context())`, and that blows up with `NullPointerException` if any value is null. The call/stream fails before the model is invoked.

## Environment
- Spring AI: main (latest)
- Java: 17 (also seen on 21)
- Vector store: not involved

## Steps to reproduce
1) Build a `ChatClientRequest` with a null context value, e.g. `context("tenantId", null)`.  
2) Use a `ChatClient` with default advisors (or directly use `ChatModelStreamAdvisor`/`ChatModelCallAdvisor`).  
3) Call `.stream(...)` or `.call(...)`.  
4) `Map.copyOf` throws NPE inside the advisor.

## Expected behavior
Either reject the request up front with a clear validation error, or avoid throwing NPE in the advisors when context contains null values.

## Minimal complete reproducible example
```java
import static org.assertj.core.api.Assertions.assertThatThrownBy;

import org.springframework.ai.chat.client.ChatClientRequest;

class ChatClientRequestNullContextTests {

    @Test
    void nullContextValueTriggersNpeInAdvisor() {
        ChatClientRequest request = ChatClientRequest.builder()
            .prompt(new Prompt("hi"))
            .context("tenantId", null) // null value
            .build();

        // Same failure point as in the advisors:
        assertThatThrownBy(() -> java.util.Map.copyOf(request.context()))
            .isInstanceOf(NullPointerException.class);
    }
}
```

## Affected code
- `spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelStreamAdvisor.java`
- `spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelCallAdvisor.java`
- `spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/SafeGuardAdvisor.java`
- `spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientRequest.java`

## Proposed resolution
Tighten `ChatClientRequest` validation to disallow null values in `context`, so invalid requests fail fast with a clear message, and the advisors can keep their immutable `Map.copyOf` calls.

Nhahan:
Fixes #4952

### Approach
- Fail fast on bad input by disallowing null context keys/values in `ChatClientRequest`.
- Extend the same guardrails to `ChatClientResponse` for consistency, so direct response creation can’t sneak in nulls.
- Keep advisor code unchanged; they can continue to use `Map.copyOf` safely.

### Scope
1) `ChatClientRequest`: constructor/builder enforce non-null keys and values (including single key/value adder).  
2) `ChatClientResponse`: constructor/builder enforce non-null keys and values for symmetry.  
3) Tests:  
   - Requests: null keys/values throw clear `IllegalArgumentException`.  
   - Responses: null keys/values throw clear `IllegalArgumentException`.  
   - Valid inputs still succeed; immutability semantics remain.

### Test plan
- `./mvnw -pl spring-ai-client-chat -DskipITs test` (module scope).  
- Unit tests in `ChatClientRequestTests` and `ChatClientResponseTests` cover null key/value rejection.

### Backward compatibility
- This change enforces non-null values in the context map.
- Existing code that passes `null` values will now throw `IllegalArgumentException` immediately at construction time, rather than causing an NPE later in Advisors.

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientRequest.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientRequest.java
@@ -37,6 +37,7 @@ public record ChatClientRequest(Prompt prompt, Map<String, Object> context) {
 		Assert.notNull(prompt, "prompt cannot be null");
 		Assert.notNull(context, "context cannot be null");
 		Assert.noNullElements(context.keySet(), "context keys cannot be null");
+		Assert.noNullElements(context.values(), "context values cannot be null");
 	}
 
 	public ChatClientRequest copy() {
@@ -68,12 +69,15 @@ public record ChatClientRequest(Prompt prompt, Map<String, Object> context) {
 
 		public Builder context(Map<String, Object> context) {
 			Assert.notNull(context, "context cannot be null");
+			Assert.noNullElements(context.keySet(), "context keys cannot be null");
+			Assert.noNullElements(context.values(), "context values cannot be null");
 			this.context.putAll(context);
 			return this;
 		}
 
 		public Builder context(String key, Object value) {
 			Assert.notNull(key, "key cannot be null");
+			Assert.notNull(value, "value cannot be null");
 			this.context.put(key, value);
 			return this;
 		}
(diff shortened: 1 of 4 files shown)


Nhahan:
Hi @ilayaperumalg — just following up on this when you have a chance.

PR #4960, opened 2025-11-25
Fix various build issues

ericbottard:
This commit addresses several build issues:
- Makes sure the checkstyle validation runs AFTER the javaformat
- Makes sure checkstyle runs only once (maven source jar generation shouldn't fork lifecylce in a reactor build)
- Makes sure javaformat runs only once (remove nonsence in the BOM project) and correctly uses 0.0.47+
- Fixes checkstyle config that will be compatable with JSpecify annotation placement.
- Clarities what the source/javadoc goals do (package, not generate new files)


ilayaperumalg (approved):

ilayaperumalg:
Rebased and merged as 3b14b4af7

PR #4975, opened 2025-11-26
Fix extraBody serialization by adding @JsonProperty annotation

### LINKED ISSUE #4966: the extraBody is not effective
```
OpenAiApi.ChatCompletionRequest request = new OpenAiApi.ChatCompletionRequest(null, null);
OpenAiChatOptions requestOptions = OpenAiChatOptions.builder()
        .model("test")
        .extraBody(Map.of("enable_thinking", true))
        .build();
request = ModelOptionsUtils.merge(requestOptions, request, OpenAiApi.ChatCompletionRequest.class);
System.out.println(request.extraBody());
```

the result is 
```
{}
```

CorgiBoyG:
## Description
Fixes the issue where `extraBody` parameters are lost during `ModelOptionsUtils.merge()` operation due to missing `@JsonProperty` annotation.

## Changes
- Added `@JsonProperty("extra_body")` annotation to `extraBody` parameter in `ChatCompletionRequest` constructor
- Added test case `testMergeWithExtraBody()` to verify `extraBody` is correctly merged from `OpenAiChatOptions` to `ChatCompletionRequest`

## Related Issue
Fixes #4966

## Checklist
- [x] Code follows project style
- [x] Commit messages include DCO sign-off
- [x] Added unit tests
- [x] All tests pass locally

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/api/OpenAiApi.java
@@ -1160,7 +1160,7 @@ public class OpenAiApi {
 			@JsonProperty("verbosity") String verbosity,
 			@JsonProperty("prompt_cache_key") String promptCacheKey,
 			@JsonProperty("safety_identifier") String safetyIdentifier,
-			Map<String, Object> extraBody) {
+			@JsonProperty("extra_body") Map<String, Object> extraBody) {
 
 		/**
 		 * Compact constructor that ensures extraBody is initialized as a mutable HashMap
--- a/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/api/ExtraBodySerializationTest.java
+++ b/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/api/ExtraBodySerializationTest.java
@@ -22,6 +22,8 @@ import java.util.Map;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.jupiter.api.Test;
 
+import org.springframework.ai.model.ModelOptionsUtils;
+import org.springframework.ai.openai.OpenAiChatOptions;
 import org.springframework.ai.openai.api.OpenAiApi.ChatCompletionRequest;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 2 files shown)


quaff:
It's not a correct fix, see https://github.com/spring-projects/spring-ai/pull/4936

CorgiBoyG:
> It's not a correct fix, see #4936

@quaff Thanks for the feedback! I appreciate you pointing to PR #4936.

However, I respectfully disagree that this PR is "not a correct fix." Let me explain why both approaches have their merits:

## Why PR #4975 (this PR) is a valid fix

### 1. **Root Cause Analysis**
The issue occurs because `ModelOptionsUtils.merge()` filters fields based on `@JsonProperty` annotations:

``
private static List<String> getJsonPropertyValues(Class<?> clazz) {
    return Arrays.stream(clazz.getDeclaredFields())
        .map(field -> field.getAnnotation(JsonProperty.class))  // Only @JsonProperty fields
        .filter(Objects::nonNull)
        .map(JsonProperty::value)
        .toList();
}
``

Problem: ChatCompletionRequest.extraBody parameter lacked @JsonProperty("extra_body"), causing it to be filtered out during merge.
Solution: Add @JsonProperty("extra_body") → merge preserves the field ✅

### 2. **Serialization is NOT affected**
@JsonAnyGetter takes precedence over @JsonProperty, so the JSON output remains flat:

``
@JsonAnyGetter
public Map<String, Object> extraBody() {
    return this.extraBody;
}
``

This ensures backward compatibility - the HTTP request body is still flattened as expected.

## Why PR #4936 is also valuable (but different scope)
PR #4936's @MergeableField annotation is an architectural improvement that:

- ✅ Provides explicit control over merge behavior
- ✅ Decouples merge logic from serialization annotations
- ✅ Could benefit future development

However, it:

- ⚠️ Changes core merge infrastructure
- ⚠️ Requires more extensive testing
- ⚠️ Better suited for a feature release (1.2.0) rather than a patch

## Proposed Path Forward
I suggest we:
1. Merge PR #4975 into 1.1.1 as an urgent hotfix (users currently cannot use Qwen3's thinking chain feature)
2. Continue discussing PR #4936 as a separate architectural improvement for 1.2.0+
3. Both PRs can coexist - #4936 can build upon #4975's fix

This way:

- ✅ Users get immediate relief
- ✅ We have more time to properly evaluate the architectural changes in #4936 
- ✅ Both contributors' work is valued

@quaff @deejay1 What are your thoughts on this approach? Thanks!

markpollack:
Thanks for the very thoughtful discussion.  The merging of options is a bit of a non-intuitive area of the code base.  I thought I fixed this but forgot about the full lifecycle.   This PR and some additional changes on top of it so that merging of the extra body map with 'default options' and 'runtime options' works correctly.   there are mock tests to ensure the wireformat is as expected.  

merged on 1.1.x 0646d1e060e28cc659e39ad862283b53ff941b77
merged on main 264fc3bbe890801ccf912adcb109c805e195e717

I agree there can be a better approach, let's keep discussing on that other PR.

PR #4988, opened 2025-11-28
Support compiling with Spring 7 in Spring AI 1.1

filiphr:
Use methods from HttpHeaders which are available both from Spring 6 and Spring 7 Make sure that the entire Spring AI compiles with Spring 7

* Add GitHub Worklfow that runs the suite with Spring 7
* Uncomment `@Override` from methods implementing from ResponseHandler
* Add custom ApiClient.mustache, original copied from https://github.com/swagger-api/swagger-codegen/blob/92fcc8196bd90fef199c2932903135e3b519f689/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache and adjusted the use of the HttpHeaders to be compliant with Spring 7
* Adjust ChatCLientExtensions.kt to be compatible with Kotlin 2

This PR is the same as #4817 which was closed due to PR #4774 which added Spring Boot 4.x compatibility to main. However, it would be good if Spring AI 1.1.x would do a best effort to be able to be used with Spring 7. The reason why I am asking this is due to the fact that we are using Spring AI, but we have also migrated to Spring Boot 4 and Spring 7. In order to achieve this we've had to copy the `spring-ai-openai` and `spring-ai-anthropic` modules and do the adjustments that are done in this PR. It would be great if this comes out-of-the-box from Spring AI itself and we can remove our copy.

Thank you in advance for considering the PR

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -24,3 +24,22 @@ jobs:
       - name: Run tests
         run: |
           ./mvnw --batch-mode test
+  spring_7:
+    name: Compile with Spring 7
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'spring-projects' }}
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v4
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'maven'
+
+      - name: Run tests
+        run: |
+          ./mvnw --batch-mode -Dspring-framework.version=7.0.1 -Dkotlin.version=2.2.0 compile
+    
--- a/auto-configurations/common/spring-ai-autoconfigure-retry/src/main/java/org/springframework/ai/retry/autoconfigure/SpringAiRetryAutoConfiguration.java
+++ b/auto-configurations/common/spring-ai-autoconfigure-retry/src/main/java/org/springframework/ai/retry/autoconfigure/SpringAiRetryAutoConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.ai.retry.autoconfigure;
 
 import java.io.IOException;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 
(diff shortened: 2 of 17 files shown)


ilayaperumalg:
@filiphr Thanks very much for the PR and the details. 

I agree on the proposed compatibility on Spring Framework at the best level basis. I will bring this up with the team and update. 
On the first look, the changes look good as well.

ilayaperumalg:
@filiphr Please be aware that our community should not get a false impression that Spring AI 1.1.x supports Spring Framework 7.x as we know, this compatibility might get broken anytime with any potential breaking changes in the code. 

Spring AI 2.x should be the only Spring Framework 7.x compatible/reliable option going forward.

I will update the team consensus on this on Monday.

filiphr:
>Please be aware that our community should not get a false impression that Spring AI 1.1.x supports Spring Framework 7.x as we know, this compatibility might get broken anytime with any potential breaking changes in the code.


I 100% agree with you here @ilayaperumalg. It's only a best effort, there are no guarantees that it works or anything like that. 

>Spring AI 2.x should be the only Spring Framework 7.x compatible/reliable option going forward.

Our plan is to upgrade to Spring AI 2.x as soon as it becomes available.

ilayaperumalg:
@filiphr Would you be fine with 2.0.0-M1 when it gets released instead of merging this PR?

filiphr:
I doubt that we are going to go with 2.0.0-M1 now, we are doing our own release a week after your release, so it will be a big tight. 

May I ask what the reason for not merging this to 1.1 would be? Are you concerned about something in particular?

ilayaperumalg:
@filiphr We have decided not to go ahead with the PR for the following reasons: 

 - This gives a false hope to the community that there is some support for Spring Framework 7 from Spring AI 1.1.1.
 - Even though your PR addresses the current breaking changes, we would be obligated to address any future PRs with any breaking changes from the community by accepting this PR. 
 - We want to make sure the community switches to the official Spring AI version 2.0 once it is available for the Spring Framework 7 and Spring Boot 4.0 compatibility. 

Sorry for the inconvenience and thanks again for the continued support!

PR #4994, opened 2025-12-01
Change compiler config to not fork

ericbottard:
This commit moves the module opening options to a jvm.config file, which applies to the JVM running Maven itself, and avoids forking a new JVM for compilation. This makes the build faster.

The toolchains plugin is thus removed in favor of the enforcer plugin to ensure a minimum Java version for the maven process itself.


markpollack:
merged in bd7d4cd08e0e8abdf23ba2552dcc827d2210ea2a

PR #5001, opened 2025-12-01
feat(vertex-gemini): Add safety ratings to response metadata

### LINKED ISSUE #687: VertexAiGeminiChatClient loses information about Safety Ratings when returning ChatResponse
**Expected Behavior**

I would like to be able to extract SafetyRatings from ChatResponse returned by VertexAiGeminiChatClient.call(Prompt prompt).
GenerateContentResponse contains list of SafetyRatings on each Candidate, so the data is already obtained in the library.

image

**Current Behavior**

VertexAiGeminiChatClient.call(Prompt prompt) returns ChatResponse which does not have information about SafetyRatings returned from Gemini.

**Context**

I'm not sure how many LLM models returns similar information, but it would be a useful feature to store it in generic format in Generation objects of ChatResponse, instead of loosing it in this abstraction layer.

markpollack:
## Summary
- Extract safety ratings from Gemini response candidates and add to AssistantMessage metadata
- Add `VertexAiGeminiSafetyRating` record with HarmCategory, HarmProbability, and HarmSeverity enums
- Update Google Cloud BOM to 26.72.0
- Add integration test for safety ratings
- Add documentation for safety settings and safety ratings

Closes #687

### DIFF
--- a/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatModel.java
+++ b/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatModel.java
@@ -86,6 +86,7 @@ import org.springframework.ai.support.UsageCalculator;
 import org.springframework.ai.tool.definition.ToolDefinition;
 import org.springframework.ai.vertexai.gemini.api.VertexAiGeminiApi;
 import org.springframework.ai.vertexai.gemini.common.VertexAiGeminiConstants;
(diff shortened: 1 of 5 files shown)


markpollack:
should be in both 1.1.x.  I am not sure about main, since vertex is going away in june 2026 - https://docs.cloud.google.com/vertex-ai/generative-ai/docs/deprecations  maybe we just put it in 1.1.x.  I think we are missing this functionality in the genai sdk module however.  I realize i made this PR against main, thought of this afterwards

sobychacko:
Merged to `main` via https://github.com/spring-projects/spring-ai/commit/2710cab12ef613ff433f4aeaabcc9f5aa5ef0ee8 and cherry-picked to `1.1.x`.

PR #5004, opened 2025-12-02
Make ToolCallAdvisor extensible with hook methods

tzolov:
- Add protected doInitializeLoop, doBeforeCall, and doAfterCall hooks to allow subclasses to customize the tool calling loop behavior.
- Update Builder to support inheritance via self-referential generics.


tzolov:
This self-contained change can be ported to the 1.1.1 as well

ilayaperumalg (approved):

tzolov:
backported to 1.1.x

PR #5005, opened 2025-12-02
Fix TokenTextSplitter for punctuation marks handling

### LINKED ISSUE #4981: TokenTextSplitter.split() splits small text into multiple chunks if there is no ./?/!/\n at the end.
spring ai version: 1.1.0

**Steps to reproduce**
```java
    TokenTextSplitter splitter = TokenTextSplitter.builder()
        .withKeepSeparator(true)
        .withChunkSize(10000)
        .withMinChunkSizeChars(10)
        .build();

    Document testDoc = new Document("Hi. This is a small text without one of the ending chars. It is splitted into multiple chunks but shouldn't");
    List<Document> splitted = splitter.split(testDoc);
    for (Document doc : splitted) {
      System.out.println("content: " + doc.getText());
    }

```

console output:
content: Hi. This is a small text without one of the ending chars.
content: It is splitted into multiple chunks but shouldn't

**Expected behavior**
only 1 chunk returned (no split because token count is small)

console output:
content: Hi. This is a small text without one of the ending chars. It is splitted into multiple chunks but shouldn't

ilayaperumalg:
- The TokenTextSplitter was incorrectly splitting small text into multiple chunks when punctuation marks were present, even when the entire text was well below the configured chunk size.

 - Only apply punctuation-based truncation when the remaining tokens exceed the chunk size (tokens.size() > chunkSize). This ensures small texts remain as single chunks while preserving correct splitting behavior for larger texts.

Fixes #4981

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/transformer/splitter/TokenTextSplitter.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/transformer/splitter/TokenTextSplitter.java
@@ -108,13 +108,18 @@ public class TokenTextSplitter extends TextSplitter {
 				continue;
 			}
 
-			// Find the last period or punctuation mark in the chunk
-			int lastPunctuation = Math.max(chunkText.lastIndexOf('.'), Math.max(chunkText.lastIndexOf('?'),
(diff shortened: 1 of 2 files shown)


markpollack:
merged 

  main:  e065965ef
  1.1.x: 8cc4ea4d1

PR #5006, opened 2025-12-02
Adapt Kotlin compiler version to using JDK 25

ericbottard:
This PR updates the kotlin compiler plugin to be compatible with JDK 25.

This harmonizes build infrastructures across branches and removes the need to switch back and forth between JDKs


ilayaperumalg:
Rebased and merged into 1.1.x via 62b32902b

PR #5009, opened 2025-12-02
Remove default temperature option from the models

ilayaperumalg:
- remove default temperature option from the models' autoconfiguration properties


markpollack:
updated docs where 0.7 was mentioned and update notes section.

merged in ce1a7fc6e9560b6f719efe00f1cb3241e8d2b398

PR #5020, opened 2025-12-03
Fix MistralAiApi to handle thinking content blocks

### LINKED ISSUE #5019: MistralAiApi fails to parse thinking content from Magistral reasoning models
**Bug description**
When using Mistral's Magistral reasoning models (e.g., magistral-medium-2509), the ChatCompletionMessage.content() method throws an IllegalStateException: The content is not a string! because Magistral models return the content field as an array of content blocks instead of a plain string.

The reasoning models return a structured response with separate thinking and text blocks, but the current implementation only handles string content.

**Environment**
- Spring AI version: 1.0.x / 1.1.0-Mx
- Java version: 21
- Mistral model: magistral-medium-latest (magistral-medium-2509)

**Steps to reproduce**
1. Configure Spring AI with Mistral AI provider
2. Use a Magistral reasoning model (magistral-medium-latest or magistral-small latest)
3. Send any chat completion request
4. Call .content() on the response message

**Expected behavior**
The content() method should extract and return the final answer from the text content block, ignoring the thinking block.

```json
  {
    "choices": [
      {
        "message": {
          "role": "assistant",
          "content": [
            {
              "type": "thinking",
              "thinking": [
                {
                  "type": "text",
                  "text": "Reasoning traces..."
                }
              ]
            },
            {
              "type": "text",
              "text": "Final answer..."
            }
          ]
        }
      }
    ]
  }
```
Leads to: 
Caused by: java.lang.IllegalStateException: The content is not a string!

org.springframework.ai.mistralai.api.MistralAiApi$ChatCompletionMessage.content(MistralAiApi.java:~1425)

KyleKreuter:
Fixes #5019

## Summary
Mistral's Magistral reasoning models return `content` as an array of content blocks
instead of a plain string.

## Changes
- Update `content()` to extract the `text` block from thinking responses

## Tested with
- Model: `magistral-medium-2509`
- Response format verified with direct API call

### DIFF
--- a/models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiAssistantMessage.java
+++ b/models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiAssistantMessage.java
@@ -1,0 +1,181 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.mistralai;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+import org.springframework.ai.chat.messages.AssistantMessage;
+import org.springframework.ai.content.Media;
+
+/**
+ * A Mistral AI specific implementation of {@link AssistantMessage} that supports
+ * additional fields returned by Magistral reasoning models.
+ *
+ * <p>
+ * Magistral models (like magistral-medium-latest and magistral-small-latest) return
+ * thinking/reasoning content alongside the regular response content. This class captures
+ * both the final response text and the intermediate reasoning process.
+ * </p>
+ *
+ * @author Kyle Kreuter
+ * @since 1.1.0
+ */
+public class MistralAiAssistantMessage extends AssistantMessage {
+
+	/**
+	 * The thinking/reasoning content from Magistral models. This contains the model's
+	 * intermediate reasoning steps before producing the final response.
+	 */
+	private String thinkingContent;
+
+	/**
+	 * Constructs a new MistralAiAssistantMessage with all fields.
+	 * @param content the main text content of the message
+	 * @param thinkingContent the thinking/reasoning content from Magistral models
+	 * @param properties additional metadata properties
+	 * @param toolCalls list of tool calls requested by the model
+	 * @param media list of media attachments
+	 */
+	protected MistralAiAssistantMessage(String content, String thinkingContent, Map<String, Object> properties,
+			List<ToolCall> toolCalls, List<Media> media) {
+		super(content, properties, toolCalls, media);
+		this.thinkingContent = thinkingContent;
+	}
+
+	/**
+	 * Returns the thinking/reasoning content from Magistral models.
+	 * @return the thinking content, or null if not available
+	 */
+	public String getThinkingContent() {
+		return this.thinkingContent;
+	}
+
+	/**
+	 * Sets the thinking/reasoning content.
+	 * @param thinkingContent the thinking content to set
+	 * @return this instance for method chaining
+	 */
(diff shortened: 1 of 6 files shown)


nicolaskrier:
Hello @KyleKreuter, thanks for your contribution!

Here are a few suggestions to improve Magistral support:
- Add a `MistralAiAssistantMessage` class (like `ZhiPuAiAssistantMessage`) to handle both `content` and `reasoning` fields.
- Replace `List`/`Map` in `MistralAiApi` with dedicated classes: `ThinkChunk`, `ReferenceChunk`, and `TextChunk` ([Mistral AI API documentation](https://docs.mistral.ai/api#operation-chat_completion_v1_chat_completions_post)).
- Support the `reasoning` prompt mode as described in the API documentation (not related to the current issue).
- Add unit/integration tests to validate these features.

Happy to discuss or clarify if needed!

KyleKreuter:
Hi, thanks for the feedback!

I've implemented the suggested improvements:

1. MistralAiAssistantMessage class
  - Created MistralAiAssistantMessage extending AssistantMessage with a thinkingContent field
  - Follows the same pattern as ZhiPuAiAssistantMessage and DeepSeekAssistantMessage
  - Includes Builder pattern, proper equals()/hashCode()/toString() implementations

2. Dedicated chunk classes
  - Added sealed interface ContentChunk with three implementations:
    - TextChunk - for text content blocks
    - ThinkChunk - for thinking/reasoning content blocks
    - ReferenceChunk - for citation references
  - Updated ChatCompletionMessage with content(), thinkingContent(), and contentChunks() methods
  - Note: The Mistral API returns thinking content in a nested format ({"type": "thinking", "thinking": [{"type": "text", "text": "..."}]}), which is now properly parsed

4. Unit/Integration tests
  - MistralAiAssistantMessageTests - 23 test cases for the message class
  - MistralAiContentParsingTests - 25 test cases for content parsing
  - MistralAiMagistralIT - 6 integration tests with real Magistral API calls

Please let me know if you have any further suggestions.

KyleKreuter:
Hi @ilayaperumalg! Just checking in on this PR – is there anything blocking the review, or anything I can help with? Would be great to know a rough timeline. Thanks!

nicolaskrier:
Hi @KyleKreuter and @ilayaperumalg, I spent some time digging into Magistral 1.2 support and made a few improvements based on my findings. You can check out the details in [this commit](https://github.com/nicolaskrier/spring-ai/commit/cf26ccd72b3f85b83edc27388e62e65db9d0c29c). FYI, I removed the unit tests from Kyle to fix quickly compilation errors and also because I believe a better approach could be found now to verify directly JSON serialization and deserialization. HTH!

KyleKreuter:
Hi @nicolaskrier,                                                                   
                                                                              
thanks for taking the time to look at this. I have a few questions about the  
changes:                                                                      
                                                                              
1. Tests: I noticed my 332 lines of unit tests were removed. I don't see any  
replacement tests for the new ContentSerializer and ContentDeserializer       
classes. Could you add tests that cover the same edge cases? Things like null 
content, empty arrays, unknown chunk types, multiple text chunks              
concatenation, etc. — these were all covered before.                          
2. Rationale: I'm curious about the motivation for the Jackson serializer     
approach. The previous implementation used simple instanceof checks in the    
accessor methods, which seemed sufficient for handling the String |           
List<ContentChunk> union type. The new approach adds ~50 lines of serializer  
code but rawContent is still typed as Object, so I'm not sure what we gain    
here. Could you help me understand the benefit?                               
3. Breaking change: ThinkChunk.thinking changed from String to                
List<ContentChunk>. Was this intentional? It would break existing consumers.  
                                                                              
Happy to discuss!

nicolaskrier:
Hi @KyleKreuter,

I spent more time on the JSON serialization/deserialization, and while the integration tests are passing, there are still a few challenges to address before I reach a robust solution:

1. Tests: As mentioned earlier, I haven’t yet added the unit tests I’d like to see (specifically, tests using real Mistral AI JSON responses and edge cases). I wanted to share my progress early (finished late last night), but I’ll add these later.
2. Rationale: The original design was described in the Javadoc (using a `String` or a list of `MediaContent` for `ChatCompletionMessage#rawContent`) but it seems it was not implemented for the deserialization. However, I noticed `MediaContent` can be replaced by `ContentChunk` now. To align closer with the [Mistral AI OpenAPI Specification](https://github.com/mistralai/platform-docs-public/blob/main/openapi.yaml), I adjusted the modeling in [this commit](https://github.com/spring-projects/spring-ai/commit/be3762932123d90937d29228c41f95775a4cc411).
3. Breaking change: This is intentional: the goal is to mirror the Mistral AI OpenAPI specification as closely as possible. You're right, there were some broken tests with my former commit. I fixed them with my latest changes.

Let me know your thoughts!

sdeleuze:
Closed as discussed in https://github.com/spring-projects/spring-ai/pull/5585#issuecomment-4556766086

Issue #5024, opened 2025-12-04
Zhipu AI model fails to retrieve usage info from ChatResponse upon chat completion


yunqilee69:
**Bug description**
Zhipu AI model fails to retrieve usage info from ChatResponse upon chat completion

**Environment**
Spring AI： 1.1.0

**Minimal Complete Reproducible example**
When invoking the Zhipu chatModel class directly via its call() or stream() methods, the usage information is unavailable in the returned response.

yunqilee69:
I’ve submitted a PR to fix the issue; this is my first contribution, so please let me know if I’ve done anything wrong.
[PR Link](https://github.com/spring-projects/spring-ai/pull/5026)

sdeleuze:
We are stopping to maintain ZhiPuAI as part of the core Spring AI project as we have no proper way to test it, and have to focus on a smaller set of models. We have not been able to find a maintainer for a support outside of this repository, but if somebody want to volunteer, please reach us on https://github.com/spring-projects/spring-ai/discussions.

PR #5026, opened 2025-12-04
fix: Add Usage with the ZhiPu model(#5024)

yunqilee69:
Fixes https://github.com/spring-projects/spring-ai/issues/5024

Due to the fact that the usage field is not correctly populated when converting ChatCompletionChunk to ChatCompletion, users always receive EmptyUsage when attempting to retrieve usage information from the ChatResponse.


YunKuiLu:
@yunqilee69 Could you please update the unit tests accordingly?

Issue #5029, opened 2025-12-04
Using OpenAI's call method for invocation, the reasoning content cannot be retrieved from the final ChatResponse


yunqilee69:
**Bug description**
Using OpenAI's call method for invocation, the reasoning content cannot be retrieved from the final ChatResponse

**Environment**
- Spring AI version: 1.1.0
- Java version: 17

**Expected behavior**
When invoking with OpenAI's call method, reasoning content can be retrieved from the ChatResponse metadata, just like with the stream method invocation.

**Minimal Complete Reproducible example**
```java
OpenAiApi openAiApi = OpenAiApi.builder()
                .baseUrl("https://open.bigmodel.cn/api/paas")
                .completionsPath("/v4/chat/completions")
                .apiKey("your ZhiPu Ai apikey")
                .build();

        // GLM-4.6 has thinking content enabled by default.
        OpenAiChatOptions chatOptions = OpenAiChatOptions.builder()
                .model("glm-4.6")
                .extraBody(Map.of(
                        "thinking", Map.of("type", "enabled")
                ))
                .build();

        OpenAiChatModel chatModel = OpenAiChatModel.builder()
                .openAiApi(openAiApi)
                .defaultOptions(chatOptions)
                .build();

        Prompt prompt = new Prompt("Tell me, who are you");

        ChatResponse chatResponse = chatModel.call(prompt);

        // The reasoning content cannot be successfully retrieved from this metadata.
        Map<String, Object> metadata = chatResponse.getResult().getOutput().getMetadata();
        System.out.println(metadata);


such as:
{role=ASSISTANT, messageType=ASSISTANT, finishReason=STOP, refusal=, annotations=[{}], index=0, id=20251204154814d4axxxxxx}
```

sobychacko:
See https://github.com/spring-projects/spring-ai/issues/5533 for more details.

PR #5030, opened 2025-12-04
fix: Add the reasoningContent field to the metadata（#5029）.

yunqilee69:
fix: https://github.com/spring-projects/spring-ai/issues/5029

In the call method of OpenAIChatModel, add a reasoningContent field to the metadata, keeping it consistent with how metadata is handled in the stream method.


yunqilee69:
@ilayaperumalg Hello, I would like to know when this PR can be merged. I see that this issue has not been resolved in version 1.1.5.I would really appreciate it if you could reply and let me know when to expect an answer.

yunqilee69:
image
In streaming calls, the returned metadata includes the reasoningContent field, but in synchronous calls, the returned metadata does not include the reasoningContent field. This PR adds that field to the metadata returned in synchronous calls.

Issue #3924, opened 2025-07-28
MethodToolCallback does not handle invalid arg gracefully


ls-rein-martha:
**Bug description**

When LLM intent to call a tool, there is possibility that the argument is invalid.
For example:
- The tool argument is an enum
- The LLM provide invalid enum
When this occur it should throw `ToolExecutionException` instead of `IllegalArgumentException`

Stack trace:
```
java.lang.IllegalArgumentException: No enum constant __INVALID_ENUM_NAME__
	at java.base/java.lang.Enum.valueOf(Enum.java:293)
	at org.springframework.ai.util.json.JsonParser.toTypedObject(JsonParser.java:152)
	at org.springframework.ai.tool.method.MethodToolCallback.buildTypedArgument(MethodToolCallback.java:150)
	at org.springframework.ai.tool.method.MethodToolCallback.lambda$buildMethodArguments$1(MethodToolCallback.java:139)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:1024)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:575)
```

**Environment**
Spring AI 1.0.0
Java 17
Azure OpenAI gpt-4o

**Steps to reproduce**
Create a tool that accepts an enum as a parameter, let's say:
- ONE_DAY
- TWO_DAY
- THREE_DAY

If you ask LLM for FOUR_DAY or similar, it might try to call the tool, even if the input schema doesn't list it so

**Expected behavior**
It should throw `ToolExecutionException` so it can be sent back to model, or handle gracefully at `ToolExecutionExceptionProcessor`. Or it might even be a good idea to change `ToolExecutionExceptionProcessor` process to all Exception instead of `ToolExecutionException` so other possible edge cases can be handled gracefully easily

Workaround (create a custom `MethodToolCallback` with a delegate):

```

@RequiredArgsConstructor
public class CustomMethodToolCallback implements ToolCallback {

    @Delegate
    private final MethodToolCallback delegate;

    @Override
    public String call(String toolInput, @Nullable ToolContext toolContext) {
        try {
[...]

Issue #4987, opened 2025-11-28
ToolExecutionException does not fully cover exception scenarios, such as IllegalStateException or JsonProcessingException


internCoder:
**Bug description**
There is a serialization issue before calling the model tool, which can cause the tool to not be called properly, and the DefaultToolExecutionExceptionProcessor cannot be processed through configuration

**Environment**
Spring AI 1.1.0
Java 21

**Steps to reproduce**
org.springframework.ai.tool.ToolCallback#call
org.springframework.ai.tool.method.MethodToolCallback#extractToolArguments
org.springframework.ai.util.json.JsonParser#fromJson
com.fasterxml.jackson.databind.ObjectMapper#readValue

The input text is not JSON, resulting in an IllegalStateException being thrown

<img width="1530" height="746" alt="Image" src="https://github.com/user-attachments/assets/1a3f03cb-1c8c-4184-83cc-92ca38b76e19" />

java.lang.IllegalStateException: Conversion from JSON to java.util.Map<java.lang.String, java.lang.Object> failed
	at org.springframework.ai.util.json.JsonParser.fromJson(JsonParser.java:97)
	at org.springframework.ai.tool.method.MethodToolCallback.extractToolArguments(MethodToolCallback.java:128)
	at org.springframework.ai.tool.method.MethodToolCallback.call(MethodToolCallback.java:105)
	at org.springframework.ai.model.tool.DefaultToolCallingManager.lambda$executeToolCall$5(DefaultToolCallingManager.java:241)
	at io.micrometer.observation.Observation.observe(Observation.java:564)
	at org.springframework.ai.model.tool.DefaultToolCallingManager.executeToolCall(DefaultToolCallingManager.java:238)
	at org.springframework.ai.model.tool.DefaultToolCallingManager.executeToolCalls(DefaultToolCallingManager.java:143)
	at org.springframework.ai.openai.OpenAiChatModel.lambda$internalStream$9(OpenAiChatModel.java:374)
	at reactor.core.publisher.FluxDeferContextual.subscribe(FluxDeferContextual.java:49)
	at reactor.core.publisher.FluxSubscribeOn$SubscribeOnSubscriber.run(FluxSubscribeOn.java:194)
	at reactor.core.scheduler.WorkerTask.call(WorkerTask.java:84)
	at reactor.core.scheduler.WorkerTask.call(WorkerTask.java:37)
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:317)
[...]

PR #3933, opened 2025-07-28
Issue 3924 Fix MethodToolCallback enum argument handling

wilocu:
## Summary
  Fixes MethodToolCallback not handling invalid enum arguments gracefully by converting
  IllegalArgumentException to ToolExecutionException.

  Closes #3924
  ## Changes Made
  - **Exception Handling**: Added try-catch block around `buildMethodArguments()` in
  `MethodToolCallback.call()`
  - **Exception Conversion**: Convert `IllegalArgumentException` to `ToolExecutionException` with
  proper context
  - **Logging**: Added debug logging for argument conversion failures following existing patterns
  - **Test Coverage**: Added comprehensive unit tests for both invalid and valid enum scenarios  

  ## Technical Details
  - **Root Cause**: `IllegalArgumentException` was thrown during enum conversion in
  `JsonParser.toTypedObject()` but not caught by `MethodToolCallback`
  - **Solution**: Wrap argument building phase with exception handling at lines 108-115 in
  `MethodToolCallback.java`
  - **Integration**: Maintains compatibility with existing `ToolExecutionExceptionProcessor` pipeline 

## Test Results
  - ✅ New unit tests pass (invalid enum throws `ToolExecutionException`)
  - ✅ Existing functionality preserved (valid enum values work normally)
  - ✅ No regressions in `MethodToolCallbackProviderTests` and `DefaultToolCallingManagerTests`

  ## Files Modified
  - `spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallback.java`
  - `spring-ai-model/src/test/java/org/springframework/ai/tool/method/MethodToolCallbackTests.java`       
  (new file)

  This change allows LLMs that provide invalid enum values to receive graceful error responses
  instead of unhandled exceptions.
Signed-off-by:  Mattia Pasetto matpat17@gmail.com

Please make sure to double check and let me know if anything is wrong, this is my first contribution.

Thank you for taking time to contribute this pull request!
You might have already read the [contributor guide][1], but as a reminder, please make sure to:

PR #5032, opened 2025-12-04
fix: Improve JSON error handling in MethodToolCallback

tzolov:
Add proper exception handling for JSON processing failures in tool argument extraction and conversion. Wrap JsonProcessingException in ToolExecutionException to provide better error context when tools receive malformed input.

Resolves: #3924 #4987 #3933

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallback.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallback.java
@@ -23,6 +23,7 @@ import java.lang.reflect.Type;
 import java.util.Map;
 import java.util.stream.Stream;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallback.java
```diff
@@ -125,11 +128,21 @@ private void validateToolContextSupport(@Nullable ToolContext toolContext) {
 	}
 
 	private Map<String, Object> extractToolArguments(String toolInput) {
-		return JsonParser.fromJson(toolInput, new TypeReference<>() {
-		});
+		try {
+			return JsonParser.fromJson(toolInput, new TypeReference<>() {
+			});
+		}
+		catch (IllegalStateException ex) {
```
ilayaperumalg, line 135: Instead of checking just the IllegalStateException, we agreed to check all exception which is similar to how MCP toolcallback error handling is done. I will make the change and update the PR.

ilayaperumalg:
Rebased and merged as 8e972a47fbed59427c2995357ea5ad37bc412bba along with 65429d92d242dfb7591b1dd8de49602eb9376d4b. Also, cherry-picked into 1.1.x via d4f22cec9f49253ef4f0e0e91c51e67fe0a85365 along with b0a97c767f88e76cbd5ff01df3e4d4ab9533d675

PR #5036, opened 2025-12-04
Revisit workflows

ericbottard:


ilayaperumalg:
Nice refactoring of the workflows! Looks great

PR #5062, opened 2025-12-08
Fine tune the BuildCache config

ericbottard:
This makes sure that the following plugins always run (because we're interested in their side effects):
 * install
 * deploy
 * flatten

Also, silence a warning about an arg of the enforcer plugin. (harmless)


ilayaperumalg (approved):

ilayaperumalg:
Rebased and merged as cd76cf923

PR #5070, opened 2025-12-09
Updated Google GenAI SDK to 1.30.0 and updated tests due to deprecation of an embeddingmodel

ddobrin:
These changes are required for
- Native Image support - mandatory upgrade to 1.30.0
- Deprecation of text-embedding-004 on Jan 14, 2026

Could this be merged into the 2.0.0 branch and backported to a 1.1.x fixes branch?

cc: @markpollack @ilayaperumalg @tzolov

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -342,8 +342,8 @@
 		<json-unit-assertj.version>4.1.0</json-unit-assertj.version>
 		
 		<!-- MCP-->
-		<mcp.sdk.version>0.17.0-SNAPSHOT</mcp.sdk.version>
-		<mcp-annotations.version>0.8.0-SNAPSHOT</mcp-annotations.version>
+		<mcp.sdk.version>0.17.0</mcp.sdk.version>
+		<mcp-annotations.version>0.8.0</mcp-annotations.version>
 
 		<!-- plugin versions -->
 		<antlr.version>4.13.1</antlr.version>


models/spring-ai-google-genai-embedding/src/main/java/org/springframework/ai/google/genai/text/GoogleGenAiTextEmbeddingOptions.java
```diff
@@ -34,7 +34,7 @@
 @JsonInclude(Include.NON_NULL)
 public class GoogleGenAiTextEmbeddingOptions implements EmbeddingOptions {
 
-	public static final String DEFAULT_MODEL_NAME = GoogleGenAiTextEmbeddingModelName.TEXT_EMBEDDING_004.getName();
+	public static final String DEFAULT_MODEL_NAME = GoogleGenAiTextEmbeddingModelName.GEMINI_EMBEDDING_001.getName();
```
markpollack: usually we should not change the default for the 1.1.x branch.  the TEXT_EMBEDDING_004 is going away next month @ddobrin so this makes it an awkward situation for 1.1.x users, in partiuclar as there are dimension changes.

markpollack:
suggestions

- Fix the typo in documentation: change 'gemini-enmbedding-001' to 'gemini-embedding-001' in the deprecation note (line 199 of the .adoc file)
- Add a prominent migration guide section to the documentation explaining how users with existing text-embedding-004 embeddings should handle the dimension change when upgrading
- Consider adding a deprecation warning log message when text-embedding-004 is explicitly specified, alerting users to the January 2026 sunset date
- Clarify in documentation which model should be used for which use case: text-embedding-005 (768 dims) vs gemini-embedding-001 (3072 dims) and the trade-offs
- Add a test case that explicitly validates backward compatibility when text-embedding-004 is still used (before Jan 2026) to ensure the transition period works correctly
- Document the batch size limitation of gemini-embedding-001 in the main properties table, not just in code comments, as this has significant performance implications

models/spring-ai-google-genai-embedding/src/main/java/org/springframework/ai/google/genai/text/GoogleGenAiTextEmbeddingOptions.java
ddobrin: I think in this case the default should change to text_embedding_005, with the same dimensions

ddobrin:
@markpollack - I'll update as per your suggestions first thing in the morning

PR #5083, opened 2025-12-11
[java doc] outdated embedding model documentation link in TokenCountBatchingStrategy Javadoc

leehaut:
Replaced the old, broken URL with the correct OpenAI documentation URL.

The original link:
``` txt
https://platform.openai.com/docs/guides/embeddings/embedding-models
```
the correct OpenAI
```txt
https://platform.openai.com/docs/guides/embeddings#embedding-models
```

No functional or behavioral changes to the code.


ilayaperumalg:
@leehaut Thanks for the fix.

PR #5087, opened 2025-12-12
Fix ChromaApi exception handling

ilayaperumalg:
- Uses HttpClientErrorException.NotFound for 404 responses instead of parsing error messages in getCollection, getTenant, and getDatabase methods


markpollack:
merged on main as e29171243
merged on 1.1.x as f7ed9d171

PR #5100, opened 2025-12-15
Add conversationHistoryEnabled option to ToolCallAdvisor

tzolov:
Introduce a configurable option to control how conversation history is managed during tool calling iterations in ToolCallAdvisor.

Resolves compatibility issues when using ToolCallAdvisor together with ChatMemory advisors that manage their own conversation history.

When conversationHistoryEnabled=true (default), the advisor maintains full conversation history internally during tool call iterations. When set to false, only the last tool response message is passed to the next iteration, which is useful when integrating with a Chat Memory advisor that already manages conversation history.

- Add conversationHistoryEnabled field and builder method to ToolCallAdvisor
- Add doGetNextInstructionsForToolCall() protected method for extensibility
- Add unit tests for the new configuration option
- Update documentation in advisors-recursive.adoc and tools.adoc


ilayaperumalg (approved):

ilayaperumalg:
Rebased, squashed and merged as [5ad8485](https://github.com/spring-projects/spring-ai/commit/5ad8485ec981fff59111d4fe777068ea6ec8d4e6). Also, back-ported the changes into 1.1.x via 583423a9cbc8c351039c6a3658d361d3cf22d847

PR #5104, opened 2025-12-15
Polish POM files

nicolaskrier:
- Use Maven transitivity to obtain test dependencies from spring-ai-test module
- Use Maven transitivity to obtain testcontainers dependency
- Add missing test scopes
- Use Oracle JDBC driver version managed by Spring Boot
- Use variable for Redis Testcontainers version
- Add missing license headers

### DIFF
--- a/spring-ai-test/pom.xml
+++ b/spring-ai-test/pom.xml
@@ -67,8 +67,8 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<exclusions>
 				<exclusion>
-					<groupId>com.vaadin.external.google</groupId>
-					<artifactId>android-json</artifactId>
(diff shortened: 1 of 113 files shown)


spring-ai-test/pom.xml
```diff
@@ -67,8 +67,8 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<exclusions>
 				<exclusion>
-					<groupId>com.vaadin.external.google</groupId>
-					<artifactId>android-json</artifactId>
+					<groupId>org.skyscreamer</groupId>
```
nicolaskrier, line 70: - I think this exclusion here is easiest to understand.
- Modules using jsonassert also require android-json. That is why I had to add jsonassert dependency to some modules.

PR #5136, opened 2025-12-22
Use explicit save/restore cache actions.

ericbottard:
This is an attempt at forcing a save, even in case of cache hit, as the contents of the directory may have changed.


ericbottard:
Merged and then reverted anyway.

PR #5139, opened 2025-12-23
Build-Cache can't update

ericbottard:
Turns out you can't reliably update a cache entry once a key has been reserved.
Somehow [this run](https://github.com/spring-projects/spring-ai/actions/runs/20432804489/job/58707106972) was a false positive solution 🤷‍♂️.

This PR uses the less effective but as good as we can get solution described [here](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache).


ericbottard:
Merged as b10443720ff6f442a1a1e2bd791734dd92816138

PR #5158, opened 2025-12-29
Make OllamaChatOptions a StructuredOutputChatOptions implementation

nicolaskrier:
- Implement corresponding methods
- Update documentation link
- Add unit tests in OllamaModelOptionsTests
- Add integration tests in OllamaApiIT and OllamaChatModelIT
- Polish existing tests for these classes
- Attempt to fix flaky tests

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaChatOptions.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaChatOptions.java
@@ -32,6 +32,7 @@ import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 import org.springframework.ai.model.ModelOptionsUtils;
+import org.springframework.ai.model.tool.StructuredOutputChatOptions;
 import org.springframework.ai.model.tool.ToolCallingChatOptions;
 import org.springframework.ai.tool.ToolCallback;
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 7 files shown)


models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaChatOptions.java
```diff
@@ -791,6 +793,18 @@ public void setToolContext(Map<String, Object> toolContext) {
 		this.toolContext = toolContext;
 	}
 
+	@Override
+	@JsonIgnore
+	public String getOutputSchema() {
```
ilayaperumalg, line 798: If the "format" is a simple "json" string value (as this can be a format option for Ollama model to generate any valid JSON string instead of a structured output) then, we shouldn't be converting it to a JSON string. I will fix this when merging.

ilayaperumalg:
Rebased and merged as 163df2e4fe51dfe79e4375aff814fd222f1c569b along with 3c4d174cd50e17e92e59bced0e8bdfc5ad03e9fa. Also backported into 1.1.x via a6ce0f604bda489d779a5bcd2889a5629f16db4d and 7b23ee2c94746c11af85da11ff0274df4a6449dd

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaChatOptions.java
nicolaskrier: Apologies for missing this part!

PR #5166, opened 2025-12-31
Fix Checkstyle plugin configuration.

ericbottard:
This commit correctly sets the location of the checktyle configuration file. This is particularly important to allow the build-cache plugin to correctly factor in the hash of that config file in the build cache key.

Remove properties that were not referenced or were implicit thanks to the maven checktyle <headerLocation> configuration.

Remove reporting section that is unused.

Auto-cherry-pick to 1.1.x


ilayaperumalg:
Rebased and merged into main as 647a13f79 and 1.1.x via 2338b256d

PR #5183, opened 2026-01-05
Jspecify part 3: chat-client + vector-store

ericbottard:


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientMessageAggregator.java
```diff
@@ -40,6 +40,7 @@ public class ChatClientMessageAggregator {
 
 	private static final Logger logger = LoggerFactory.getLogger(ChatClientMessageAggregator.class);
 
+	@SuppressWarnings("NullAway") // https://github.com/uber/NullAway/issues/1290
```
sdeleuze: I see you have upgraded to NullAway 0.12.13, and that https://github.com/uber/NullAway/issues/1290#issuecomment-3597659757 mentions that this bug should be fixed for non wildcard use cases. Also I don't see a `@Nullable` in those lambas.

So I am not 100% sure this is related to https://github.com/uber/NullAway/issues/1290, could you please share the error you get when you remove `@SuppressWarnings("NullAway")` ?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientRequest.java
```diff
@@ -66,19 +68,20 @@ public Builder prompt(Prompt prompt) {
 			return this;
 		}
 
-		public Builder context(Map<String, Object> context) {
+		public Builder context(Map<String, @Nullable Object> context) {
```
sdeleuze, line 71: If we want to accept both `Map<String, @Nullable Object>` and `Map<String, Object>` in null-marked code, it should be `Map<String, ? extends @Nullable Object>`. Is forcing `Map<String, @Nullable Object>` done on purpose here?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientResponse.java
```diff
@@ -31,7 +32,7 @@
  * @author Thomas Vitale
  * @since 1.0.0
  */
-public record ChatClientResponse(@Nullable ChatResponse chatResponse, Map<String, Object> context) {
+public record ChatClientResponse(@Nullable ChatResponse chatResponse, Map<String, @Nullable Object> context) {
```
sdeleuze, line 35: Same question as above, do we want `Map<String, ? extends @Nullable Object>` or `Map<String, @Nullable Object>` here?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientMessageAggregator.java
ericbottard: Changed the comment to refer to https://github.com/uber/NullAway/issues/1350 instead of https://github.com/uber/NullAway/issues/1290

The error is
```
[ERROR] /Users/bottarde/Documents/Projects/spring-ai-umbrella/spring-ai/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientMessageAggregator.java:[51,25] [NullAway] returning @Nullable expression from method with @NonNull return type
    (see http://t.uber.com/nullaway )
``` 
referring to the return of `mapNotNull()`

Issue #1608, opened 2024-10-27
Stability AI does not produce observability metrics


habuma:
When I generate images with OpenAI, I can see the metrics for those generations in the Actuator's metrics endpoint. But not when generating images with Stability AI. A quick search through the Stability AI code suggests that observability has not been introduced into that model yet.

RodrigoDiasDeOliveira:
Hello whats up?! i think it maybe help, an friend of mine had that same issue, to address this issue, observability needs to be integrated into the Stability AI model to track image generation metrics, similar to how OpenAI's model metrics are visible in the Actuator's metrics endpoint.

Steps to Enable Observability for Stability AI Image Generation
Implement Metrics Logging:

Add custom metrics logging for Stability AI image generation requests. This involves defining key performance indicators (KPIs) such as generation time, number of requests, errors, and other relevant metrics to track.
Use a monitoring framework like Micrometer or Prometheus, which can be configured to expose metrics through the Actuator’s metrics endpoint.
Integrate with Actuator's Metrics Endpoint:

Update the Stability AI service code to push these metrics to the Actuator's metrics endpoint. Ensure that these metrics are tagged appropriately (e.g., stability-ai.image-generation) to distinguish them from other services.
Test Observability:

Once observability is implemented, verify that Stability AI image generation metrics appear correctly in the Actuator's endpoint. This will involve generating test images with Stability AI and confirming that metrics data is recorded and visible in your monitoring dashboard.
Documentation Update:

Update the project documentation to include the new metrics being tracked and instructions on accessing them. Specify any new configuration steps required for enabling observability in the Stability AI service.
By following these steps, you can monitor image generation metrics from Stability AI in a way that aligns with the existing observability setup for OpenAI, providing a comprehensive view of performance across both models.

i hope it can helps

PR #5188, opened 2026-01-05
Add observability metrics support for Stability AI image generation

jee14:
Relates to: #1608

## Summary
- Fix deprecated `DEFAULT_IMAGE_MODEL` (`stable-diffusion-v1-6` → `stable-diffusion-xl-1024-v1-0`)
- Add `ObservationRegistry` support to `StabilityAiImageModel` for metrics/tracing
- Add `STABILITY_AI` to `AiProvider` enum
- Add `ImageModel` enum to `StabilityAiApi` for type-safe model selection
- Add integration test `StabilityAiImageModelObservationIT`

## Changes
Following the same pattern as `OpenAiImageModel`:
- `StabilityAiImageModel`: Wrap API call with observation instrumentation
- `StabilityAiImageAutoConfiguration`: Inject `ObservationRegistry` and `ImageModelObservationConvention`
- `AiProvider`: Add `STABILITY_AI("stability_ai")` enum value

## Test
- Added `StabilityAiImageModelObservationIT` (requires `STABILITYAI_API_KEY` env var)


sobychacko:
Thanks for this @jee14, and for getting to #1608 first. It overlaps a lot with #6229, which is scoped just to the observability wiring and has a unit test that mocks `StabilityAiApi` (yours is an integration test that needs a real API key). #6229 also uses `stabilityAiImageProperties.toOptions()`, which matches current `main`, so we are going to go with that one and close this.

One thing worth keeping from your PR though: the deprecated default model change (`stable-diffusion-v1-6` to `stable-diffusion-xl-1024-v1-0`) plus the docs update. That is worth keeping and separate from the observability issue, so if you want to open a focused PR for just that, we can review it separately. Thanks again for contributing, and hope to see more from you.

PR #5207, opened 2026-01-08
Jspecify 6: Retry, MCP, Cloud bindings modules.

ericbottard:


mcp/common/src/main/java/org/springframework/ai/mcp/McpConnectionInfo.java
```diff
@@ -28,9 +29,9 @@
  * @author Christian Tzolov
  */
 public record McpConnectionInfo(// @formatter:off
-	McpSchema.ClientCapabilities clientCapabilities,
-	McpSchema.Implementation clientInfo,
-	McpSchema.InitializeResult initializeResult) { // @formatter:on
+	McpSchema.@Nullable ClientCapabilities clientCapabilities,
+	McpSchema.@Nullable Implementation clientInfo,
```
Kehrlann: `clientCapababilities` and `clientInfo` are defined client-side, and IIRC, should not be null.

`initializeResult`, on the other hand, can be null.

Kehrlann (changes requested):
Looks good on the MCP side of things, one change.
I have not reviewed Retry.

mcp/common/src/main/java/org/springframework/ai/mcp/McpConnectionInfo.java
ericbottard: Fixed by doing the correct thing and updating tests: e4b590b40005772cb51c606f06c085372e49e598

PR #5223, opened 2026-01-12
[Minor enhancement] Add Prompt#getSystemMessages() for convenience

sebin1213:
Small convenience enhancement for `org.springframework.ai.chat.prompt.Prompt`.

`Prompt` already provides:
- `getSystemMessage()` (returns the first system message, or an empty one)
- `getUserMessage()` (returns the last user message, or an empty one)
- `getUserMessages()` (returns all user messages)

But there is no accessor to retrieve **all** `SystemMessage`s when multiple are present.

## Proposed change
Add:

```java
public List<SystemMessage> getSystemMessages()
````

## Notes / Alternative
If you prefer a strict Harmony-style approach (single `SystemMessage` only), please comment. In that case, I will open a follow-up PR that adds logic to **enforce/normalize a single system message** upstream instead of introducing `getSystemMessages()`.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/Prompt.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/Prompt.java
@@ -140,6 +140,20 @@ public class Prompt implements ModelRequest<List<Message>> {
 	}
 
 	/**
+	 * Get all system messages in the prompt.
+	 * @return a list of all system messages in the prompt
+	 */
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
@sebin1213 Thanks for the PR! Rebased and merged as 8a7939041e12b3f14a32b9d69a86da4527f8ef2c along with the doc updates via 190fe98aefbf1e3fb4585d5c4f5053208be26db3. Back ported into 1.1.x as well.

PR #5224, opened 2026-01-12
Improve model provider detection for Microsoft Foundry integration

jdubois:
- Remove Azure OpenAI references
- Add a few more tests for detecting the base URL

### DIFF
--- a/models/spring-ai-openai-sdk/src/test/java/org/springframework/ai/openaisdk/setup/OpenAiSdkSetupTests.java
+++ b/models/spring-ai-openai-sdk/src/test/java/org/springframework/ai/openaisdk/setup/OpenAiSdkSetupTests.java
@@ -99,23 +99,12 @@ public class OpenAiSdkSetupTests {
 	}
 
 	@Test
-	void detectModelProvider_returnsMicrosoftFoundry_whenMicrosoftFoundryEndpointProvided() {
-		String endpoint = "https://xxx.openai.azure.com/openai/v1/";
-		String deploymentName = ChatModel.GPT_5_2.asString();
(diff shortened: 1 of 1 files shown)


models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/setup/OpenAiSdkSetup.java
```diff
@@ -43,7 +43,7 @@ public final class OpenAiSdkSetup {
 
 	static final String OPENAI_URL = "https://api.openai.com/v1";
 	static final String OPENAI_API_KEY = "OPENAI_API_KEY";
-	static final String AZURE_OPENAI_KEY = "AZURE_OPENAI_KEY";
+	static final String MICROSOFT_FOUNDRY_API_KEY = "MICROSOFT_FOUNDRY_API_KEY";
```
ilayaperumalg, line 46: @jdubois Thanks for the PR! Given this was part of 1.1.2 release, Do you think we need to add support for backwards compatibility?

jdubois: People will need to migrate to the new key, then the rebranding has already be done on our side, so I don't think there's any backward compatibility to have as from now on you'll only have Microsoft Foundry.

ilayaperumalg:
Rebased, squashed and merged as 5c18a40c4258592ee2d4aed7fc626f20dd27b057 and backported into 1.1.x via d4db91764

PR #5225, opened 2026-01-12
Fix filter expr parser

ericbottard:
This PR fixes two issues with `FilterExpressionTextParser`:
* At the grammar level, `constant` was incorrectly defined as `QUOTED_STRING+` (ie one or more)
* At the parser level, handling was only stripping the outer quotes, forgetting to un-escape the potential inner quotes.

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
@@ -193,12 +193,24 @@ public class FilterExpressionTextParser {
 
 		@Override
 		public Filter.Operand visitTextConstant(FiltersParser.TextConstantContext ctx) {
-			String onceQuotedText = removeOuterQuotes(ctx.getText());
+			String onceQuotedText = unescapeStringValue(ctx.getText());
 			return new Filter.Value(onceQuotedText);
 		}
 
-		private String removeOuterQuotes(String in) {
-			return in.substring(1, in.length() - 1);
+		/**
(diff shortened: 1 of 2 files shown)


spring-ai-vector-store/src/main/java/Filters.tokens
```diff
@@ -0,0 +1,43 @@
+LONG_SUFFIX=1
```
ilayaperumalg, line 1: Are these necessary to be committed or can be added to .gitignore files?

ilayaperumalg (approved):
LGTM. We need to decide if we should add these tokens into .gitignore or move them under antlr4 package.

ericbottard:
> LGTM. We need to decide if we should add these tokens into .gitignore or move them under antlr4 package.

They are part of the generated files and it is usual practice to commit them (even if we don't need them. Similarly, we don't use the *Listener classes, but we did commit them). We shouldn't add complexity (add them to .gitignore) when none is needed in my opinion

spring-ai-vector-store/src/main/java/Filters.tokens
ericbottard: They're not _necessary_ but it is common practice to version the whole set of things antlr generates. They are stable though (ie if we don't change the .g4 file, they won't change, so they don't create git noise)

ilayaperumalg:
Rebased, squashed and merged as 8ecc334b1

ilayaperumalg:
Back ported into 1.1.x via fe5c3f908d3e01526783a86c7560b034f15f28f0

ericbottard:
Back ported into 1.0.x as f6204b5952bd236ec311864184673d46d6d0300d

Issue #5116, opened 2025-12-18
Streaming tool_calls with Qwen models cause function.arguments Field required in Spring AI 2.0.0


xakepbean:
Qwen follows the OpenAI-compatible streaming protocol

Environment
Spring AI version: 2.0.0
Java version: 17
Model: qwen3
Mode: stream=true

// //ok
// ChatClient chatClient = ChatClient.builder(chatModel)
// 		.defaultToolCallbacks(MethodToolCallbackProvider.builder().toolObjects(new MyTools()).build())
// 	.build(); 
// var answer = chatClient
// 		.prompt("What day is tomorrow?")
// 		.call().content();
// System.out.println(answer);

// error
ChatClient chatClient = ChatClient.builder(chatModel)
		.defaultToolCallbacks(MethodToolCallbackProvider.builder().toolObjects(new MyTools()).build())
	.build(); 
var answer = chatClient
		.prompt("What day is tomorrow?")
		.stream().content();
		String content = answer.collectList().block().stream().collect(Collectors.joining());
System.out.println(content);

static class MyTools {

    @Tool(description = "Get the current date and time in the user's timezone")
    String getCurrentDateTime() {
        return LocalDateTime.now().atZone(LocaleContextHolder.getTimeZone().toZoneId()).toString();
    }

}

logs：

2025-12-18T10:11:40.888+08:00 DEBUG 27332 --- [reactor-http-nio-2] reactor.netty.http.client.HttpClient     : [e2db0bd5-1,  READ COMPLETE
2025-12-18T10:11:40.966+08:00 DEBUG 27332 --- [reactor-http-nio-2] reactor.netty.http.client.HttpClient     : [e2db0bd5-1,  READ: 607B c4
data: {"id":"chatcmpl-b3d783b2b520495fb6e3f4f8f4421e3f","object":"chat.completion.chunk","created":1766023897,"model":"Qwen3","choices":[{"index":0,"delta":{"content":" call","tool_calls":[]}}]}

c2
data: {"id":"chatcmpl-b3d783b2b520495fb6e3f4f8f4421e3f","object":"chat.completion.chunk","created":1766023897,"model":"Qwen3","choices":[{"index":0,"delta":{"content":".\n","tool_calls":[]}}]}

c7
[...]

PR #5232, opened 2026-01-13
GH-5116: Fix openai default null tool arguments to empty JSON string

thisishwan2:
Fixes https://github.com/spring-projects/spring-ai/issues/5116

## Description
This PR fixes a compatibility issue with strict OpenAI-compatible models (e.g., Qwen) where tool_calls requests failed due to a missing function.arguments field.
Previously, when arguments was null, the field was omitted during serialization due to the @JsonInclude(Include.NON_NULL) annotation. While the official OpenAI API is lenient, strict providers reject this as invalid protocol usage, returning 400 Bad Request.

## Changes
Updated the ChatCompletionFunction record in OpenAiApi by adding a compact constructor. This ensures that arguments defaults to "{}" (empty JSON object string) when null, guaranteeing the field is always present in the serialized JSON.

## Verification
- I have verified that the code compiles successfully.
- Added a new unit test nullArgumentsDefaultsToEmpty in `OpenAiApiT.java` to verify that null arguments are correctly converted to "{}" and that the serialized JSON explicitly contains "arguments": "{}".

### DIFF
--- a/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/api/OpenAiApiIT.java
+++ b/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/api/OpenAiApiIT.java
@@ -308,7 +308,7 @@ class OpenAiApiIT {
 
 		assertThat(function.arguments()).isEqualTo("{}");
 
-		// 검증 2: JSON 문자열에 arguments 필드가 포함되었는지 확인 ("arguments": "{}")
+		// Verify that the JSON string contains the "arguments" field ("arguments": "{}")
 		assertThat(json).contains("\"arguments\":\"{}\"");
 	}
 


models/spring-ai-openai/src/test/java/org/springframework/ai/openai/api/OpenAiApiIT.java
```diff
@@ -295,4 +297,19 @@ void userAgentHeaderIsSentInChatCompletionRequests() throws Exception {
...
+		final String TOOL_FUNCTION_NAME = "CurrentWeather";
+
+		var function = new ChatCompletionMessage.ChatCompletionFunction(TOOL_FUNCTION_NAME, null);
+
+		ObjectMapper mapper = new ObjectMapper();
+		String json = mapper.writeValueAsString(function);
+
+		assertThat(function.arguments()).isEqualTo("{}");
+
+		// 검증 2: JSON 문자열에 arguments 필드가 포함되었는지 확인 ("arguments": "{}")
```
dev-jonghoonpark: Please convert the above comments into English.

thisishwan2: I've updated the comments to English. Thank you for pointing this out.

PR #5233, opened 2026-01-13
Breaking Change: Move Couchbase vector store classes to a dedicated package

ericbottard:


ilayaperumalg:
@ericbottard As a side note, let's add this change into the upgrade notes section.

ilayaperumalg (approved):

PR #5258, opened 2026-01-15
Fix mutating global `RestClient` and `WebClient` builders

shenjianeng:
Since the builders for HTTP clients are mutable and shared, they should only be configured with globally applicable settings. The current use leaks specific details into other usages and affects newly instantiated clients


quaff:
https://github.com/spring-projects/spring-boot/blob/6eaeeccd383662dad162d207208af31fd177eea4/module/spring-boot-webclient/src/main/java/org/springframework/boot/webclient/autoconfigure/WebClientAutoConfiguration.java#L59

It's safe to mutate since it's prototype.

shenjianeng:
> https://github.com/spring-projects/spring-boot/blob/6eaeeccd383662dad162d207208af31fd177eea4/module/spring-boot-webclient/src/main/java/org/springframework/boot/webclient/autoconfigure/WebClientAutoConfiguration.java#L59
> 
> It's safe to mutate since it's prototype.

But We don't know how the users will use it. 

the defaultHeaders have been modified. Using clone is a safer approach.

PR #5259, opened 2026-01-15
Instrument observation and retry for ImageModel implementations

quaff:


quaff:
@jonatan-ivanov Could you review the observation parts?

PR #5260, opened 2026-01-15
Make RetryTemplate as optional dependency

quaff:
`SpringAiRetryAutoConfiguration` is no longer required for model auto-configurations.


ilayaperumalg:
@quaff Thanks for the PR! LGTM. Having the PR description as "RetryTemplate as optional dependency" may be confused with the maven dependency. I will rename this to use "RetryTemplate as optional bean" to make it better. Hope that makes sense.

ilayaperumalg:
Rebased and merged via 4f121c6f76fd934b246def182c960e446f49b0d7

PR #5262, opened 2026-01-15
GH-3797 change: make QdrantVectorStore contentFieldName configurable

### LINKED ISSUE #3797: Feature Request: Allow configuration of payload content key for QdrantVectorStore
**The Problem**
The QdrantVectorStore implementation hardcodes the payload field name for document content as `"doc_content".` This makes it difficult to integrate Spring AI with pre-existing Qdrant collections that use a different naming convention for the primary text field (e.g., "page_content", "text", etc.).

The key is defined as a private static final String and used in private methods (toDocument, toPayload), preventing simple extension or configuration.

**The Desired Solution**
Introduce a configuration property to allow users to specify the content field name. This would provide a declarative and straightforward way to align Spring AI with existing data schemas.

**A clear example of the desired solution**
A user should be able to configure the content key in their application.properties or application.yml file:

The QdrantVectorStore should then use this value when reading from and writing to the Qdrant payload, falling back to the default "doc_content" if the property is not set.

The value of the solution
This feature would significantly improve the out-of-the-box experience for users connecting to existing Qdrant databases, removing the need for boilerplate code (subclassing QdrantVectorStore and overriding methods) just to handle a different field name. It makes the component more flexible and easier to adopt in established environments.

destiny3912:
## Description
This Pull Request changes the content field name of the Qdrant Vector Store to enable setting via auto configuration in hard-coded "doc_contents"
fix #3797 

## Changes
- Add contentFieldName in QdrantVecotrStoreProperties

## Usage
Add property spring.ai.vectorstore.qdrant.content-field-name to application.yml

### DIFF
--- a/vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantVectorStore.java
+++ b/vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantVectorStore.java
@@ -129,11 +129,13 @@ public class QdrantVectorStore extends AbstractObservationVectorStore implements
 
 	public static final String DEFAULT_COLLECTION_NAME = "vector_store";
 
-	private static final String CONTENT_FIELD_NAME = "doc_content";
+	public static final String DEFAULT_CONTENT_FIELD_NAME = "doc_content";
 
 	private final QdrantClient qdrantClient;
 
 	private final String collectionName;
+
+	private final String contentFieldName;
 
 	private final QdrantFilterExpressionConverter filterExpressionConverter = new QdrantFilterExpressionConverter();
 
@@ -155,6 +157,7 @@ public class QdrantVectorStore extends AbstractObservationVectorStore implements
 		this.qdrantClient = builder.qdrantClient;
 		this.collectionName = builder.collectionName;
 		this.initializeSchema = builder.initializeSchema;
+		this.contentFieldName = builder.contentFieldName;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantVectorStore.java
```diff
@@ -280,7 +283,7 @@ private Document toDocument(ScoredPoint point) {
 			var metadata = QdrantObjectFactory.toObjectMap(point.getPayloadMap());
 			metadata.put(DocumentMetadata.DISTANCE.value(), 1 - point.getScore());
 
-			var content = (String) metadata.remove(CONTENT_FIELD_NAME);
+			var content = (String) metadata.remove(DEFAULT_CONTENT_FIELD_NAME);
```
ilayaperumalg: Now that the content field name is configurable, we should be using the content field name instead of the default one.

destiny3912: Thank you for the feedback. The intention was to handle null safety when contentFieldName was not configured, but it was wrong. It will be corrected

ilayaperumalg: @destiny3912 Are you planning to update the PR to fix this? or, do you want me to handle when merging? Thank you

destiny3912: @ilayaperumalg I  fix it now, please check it, Thank you

ilayaperumalg:
@destiny3912 Thanks for the PR! Rebased and merged via f65756ee20e16ca49385d4deea75d8ebbc38c283 along with the doc updates via 002a20889c5e63cb4e263483093dbf0a73847036

destiny3912:
Thanks!

PR #5269, opened 2026-01-17
Fix #5250: Add ambiguity check and improve Google GenAI auth logic

### LINKED ISSUE #5250: spring-ai-starter-model-google-genai auth failure
**Bug description**
When using spring-ai-starter-model-google-genai version 1.1.2 and Gemini, ChatClient returns an auth error:
This is the config
```
spring.ai.google.genai.api-key=${GOOGLE_GEMINI_API_KEY}
spring.ai.google.genai.chat.options.model=gemini-3-flash-preview
```
this is the error
```
com.google.genai.errors.ClientException: 401 . Authentication Error, LiteLLM Virtual Key expected. Received=AI***********************************, expected to start with 'sk-'.
```
While:
Using the same code with `spring-ai-starter-model-openai` version 1.1.2 and changing the configuration make ChatClient work fine:
```
spring.ai.openai.chat.base-url=https://generativelanguage.googleapis.com
spring.ai.openai.chat.completions-path=/v1beta/openai/chat/completions
spring.ai.openai.api-key=${GOOGLE_GEMINI_API_KEY}
spring.ai.openai.chat.options.model=gemini-3-flash-preview
```

**Environment**
Spring AI Version: 1.1.2
Java Version: 21
Spring Boot Version: 3.5.9
Language: Java
Vector Store: None
Chat/Embedding Model: Google Gemini (gemini-3-flash-preview)

**Steps to reproduce**
The `main` branch of this repository works with openapi: https://github.com/bu3/spring-ai-gemini-example/tree/main
The `starter-failure` branch with spring-ai-starter-model-google-genai fails with an auth error : https://github.com/bu3/spring-ai-gemini-example/tree/starter-failure

**Expected behavior**
I would like to use Gemini with both starters to gather information about dog breeds

```
mvn clean spring-boot:run
curl -X POST http://localhost:8080/api/dog-breeds/info \
  -H "Content-Type: application/json" \
  -d '{"dog_breed": "Labrador"}'
```

snosrati200:
This PR addresses issue #5250 by adding a check for ambiguous configurations in `GoogleGenAiChatAutoConfiguration`.

Key changes:
- Added logic to throw an `IllegalStateException` when both `api-key` and Vertex AI properties (`project-id`, `location`) are provided.
- Added unit tests to verify Vertex AI configuration, API Key configuration, and the ambiguity check.
- Verified the build and checkstyle with Java 25.

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-google-genai/src/main/java/org/springframework/ai/model/google/genai/autoconfigure/chat/GoogleGenAiChatAutoConfigurationTests.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-google-genai/src/main/java/org/springframework/ai/model/google/genai/autoconfigure/chat/GoogleGenAiChatAutoConfigurationTests.java
@@ -23,17 +23,32 @@ import org.springframework.boot.test.context.runner.ApplicationContextRunner;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-class GoogleGenAiChatAutoConfigurationTests {
+public class GoogleGenAiChatAutoConfigurationTests {
 
 	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
 		.withConfiguration(AutoConfigurations.of(GoogleGenAiChatAutoConfiguration.class));
 
 	@Test
-	void shouldFailOnAmbiguousConfiguration() {
+	void shouldNotFailOnAmbiguousConfigurationButPrioritizeApiKey() {
 		this.contextRunner
 			.withPropertyValues("spring.ai.google.genai.api-key=test-key",
 					"spring.ai.google.genai.project-id=test-project", "spring.ai.google.genai.location=us-central1")
-			.run(context -> assertThat(context).failed().hasRootCauseInstanceOf(IllegalStateException.class));
+			.run(context -> {
+				assertThat(context).hasSingleBean(Client.class);
+				// We no longer expect a failure here, just a warning in the logs
+			});
+	}
+
+	@Test
+	void shouldFailWhenVertexAiEnabledButConfigMissing() {
+		this.contextRunner.withPropertyValues("spring.ai.google.genai.vertex-ai=true") // Explicitly
+																						// enabled
+																						// but
+																						// no
+																						// project/location
+			.run(context -> assertThat(context).failed()
+				.hasRootCauseInstanceOf(IllegalStateException.class)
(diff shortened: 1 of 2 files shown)


ddobrin:
Hi @snosrati200 
While you should generally not provide both API Key and credentials, and should know precisely in which env you are setting which credentials. 

**Prod** - recommend always setting project/location
**Local Dev** - choices are available, and API Key provides easier onboarding

While the Java SDK builders allow you to syntactically set both, the SDK has a strict precedence rule: 
- if you provide a Project ID and Location, the SDK assumes you are using standard Google Cloud Authentication (ADC/Service Accounts) and will ignore the API Key
- SDK will attempt to find Google Cloud credentials on your machine. If you don't have them set up (say local env, and were relying on the API key), the call will fail with a 401 Unauthenticated or No credentials found error

The `.vertexAI(true);` can be used to toggle between the two.

I would not recommend merging the PR as is and can lend support to help fix this

CC: @markpollack @ilayaperumalg

snosrati200:
Hi @ddobrin, thanks again for the guidance — I’ve updated the implementation based on your feedback.

The configuration logic now:

Prioritizes Vertex AI when vertex-ai=true is set.

Uses isVertexAi() as the explicit primary toggle for the authentication mode.

Adds fail-fast validation: It now throws an IllegalStateException if Vertex AI is enabled but either project-id or location is missing, preventing the 401 issues caused by implicit credential fallback.

I also updated the test suite to cover the new precedence and validation behavior. Ready for another review when you have time! 👍

ilayaperumalg:
@snosrati200 Thanks for the PR and following it up with the updates. Could you please fix the DCO compliance? You can do this by "git commit -s --amend" and force pushing the changes to the PR.

snosrati200:
Sorry for the delay! I've just updated the PR with the DCO sign-off. Everything should be in order now. Thanks for your patience.

auto-configurations/models/spring-ai-autoconfigure-model-google-genai/src/main/java/org/springframework/ai/model/google/genai/autoconfigure/chat/GoogleGenAiChatAutoConfigurationTests.java
```diff
@@ -0,0 +1,68 @@
+/*
```
ilayaperumalg, line 1: This needs to be moved to src/test. will fix when merging.

ilayaperumalg:
Rebased, squashed, moved the test to src/test, fixed the test assertion issue and merged via c974b21a8. Backported into 1.1.x via 29ca4a656

Issue #5274, opened 2026-01-18
Migrate MongoDB Atlas container support to the new native feature in Spring Boot 4


ThomasVitale:
Currently, Spring AI defines a series of `ContainerConnectionDetailsFactory` to support Testcontainers with several vector stores and inference services. One of them is [`MongoDbAtlasLocalContainerConnectionDetailsFactory`](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-spring-boot-testcontainers/src/main/java/org/springframework/ai/testcontainers/service/connection/mongo/MongoDbAtlasLocalContainerConnectionDetailsFactory.java).

As of Spring Boot 4, there is a native [`MongoDbAtlasLocalContainerConnectionDetailsFactory`](https://github.com/spring-projects/spring-boot/blob/main/module/spring-boot-mongodb/src/main/java/org/springframework/boot/mongodb/testcontainers/MongoDbAtlasLocalContainerConnectionDetailsFactory.java) directly in the Spring MongoDB module. 

For Spring AI 2.0, I recommend migrating to the native support in Spring Boot. Otherwise, if an application is using MongoDB Atlas both from Spring Data and Spring AI Vector Store, two different configurations will be needed.

ThomasVitale:
PR to implement the change: https://github.com/spring-projects/spring-ai/pull/5277

PR #5277, opened 2026-01-18
feat: Use MongoDB Atlas container support from Spring Boot 4

ThomasVitale:
* Remove custom support for MongoDB Atlas in Docker Compose and Testcontainers.
* Document the service is now supported natively in Spring Boot.

The migration should be transparent and not require any code change.

Fixes gh-5274


ilayaperumalg:
@ThomasVitale Thanks for the PR! Rebased and merged via 5b39fbb5e

PR #5280, opened 2026-01-19
Make ResponseErrorHandler as optional bean

quaff:
Continue of commit 4f121c6f76fd934b246def182c960e446f49b0d7

PR #5281, opened 2026-01-19
Remove reference to SpringAiRetryAutoConfiguration from other auto-configurations

quaff:
`AutoConfigureAfter`/`AutoConfigureBefore` is hint for ordering `auto-configuration`s, it affects condition evaluation (`@ConditionalOn[Missing]Bean`), it's abused in Spring AI's codebase, let's clean only `SpringAiRetryAutoConfiguration` now for safety.


quaff:
It depends on #5280

quaff:
https://github.com/spring-projects/spring-ai/blob/4f121c6f76fd934b246def182c960e446f49b0d7/spring-ai-test/src/main/java/org/springframework/ai/utils/SpringAiTestAutoConfigurations.java#L45-L62

Tests failed due to that `SpringAiTestAutoConfigurations` expands `@AutoConfiguration.after()`, I think changing the standard behavior is bad since it break consistency between production and test codes.

quaff:
@ilayaperumalg Please review.

quaff:
This PR is outdated due to https://github.com/spring-projects/spring-ai/commit/5b5631adf9fd54fe0905e3c6336abff8986953d7

PR #5282, opened 2026-01-19
Make `spring-ai-anthropic` null-safe

sdeleuze:
Check TODOs in the changes.

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatOptions.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatOptions.java
@@ -29,6 +29,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.ai.anthropic.api.AnthropicApi;
 import org.springframework.ai.anthropic.api.AnthropicApi.ChatCompletionRequest;
@@ -39,7 +40,6 @@ import org.springframework.ai.model.ModelOptionsUtils;
 import org.springframework.ai.model.tool.StructuredOutputChatOptions;
 import org.springframework.ai.model.tool.ToolCallingChatOptions;
 import org.springframework.ai.tool.ToolCallback;
-import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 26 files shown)


spring-ai-model/src/main/java/org/springframework/ai/model/ModelOptionsUtils.java
```diff
@@ -458,7 +459,8 @@ else if (node.isArray()) {
 	/**
 	 * Return the runtime value if not empty, or else the default value.
 	 */
-	public static <T> T mergeOption(T runtimeValue, T defaultValue) {
+	@Contract("_, !null -> !null")
```
ericbottard, line 462: Can't we also add a contract for when runtimeValue is not null?

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatOptions.java
```diff
@@ -293,12 +293,14 @@ public void setInternalToolExecutionEnabled(@Nullable Boolean internalToolExecut
 	}
 
 	@Override
+	@Nullable
```
ericbottard: should move

spring-ai-model/src/main/java/org/springframework/ai/model/ModelOptionsUtils.java
sdeleuze: I tried, but I don't think we can add a useful one. Please either propose one or resolve this discussion.

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/StreamHelper.java
```diff
@@ -125,25 +128,27 @@ else if (event.type() == EventType.CONTENT_BLOCK_STOP) {
 	 * multiple events.
 	 * @return A ChatCompletionResponse representing the processed chunk.
 	 */
+	// TODO How to deal with the use cases where id is null since it is now mandatory in
```
sdeleuze, line 131: @tzolov I would need your input here.

I have tried to implement in this PR the same optionality of properties than specified on https://platform.claude.com/docs/en/api/overview. That means for example that in `ChatCompletionResponse` the `id`, `content` and `model` are not null. For certain kind of events, we can have a message id, but for some others we don't, hence the following tests are broken:
```
[ERROR] Errors: 
[ERROR]   StreamHelperTests.testErrorEventTypeWithEmptyContentBlock:55 » IllegalState The id must not null
[ERROR]   StreamHelperTests.testMultipleErrorEventsHandling:70 » IllegalState The id must not null
[ERROR]   StreamHelperTests.testPingEventHandling:493 » IllegalState The id must not null
```
I think we should really stick to those Anthropic API optionanality definitions, but then I am not sure how to resolve those errors. Do we really have to support those use cases when the event does not provide a message id?

spring-ai-model/src/main/java/org/springframework/ai/model/ModelOptionsUtils.java
ericbottard: You're right, this simple one catches all the cases we can reason about anyway.

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AssistantMessage.java
```diff
@@ -44,7 +44,7 @@ public class AssistantMessage extends AbstractMessage implements MediaContent {
 
 	protected final List<Media> media;
 
-	public AssistantMessage(String content) {
+	public AssistantMessage(@Nullable String content) {
```
ericbottard, line 47: These changes to core APIs seem "malheureuses", as we say in French.
This one in particular.
Is there a way to avoid allowing more nulls creeping in?

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/StreamHelper.java
sdeleuze: Fixed by 66dc2ea4f05fff69a6b9368e1f781e8ea7b92c7c

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AssistantMessage.java
sdeleuze: In theory I agree, in practice `textContent` is nullable in `AbstractMessage`, so unless we do a wider refactoring (if we do, should probably happen after this PR), I think we should keep it nullable.

sdeleuze:
@tzolov This PR is ready for review.

sdeleuze:
@ericbottard Compilation works locally so I think https://github.com/spring-projects/spring-ai/actions/runs/21394650669/job/61589846961?pr=5282 failed due to the compilation cache enabled on the CI.

tzolov (approved):
Changes LGTM, 
All tests except (the AnthropicSkillsIT#shouldGeneratePowerPointWithPptxSkill and shouldUseMultipleSkills) pass. 
Later though are failing for me on main as well. I will raise a separate issue for those after this PR is merged.

tzolov:
Squashed, rebased and merged at 1f3b0dc1410a8c97802956afe765cdf38189bd37

PR #5284, opened 2026-01-19
Jspecify semantic cache

ericbottard:


kkshanib:
Can we have optional / condition based caching instead of caching all response , we should have an option to cache universal / generic request and responses . May be an llm call to check if it's eligible for cache .

Issue #5220, opened 2026-01-11
ZhiPuAI Default model leads to "error code:1211 Model does not exist" error & cleanup deprecated models


thisishwan2:
**Bug description**
When using the `spring-ai-zhipuai` starter with default settings ([Spring AI ZhiPu AI Chat official documentation examples](https://docs.spring.io/spring-ai/reference/api/chat/zhipuai-chat.html)), the application fails to generate a response. 
The ZhipuAI API returns error code 1211, indicating that the configured default model does not exist or is no longer supported.([bigmodel api description](https://docs.bigmodel.cn/cn/faq/api-code))

Additionally, the ZhiPuAiApi.ChatModel enum seems to contain several models that are deprecated or invalid in the current ZhipuAI API.

**Environment**
Please provide as many details as possible: Spring AI version, Java version, which vector store you use if any, etc
Spring AI: 1.0.0
Java: 17

**Steps to reproduce**
1. Configure a Spring Boot application with spring-ai-zhipuai-spring-boot-starter.
2. Do not specify any spring.ai.zhipuai.chat.options.model in application.yml (rely on defaults).
3. Implement a simple ChatController following the [official documentation](https://www.google.com/url?sa=E&q=https%3A%2F%2Fdocs.spring.io%2Fspring-ai%2Freference%2Fapi%2Fchat%2Fzhipuai-chat.html).
Send a request to the chat endpoint.

**Expected behavior**
The application should successfully connect to the ZhipuAI API and return a chat response using a valid default model.

**Minimal Complete Reproducible example**
```java
@RestController
public class ChatController {

    private final ZhiPuAiChatModel chatModel;

    @Autowired
    public ChatController(ZhiPuAiChatModel chatModel) {
        this.chatModel = chatModel;
    }

    @GetMapping("/ai/generate")
    public Map generate(@RequestParam(value = "message", defaultValue = "Tell me a joke") String message) {
        return Map.of("generation", this.chatModel.call(message));
    }

    @GetMapping(value = "/ai/generateStream", produces = MediaType.TEXT_EVENT_STREAM_VALUE)
    public Flux<ChatResponse> generateStream(@RequestParam(value = "message", defaultValue = "Tell me a joke") String message) {
        var prompt = new Prompt(new UserMessage(message));
        return this.chatModel.stream(prompt);
    }
}
```

**Analysis**
The current default model (appears to be glm-4-air or similar) is causing the 1211 error.
[...]

PR #5285, opened 2026-01-19
GH-5220: Update ZhiPuAI default model to the latest available model

thisishwan2:
Fixes [GH-5220](https://github.com/spring-projects/spring-ai/issues/5220)

## Description
This PR updates the default ZhiPuAI model to **glm-4.7**, which is the latest supported model, to ensure compatibility with the current ZhiPuAI API.

Several tests were previously failing due to references to deprecated or non-existent model identifiers. This change aligns the default configuration with supported models and stabilizes the test suite.

## Changes
- Update the default ZhiPuAI chat model to `glm-4.7`
- Fix failing tests caused by deprecated or unavailable model identifiers
- Add tests to verify behavior when using the default model configuration

## Verification
- All existing tests pass
- New tests validate chat completion behavior using the default model

## Question
1. There are several ZhiPuAI model identifiers referenced in the codebase that appear to be deprecated or no longer supported according to this issue(https://github.com/spring-projects/spring-ai/issues/5220)
   Would it be preferable to remove enums for models that are no longer available, or should we keep them for backward compatibility?

2. Although the official ZhiPuAI documentation indicates that Embedding Model 3 is supported, actual API calls fail with "model not found" errors.  
   This issue affects both `embedding-2` and `embedding-3`, and currently causes CI test failures.

   Would it be better to remove or temporarily disable embedding model tests until the API behavior is clarified?

### DIFF
--- a/models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/api/ZhiPuAiApi.java
+++ b/models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/api/ZhiPuAiApi.java
@@ -69,7 +69,7 @@ public class ZhiPuAiApi {
 		return new Builder();
 	}
 
-	public static final String DEFAULT_CHAT_MODEL = ChatModel.GLM_4_Air.getValue();
+	public static final String DEFAULT_CHAT_MODEL = ChatModel.GLM_4_7.getValue();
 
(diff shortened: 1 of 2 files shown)


sdeleuze:
We are stopping to maintain ZhiPuAI as part of the core Spring AI project as we have no proper way to test it, and have to focus on a smaller set of models. We have not been able to find a maintainer for a support outside of this repository, but if somebody want to volunteer, please reach us on https://github.com/spring-projects/spring-ai/discussions.

PR #5295, opened 2026-01-20
Fix condition at Infinispan AutoConfiguration

eddumelendez:
Signed-off-by: Eddú Meléndez <eddu.melendez@gmail.com>


nicolaskrier (approved):
Good catch!

ericbottard (approved):

PR #5301, opened 2026-01-21
Fix ToolCallAdvisor to preserve system message when memory disabled

tzolov:
When conversation history is disabled in ToolCallAdvisor, the system message was being lost on subsequent tool call iterations, causing the LLM to lose its system instructions.

- Include system message in the messages sent to LLM when conversationHistoryEnabled is false
- Add disableMemory() builder method for clearer API. Shortcut for the conversationHistoryEnabled(false)
- Fix MessageChatMemoryAdvisor to ensure system message appears first after merging with memory messages
- Add tests for both fixes

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
@@ -33,6 +33,7 @@ import org.springframework.ai.chat.client.advisor.api.BaseChatMemoryAdvisor;
 import org.springframework.ai.chat.client.advisor.api.StreamAdvisorChain;
 import org.springframework.ai.chat.memory.ChatMemory;
 import org.springframework.ai.chat.messages.Message;
+import org.springframework.ai.chat.messages.SystemMessage;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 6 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
```diff
@@ -85,6 +86,15 @@ public ChatClientRequest before(ChatClientRequest chatClientRequest, AdvisorChai
 		List<Message> processedMessages = new ArrayList<>(memoryMessages);
 		processedMessages.addAll(chatClientRequest.prompt().getInstructions());
 
+		// 2.1. Ensure system message, if present, appears first in the list.
+		for (int i = 0; i < processedMessages.size(); i++) {
+			if (processedMessages.get(i) instanceof SystemMessage) {
+				Message systemMessage = processedMessages.remove(i);
+				processedMessages.add(0, systemMessage);
```
ilayaperumalg, line 93: Given there can be multiple System messages, should we include all of them?

tzolov: No, AFAIK you always should have only one system message

ilayaperumalg:
Rebased and merged as 1dac9d09b442c45751f0e29736413d16268629df and backported into 1.1.x via de32dd257fe7373ce2e8a5f11497f916154709a2

PR #5305, opened 2026-01-22
Fix: Prevent content swallowing when tool calls occur in streaming

### LINKED ISSUE #2575: Content field is empty in streaming response after tool call in MCP client
**Bug description**
When using the Spring AI MCP client for handling tool calls, we've observed inconsistent behavior between streaming and non-streaming modes regarding the `content` field in the request payload.

1. **Streaming Mode** (stream: true):
   ```json
   {
     "messages": [
       {
         "content": "What data does the database contain?",
         "role": "user"
       },
       {
         "content": "",  // 👈 Content is unexpectedly empty
         "role": "assistant",
         "tool_calls": [
           {
             "id": "...",
             "type": "function",
             "function": {"name": "list_tables"}
           }
         ]
       },
       // ... subsequent tool response
     ],
     "stream": true
   }
   ```

2. **Non-streaming Mode** (stream: false):
   ```json
   {
     "messages": [
       {
         "content": "What data does the database contain?",
         "role": "user"
       },
       {
         "content": "\n<think>\n ... \n</think>\n\n",  // Content contains intermediate thought steps
         "role": "assistant",
         "tool_calls": [
           // ... same tool call structure
         ]
       },
       // ... subsequent tool response
     ],
     "stream": false
   }
   ```

**Environment**
- Spring AI version: 1.0.0-M6
- Java version: 17
- MCP API version: 0.7.0-SNAPSHOT

**Steps to reproduce**
1. Send a tool call request with `stream: true`
2. Observe empty `content` field in assistant message
3. Compare with non-streaming request where `content` contains text content

**Expected behavior**
- Both streaming and non-streaming modes should maintain consistent payload structure
- The `content` field should either:
  1. Contain intermediate thought steps in **both** modes, or
  2. Be omitted in streaming mode while preserving tool call metadata

EH-OI:
### Motivation
In OpenAI streaming mode, if a single chunk contains both textual content (e.g., "Thinking...") and a tool call, the current implementation swallows the text content. This happens because the reactive stream switches directly to tool execution logic without emitting the current response.

Resolves #2575

### Modifications
* Modified `OpenAiChatModel.java` to use `Flux.concat`.
* Ensuring that the current response (containing content) is emitted to the downstream subscriber *before* triggering the tool execution flux.

### Result
Text content is no longer lost when it appears alongside tool calls in the same chunk during streaming.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      https://www.apache.org/licenses/LICENSE-2.0
+ *		https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


EH-OI:
I have added the reproduction test case (`OpenAiChatModelStreamingToolTests.java`) as requested.
The test verifies that the content is not swallowed when it appears alongside tool calls.
Also, I fixed the checkstyle issues, and the build passes locally.

PR #5336, opened 2026-01-27
Allow TokenTextSplitter to accept EncodingType

JM-Lab:
Replace the hardcoded CL100K_BASE encoding type with a configurable EncodingType.

This allows TokenTextSplitter to support different encoding types (e.g. CL100K_BASE, P50K_BASE, O200K_BASE).


sobychacko:
@JM-Lab This has been merged with one small fix — the existing 6-argument constructor was accidentally delegating with default constants instead of the actual parameter values, which would have silently ignored custom settings for callers using that constructor. That's been corrected in the squashed commit. Thanks for the contribution!

sobychacko:
@JM-Lab I added some more improvements on top of your PR here - https://github.com/spring-projects/spring-ai/pull/5493.

JM-Lab:
@sobychacko Thank you for the review and merge.

I appreciate the follow-up fix and enhancements in #5493.  
Let me know if anything else is required.

PR #5339, opened 2026-01-27
Fix message ordering in MongoChatMemoryRepository

### LINKED ISSUE #5330: MongoChatMemoryRepository is changing the order of the messages to descending during fetch and consequently the LLM is behaving unexpectedly
**Bug description**
`org.springframework.ai.chat.memory.repository.mongo.MongoChatMemoryRepository` is changing the order of the messages to descending during fetch and consequently the LLM is behaving unexpectedly.

I believe this happening because of the descending sort in the below code:

```java
@Override
public List<Message> findByConversationId(String conversationId) {
	var messages = this.mongoTemplate.query(Conversation.class)
		.matching(Query.query(Criteria.where("conversationId").is(conversationId))
			.with(Sort.by("timestamp").descending()));
	return messages.stream().map(MongoChatMemoryRepository::mapMessage).collect(Collectors.toList());
}
```
If we compare the sort with `org.springframework.ai.chat.memory.repository.jdbc.PostgresChatMemoryRepositoryDialect`, it uses an ascending sort rather than a descending one.
```java
@Override
public String getSelectMessagesSql() {
	return "SELECT content, type FROM SPRING_AI_CHAT_MEMORY WHERE conversation_id = ? ORDER BY \"timestamp\"";
}
```
**Environment**
Spring AI 1.1.2
Azure CosmosDB with Mongo API

**Expected behavior**
The order of conversation history should maintain the natural flow of the conversation (USER, ASSISTANT, USER, ASSISTANT.....)

chanani:
## Description
Fixed message ordering bug where `MongoChatMemoryRepository.findByConversationId()` returned messages in reverse chronological order, causing incorrect conversation flow for LLMs.

## Changes
- Changed sort order from `descending()` to `ascending()` in `findByConversationId()`

## Why
**Before (Bug):**
```java
.with(Sort.by("timestamp").descending())  // Returns: [newest...oldest]
```

**After (Fixed):**
```java
.with(Sort.by("timestamp").ascending())   // Returns: [oldest...newest]
```

This aligns with `PostgresChatMemoryRepositoryDialect`:
```sql
ORDER BY "timestamp"  -- ascending by default
```

## Testing
- Existing test `findByConversationId()` validates message order with `assertThat(results).isEqualTo(messages)`
- Unable to run locally (Docker constraints), but `CI` will validate

Fixes #5330

### DIFF
--- a/memory/repository/spring-ai-model-chat-memory-repository-mongodb/src/test/java/org/springframework/ai/chat/memory/repository/mongo/MongoChatMemoryRepositoryIT.java
+++ b/memory/repository/spring-ai-model-chat-memory-repository-mongodb/src/test/java/org/springframework/ai/chat/memory/repository/mongo/MongoChatMemoryRepositoryIT.java
@@ -125,6 +125,19 @@ public class MongoChatMemoryRepositoryIT {
 	}
 
 	@Test
+	void messagesAreReturnedInChronologicalOrder() {
+		var conversationId = UUID.randomUUID().toString();
+		var messages = List.<Message>of(new UserMessage("First message"),
+				new AssistantMessage("Second message"),
+				new UserMessage("Third message"));
+
+		this.chatMemoryRepository.saveAll(conversationId, messages);
+
+		var results = this.chatMemoryRepository.findByConversationId(conversationId);
+		assertThat(results).isEqualTo(messages);
+	}
+
+	@Test
 	void deleteMessagesByConversationId() {
 		var conversationId = UUID.randomUUID().toString();
 		var messages = List.<Message>of(new AssistantMessage("Message from assistant - " + conversationId),


chanani:
@tinomthomas, Added test case `messagesAreReturnedInChronologicalOrder()` to explicitly verify chronological ordering.

The test validates that messages are returned in the same order they were saved, ensuring the fix maintains proper conversation flow.

Let me know if you need any changes !

ilayaperumalg:
@chanani Thanks for the fix!

deejay1:
@ilayaperumalg LGTM, except the javaformat. Do we have a way to mark this change in release notes as it may break someones behaviour?

ilayaperumalg:
@deejay1 Thanks for the review. Good point on the notes related to the breaking change.

Rebased and merged via c65cb218adbeae6a00613e5c39940947d8186bc7 along with the update to docs to clarify the breaking change and the message ordering via dbde175a2cec5129ee18db4d7898ab1ca60c0e9a. 

Backported the changes to 1.1.x as well.

PR #5343, opened 2026-01-27
Fix locale-sensitive uppercasing in JSON schema type normalization

### LINKED ISSUE #5340: Locale-sensitive toUpperCase breaks Gemini tool type normalization (tr_TR → İNTEGER)
**Bug description**
When using the Google Gemini provider with Spring AI’s tool calling support, the tool definitions submitted to Gemini contain incorrectly uppercased field type values (e.g., `İNTEGER`, `STRİNG`) due to locale-sensitive string normalization. This causes Gemini API validation errors such as:

> Invalid value at 'tools[0].function_declarations[1].parameters.properties[0].value.type'
(type.googleapis.com/google.ai.generativelanguage.v1beta.Type), "İNTEGER"

The root cause is the use of `String.toUpperCase()` without specifying a locale, which on Turkish JVM locales (e.g., tr_TR) transforms `integer → İNTEGER` and `string → STRİNG`. The Gemini tool calling protocol expects strict enum values (`INTEGER`, `STRING`, etc.), so this locale conversion breaks the generated definition.

**Environment**
Spring AI version: 1.1.0
Java version: 21.0.5 2024-10-15 LTS (on tr_TR locale)
OS / Locale: Windows 10 / tr_TR.UTF-8
Provider: Google Gemini via spring-ai-google
Tool calling enabled: yes

**Steps to reproduce**
- Configure Spring AI with the Gemini provider and enable tool calling.
- Define a custom tool with parameter schemas that include basic field types (`integer`, `string`, etc.).
- Run a request that triggers the tool definition to be sent to the Gemini API.
- Observe the request payload and Gemini API error response showing type values incorrectly uppercased with Turkish locale characters (`İNTEGER`, `STRİNG`).

**Expected behavior**
Tool definition field types should be normalized to uppercase English enum values (`INTEGER`, `STRING`, `BOOLEAN`, etc.) regardless of JVM locale. No locale-specific characters should be introduced.

```java
@SpringBootTest
class GeminiToolLocaleTest {

    @Test
    void toolDefinitionLocaleTest() {
        final String inputSchema = """
                {
                  "type": "object",
                  "properties": {
                    "id": {
                      "type": "integer"
                    }
                  },
                  "required": [
                    "id"
                  ]
                }
                """;

        ToolDefinition td = ToolDefinition.builder()
                .name("test-tool")
                .description("test-tool description")
                .inputSchema(inputSchema)
                .build();

        // resolve tool definition like GoogleGenAiToolCallingManager class
        ObjectNode jsonSchema = JsonSchemaConverter.fromJson(td.inputSchema());
        ObjectNode openApiSchema = JsonSchemaConverter.convertToOpenApiSchema(jsonSchema);
        JsonSchemaGenerator.convertTypeValuesToUpperCase(openApiSchema);

        // print the resolved tool definition
        System.out.println(openApiSchema.toPrettyString());

        // assertions
        Assertions.assertNotNull(openApiSchema.get("properties"));
        Assertions.assertNotNull(openApiSchema.get("properties").get("id"));
        Assertions.assertNotNull(openApiSchema.get("properties").get("id").get("type"));
        Assertions.assertEquals("İNTEGER", openApiSchema.get("properties").get("id").get("type").asText());
    }
}
```

**Proposed Solution**
The issue can be resolved by making the type normalization logic locale-independent.

Currently, `JsonSchemaGenerator.convertTypeValuesToUpperCase()` relies on `String.toUpperCase()` without specifying a locale, which makes the behavior locale-sensitive and causes incorrect results on JVMs running with locales such as `tr_TR`.

A proposed fix is to overload `convertTypeValuesToUpperCase()` to accept a `Locale` parameter and use it explicitly during normalization. For example:
```java
convertTypeValuesToUpperCase(schema, Locale.ROOT)
```
Internally, the implementation should call:
```java
oldValue.toUpperCase(locale)
```
and default to `Locale.ROOT` (or `Locale.ENGLISH`) when no locale is explicitly provided.

This ensures that JSON schema type values (`string`, `integer`, etc.) are consistently converted to valid, locale-agnostic enum values (`STRING`, `INTEGER`, etc.) and prevents tool definition validation errors when interacting with providers like Google Gemini that strictly validate enum types.

vlknakkaya:
### Summary
Fixes locale-sensitive uppercasing when normalizing JSON schema type values.
On Turkish locales (tr_TR), `String.toUpperCase()` produces invalid values
(e.g. `İNTEGER`, `STRİNG`), which breaks Gemini tool calling.

### Changes
- Overloaded JsonSchemaGenerator.convertTypeValuesToUpperCase to accept Locale
- Defaulted normalization to `Locale.ROOT`
- Added regression test for Turkish locale

Fixes #5340

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.stream.Stream;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
@@ -258,29 +259,33 @@ public final class JsonSchemaGenerator {
 
 	// Based on the method in ModelOptionsUtils.
 	public static void convertTypeValuesToUpperCase(ObjectNode node) {
+		convertTypeValuesToUpperCase(node, Locale.ROOT);
+	}
+	
+	public static void convertTypeValuesToUpperCase(ObjectNode node, Locale locale) {
 		if (node.isObject()) {
 			node.fields().forEachRemaining(entry -> {
 				JsonNode value = entry.getValue();
 				if (value.isObject()) {
-					convertTypeValuesToUpperCase((ObjectNode) value);
+					convertTypeValuesToUpperCase((ObjectNode) value, locale);
 				}
 				else if (value.isArray()) {
 					value.elements().forEachRemaining(element -> {
 						if (element.isObject() || element.isArray()) {
-							convertTypeValuesToUpperCase((ObjectNode) element);
+							convertTypeValuesToUpperCase((ObjectNode) element, locale);
 						}
 					});
 				}
 				else if (value.isTextual() && entry.getKey().equals("type")) {
 					String oldValue = node.get("type").asText();
-					node.put("type", oldValue.toUpperCase());
+					node.put("type", oldValue.toUpperCase(locale));
 				}
 			});
 		}
 		else if (node.isArray()) {
 			node.elements().forEachRemaining(element -> {
 				if (element.isObject() || element.isArray()) {
-					convertTypeValuesToUpperCase((ObjectNode) element);
+					convertTypeValuesToUpperCase((ObjectNode) element, locale);
 				}
 			});
 		}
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks for this fix, @vlknakkaya, and for the clear regression test. This targets the same locale-sensitivity issue (#5340) as #6048, and both change the same line in `JsonSchemaGenerator.convertTypeValuesToUpperCase`. We are going to move forward with #6048 since it also covers two additional affected call sites (`GoogleGenAiModalityTokenCount` and `GoogleGenAiTrafficType`) and fixes the site inline without adding a new public overload. One thing worth noting for future reference: this branch predates the move to the `tools.jackson` package used on current `main`, so it may need a rebase to apply cleanly. We really appreciate you catching this and taking the time to write it up, and we hope you will consider contributing again.

PR #5354, opened 2026-01-28
Make `spring-ai-mistral-ai` null-safe

sdeleuze:
Please check integration tests before merging since I don't have a `MISTRAL_AI_API_KEY`.


models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/api/MistralAiApi.java
```diff
@@ -1208,8 +1214,8 @@ public record ChatCompletionFunction(@JsonProperty("name") String name,
 		public record MediaContent(
```
nicolaskrier, line 1218: You could consider not adding `@Nullable` annotations in this record and adapting the Jackson mapping accordingly to [Mistal AI OpenAPI Specifications](https://github.com/mistralai/platform-docs-public/blob/main/openapi.yaml).

I think a [`ContentChunk`](https://github.com/mistralai/platform-docs-public/blob/main/openapi.yaml#L8048-L8056) interface could be introduced as `MediaContent` replacement with the two following records implementing it:
- [`TextChunk`](https://github.com/mistralai/platform-docs-public/blob/main/openapi.yaml#L3820-L3835),
- [`ImageURLChunk`](https://github.com/mistralai/platform-docs-public/blob/main/openapi.yaml#L3527-L3545).

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/api/MistralAiApi.java
nicolaskrier, line 1097:
```suggestion
	 * @param rawContent The contents of the message. Can be either a list of {@link MediaContent}
```

nicolaskrier:
Hi @sdeleuze,
Regarding my `ContentChunk` proposal, it might be tricky to apply directly without a Mistral AI API key. If you’d prefer, I can submit a dedicated PR (draft available [here](https://github.com/nicolaskrier/spring-ai/tree/GH-5019) — note that `ReferenceChunk`, `ThinkChunk`, `ContentSerializer` and `ContentDeserializer` are not part of this scope and can be ignored).
I’ll also verify the integration tests tomorrow evening if this PR is still open.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/api/MistralAiApi.java
sdeleuze: Yes, this is our goal but so far we were on a strategy to step 1 first annotate the current semantics, and then step 2 (before GA) refactor to match the API vendor spec on nullability. I will discuss that with @ericbottard and @tzolov.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/api/MistralAiApi.java
ericbottard: Don't know the Mistral API and this implementation much, but from what I read of the `rawContent` field, the only read access of the field is in content(), and it expects a String. Or did I miss something?

nicolaskrier: You're right, this Javadoc needed more efforts to be understandable than what I have proposed above. I have created #5364 with this intent.

nicolaskrier:
@sdeleuze, I’ve attached [Spring AI Mistral AI integration tests logs](https://github.com/user-attachments/files/24949413/spring-ai-mistral-ai_integration-tests.log) for your investigation. I noticed some failures and put together [a proposal](https://github.com/spring-projects/spring-ai/commit/5b0ca9171b0179e57a1c35d25892fb645a318cd8) to address them. Let me know if this approach is fine or if you’d like to adjust it.

sdeleuze:
@nicolaskrier Thanks, added to my PR.

nicolaskrier:
I have forgotten to mention that `MistralAiChatModelIT.multiModalityEmbeddedImage` integration test is failing for me but it is also the case with the main branch. I believe it is due to the limitations with my subscription.

Here is the exception thrown:
```
org.springframework.ai.retry.NonTransientAiException: 429 - {"object":"error","message":"Rate limit exceeded","type":"rate_limited","param":null,"code":"1300"}
```

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/api/MistralAiApi.java
```diff
@@ -1389,10 +1399,10 @@ public record ChatCompletionChunk(
 		@JsonIgnoreProperties(ignoreUnknown = true)
 		public record ChunkChoice(
 		// @formatter:off
-			@JsonProperty("index") Integer index,
-			@JsonProperty("delta") ChatCompletionMessage delta,
-			@JsonProperty("finish_reason") ChatCompletionFinishReason finishReason,
-			@JsonProperty("logprobs") LogProbs logprobs) {
+			@SuppressWarnings("NullAway.Init") @JsonProperty("index") Integer index,
```
ericbottard: @sdeleuze Can you eleborate on the need for those SuppressWarnings? It seems the only constructor call we have passes non-null properties. And a record can't be half initialized, can it?

sdeleuze: Good catch, removed.

sdeleuze:
The build is green locally for me, so that's probably another compiler cache issue.

ericbottard:
Rebased, squashed on main as 3032521797561f2b0e079099a2028c0408f4d60d

PR #5355, opened 2026-01-28
Fix Bedrock cache metrics not reported in Usage native object

### LINKED ISSUE #4930: Bedrock token cache read/write metric
**Expected Behavior**

To support native AWS Bedrock `cacheReadInputTokens` and `cacheWriteInputTokens` metric count

**Current Behavior**
Spring AI 1.1.0

The cache write/read token is not being calculated, from what I see at ChatModelMeterObservationHandler, the chatResponse metadata token usage does not has the native usage. I believe need to add native usage at the chat response mapping [here](https://github.com/spring-projects/spring-ai/blob/v1.1.0/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java#L727) so it also can be handled at the obs handler.

At ConverseChatResponseStream.mergeNativeTokenUsage it already handle the cache read/write input

**Context**

On applications that use cache, the input token metric will not represent the whole input

Check the debug value below on how it's being shown when using cache:

on chat response stream

Image

On the obs handler, the total token calculates all, including the cache read/write
Image

As shown, it only `4` token input (prob metadata), which will lead to a wrong assumption and it is not possible to calculate  wether it is read/write cache based on the 3 existing fields

ahmed-sekka:
## Summary
Fixes the issue where Bedrock prompt cache metrics (`cacheReadInputTokens` and `cacheWriteInputTokens`) were not being properly exposed in the `Usage` object's native usage field, making them unavailable for observability handlers and metrics collection.

## Problem
As reported in #4930, the cache token metrics were being added to metadata as a separate map, but not included in the native `TokenUsage` object within the `Usage`. This caused metrics handlers to miss these important cache statistics.

## Solution
- Include cache metrics in the native `TokenUsage` object when creating `DefaultUsage`
- Properly merge cache metrics when aggregating usage across multiple tool-call responses
- Remove the redundant metadata map approach that was previously used

## Changes
- Add `TokenUsage` to imports in `BedrockProxyChatModel`
- Create native `TokenUsage` with cache metrics fields populated
- Merge cache metrics from previous responses during tool execution aggregation
- Remove the workaround code that added cache metrics to metadata

## Testing
- All existing unit tests pass (17/17)
- Verified compilation succeeds
- Cache metrics are now properly accessible via `usage.getNativeUsage()`

Fixes #4930

### DIFF
--- a/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
+++ b/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
@@ -24,7 +24,6 @@ import java.net.URLConnection;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Base64;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -62,6 +61,7 @@ import software.amazon.awssdk.services.bedrockruntime.model.Message;
 import software.amazon.awssdk.services.bedrockruntime.model.S3Location;
 import software.amazon.awssdk.services.bedrockruntime.model.StopReason;
 import software.amazon.awssdk.services.bedrockruntime.model.SystemContentBlock;
+import software.amazon.awssdk.services.bedrockruntime.model.TokenUsage;
 import software.amazon.awssdk.services.bedrockruntime.model.Tool;
 import software.amazon.awssdk.services.bedrockruntime.model.ToolConfiguration;
 import software.amazon.awssdk.services.bedrockruntime.model.ToolInputSchema;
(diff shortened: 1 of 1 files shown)


sobychacko:
@a-sekka The PR has been merged upstream. Thanks for the fix. 

After looking at the changes more closely, I realized removing the metadata map for cache tokens would break existing users who access them via `response.getMetadata().get("cacheReadInputTokens")`. So I've kept that code in place - now cache metrics are available both ways:

  1. Via `getNativeUsage()` - for observability handlers and metrics collection
  2. Via metadata map - for anyone already using the old approach

  Also added a couple unit tests to cover the new behavior:
  - One that checks cache metrics show up in both places
  - One that verifies cache tokens get aggregated correctly when tool calling happens (multiple round trips to the model)

  Updated the docs too - the Token Usage Tracking section now shows both ways to access cache metrics.

sobychacko:
Also, the fixes have been back-ported to `1.1.x`.

PR #5364, opened 2026-01-29
Improve Mistal AI API Javadoc

nicolaskrier:
## Description

- Clarify supported content types for request (string or list of media content),
- Clarify supported content types for response (string only).

## Explanations

Spring AI supports different content types for requests and for responses. It is a limitation on Spring AI side and it should be well documented in the Javadoc.
Here are interesting highlights of the different types used:
- [Request writing](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatModel.java#L460-L468),
- [Response reading](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/api/MistralAiApi.java#L1144-L1152). 

## Targeted versions

1.1.x and 2.0.x.

## Related work

Spring AI doesn't support Magistral 1.2 models family and the restricted content type for the response is expected to be removed for this feature. More details can be found in #5020.

### DIFF
--- a/models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/api/MistralAiApi.java
+++ b/models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/api/MistralAiApi.java
@@ -1094,8 +1094,9 @@ public class MistralAiApi {
 	/**
 	 * Message comprising the conversation.
 	 *
-	 * @param rawContent The contents of the message. Can be either a {@link MediaContent}
-	 * or a {@link String}. The response message content is always a {@link String}.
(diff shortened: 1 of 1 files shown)


nicolaskrier:
Hi @ilayaperumalg, would you mind adding a milestone to this PR? Thanks.

ilayaperumalg:
@nicolaskrier Thanks for the PR!

ilayaperumalg:
Rebased and merged as cb967d10c9bf7daf03b648ffb210d315f2f1888b and backported into 1.1.x via 18a661b50

PR #5366, opened 2026-01-30
Initial Support for the Anthropic Java SDK

sobychacko:
Initial support for the Anthropic Java SDK. Current support in the PR includes both sync and async chat model interactions with tool calling support in both cases.


models/spring-ai-anthropic-sdk/src/main/java/org/springframework/ai/anthropicsdk/AnthropicSdkChatOptions.java
```diff
@@ -0,0 +1,609 @@
...
+
+	/**
+	 * Default max tokens for chat completions.
+	 */
+	public static final Integer DEFAULT_MAX_TOKENS = 4096;
+
+	/**
+	 * Maximum number of tokens to generate in the response.
+	 */
+	@Nullable private Integer maxTokens;
```
ericbottard: All of those JSpecify annotations are out of place.
Should be

```suggestion
	private @Nullable Integer maxTokens;
```

models/spring-ai-anthropic-sdk/src/main/java/org/springframework/ai/anthropicsdk/AnthropicSdkChatOptions.java
```diff
@@ -0,0 +1,609 @@
...
+				+ ", internalToolExecutionEnabled=" + this.internalToolExecutionEnabled + ", toolContext="
+				+ this.toolContext + '}';
+	}
+
+	/**
+	 * Builder for creating {@link AnthropicSdkChatOptions} instances.
+	 */
+	public static final class Builder {
+
+		private final AnthropicSdkChatOptions options = new AnthropicSdkChatOptions();
```
ericbottard: Builders should be re-usable. Hence, they should not act on an pre-created instance of options, otherwise if you call build(), then mutate the builder, then build() again, you actually get the same instance and the result of the first call is mutated while the user may not realise.

models/spring-ai-anthropic-sdk/src/main/java/org/springframework/ai/anthropicsdk/metadata/package-info.java
```diff
@@ -0,0 +1,29 @@
...
+ * <p>
+ * This package contains classes that represent metadata extracted from Anthropic API
+ * responses, including usage statistics, model information, and response attributes.
+ *
+ * @since 2.0.0
+ */
+@NullMarked
+package org.springframework.ai.anthropicsdk.metadata;
+
+import org.jspecify.annotations.NullMarked;
```
ericbottard, line 23: This package is empty

models/spring-ai-anthropic-sdk/src/main/java/org/springframework/ai/anthropicsdk/AbstractAnthropicSdkOptions.java
```diff
@@ -0,0 +1,155 @@
...
+	 * Proxy settings for the Anthropic client.
+	 */
+	@Nullable private Proxy proxy;
+
+	/**
+	 * Custom HTTP headers to add to Anthropic client requests.
+	 */
+	private Map<String, String> customHeaders = new HashMap<>();
+
+	@Nullable public String getBaseUrl() {
```
ericbottard: Similarly, all the method annotations are out of place.

Should be

```suggestion
	public @Nullable String getBaseUrl() {
```

models/spring-ai-anthropic-sdk/src/main/java/org/springframework/ai/anthropicsdk/AnthropicSdkChatModel.java
```diff
@@ -0,0 +1,1197 @@
...
+	private final ToolCallingManager toolCallingManager;
+
+	private final ToolExecutionEligibilityPredicate toolExecutionEligibilityPredicate;
+
+	private ChatModelObservationConvention observationConvention = DEFAULT_OBSERVATION_CONVENTION;
+
+	/**
+	 * Creates a new AnthropicSdkChatModel with default options.
+	 */
+	public AnthropicSdkChatModel() {
```
ericbottard: There should be only ONE constructor, which should be private.
Users construct instance via the builder.

sobychacko:
@ericbottard Thanks for the review. The initial focus was to get the basic functionality done and have feature parity with the regular anthropic module for sync and async calls. I will go through this review and address them soon. Thanks again!

sobychacko:
@ericbottard Addressed most of the comments and issues you raised. Could you take a look again? Thanks!

sdeleuze:
@sobychacko How should we proceed for this PR versus https://github.com/spring-projects/spring-ai/pull/5282 ? I guess we should decide which one is merged first.

Edit: #5282 has been merged so please let me know if you need my inputs for the rebase.

sobychacko:
@sdeleuze I used jackson2 in this PR. I think I will make it jackson 3 based.

sobychacko:
@sdeleuze I think it might be better to wait for the jackson 3 changes with the broader upgrade in the framework. What do you think?

auto-configurations/models/spring-ai-autoconfigure-model-anthropic-legacy/src/main/java/org/springframework/ai/model/anthropic/legacy/autoconfigure/AnthropicChatAutoConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+ * {@link AutoConfiguration Auto-configuration} for Anthropic Chat Model.
+ *
+ * @author Christian Tzolov
+ * @author Thomas Vitale
+ * @author Ilayaperumal Gopinathan
+ * @author Hyoseop Song
+ * @author Yanming Zhou
+ * @since 1.0.0
+ */
+@AutoConfiguration(after = { RestClientAutoConfiguration.class, WebClientAutoConfiguration.class,
```
ericbottard: This configuration only defines 2 beans, one of type AnthropicApi and one of type AnthropicChatModel (ChatModel).

None of those are defined in the configurations cited in `after`, so that attribute should be removed

auto-configurations/models/spring-ai-autoconfigure-model-anthropic/src/main/java/org/springframework/ai/model/anthropic/autoconfigure/AnthropicChatAutoConfiguration.java
```diff
@@ -16,88 +16,73 @@
...
- * @author Yanming Zhou
- * @since 1.0.0
+ * @author Soby Chacko
+ * @since 2.0.0
  */
-@AutoConfiguration(after = { RestClientAutoConfiguration.class, WebClientAutoConfiguration.class,
-		ToolCallingAutoConfiguration.class, SpringAiRetryAutoConfiguration.class })
-@EnableConfigurationProperties({ AnthropicChatProperties.class, AnthropicConnectionProperties.class })
-@ConditionalOnClass(AnthropicApi.class)
+@AutoConfiguration(after = { ToolCallingAutoConfiguration.class })
```
ericbottard, line 45: See my comment above about after

auto-configurations/models/spring-ai-autoconfigure-model-anthropic-legacy/src/main/java/org/springframework/ai/model/anthropic/legacy/autoconfigure/AnthropicChatAutoConfiguration.java
```diff
@@ -0,0 +1,108 @@
+/*
```
ericbottard: What's with that "legacy" module?

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatOptions.java
```diff
@@ -418,366 +487,489 @@ public boolean equals(Object o) {
...
+				+ this.toolContext + ", citationDocuments=" + this.citationDocuments + ", cacheOptions="
+				+ this.cacheOptions + ", outputConfig=" + this.outputConfig + ", httpHeaders=" + this.httpHeaders
+				+ ", skillContainer=" + this.skillContainer + '}';
+	}
 
-		private ChatCompletionRequest.@Nullable Metadata metadata;
+	/**
+	 * Builder for creating {@link AnthropicChatOptions} instances.
+	 */
+	public static final class Builder {
```
ericbottard: Please use the builder pattern found eg in 88fbe6bc5d7aa02c135808d55adc25d3c75b7e8a

models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/AnthropicChatOptionsTests.java
```diff
@@ -16,163 +16,173 @@
...
+ * copy, merge, equals/hashCode, and validation.
  *
- * @author Alexandros Pappas
  * @author Soby Chacko
- * @author Austin Dase
- * @author Filip Hrisafov
  */
-class AnthropicChatOptionsTests<B extends AnthropicChatOptions.Builder<B>>
-		extends AbstractChatOptionsTests<AnthropicChatOptions, B> {
+class AnthropicChatOptionsTests {
```
ericbottard: See 88fbe6bc5d7aa02c135808d55adc25d3c75b7e8a for added builder tests

auto-configurations/models/spring-ai-autoconfigure-model-anthropic/src/main/java/org/springframework/ai/model/anthropic/autoconfigure/AnthropicChatAutoConfiguration.java
sobychacko: The `anthropicChatModel` bean directly injects `ToolCallingManager`, which is provided by `ToolCallingAutoConfiguration`. Removing the `after` attribute breaks various auto-config tests since `SpringAiTestAutoConfigurations.of()` rely on it to include dependency configs in the test context. `OpenAiSdkChatAutoConfiguration` also follows the same pattern. Therefore I am keeping the `after` as is, unless you see further issues with that.

ericbottard: This is not correct. `SpringAiTestAutoConfigurations` is wrong too (it adding any config that is in `after` to the list of configs to load is wrong and does not reflect what happens at runtime). Please remove the `after`s from this PR and I'll also create a PR to get rid of `SpringAiTestAutoConfigurations` for M3

sobychacko: @ericbottard Based on our discussion on this matter, we are good here, right?

ericbottard: yes, let's postpone

tzolov:
Great contribution. Thanks @sobychacko 
Rebased and merged at b0ce25aed319d4fb0c80535375a81d8ce57bdb80

PR #5371, opened 2026-01-30
Move swagger files out of resources directory.

ericbottard:
Template files and OpenAPI specs do not belong to application resources and should not be packaged in the generated JAR.

Auto-cherry-pick to 1.1.x


ilayaperumalg:
Rebased and merged as c85105c6b and backported as well.

Issue #5256, opened 2026-01-15
[Redis] `SemanticCacheAdvisor`: Cache key only considers user text, ignores system prompt and context


sobychacko:
The cache key is derived only from the user's message text, ignoring system prompts, conversation history, and model parameters.

  ## Current Behavior
  ```java
  private String extractUserTextFromRequest(ChatClientRequest request) {
      return request.prompt().getUserMessage().getText();
  }
```

 ## Problem

  Two identical questions with different system prompts will hit the same cache entry:

  - Query: "Summarize this" + System: "You are a technical writer" → Response A
  - Query: "Summarize this" + System: "You are a poet" → Returns Response A (incorrect!)

  Similarly, different model parameters (temperature, etc.) are not considered.

  ## Suggested Approach

  Consider including in the cache key:
  1. System prompt (or its hash)
  2. Relevant model parameters (temperature, model name)
  3. Optionally, conversation history context

PR #5392, opened 2026-02-03
GH-5256: Add system prompt isolation to semantic cache

sobychacko:
Fixes: https://github.com/spring-projects/spring-ai/issues/5256

Previously, the semantic cache only used the user message text as the cache key, ignoring the system prompt. This caused incorrect cache hits when the same user question was asked with different system prompts (e.g., different AI personas or contexts).

This change introduces context-aware caching that isolates cached responses based on the system prompt:

- Add context-aware methods to SemanticCache interface:
  - set(query, response, contextHash)
  - get(query, contextHash)

- Implement metadata-based filtering in DefaultSemanticCache:
  - Store context_hash as a TAG field in document metadata
  - Filter by exact hash match during retrieval

- Update SemanticCacheAdvisor to compute deterministic SHA-256 hash of system prompt and pass it to cache operations

 - Add comprehensive tests:
  - Unit tests for hash computation and context isolation
  - Integration tests for system prompt isolation scenarios

- Update reference documentation with new sections:
  - System Prompt Isolation
  - Context-Aware Cache API

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/vectordbs/redis.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/vectordbs/redis.adoc
@@ -600,6 +600,74 @@ SemanticCacheAdvisor user1Advisor = SemanticCacheAdvisor.builder()
     .build();
 ----
 
+=== System Prompt Isolation
+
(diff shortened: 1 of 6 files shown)


markpollack:
rebased and merged in 7b30b500943bec086548b21938394b93c4c2e8c9

PR #5397, opened 2026-02-04
Add Builder pattern to OpenAiSdkChatModel

sobychacko:
- Add static builder() factory method and nested Builder class
- Deprecate all public constructors (since 2.0.0-M3, for removal in 2.1.x)
- Update auto-configuration to use builder pattern
- Update tests to use builder pattern
- Update documentation examples to use builder pattern


ilayaperumalg:
Rebased and merged into main via e58dd56fd and 1.1.x via 982bc1ea1

PR #5403, opened 2026-02-04
Make `ModelRequest#getOptions` non-null

sdeleuze:
Before this commit, `ModelRequest#getOptions` was nullable but almost all usages were expecting mandatory options.

So this commit turns it to non-null and adapt usages accordingly, using default/empty options where relevant.

### DIFF
--- a/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatOptions.java
+++ b/models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatOptions.java
@@ -80,7 +80,7 @@ public class VertexAiGeminiChatOptions implements ToolCallingChatOptions, Struct
 	 * Optional. If true, returns the log probabilities of the tokens that were chosen by the model at each step.
 	 * By default, this parameter is set to false.
 	 */
-	private @JsonProperty("responseLogprobs") boolean responseLogprobs;
+	private @JsonProperty("responseLogprobs") Boolean responseLogprobs;
 
 	/**
 	 * Optional. If specified, nucleus sampling will be used.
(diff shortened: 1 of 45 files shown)


ilayaperumalg (approved):
LGTM

models/spring-ai-vertex-ai-gemini/src/main/java/org/springframework/ai/vertexai/gemini/VertexAiGeminiChatOptions.java
```diff
@@ -80,7 +80,7 @@ public class VertexAiGeminiChatOptions implements ToolCallingChatOptions, Struct
 	 * Optional. If true, returns the log probabilities of the tokens that were chosen by the model at each step.
 	 * By default, this parameter is set to false.
 	 */
-	private @JsonProperty("responseLogprobs") boolean responseLogprobs;
+	private @JsonProperty("responseLogprobs") Boolean responseLogprobs;
```
ericbottard, line 83: Seems like an unrelated change (which contradicts the javadoc)

sdeleuze: I had to refine this to fix the merge logic that was breaking a test otherwise, in order to differentiate cases where this is not set explicitly versus explicitly set to `false`, [because this property is optional](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#responseLogprobs) and in order to be consistent with other properties that are not using primitive type.

The Javadoc is generated from https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference so it needs to be read as, if you don't define it on Java side (`null` value which will be annotated when I will make this module null-safe), the model will consider it as `false`.

I understand it could be slightly confusing, but since this is the global documentation principle that has been made for the whole class, so I tend to think it is ok to keep it as it is in my PR, but please let me know if you see a better option.

sdeleuze:
Due to various unwanted side effects, this change has been reverted.

PR #5407, opened 2026-02-05
Add design document acting as a TLDR for using JSpecify.

ericbottard:


design/01-null-safety.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+== Problems
+The "billion dollar mistake" is well known, and Spring AI 2.0 aligns with the rest of the Spring portfolio in its use of JSpecify annotations to express (non-)nullability of its APIs.
+
+This document is a quick reference for contributors to learn how to properly annotate the code base with JSpecify annotations.
+
+For further reference, please consult the following:
+
+* JSpecify https://jspecify.dev/docs/user-guide/[user-guide]
+* Spring https://spring.io/blog/2025/03/10/null-safety-in-spring-apps-with-jspecify-and-null-away[Blog Post] about null safety
```
sdeleuze: Probably worth to add https://spring.io/blog/2025/11/12/null-safe-applications-with-spring-boot-4 and https://docs.spring.io/spring-framework/reference/core/null-safety.html.

design/01-null-safety.adoc
```diff
@@ -0,0 +1,145 @@
...
+* https://github.com/uber/NullAway[NullAway] and https://errorprone.info/[ErrorProne]
+
+== Solution
+Spring AI uses JSpecify + NullAway + ErrorProne to enforce nullability checks.
```
sdeleuze: ErrorProne is an implementation detail, I would not mention it except initially as you did already above.

design/01-null-safety.adoc
```diff
@@ -0,0 +1,145 @@
...
+Any new package should be annotated with `@NullMarked`, like so:
+
+[source,java]
+----
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+package org.springframework.ai.foo;
+----
+From there on, any element in that package assumes non nullability by default (for its own members).
```
sdeleuze: "assumes non-null type usage by default" is maybe more accurate than "assumes non nullability by default"

design/01-null-safety.adoc
```diff
@@ -0,0 +1,145 @@
...
+[source, java]
+----
+import org.springframework.util.Assert;
+
+// in a @NullMarked package
+public class MyThing {
+
+	// bar and foo are assumed to be non-null
+    public MyThing(String bar, Foo foo) {
+        Assert.notNull(bar, "bar should not be null");
```
sdeleuze, line 38: Maybe worth to decide if we use consistently "should not be null" or "must not be null", I think I prefer the later.

design/01-null-safety.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+    @Nullable
+    public Foo something() {
+        ...
+        return null;
+    }
+}
+----
+
+As a rule of thumb, the annotation should be placed closest to the thing it expresses the nullability of:
```
sdeleuze, line 117: The is a global remark about the doc that is very visible in that section, but IMO if we link https://docs.spring.io/spring-framework/reference/core/null-safety.html earlier in the doc, maybe we can focus on topics that are Spring AI specific or that we want to highlight. Not sure we have to re-explain everything as the Spring Framework null-safety documentation is expected to apply to the whole portfolio, and we can potentially refine it with points that Spring AI wants to clarify. That could avoid maintenance issue.

sdeleuze (changes requested):

design/01-null-safety.adoc
ericbottard: I would agree with you when it comes to this particular case (constructor enforcement).

Note that there is also another case, code paths where we expect things to not be null because of domain specific rules / previous if-then that the compiler can not reason about / moon phase alignment / etc.
In that case, I tend to use `... should not be null`

design/01-null-safety.adoc
sdeleuze: Discussed via Zoom, does not hurt to highlight the most important points here.

PR #5412, opened 2026-02-05
Support dynamic disabling of Structured Output Native

filiphr:
With this PR it allows users to dynamically disable Structured Output Native, even if it is globally enabled.

PR #5417, opened 2026-02-06
Return correct output schema for OpenAi(Sdk)ChatOptions

filiphr:
- Fix NPE in OpenAi(Sdk)ChatOptions and return correct schema
- Add output schema test cases across for different models

### DIFF
--- a/models/spring-ai-anthropic/pom.xml
+++ b/models/spring-ai-anthropic/pom.xml
@@ -106,6 +106,13 @@
 		<artifactId>jackson-dataformat-xml</artifactId>
 		<scope>test</scope>
 	</dependency>
+
+		<dependency>
+			<groupId>net.javacrumbs.json-unit</groupId>
+			<artifactId>json-unit-assertj</artifactId>
+			<version>${json-unit-assertj.version}</version>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 
 </project>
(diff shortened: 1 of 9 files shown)


filiphr:
Hey @ilayaperumalg, have you had the time to review this PR? Is there something more that I should do for it?

models/spring-ai-anthropic/pom.xml
```diff
@@ -106,6 +106,11 @@
 			<artifactId>jackson-dataformat-xml</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
ilayaperumalg, line 110: Can this dependency be avoided? We use `org.skyscreamer.jsonassert.JSONAssert` for JSON assertions in other tests. As a side note, this should have been on "test" scope.

filiphr: Thanks @ilayaperumalg. I've fixed the scope here, it should be test scope. I decided to use this since the same dependency is already used in other modules as well, e.g.

* spring-ai-openai
* spring-ai-ollama
* spring-ai-chrome-store
etc.

However, if you prefer `org.skyscreamer.jsonassert.JSONAssert`, I can look into adjusting it. It's used less though.

filiphr:
Thanks for the review @ilayaperumalg. I've fixed the scope of the dependency and I've replied to your comment. I also rebased and fixed the conflicts.

models/spring-ai-anthropic/pom.xml
ilayaperumalg: Yeah, I see this module is being used in other modules as well. Thanks for pointing this out. Let's keep yours for now.

ilayaperumalg:
@filiphr Thanks for the quick follow-up! Rebased and merged as 0f91e2164

filiphr:
Thanks a lot for merging this @ilayaperumalg. By the way, I have some other open PRs, e.g. #5412, I have more things prepared to push as well related to that one, but I first wanted to see if you even agree with that one. Is it possible to get some feedback on it as well?

PR #5421, opened 2026-02-06
Make `spring-ai-deepseek` null-safe

sdeleuze:


models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/api/DeepSeekApi.java
```diff
@@ -370,27 +364,15 @@ public enum Type {
...
-			@JsonIgnore
-			private String jsonSchema;
-
-			/**
-			 * NOTE: Required by Jackson, JSON deserialization!
-			 */
-			@SuppressWarnings("unused")
-			private Function() {
-			}
+			private @Nullable String jsonSchema;
```
ericbottard: `getJsonSchema()` is never used.
As such, there does not need to be a field for `jsonSchema` (only one for `parameters`) and `setJsonSchema()` sets `parameters`.

models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/api/DeepSeekApi.java
```diff
@@ -406,7 +388,11 @@ private Function() {
 			 * defined in the parameters field. Only a subset of JSON Schema is supported
 			 * when strict is true.
 			 */
-			public Function(String description, String name, Map<String, Object> parameters, Boolean strict) {
+			@JsonCreator
```
ericbottard, line 389: If I understand correctly, `Function` is never de-serialized, only ever serialized.

models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/api/DeepSeekApi.java
sdeleuze: Good point, I have refined that as part of 4589c4d254f1562c0786535c08646f875f2c7b7d.

models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/api/DeepSeekApi.java
sdeleuze: Indeed, but `@JsonCreator` allows Jackson to know where to find the `@JsonProperty` annotations so I tend to think the code is fine as it is.

ericbottard (approved):

PR #5422, opened 2026-02-07
Improve chat client entity with structured output integration tests

nicolaskrier:
## Description

- Use `MistralAiChatOptions` instead of `DefaultChatOptions` to fix `chatClientEntityWithStructuredOutput` test,
- Polish `MistralAiChatModelIT` integration tests.

## Explanations

I have noticed a failing integration for Mistral AI on [the latest integration tests](https://github.com/spring-projects/spring-ai-integration-tests/actions/runs/21738860431/job/62709924400).
The test is failing due to [this assertion](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-mistral-ai/src/test/java/org/springframework/ai/mistralai/MistralAiChatModelIT.java#L539-L542) that is not true. The `ChatOptions` is not an instance of `MistralAiChatOptions` [here](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-mistral-ai/src/test/java/org/springframework/ai/mistralai/MistralAiChatModelIT.java#L508) so `nativeStructuredOutputUsed` flag is not set to true.

`ChatOptions` used to be nullable and a vendor specific implementation like `MistralAiChatOptions` was created in the end when it was null. `ChatOptions` isn't nullable anymore and `DefaultChatOptions` is used by default. Consequently, the caller is forced to provide the `ChatOptions` specific to the vendor if it wants to benefit from its behavior.

I think it is acceptable to do so here to fix this integration test.

## Related work

The non-nullability behavior has been introduced with #5403.

### DIFF
--- a/models/spring-ai-mistral-ai/src/test/java/org/springframework/ai/mistralai/MistralAiChatModelIT.java
+++ b/models/spring-ai-mistral-ai/src/test/java/org/springframework/ai/mistralai/MistralAiChatModelIT.java
@@ -27,8 +27,6 @@ import java.util.stream.Collectors;
 
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.ValueSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
@@ -89,22 +87,10 @@ class MistralAiChatModelIT {
 	private static final Logger logger = LoggerFactory.getLogger(MistralAiChatModelIT.class);
 
 	@Autowired
-	protected ChatModel chatModel;
+	private ChatModel chatModel;
 
 	@Autowired
-	protected StreamingChatModel streamingChatModel;
-
-	@Value("classpath:/prompts/eval/qa-evaluator-accurate-answer.st")
-	protected Resource qaEvaluatorAccurateAnswerResource;
-
-	@Value("classpath:/prompts/eval/qa-evaluator-not-related-message.st")
-	protected Resource qaEvaluatorNotRelatedResource;
-
-	@Value("classpath:/prompts/eval/qa-evaluator-fact-based-answer.st")
-	protected Resource qaEvaluatorFactBasedAnswerResource;
-
-	@Value("classpath:/prompts/eval/user-evaluator-message.st")
-	protected Resource userEvaluatorResource;
+	private StreamingChatModel streamingChatModel;
 
 	@Value("classpath:/prompts/system-message.st")
 	private Resource systemResource;
@@ -140,7 +126,6 @@ class MistralAiChatModelIT {
 			.build();
 		Prompt prompt = new Prompt(promptTemplate.createMessage());
 		Generation generation = this.chatModel.call(prompt).getResult();
-
 		List<String> list = outputConverter.convert(generation.getOutput().getText());
 		assertThat(list).hasSize(5);
 	}
@@ -208,8 +193,9 @@ class MistralAiChatModelIT {
 
 		String generationTextFromStream = this.streamingChatModel.stream(prompt)
 			.collectList()
-			.block()
+			.blockOptional()
 			.stream()
+			.flatMap(List::stream)
 			.map(ChatResponse::getResults)
 			.flatMap(List::stream)
 			.map(Generation::getOutput)
@@ -217,7 +203,7 @@ class MistralAiChatModelIT {
 			.collect(Collectors.joining());
 
 		ActorsFilmsRecord actorsFilms = outputConverter.convert(generationTextFromStream);
-		logger.info("" + actorsFilms);
+		logger.info(actorsFilms.toString());
 		assertThat(actorsFilms.actor()).isEqualTo("Tom Hanks");
 		assertThat(actorsFilms.movies()).hasSize(5);
 	}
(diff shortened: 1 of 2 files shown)


models/spring-ai-mistral-ai/src/test/java/org/springframework/ai/mistralai/MistralAiChatModelIT.java
```diff
@@ -25,10 +25,9 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
+import org.jspecify.annotations.NonNull;
```
ericbottard: Tests are not (and thus should not be) using JSpecify. Not really sure what annotating parameters does here.

nicolaskrier: My intent was to get rid of the following warning from my IntelliJ IDEA: _Not annotated method overrides method annotated with @NullMarked._ I have also tried using `@SuppressWarnings("NullAway")` but the warning remained with this approach.
How do you usually handle this kind of warning in your unit and/or integration tests?

ericbottard: I see. Our current motto is that we do not use JSpecify annotations on test code, at all. This may change in the future and we would take a global approach then. For the time being, don't bother :)

sdeleuze: Agreed, for now please remove JSpecify imports and annotation in tests.

sdeleuze (commented):
@nicolaskrier After a second look, I am wondering if we can avoid this kind of breakage without changing the code on user site, could you please share the error you see on integration tests?

ilayaperumalg:
@sdeleuze Here is the integration test failure: 

> Error:  Tests run: 16, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 22.10 s <<< FAILURE! -- in org.springframework.ai.mistralai.MistralAiChatModelIT
Error:  org.springframework.ai.mistralai.MistralAiChatModelIT.chatClientEntityWithStructuredOutput -- Time elapsed: 0.552 s <<< FAILURE!
org.opentest4j.AssertionFailedError: 
[Native structured output should be used with ResponseFormat.Type.JSON_SCHEMA] 
Expecting value to be true but was false
	at org.springframework.ai.mistralai.MistralAiChatModelIT.chatClientEntityWithStructuredOutput(MistralAiChatModelIT.java:542)

You can find the complete test run [here](https://github.com/spring-projects/spring-ai-integration-tests/actions/runs/21738860431/job/62709924400)

nicolaskrier:
> @nicolaskrier After a second look, I am wondering if we can avoid this kind of breakage without changing the code on user site, could you please share the error you see on integration tests?

@sdeleuze:
- I have updated the description with better explanations. I hope it is clearer now. 
- I am not sure if we should modify the test or the actual behavior. Providing by default a non-vendor specific `ChatOptions` is debatable. As a Spring AI user, I will expect this behavior at least to be documented (maybe in `Prompt` class Javadoc?) and highlighted as a breaking change for Spring AI 2 migration.
- It is not handy to investigate these modifications without a Mistrai AI API key that is why I have implemented an integration test similar to Mistral AI's one for Ollama in a second commit of this pull request. I hope it will help you.

@ilayaperumalg: Thank you for quoting the test failure, I should have done it initially.

sdeleuze:
@nicolaskrier Thanks for adding the Ollama test, indeed that helps a lot.

Looks like indeed a side effect of #5403 due to `ChatModelCallAdvisor#augmentWithFormatInstructions` storing the output schema in the prompt option only if it is an instance of `StructuredOutputChatOptions`.

IMO it is not acceptable to force users to do an explicit `.chatOptions(OllamaChatOptions.builder().build())` since it should provide the same behavior than not invoking it.

nicolaskrier:
I had another idea in mind before seeing your suggestion. I added a third commit containing my proposal. There are still several unit tests failing but `chatClientEntityWithStructuredOutput` integration tests for Mistral AI and Ollama are passing on my computer. What do you think about this approach?

If you prefer yours, I propose that you create another PR with your approach and then I will modify this PR to polish `MistralAiChatModelIT` integration tests and to add `chatClientEntityWithStructuredOutput` integration test to `OllamaChatModelIT` because I will be less available the coming days.

sdeleuze:
I think the approach you proposed will still induce side effects given current Spring AI design, so maybe yeah please polish this PR without the third commit (it is ok to force push) and I will create another PR to fix the regression I introduced with #5403 and merge both PRs.

nicolaskrier:
Thank you for having taken time to review my proposal. I think the incoming changes are quite impactful, so I appreciate you spotting the side effects early on.
I will take a closer look at your PR, and in the meantime I have made the adjustments we discussed.

sdeleuze:
Merged in `main` via 617c290d5503796e35e330e492d781fe4c724b6c, I have chosen to revert #5403 as at prompt level, the options should in fact remain nullable, otherwise too much side effects.

I will monitor CI runs of `OllamaChatModelIT` as it works in IDEA but fails on command line for reasons I am for now unable to explain (I cleaned the compilation cache and tried various other things).

sdeleuze:
So with @ilayaperumalg we confirm that in IDEA, `OllamaChatModelIT#chatClientEntityWithStructuredOutput` works because the options is an instance of `OllamaChatOptions` but on command line, it fails because it is a `DefaultChatOptions`! We are trying to figure out why.

nicolaskrier:
This is weird! I have seen #5441 planned for M4. You may consider disabling this test until this issue is resolved if it becomes a blocking point.

sdeleuze:
Seems like this was due to the command documented to run integration tests in https://github.com/spring-projects/spring-ai?tab=readme-ov-file#building missing the `-am` flag (to build also dependent modules in the Maven Reactor) and the `-Dfailsafe.failIfNoSpecifiedTests=false`one  (to not fail due to the test specified not being in the dependent modules). Took me some time to figure out as the test running was the correct one but with an old implementation.

I will confirm tomorrow with @ilayaperumalg and @ericbottard, and update the README accordingly if confirmed.

PR #5431, opened 2026-02-09
fix(vectorstore): parse long integer literals in text filters

### LINKED ISSUE #4705: FilterExpressionTextParser treats all number literals as Integer, failing on Long values
**Bug description**

```java
@Override
}
```

Example exception:

```
```

---

**Environment**

* Spring AI version: 1.0.3
* Java version: 21
* OS: Linux (Ubuntu 24.04)

---

**Steps to reproduce**

```java

public class ParserExample {

    }
}
```

**Actual behavior:**

```
```

---

**Expected behavior**

---

```java
```

**Expected:** no exception

---

**Proposed fix**

```java
@Override
  String text = ctx.getText();
  try {
}
```

weiguangli-io:
## Summary

Fixes #4705.

`FilterExpressionTextParser` currently parses all integer tokens as `Integer`, which throws `NumberFormatException` for values outside the 32-bit range (for example `Long.MAX_VALUE`).

## What changed

- Updated integer constant parsing in `FilterExpressionTextParser`:
  - First try `Integer.valueOf(...)`
  - On overflow, fallback to `Long.valueOf(...)`
- Extended parser tests to cover large signed long literals:
  - `Long.MAX_VALUE`
  - `Long.MIN_VALUE`

## Validation

Executed locally:

```bash
export JAVA_HOME=/usr/local/opt/openjdk@21/libexec/openjdk.jdk/Contents/Home
./mvnw -pl spring-ai-vector-store -Dtest=FilterExpressionTextParserTests test
```

Result: `BUILD SUCCESS` (`15` tests, `0` failures)

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
@@ -215,7 +215,13 @@ public class FilterExpressionTextParser {
 
 		@Override
 		public Filter.Operand visitIntegerConstant(FiltersParser.IntegerConstantContext ctx) {
-			return new Filter.Value(Integer.valueOf(ctx.getText()));
+			String value = ctx.getText();
+			try {
+				return new Filter.Value(Integer.valueOf(value));
+			}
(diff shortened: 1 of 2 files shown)


ericbottard:
Closing this PR as there already is support for explicitly using Longs in literals (use `123L`)

PR #5433, opened 2026-02-09
fix(mcp): warn when stateless MCP tools are skipped

### LINKED ISSUE #5373: Tools with McpSyncServerExchange/McpSyncRequestContext parameters are silently ignored in stateless mode
### Problem
The McpSyncServerExchange or McpSyncRequestContext are not supported in stateless mode, they make sense only with sessions.
However, when using stateless mode (spring.ai.mcp.server.protocol=stateless), tool methods with special parameters like McpSyncServerExchange or McpSyncRequestContext are silently ignored - the tool is not registered and never called. There's no warning or error at startup.

### Steps to reproduce:

1. Configure stateless MCP server:
`spring.ai.mcp.server.protocol=stateless`

2. Create a tool with special parameter:
```
@McpTool(description = "Get items")
public List<Item> getItems(McpSyncServerExchange exchange) {
    return items;
}
```

3. Configure client with streamable-http:
`spring.ai.mcp.client.streamable-http.connections.server1.url=http://localhost:8081`

4. Call the tool - it's not found/registered. No warning in logs.

[unicorn-mcp-server.zip](https://github.com/user-attachments/files/24970060/unicorn-mcp-server.zip)
[unicorn-spring-ai-agent.zip](https://github.com/user-attachments/files/24970059/unicorn-spring-ai-agent.zip)

### Current behavior
Tool is silently not registered.

### Expected behavior (options)

1. Preferred: Register the tool and invoke it with null injected for unsupported parameters. Let the developer handle the null case:
  
java
```
   @McpTool(description = "Get items")
   public List<Item> getItems(McpSyncServerExchange exchange) {
       if (exchange != null) {
           // stateful-specific logic
       }
       return items;
   }
```

2. Alternative: Log a clear warning at startup:
   WARN: Tool 'getItems' uses McpSyncServerExchange which is not available in stateless mode. 
         Tool will not be registered. Remove the parameter or switch to stateful mode.
   
The current silent behavior makes debugging difficult - developers don't know why their tool isn't being called.

### Environment

- Spring AI version: 1.1.2
- Spring Boot version: 3.5.3
- Java version: 25

weiguangli-io:
## Summary
This PR improves the stateless MCP experience when `@McpTool` methods use context parameters that are unsupported in stateless mode.

Currently those tools are silently dropped by the underlying provider predicate, which makes the behavior hard to diagnose.

This change adds an explicit warning during stateless tool specification discovery so users can immediately see why a tool was skipped.

Closes #5373

## Changes
- Added stateless warning logic in `StatelessServerSpecificationFactoryAutoConfiguration` for `@McpTool` methods using:
  - `McpSyncServerExchange`
  - `McpAsyncServerExchange`
  - `McpSyncRequestContext`
  - `McpAsyncRequestContext`
- Reused the same warning logic for both sync and async stateless tool specification paths.
- Added integration tests in `McpStatelessServerAutoConfigurationIT` to verify:
  - unsupported stateless tools are not registered (existing behavior remains)
  - warning logs are emitted with actionable guidance

## Why this helps
Users now get a direct warning explaining that the tool was skipped in stateless mode and how to proceed (`McpTransportContext` or stateful protocol), instead of having to infer the cause from missing tool registration.

## Validation
Local tests:
- `./mvnw -Dmaven.build.cache.enabled=false -pl auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common -Dtest=McpStatelessServerAutoConfigurationIT,StatelessToolCallbackConverterAutoConfigurationIT test`

Changed-line coverage (production file):
- measurable added lines: 21
- covered lines: 21
- coverage: 100%

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/annotations/StatelessServerSpecificationFactoryAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/annotations/StatelessServerSpecificationFactoryAutoConfiguration.java
@@ -16,13 +16,21 @@
 
 package org.springframework.ai.mcp.server.common.autoconfigure.annotations;
 
+import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
+import io.modelcontextprotocol.server.McpAsyncServerExchange;
 import io.modelcontextprotocol.server.McpStatelessServerFeatures;
+import io.modelcontextprotocol.server.McpSyncServerExchange;
 import org.springaicommunity.mcp.annotation.McpComplete;
 import org.springaicommunity.mcp.annotation.McpPrompt;
 import org.springaicommunity.mcp.annotation.McpResource;
 import org.springaicommunity.mcp.annotation.McpTool;
+import org.springaicommunity.mcp.context.McpAsyncRequestContext;
+import org.springaicommunity.mcp.context.McpSyncRequestContext;
 
 import org.springframework.ai.mcp.annotation.spring.AsyncMcpAnnotationProviders;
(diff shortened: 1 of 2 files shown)


weiguangli-io:
Closing this in favor of #5434 because I had to re-open from a branch that includes a DCO-signed commit.

PR #5438, opened 2026-02-10
GH-5437: Handle non-JSON tool responses in GoogleGenAiChatModel.parseJsonToMap()

### LINKED ISSUE #5437: GoogleGenAiChatModel.parseJsonToMap() crashes on non-JSON tool responses and silently loses data
## Bug description

`GoogleGenAiChatModel.parseJsonToMap()` has two defects when handling tool response strings:

1. **Crash**: Any tool that returns plain text (not valid JSON) causes an unrecoverable `RuntimeException`, terminating the entire agent loop.
2. **Silent data loss**: When a tool response happens to start with a JSON-valid token (e.g., `"3.0 + 5.0 = 8.0"`), Jackson parses only the first token (`3.0`) and silently discards the rest.

The root cause is in `messageToGeminiParts()` — when converting a `ToolResponseMessage` into Gemini API format, it calls `parseJsonToMap(response.responseData())` which blindly passes the tool result to `ObjectMapper.readValue()`. This assumes all tool responses are valid JSON, but Spring AI's `ToolCallback` interface returns `String`, which is commonly plain text.

## Environment

- **Spring AI**: 1.1.2 (also verified the code is identical in 2.0.0-M1)
- **Spring Boot**: 3.5.9
- **Java**: 21 (Kotlin 2.3.10)
- **Gemini Model**: gemini-2.0-flash

### 1. Define a tool that returns plain text

```java
@Component
public class DateTimeTool implements ToolCallback {
    @Override
    public String getName() { return "current_datetime"; }

    @Override
    public String getDescription() { return "Get the current date and time"; }

    @Override
    public String getInputSchema() {
        return """
            {
              "type": "object",
              "properties": {
                "timezone": {
                  "type": "string",
                  "description": "Timezone ID (e.g., Asia/Seoul, UTC)"
                }
              }
            }
            """;
    }

    @Override
    public Object call(Map<String, Object> arguments) {
        String tz = (String) arguments.getOrDefault("timezone", "UTC");
        ZonedDateTime now = ZonedDateTime.now(ZoneId.of(tz));
        DateTimeFormatter fmt = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss (EEEE)");
        return now.format(fmt) + " [" + tz + "]";
        // Returns: "2026-02-11 00:14:54 (Tuesday) [UTC]"
    }
}
```

### 2. Send a prompt that triggers the tool

```
POST /api/chat
{ "message": "What time is it now? Use the current_datetime tool." }
```

### 3. Observe the crash

The first LLM call succeeds — Gemini invokes the tool. The tool executes and returns the datetime string. But the **second** LLM call (which includes the tool response in conversation history) crashes because `messageToGeminiParts` passes the plain text result to `parseJsonToMap()`.

## Test matrix: `parseJsonToMap()` behavior on various inputs

| Input | Result | Issue |
|-------|--------|-------|
| `2026-02-11 00:14:54 (Tuesday) [UTC]` | **RuntimeException** | Crash — `2026` parsed as number, `-` unexpected |
| `Hello, the current time is 3pm` | **RuntimeException** | Crash — `Hello` is not a JSON token |
| `Error: Invalid timezone` | **RuntimeException** | Crash — `Error` is not a JSON token |
| `3.0 + 5.0 = 8.0` | Returns `{"result": 3.0}` | **Silent data loss** — `+ 5.0 = 8.0` discarded |
| `42 is the answer` | Returns `{"result": 42}` | **Silent data loss** — `is the answer` discarded |
| `42` | Returns `{"result": 42}` | OK |
| `true` | Returns `{"result": true}` | OK |
| `{"key": "value"}` | Returns `{"key": "value"}` | OK |

**9 out of 17 tested inputs fail, and 2 more silently lose data.**

## Stack trace

```
java.lang.RuntimeException: Failed to parse JSON: 2026-02-11 00:14:54 (Tuesday) [UTC]
    at o.s.ai.google.genai.GoogleGenAiChatModel.parseJsonToMap(GoogleGenAiChatModel.java:397)
    at o.s.ai.google.genai.GoogleGenAiChatModel.messageToGeminiParts(GoogleGenAiChatModel.java:337)
    at o.s.ai.google.genai.GoogleGenAiChatModel.toGeminiContent(GoogleGenAiChatModel.java:885)
    at o.s.ai.google.genai.GoogleGenAiChatModel.createGeminiRequest(GoogleGenAiChatModel.java:830)
    at o.s.ai.google.genai.GoogleGenAiChatModel.internalCall(GoogleGenAiChatModel.java:437)
    ...
Caused by: com.fasterxml.jackson.core.JsonParseException:
    Unexpected character ('-' (code 45)): Expected space separating root-level values
    at [Source: REDACTED; line: 1, column: 5]
```

## Root cause analysis

In `GoogleGenAiChatModel.java`, `messageToGeminiParts` handles `ToolResponseMessage`:

```java
if (message instanceof ToolResponseMessage toolResponseMessage) {
    return toolResponseMessage.getResponses().stream()
        .map(response -> Part.builder()
            .functionResponse(FunctionResponse.builder()
                .name(response.name())
                .response(parseJsonToMap(response.responseData()))  // ← crashes here
                .build())
            .build())
        .toList();
}
```

And `parseJsonToMap` has no fallback:

```java
private static Map<String, Object> parseJsonToMap(String json) {
    try {
        Object parsed = OBJECT_MAPPER.readValue(json, Object.class);
        if (parsed instanceof List) return Map.of("result", parsed);
        if (parsed instanceof Map) return (Map) parsed;
        return Map.of("result", parsed);
    } catch (Exception e) {
        throw new RuntimeException("Failed to parse JSON: " + json, e);  // no fallback
    }
}
```

## Suggested fix

Wrap non-JSON text in the catch block instead of throwing:

```java
private static Map<String, Object> parseJsonToMap(String json) {
    try {
        Object parsed = OBJECT_MAPPER.readValue(json, Object.class);
        if (parsed instanceof List) return Map.of("result", parsed);
        if (parsed instanceof Map) return (Map) parsed;
        return Map.of("result", parsed);
    } catch (Exception e) {
        // Gracefully handle non-JSON tool responses by wrapping as {"result": "..."}
        Map<String, Object> wrapper = new HashMap<>();
        wrapper.put("result", json);
        return wrapper;
    }
}
```

This is consistent with:
- The Gemini API's `FunctionResponse`, which accepts any `Map<String, Object>` as the response
- The method's existing behavior for non-Map values (List, primitives), which are already wrapped as `{"result": value}`

## Impact

- **All tools returning plain text are broken** when using Google GenAI / Gemini
- This is language-agnostic — English text fails identically to non-English text
- Other model providers (OpenAI, Anthropic) are **not affected**
- **Workaround**: Return tool results as JSON strings (e.g., `{"result": "..."}`)

wlsdks:
## Summary

`GoogleGenAiChatModel.parseJsonToMap()` throws `RuntimeException` when a tool returns plain text (not valid JSON), crashing the entire agent loop. This PR fixes the issue by gracefully wrapping non-JSON text as `{"result": "<text>"}` in the catch block, consistent with how the method already handles non-Map parsed values (List, primitives).

- Handle non-JSON tool responses gracefully instead of throwing RuntimeException
- Add 4 unit tests covering plain text, error text, valid JSON, and JSON array tool responses
- All 184 module tests pass

Fixes https://github.com/spring-projects/spring-ai/issues/5437

## Test plan

- [x] `messageToGeminiPartsShouldHandlePlainTextToolResponse` — datetime string like `"2026-02-11 00:14:54 (Tuesday) [UTC]"`
- [x] `messageToGeminiPartsShouldHandleErrorTextToolResponse` — error message like `"Error: Invalid timezone"`
- [x] `messageToGeminiPartsShouldHandleValidJsonToolResponse` — valid JSON still works as before
- [x] `messageToGeminiPartsShouldHandleJsonArrayToolResponse` — JSON arrays still wrapped correctly

### DIFF
--- a/models/spring-ai-google-genai/src/main/java/org/springframework/ai/google/genai/GoogleGenAiChatModel.java
+++ b/models/spring-ai-google-genai/src/main/java/org/springframework/ai/google/genai/GoogleGenAiChatModel.java
@@ -394,7 +394,12 @@ public class GoogleGenAiChatModel implements ChatModel, DisposableBean {
 			}
 		}
 		catch (Exception e) {
-			throw new RuntimeException("Failed to parse JSON: " + json, e);
+			// Gracefully handle non-JSON tool responses (e.g. plain text)
+			// by wrapping them as {"result": "<text>"}, consistent with
+			// how non-Map parsed values (List, primitives) are already wrapped.
+			Map<String, Object> wrapper = new HashMap<>();
+			wrapper.put("result", json);
+			return wrapper;
 		}
 	}
 
--- a/models/spring-ai-google-genai/src/test/java/org/springframework/ai/google/genai/GoogleGenAiChatModelParseJsonTests.java
+++ b/models/spring-ai-google-genai/src/test/java/org/springframework/ai/google/genai/GoogleGenAiChatModelParseJsonTests.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2023-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.google.genai;
+
+import java.util.List;
+import java.util.Map;
+
+import com.google.genai.types.Part;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.ai.chat.messages.ToolResponseMessage;
+import org.springframework.ai.chat.messages.ToolResponseMessage.ToolResponse;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for
+ * {@link GoogleGenAiChatModel#messageToGeminiParts(org.springframework.ai.chat.messages.Message)}
+ * verifying that non-JSON tool responses are handled gracefully instead of throwing
+ * RuntimeException.
+ *
+ * @author Jinan Choi
+ * @since 2.0.0
+ * @see <a href="https://github.com/spring-projects/spring-ai/issues/5437">GH-5437</a>
+ */
+class GoogleGenAiChatModelParseJsonTests {
+
+	@Test
+	void messageToGeminiPartsShouldHandlePlainTextToolResponse() {
+		// A tool that returns a datetime string (not valid JSON)
+		String plainTextResult = "2026-02-11 00:14:54 (Tuesday) [UTC]";
+		ToolResponseMessage message = ToolResponseMessage.builder()
+			.responses(List.of(new ToolResponse("call-1", "current_datetime", plainTextResult)))
+			.build();
+
+		List<Part> parts = GoogleGenAiChatModel.messageToGeminiParts(message);
+
+		assertThat(parts).hasSize(1);
+		Map<String, Object> response = parts.get(0).functionResponse().get().response().get();
+		assertThat(response).containsEntry("result", plainTextResult);
+	}
+
+	@Test
+	void messageToGeminiPartsShouldHandleErrorTextToolResponse() {
+		// A tool that returns an error message (not valid JSON)
+		String errorResult = "Error: Invalid timezone 'foo/bar'";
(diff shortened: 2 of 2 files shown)


sobychacko:
Thanks for the fix, @wlsdks, and sorry this sat open for a while.

This overlaps with #6049, which targets the same issue (#5437) and applies the same core fix (wrapping non-JSON tool responses as {"result": "<text>"} instead of throwing). #6049 additionally handles a related case - partial-JSON input like "3.0 + 5.0 = 8.0", which the lenient parser would otherwise truncate to 3.0, by using a `FAIL_ON_TRAILING_TOKENS`
mapper. Its tests also line up with the current `messageToGeminiParts` instance signature on main, whereas this PR would need a rebase to compile.

Because of that, we're going to move forward with #6049 and close this one. Really appreciate the contribution and the clear test coverage, please keep them coming!

PR #5449, opened 2026-02-12
Make spring-ai-openai-sdk model null-safe

ilayaperumalg:
- All null-safety configurations for the OpenAI SDK model and autoconfiguration classes

### DIFF
--- a/models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/AbstractOpenAiSdkOptions.java
+++ b/models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/AbstractOpenAiSdkOptions.java
@@ -28,6 +28,16 @@ import org.jspecify.annotations.Nullable;
 public class AbstractOpenAiSdkOptions {
 
 	/**
+	 * Default request timeout for the OpenAI client.
+	 */
+	public static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(60);
+
+	/**
+	 * Default maximum number of retries for the OpenAI client.
+	 */
+	public static final int DEFAULT_MAX_RETRIES = 3;
+
+	/**
 	 * The deployment URL to connect to OpenAI.
 	 */
 	private @Nullable String baseUrl;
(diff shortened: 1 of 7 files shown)


ericbottard:
Looks good from a practical standpoint. 
However, I find that there are a lot of things that are nullable (I understand that this may not be your doing). Maybe we can tighten this a bit and move nullability more towards "the center" so that nullability doesn't bleed too much to calling code.

Also, and this is unrelated, we need to
* fix the builder patterns to not use an embedded instance
* hide all those constructors and force the use of the builder pattern (for OpenAiSdkImageModel for ex)

ericbottard:
Also, please delete line 84 from src/checkstyle/checkstyle-suppressions.xml

models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/AbstractOpenAiSdkOptions.java
```diff
@@ -23,46 +23,47 @@
 
 import com.openai.azure.AzureOpenAIServiceVersion;
 import com.openai.credential.Credential;
+import org.jspecify.annotations.Nullable;
 
 public class AbstractOpenAiSdkOptions {
```
sdeleuze, line 28: Please check in the OpenAI SDK what options are mandatory and make related field non-null with additional related assertions. For example, is the `apiKey` really optional or is it mandatory?

ilayaperumalg: `apiKey` is intentionally nullable as OpenAiSdkSetup already implements a full fallback chain per provider:
  explicit apiKey → env var (OPENAI_API_KEY / GITHUB_TOKEN / MICROSOFT_FOUNDRY_API_KEY) → explicit credential → Azure
  passwordless authentication (via azure-identity).

ericbottard: That what I alluded to in a meeting: IMO, the field should be NON nullable and it's not the responsibility of deep, inner code to fallback to other values. It is ok to provide a helper (in "outer" layer code though).
When options are created, apiKey should be set (if it's the only auth mechanism)

ericbottard:
Squashed, rebased and resolved conflicts on main as 73f020f5aa3037049a551b5c0cb07d2d49d1a681

PR #5450, opened 2026-02-12
DeepSeek: avoid repeated output after streaming tool calls (#2575)

### LINKED ISSUE #2575: Content field is empty in streaming response after tool call in MCP client

Okunoda:
## Summary

Fixes #2575

This PR addresses the behavior reported in #2575, where the `content` field in the assistant message can be empty/null in **streaming mode** when the model triggers a tool call. In practice (especially for tool-call chunks), the provider may omit textual content and only emit tool-call metadata, which caused downstream handling issues.

My change makes the streaming tool-call path more robust by treating tool-call chunks with missing content as a valid/expected case, so the conversation can continue without repeated or restarted output.

## Related Issue

- Fixes / Relates to: #2575

## What I changed

- Handle streaming tool-call chunks where `content` is missing/empty (i.e., tool calls are present but textual content is not).
- Preserve/restore the already-streamed assistant prefix when continuing after tool execution, so the model can continue instead of starting over (and potentially repeating content).
- Add/adjust tests to cover the streaming tool-call continuation behavior.

## Why this change

In streaming mode, some providers produce tool-call deltas with `content == null` or `content == ""` while still providing valid `tool_calls`. This makes the payload structure differ from non-streaming mode and can lead to repeated/restarted generation when we resume after tool execution.

This PR aims to make the streaming path tolerant to that behavior and keep the conversation flow consistent.

## Tests

- Added/updated regression tests for streaming tool-call continuation to ensure we do not repeat previously streamed content after tool execution.

## Review request

I would greatly appreciate a review from someone experienced with Spring AI’s streaming/tool-calling internals and the expected API semantics here. I’m not 100% certain this is the best long-term approach, and I’d welcome guidance on whether the behavior should be standardized differently (e.g., omitting `content` vs. enforcing empty string).

If this approach looks correct, I’m happy to follow up with additional PRs to apply the same pattern to other chat model implementations as needed (including, but not limited to, `OpenAiChatModel`).

## Notes

I’m new to contributing in the open-source ecosystem. If anything in this PR (structure, naming, tests, commit message, DCO sign-off, etc.) is not aligned with the project’s contribution guidelines, please let me know—any feedback is very welcome. Thank you for taking the time to review this PR.


sobychacko:
Thanks for the PR!

Closing this one as obsolete. Commit `1f1f5c824` removed internal tool execution from `DeepSeekChatModel` under #6251, so the `internalCall` / `internalStream` tool-pushback branch this patches no longer exists. The `DeepSeekApi.ChatCompletionMessage.content()` accessor it rewrites is also gone; `content` is now a `@Nullable String` record component.

The `continuousStream` idea in `spring-ai-model` has no consumer left in the tree, so if it is still wanted it needs a fresh design discussion on #2575 rather than a rebase of this branch.

Issue #5442, opened 2026-02-11
MetadataMode is ignored in batched embedding path used by vector stores


adolfoj-empathy:
**Bug description**

The [`EmbeddingModel`](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-model/src/main/java/org/springframework/ai/embedding/EmbeddingModel.java#L92) interface's default `embed(List<Document>, EmbeddingOptions, BatchingStrategy)` method extracts text via `Document::getText`, completely ignoring the `MetadataMode` configured on the embedding model. This means that when vector stores call this batched embedding path (which all built-in vector stores do), metadata is never included in the embedding text, regardless of the `MetadataMode` configuration.

There is an inconsistency between the single-document and batched embedding paths:

- `OpenAiEmbeddingModel.embed(Document)` correctly uses `document.getFormattedContent(this.metadataMode)`.
- `EmbeddingModel.embed(List<Document>, EmbeddingOptions, BatchingStrategy)` uses `Document::getText`, ignoring `metadataMode` entirely.

Since all built-in vector stores (e.g., `PgVectorStore.doAdd()`) call the batched path, the `MetadataMode` parameter accepted by embedding model constructors (e.g., `new OpenAiEmbeddingModel(api, MetadataMode.EMBED, options)`) is effectively dead code when used with vector stores.

Additionally, the `TokenCountBatchingStrategy` accepts a `ContentFormatter` and `MetadataMode`, but only uses them for **token counting** to size sub-batches -- not for formatting the actual text sent to the embedding API. This creates a misleading API surface: users may configure `MetadataMode.EMBED` expecting metadata to be included in embeddings, but it is silently ignored.

**Environment**

- Spring AI version: 2.0.0-M2 (also reproducible on `main` branch at time of writing)
- Java version: 25
- Vector store: PgVectorStore (but applies to all vector stores using the default `EmbeddingModel.embed(documents, options, batchingStrategy)`)

**Steps to reproduce**

1. Create an `OpenAiEmbeddingModel` with `MetadataMode.EMBED`:
   ```java
   var embeddingModel = new OpenAiEmbeddingModel(openAiApi, MetadataMode.EMBED, options);
   ```

2. Create a `PgVectorStore` (or any vector store) with this embedding model:
   ```java
   var vectorStore = PgVectorStore.builder(jdbcTemplate, embeddingModel)
       .batchingStrategy(new TokenCountBatchingStrategy(
[...]

PR #5452, opened 2026-02-12
Fix MetadataMode being ignored in batched embedding path

sobychacko:
Fixes https://github.com/spring-projects/spring-ai/issues/5442

The default EmbeddingModel.embed(List<Document>, EmbeddingOptions, BatchingStrategy) method used Document::getText, ignoring MetadataMode. Since all vector stores use this batched path, MetadataMode configuration on embedding models was effectively dead code.

Add getEmbeddingContent(Document) default method to EmbeddingModel that returns document.getText() for backward compatibility. Models with MetadataMode support (OpenAI, Azure OpenAI, MistralAI, ZhiPuAI, MiniMax, Transformers, OpenAI SDK, PostgresML) override it to return document.getFormattedContent(metadataMode). The batched embed method now delegates to getEmbeddingContent() instead of getText().


ilayaperumalg:
Rebased and merged into main via 15e39fe44. Backported into 1.1.x via 6fa8730c1

PR #5462, opened 2026-02-14
Remove adding toolcall history into toolcontext

### LINKED ISSUE #5455: toolContext: cleanup TOOL_CALL_HISTORY from McpMeta context
Spring AI Tool support provide nice opportunity to transfer custom metadata from chatClient into tool method using `toolContext`:
```
        var chatResponse = chatClient
                .prompt()
                .user(promptRequest.prompt())
                .toolContext(Map.of("my-context-key", "my-context-value"))
                .call().chatResponse();
```
On the server (tool) side, it is possible to read these values using special McpMeta type:
```
    @McpTool(description = "Get a list of all unicorns in the unicorn store")
    public List<Unicorn> getAllUnicorns(McpMeta meta) {
        logger.info("=== getAllUnicorns called ===");
        logger.info(meta.get("my-context-key").toString());
        return unicornsStorage.values().stream().toList();
    }
```
Unfortunately, appart from custom values, McpMeta contains additionally the whole tool call history:
```
McpMeta[meta={TOOL_CALL_HISTORY=[{messageType=SYSTEM, metadata={messageType=SYSTEM}, text=I assistant helps to recognize image and get information from here
}, {messageType=USER, metadata={messageType=USER}, media=[], text=Get a list of all unicorns in the unicorn store?}, {messageType=ASSISTANT, metadata={messageType=ASSISTANT}, toolCalls=[{id=tooluse_4lTmXExO198z8Eqrs7GkVV, type=function, name=getAllUnicorns, arguments={"context": {}}}], media=[], text=}], my-context-key=my-context-value}]
```
In case of long conversation this can be quite expensive. If there are no need and important use cases for that, can the TOOL_CALL_HISTORY be removed form the McpMeta?

ilayaperumalg:
- By default, the toolcall history is added into the toolcontext under TOOL_CALL_HISTORY key when the tool execution is handled by Spring AI's default toolcalling manager (via the implementing ChatModel's internal tool execution). This caused memory growth issues in long conversations and created an unnecessary coupling between tool execution and conversation management.
    - Given the toolcall history can be enabled using the advisor chain via the newly added ToolCallAdvisor, removing the tool call history from ToolContext avoids overloading the Spring AI's internal tool call response with the conversation history.
  - Update upgrade notes

Fixes #5455

### DIFF
--- a/models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/client/AnthropicChatClientMethodInvokingFunctionCallbackIT.java
+++ b/models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/client/AnthropicChatClientMethodInvokingFunctionCallbackIT.java
@@ -16,7 +16,6 @@
 
 package org.springframework.ai.anthropic.client;
 
-import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


ashakirin:
Would it be possible to backport this fix into Spring AI 1.X?

Issue #5444, opened 2026-02-11
gen_ai.client.token.usage Missing in Spring Boot 4


jamesward:
When I upgrade to Spring Boot 4, I no longer get the `gen_ai.client.token.usage` actuator metric.

Tested with Spring AI `1.1.2` and `2.0.0-M2` and with the Bedrock Converse & OpenAI SDK chat clients.

sobychacko:
@jamesward I think this is whats causing this issue. In Spring Boot 3.x, the `ObservationAutoConfiguration` class lived at `org.springframework.boot.actuate.autoconfigure.observation.ObservationAutoConfiguration` and declared an
  explicit ordering dependency `(after = CompositeMeterRegistryAutoConfiguration)`, which guaranteed that a `MeterRegistry` bean was available by the time it ran.

  Spring AI's observation auto-configurations (for chat, embedding, image, and vector store) used `@AutoConfiguration(afterName =
  "org.springframework.boot.actuate.autoconfigure.observation.ObservationAutoConfiguration")` to ensure they processed after Boot's observation setup, which transitively ensured `MeterRegistry` was present. In Spring Boot 4, this class was moved to `org.springframework.boot.micrometer.observation.autoconfigure.ObservationAutoConfiguration` as part of the actuator module restructuring. The old class no longer exists. When Spring Boot's `AutoConfigurationSorter` encounters an `afterName` reference to a non-existent class, it silently drops the ordering constraint. This means Spring AI's observation auto-configurations can now process before the metrics auto-configurations have created a `MeterRegistry` bean. When that happens, the `@ConditionalOnBean(MeterRegistry.class)` condition on `ChatModelMeterObservationHandler` (and similar handler beans) evaluates to `false`, and the handler bean is never created. Without this handler, no `gen_ai.client.token.usage` counter metrics are ever generated.

  We need to update the `afterName` references in all affected auto-configuration classes to point to the Boot 4 class locations:

```  
@AutoConfiguration(afterName = {
      "org.springframework.boot.micrometer.observation.autoconfigure.ObservationAutoConfiguration",
      "org.springframework.boot.micrometer.metrics.autoconfigure.CompositeMeterRegistryAutoConfiguration" })
```

The second entry is the key one — it ensures the auto-configuration runs after `CompositeMeterRegistryAutoConfiguration`, which is where the `MeterRegistry` bean is created. The first entry preserves the original semantic intent of running after the observation infrastructure is set up.

[...]

PR #5472, opened 2026-02-18
Fix auto-configuration ordering for observation metrics in Boot 4

sobychacko:
Fixes https://github.com/spring-projects/spring-ai/issues/5444

Update `afterName` references in all observation auto-configuration classes to use the Boot 4 class locations. The previous references pointed to `org.springframework.boot.actuate.autoconfigure.observation.ObservationAutoConfiguration`, which no longer exists in Boot 4, causing `AutoConfigurationSorter` to silently drop the ordering constraint.

Without correct ordering, `@ConditionalOnBean(MeterRegistry.class)` evaluates before `CompositeMeterRegistryAutoConfiguration` runs, so `ChatModelMeterObservationHandler` and `EmbeddingModelMeterObservationHandler` beans are never created, and the `gen_ai.client.token.usage` metric is never emitted.

The fix updates afterName in `ChatObservationAutoConfiguration`, `EmbeddingObservationAutoConfiguration`, `ImageObservationAutoConfiguration`, `VectorStoreObservationAutoConfiguration`, and `ChatClientAutoConfiguration` to reference both `ObservationAutoConfiguration` and `CompositeMeterRegistryAutoConfiguration` at their new `org.springframework.boot.micrometer` package locations.

Also fixes a pre-existing bug in `ImageObservationAutoConfiguration` where the `afterName` string had a spurious .class suffix.

### DIFF
--- a/auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client/src/main/java/org/springframework/ai/model/chat/client/autoconfigure/ChatClientAutoConfiguration.java
+++ b/auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client/src/main/java/org/springframework/ai/model/chat/client/autoconfigure/ChatClientAutoConfiguration.java
@@ -59,7 +59,8 @@ import org.springframework.context.annotation.Scope;
  * @since 1.0.0
  */
 @AutoConfiguration(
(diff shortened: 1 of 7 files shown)


ilayaperumalg:
Rebased and merged via c5d96da22

PR #5485, opened 2026-02-22
Redeisgn Bedrock autoconfiguration and change BedrockProxyChatModel builder design

### LINKED ISSUE #823: SpringAI does not follow the AWS SDK's default region resolution rules

### LINKED ISSUE #5108: Confusing WARN log while AWS region resolving
**Bug description**
[Current documentation](https://docs.spring.io/spring-ai/reference/api/bedrock.html#_connect_to_aws_bedrock) states:

> AWS region is resolved in the following order:
> - Spring-AI Bedrock spring.ai.bedrock.aws.region property.
> - Java System Properties - aws.region.
> - Environment Variables - AWS_REGION.
> - Credential profiles file at the default location (~/.aws/credentials) shared by all AWS SDKs and the AWS CLI.
> - Instance profile region delivered through the Amazon EC2 metadata service.

In case when region configuration provided only using **spring.ai.bedrock.aws.region** property (no Java System Properties, Env Vars, etc) Folowing WARN log appears:
```
o.s.a.b.c.BedrockProxyChatModel - Failed to load region from DefaultAwsRegionProviderChain, using US_EAST_1
2025-12-16T14:51:07.757422240Z software.amazon.awssdk.core.exception.SdkClientException: Unable to load region from any of the providers in the chain software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain@f989d46: [software.amazon.awssdk.regions.providers.SystemSettingsRegionProvider@1dca91d6: Unable to load region from system settings. Region must be specified either via environment variable (AWS_REGION) or  system property (aws.region)., software.amazon.awssdk.regions.providers.AwsProfileRegionProvider@4ebbee76: No region provided in profile: default, software.amazon.awssdk.regions.providers.InstanceProfileRegionProvider@1c7385d: Unable to contact EC2 metadata service.]
```
It happens in **BedrockConverseProxyChatAutoConfiguration** while [BedrockProxyChatModel.builder()](https://github.com/spring-projects/spring-ai/blob/main/auto-configurations/models/spring-ai-autoconfigure-model-bedrock-ai/src/main/java/org/springframework/ai/model/bedrock/converse/autoconfigure/BedrockConverseProxyChatAutoConfiguration.java#L74C19-L74C50) execution. Builder in its constructor [checks DefaultAwsRegionProviderChain](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java#L892-L899) in order to find region and produces WARN log if nothing was found without consideration of **spring.ai.bedrock.aws.region** property. 

Hopefully autoconfiguration [provides region to the builder after creation](https://github.com/spring-projects/spring-ai/blob/main/auto-configurations/models/spring-ai-autoconfigure-model-bedrock-ai/src/main/java/org/springframework/ai/model/bedrock/converse/autoconfigure/BedrockConverseProxyChatAutoConfiguration.java#L76), so the resulting ChatModel will use configured region, but WARN log doe's not represent current state of the application and looks confusing.

**Environment**
Spring AI version: 1.1.2

**Expected behavior**
No WARN logs in case **spring.ai.bedrock.aws.region** property exists.

MatejNedic:
Made changes in Credentials auto configuration by providing new CredentialsProviders:
1. StsAssumeRoleCredentialsProvider
2. StsWebIdentityTokenFileCredentialsProvider
3. InstanceProfileCredentialsProvider

Support for multiple CredentialProviders in chain is now also possible.

Change autoConfiguration for BedrockProxyChatModel by AutoConfiguring SDK clients as well.

BedrockProxyChatModel builder is now changed to accept Clients only and failing if clients were not configured.

Closes #5108 
Closes #823


MatejNedic:
Test Failures are in -> [INFO] WebFlux transports ................................. FAILURE [  2.920 s] 

Which is not affected by this PR. Locally for me all tests are passing.

MatejNedic:
Hey @sobychacko ,

sorry for ping I know you are extremely busy. Could I get PR review since this helps with better region/credentials management. It also improves build of BedrockChat.

One thing to note there was a bug where wrong class was used in `@EnableConfigurationProperties`.


Also I am afraid there will be a lot of conflicts soon.

sobychacko:
@MatejNedic Some initial feedback. Looks like you removed a few api methods on the builder which might cause backward compatiblity issues. Any public API methods that we provide today should be deprecated first and suggest users to migrate to the new alternatives. For example, after the PR -  any code that builds `BedrockProxyChatModel` programmatically without passing in `BedrockRuntimeClient` and `BedrockRuntimeAsyncClient` will break, right? 

Right now the PR removes `credentialsProvider()`, `region()`, `timeout()`, `connectionTimeout()`, `asyncReadTimeout()`, `connectionAcquisitionTimeout()`, and `socketTimeout()` from the builder, which breaks programmatic use. I suggest keeping them on the builder, annotate them with `@Deprecated`. If `bedrockRuntimeClient` / `bedrockRuntimeAsyncClient` are not set, create them from the deprecated credentials/region/timeout settings (same as today). New code can use the client-based builder; existing code keeps working and gets a deprecation warning. Remove the deprecated path in a later major/minor release once callers have migrated. What do you think about that?

MatejNedic:
Hey @sobychacko, 

Tnx on review!

Agree 100%, thought mb if this got into 2.0.0 that api could be broken 😅 

Will amend with suggestions, also I will adjust similar to Spring Boot WebClient approach and will make SDK Builders bean as well.

auto-configurations/models/spring-ai-autoconfigure-model-bedrock-ai/src/main/java/org/springframework/ai/model/bedrock/autoconfigure/BedrockAwsConnectionProperties.java
```diff
@@ -54,37 +59,61 @@ public class BedrockAwsConnectionProperties {
...
+	 */
+	private boolean instanceProfile = false;
+
 	/**
 	 * Aws profile. (optional) When the {@link #accessKey} and {@link #secretKey} are not
 	 * declared. Otherwise, the AwsBasicCredentials are used.
 	 */
 	@NestedConfigurationProperty
-	private ProfileProperties profile;
+	@Nullable private ProfileProperties profile;
```
ericbottard: Those annotations are out of place.
See https://github.com/spring-projects/spring-ai/blob/657252ae3be5e7d97e45319a838b88de949a2f0d/design/01-null-safety.adoc

MatejNedic: I have pushed fix for all nullable fields, also resolved conflicts

sdeleuze:
Sorry for the delay, there is too much conflict now, but the good news if that the 2 related issues should be fixed by my commit 92b2785b4bfc28796be11311e0ad86a4c23ed295.

Please check `2.0.1-SNAPSHOT` when the build is finished, and create an issue/PR if something needs to be refined or added.

As a consequence, I am closing this PR unmerged.

PR #5486, opened 2026-02-22
Ensure CosmosDB chat memory auto-configuration is applied first

buvidk1234:
Apply explicit ordering so CosmosDBChatMemoryRepositoryAutoConfiguration runs before ChatMemoryAutoConfiguration.

This prevents premature default bean creation that would otherwise override or bypass the CosmosDB repository setup.


ericbottard:
Rebased and clarifed as d57da4bbb.

Thanks for your contribution @buvidk1234 !

Issue #5481, opened 2026-02-20
Neo4jVectorstore: Make filterExpressionConverter configurable via builder in Neo4jVectorStore


pascalwilbrink:
## Summary

The `filterExpressionConverter` field in `Neo4jVectorStore` is currently hardcoded as a final field initialized inline, making it impossible to customize filter expression conversion behavior without subclassing the store itself.

## Current Behavior
```java
private final Neo4jVectorFilterExpressionConverter filterExpressionConverter = new Neo4jVectorFilterExpressionConverter();
```

There is no way to override this converter through the builder, even though the builder pattern is used for all other configuration.

## Desired Behavior

The builder should expose a `filterExpressionConverter(...)` method accepting a `FilterExpressionConverter`, allowing users to plug in a custom implementation:
```java
Neo4jVectorStore vectorStore = Neo4jVectorStore.builder(driver, embeddingModel)
    .filterExpressionConverter(new MyCustomFilterExpressionConverter())
    .build();
```

## Proposed Changes

- Add a `filterExpressionConverter` field to `Neo4jVectorStore.Builder` defaulting to `new Neo4jVectorFilterExpressionConverter()`
- Add a `filterExpressionConverter(FilterExpressionConverter)` setter on the builder
- Change the field type in `Neo4jVectorStore` from `Neo4jVectorFilterExpressionConverter` to `FilterExpressionConverter` to keep the extension point open
- Initialize the field from the builder in the constructor

## Motivation

Users with custom metadata structures or Neo4j-specific query requirements may need to adapt filter expression conversion without forking the class. Using the `FilterExpressionConverter` interface rather than the concrete class also aligns with the open/closed principle and is consistent with how other extension points in Spring AI are designed.

PR #5490, opened 2026-02-23
feat(vector-stores): configurable `filterExpressionConverter` via builder in `Neo4jVectorStore`

yaner-here:
Close https://github.com/spring-projects/spring-ai/issues/5481

`filterExpressionConverter` is expected to be configurable, rather than just a fixed value (restricted by `final`).


sobychacko:
@yaner-here The PR has been merged upstream to `main` and back-ported to `1.1.x`. Thanks for the contribution!

PR #5493, opened 2026-02-23
Improve TokenTextSplitter EncodingType support

sobychacko:
Follow-up to cfbe546e04 (Allow TokenTextSplitter to accept EncodingType).

  - Make the full constructor private, accessible only via builder
  - Deprecate parameterized constructors in favor of builder
  - Add null validation for encodingType
  - Update docs to use builder pattern and document encodingType
  - Add test for null encodingType rejection


ilayaperumalg:
The PR check failure is due to another issue which gets fixed after rebasing. Rebased and merged via 045e8c149

Issue #5494, opened 2026-02-23
Anthropic: support multi-block system messages with independent cache_control per block


CalamityAdam:
**Expected Behavior**

Developers should be able to designate portions of the system message as static (cacheable) vs dynamic (uncacheable), with `buildSystemContent()` emitting separate Anthropic `ContentBlock`s accordingly:

```json
{
  "system": [
    {"type": "text", "text": "static instructions...", "cache_control": {"type": "ephemeral"}},
    {"type": "text", "text": "dynamic per-request context..."}
  ]
}
```

Anthropic's API already supports arrays of content blocks in the `system` field, each with independent `cache_control` (up to 4 breakpoints per request). The static block would get cache hits on subsequent requests while the dynamic block is processed fresh.

One possible API shape — a cache policy on `SystemMessage`:

```java
// Static behavioral instructions — cached
new SystemMessage("safety guardrails...", CachePolicy.CACHEABLE);

// Dynamic advisor-injected context — not cached
new SystemMessage("product recommendations...", CachePolicy.NO_CACHE);
```

Another option would be at the advisor level, letting advisors declare whether their `augmentSystemMessage()` output belongs in a cached or uncached block.

**Current Behavior**

`AnthropicChatModel.buildSystemContent()` concatenates all system messages into a single string via `Collectors.joining()`, then wraps the result in one `ContentBlock` with one `cache_control` marker.

This means if *any* part of the system message is dynamic, the entire system cache misses every request. With `SYSTEM_ONLY` strategy enabled, each miss pays Anthropic's 1.25x cache-write cost with zero reads — making it **more expensive** than disabling caching entirely.

There is no way to split static from dynamic content within the system message using the current API. The only workaround is moving dynamic content out of the system message entirely.

**Context**

Our application uses Anthropic Sonnet for conversational AI agents. The system message has two layers:

- **Static behavioral instructions** (~2,280 tokens): safety guardrails, domain principles, brand tone, mode instructions, response formatting. Identical across all requests for a given tenant + mode.
- **Dynamic context injected by Spring AI advisors**: RAG-matched product recommendations (change every query), intent/slot state (change every turn), temporal context (changes daily).

Because `buildSystemContent()` concatenates everything into one block, we can't cache the static prefix independently. Enabling `SYSTEM_ONLY` with any dynamic advisor content means 100% cache misses at 1.25x write cost.

[...]

PR #5495, opened 2026-02-23
GH-5494: Fix multi-block system message caching for Anthropic prompt …

sobychacko:
…caching

Fixes https://github.com/spring-projects/spring-ai/issues/5494

When prompt caching is enabled, `buildSystemContent()` joins all `SystemMessage` instances into a single `ContentBlock` with one `cache_control` marker. If any part is dynamic (e.g., advisor-injected RAG context), the entire block changes every request, causing 100% cache misses at 1.25x write cost.

Add `multiBlockSystemCaching` flag to `AnthropicCacheOptions`. When enabled, `buildSystemContent()` emits one `ContentBlock` per `SystemMessage` with `cache_control` on the second-to-last block, allowing the static prefix to be cached while the last block changes freely. Defaults to false for backward compatibility. Single `SystemMessage` behavior is unchanged.

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/anthropic-chat.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/anthropic-chat.adoc
@@ -470,6 +470,8 @@ This produces the following wire format:
 NOTE: If only a single `SystemMessage` is present, it still gets `cache_control` applied (same as the default behavior).
 The `multiBlockSystemCaching` option only changes behavior when there are two or more system messages.
 
+IMPORTANT: Message ordering matters. The `cache_control` marker is always placed on the second-to-last system block, so static content must come before dynamic content in the message list. If you register your base instructions as the first `SystemMessage` and let advisors append dynamic context after, this ordering is naturally satisfied.
+
 === Usage Example
 
 Here's a complete example demonstrating prompt caching with cost tracking:


CalamityAdam:
Friendly bump on this one 👋 — this issue is currently blocking our team from using Anthropic models in production.

We have a multi-block system prompt setup where static instructions are paired with dynamic RAG-injected context. Without `multiBlockSystemCaching`, the entire system prompt cache misses on every request because everything gets joined into a single block. This makes Anthropic models cost-prohibitive compared to alternatives that don't have this limitation in the integration layer.

Really appreciate the work here from @sobychacko — the approach of keeping backward compat via an opt-in flag is solid. Is there anything we can do to help move this forward (testing, review feedback, etc.)?

sobychacko:
@CalamityAdam Apologies for the delay in merging the PR - things were a bit hectic on our end. Are you using `1.1.x`? The reason I am asking is because we already made some significant changes with `spring-ai-anthropic` module in `2.0.x` line in that the old `RestClient` based support is completely revamped with the new official Java SDK from Anthropic. There we already support the multi-block system message caching. So, we only need to merge this PR into the `1.1.x` branch. That's why I am curious to know which version of Spring AI you are using currently?

sobychacko:
Merged via https://github.com/spring-projects/spring-ai/commit/11bd45ebd41eb7f427b5e227b538870524db9fbb.

CalamityAdam:
@sobychacko Thanks so much for merging this, and no worries on the timing — we really appreciate you backporting it to 1.1.x!

To answer your question: yes, we're on **Spring AI 1.1.0** in production. We'll be upgrading to 1.1.4 as soon as it's released to pick up this fix. The multi-block system prompt caching is a big deal for us — we're using Anthropic's extended thinking with Sonnet and the single-block limitation was preventing us from leveraging prompt caching on system messages.

We do plan to move to 2.0.x eventually, but 1.1.4 unblocks us in the short term without a major version migration. Thanks again! 🙏

Issue #5488, opened 2026-02-23
Can't attach XLSX file to Amazon Bedrock Converse prompt


k0staa:
### **Bug description**

When I want to add Media (`org.springframework.ai.content.Media#Media(org.springframework.util.MimeType, org.springframework.core.io.Resource)`) to the UserMessage with XLSX file and use this message in Amazon Bedrock Converse API call (Claude Opus), I've get this runtime error:
```
software.amazon.awssdk.services.bedrockruntime.model.ValidationException: The document file name can only contain alphanumeric characters, whitespace characters, hyphens, parentheses, and square brackets. The name can't contain more than one consecutive whitespace character. Check that the file name conforms to these restrictions and retry your request. 
```

The problem is that mime type of this file is `vnd.openxmlformats-officedocument.spreadsheetml.sheet` and when I'm using `Media` class constructor that take `MimeType` and `Resource` (`UrlResource` in my case) then the name of this object is created as follows:

```
this.name = generateDefaultName(mimeType);
```
generateDefaultName (https://github.com/spring-projects/spring-ai/blob/628b020041f94fb223ffd873e94c9304c42bc10f/spring-ai-commons/src/main/java/org/springframework/ai/content/Media.java#L161) :

```
private static String generateDefaultName(MimeType mimeType) {
		return NAME_PREFIX + mimeType.getSubtype() + "-" + java.util.UUID.randomUUID();
	}
```
This name is passed to Amazon Bedrock and dots are causing runtime exception. 

### **Environment**

1.0.2 but the newest version is also affected

### **Steps to reproduce**

Described in description. 

### **Expected behavior**

The name should be stripped from dots when passed to Amazon Bedrock converse call. 

**Minimal Complete Reproducible example**

1. Use Amazon Bedrock ChatClient:
```
 val options = BedrockChatOptions.builder()
            .model("eu.anthropic.claude-opus-4-6-v1")
            .build()

        val chatModel = bedrockModelBuilder()
            .asyncReadTimeout(ofSeconds(120))
            .socketTimeout(ofSeconds(120))
            .connectionTimeout(ofSeconds(120))
            .connectionAcquisitionTimeout(ofSeconds(20))
            .defaultOptions(options)
            .build()
        return ChatClient.builder(chatModel).defaultAdvisors(memory).build()
```
2.  Create `UserMessage` with at least one `Media` of XLSX and `Prompt` object:
```
[...]

PR #5498, opened 2026-02-24
GH-5488: Sanitize document name for Bedrock Converse API compliance

sobychacko:
Fixes https://github.com/spring-projects/spring-ai/issues/5488

The auto-generated Media name includes the MIME subtype directly (e.g., vnd.openxmlformats-officedocument.spreadsheetml.sheet), which contains dots that Bedrock rejects. The fix sanitizes the name in the Bedrock layer before passing it to the AWS SDK.


ilayaperumalg:
Rebased and merged into main via 6a28b39af7112cd53d94cd60a68a1ed2d2ff865f and backported into 1.1.x via 924478bf4

PR #5500, opened 2026-02-24
Add private Builder-accepting constructor to `OpenAiSdkChatModel`

sobychacko:
All public constructors are deprecated in favor of the builder pattern, but the `Builder.build(`) method was forced to use `@SuppressWarnings(deprecation)` to call the deprecated all-args constructor. Add a private constructor that takes a `Builder`, allowing `build()` to bypass the deprecated constructors cleanly.

### DIFF
--- a/models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkChatModel.java
+++ b/models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkChatModel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025-2025 the original author or authors.
+ * Copyright 2025-2026 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


ericbottard:
Please make the builder have its own fields rather than an embedded instance of what it constructs.

Ideally, builders should pass a test inspired by https://github.com/spring-projects/spring-ai/pull/5506/changes#diff-3fe453d26b01d56d240916b744c4938a9d362995651e350503a5740cd3f2f9b6R1

sobychacko:
@ericbottard You mean for the `OpenAiSdkChatOptions.Builder`, right? This PR is only for the `OpenAiSdkChatModel`, which already holds its own fields and creates a new instance on each build() call. I can tackle the `OpenAiSdkChatOptions.Builder` in another PR following the same pattern from #5506. Thanks!

EDIT - Actually I wonder if we should wait until #5506 is merged to make these types of changes in the `OpenAiSdkChatOptions`? In any case, I think this PR can be merged independently though.

ericbottard:
Yes. I was thrown off by quickly reading that constructor accepting the builder. As discussed elsewhere, this can lead to unintended side effects quickly when people get creative ;-).
We'll do a complete pass on building before 2.0 anyway

models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkChatModel.java
```diff
@@ -270,6 +270,35 @@ public OpenAiSdkChatModel(@Nullable OpenAIClient openAiClient, @Nullable OpenAIC
 				new DefaultToolExecutionEligibilityPredicate());
 	}
 
+	private OpenAiSdkChatModel(Builder builder) {
+		if (builder.options == null) {
```
ericbottard, line 274: can use Objects.requireNonNullElse() for consistency

ilayaperumalg (approved):

ilayaperumalg:
Rebased and merged via fbcb8a88d after the fix to include Objects.requireNonNullElse for assertion.

Issue #5441, opened 2026-02-11
Refactor Model|ChatOptions handling


ericbottard:
The current logic for handling `ChatOptions` merging between default and per-prompt options is very cumbersome and error-prone. It relies on the options classes being Jackson annotated, even though options have no point being serialized on the wire (directly).
Moreover, this behavior conflicts with easily stating which fields are required and which fields are not, because the current logic deals with half-populated objects (the per-prompt overrides).

This issue is about refactoring this from the ground up, with the following plan:
* Option classes can clearly state which fields are required or not. Valid construction is enforced by a Builder.
* Builders are used as a vehicle for stating options overrides, as they allow partial definition of state
* `ModelOptionsUtils.merge()` and similar methods become irrelevant

nicolaskrier:
`ChatOptions` are also used as nested configuration property (see [`OllamaChatProperties`](https://github.com/spring-projects/spring-ai/blob/main/auto-configurations/models/spring-ai-autoconfigure-model-ollama/src/main/java/org/springframework/ai/model/ollama/autoconfigure/OllamaChatProperties.java) for example) and all the conversions between string and the different properties types should be handled to be configurable in properties files (see #4854).

It is a bit out of context here but it is yet another conversion that has to be done correctly.

PR #5506, opened 2026-02-25
Incremental ChatOptions rework

ericbottard:
This is a first stepping stone for the options rework (#5441), NOT changing anything yet to the way options are being merged but introducing the Builder changes we know we'll want to be using.

Please merge as two separate commits @ilayaperumalg

### DIFF
--- a/models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaChatOptionsTests.java
+++ b/models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaChatOptionsTests.java
@@ -24,6 +24,7 @@ import java.util.Set;
 import org.junit.jupiter.api.Test;
 import tools.jackson.core.JacksonException;
 
+import org.springframework.ai.test.options.AbstractChatOptionsTests;
 import org.springframework.ai.util.ResourceUtils;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -34,7 +35,18 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
  * @author Mark Pollack
  * @author Nicolas Krier
  */
-class OllamaChatOptionsTests {
+class OllamaChatOptionsTests<B extends OllamaChatOptions.Builder<B>>
+		extends AbstractChatOptionsTests<OllamaChatOptions, B> {
+
+	@Override
+	protected Class<OllamaChatOptions> getConcreteOptionsClass() {
+		return OllamaChatOptions.class;
+	}
+
+	@Override
+	protected B readyToBuildBuilder() {
+		return (B) OllamaChatOptions.builder();
+	}
 
 	@Test
 	void testBasicOptions() {
@@ -82,16 +94,16 @@ class OllamaChatOptionsTests {
 		assertThat(optionsMap).containsEntry("num_predict", 100);
 		assertThat(optionsMap).containsEntry("top_k", 40);
 		assertThat(optionsMap).containsEntry("top_p", 0.9);
-		assertThat(optionsMap).containsEntry("tfs_z", 1.0);
-		assertThat(optionsMap).containsEntry("typical_p", 1.0);
+		assertThat(optionsMap).containsEntry("tfs_z", 1.0f);
+		assertThat(optionsMap).containsEntry("typical_p", 1.0f);
 		assertThat(optionsMap).containsEntry("repeat_last_n", 64);
 		assertThat(optionsMap).containsEntry("temperature", 0.7);
 		assertThat(optionsMap).containsEntry("repeat_penalty", 1.1);
 		assertThat(optionsMap).containsEntry("presence_penalty", 0.0);
 		assertThat(optionsMap).containsEntry("frequency_penalty", 0.0);
 		assertThat(optionsMap).containsEntry("mirostat", 2);
-		assertThat(optionsMap).containsEntry("mirostat_tau", 5.0);
-		assertThat(optionsMap).containsEntry("mirostat_eta", 0.1);
+		assertThat(optionsMap).containsEntry("mirostat_tau", 5.0f);
+		assertThat(optionsMap).containsEntry("mirostat_eta", 0.1f);
 	}
 
 	@Test
--- a/models/spring-ai-ollama/pom.xml
+++ b/models/spring-ai-ollama/pom.xml
@@ -80,6 +80,13 @@
             <scope>test</scope>
         </dependency>
 
+		<dependency>
+			<groupId>org.springframework.ai</groupId>
+			<artifactId>spring-ai-test</artifactId>
+            <version>${project.parent.version}</version>
+			<scope>test</scope>
+		</dependency>
+
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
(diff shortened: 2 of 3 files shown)


spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
```diff
@@ -184,133 +185,152 @@ public void setTopP(@Nullable Double topP) {
...
+			return options;
 		}
 
 		@Override
-		public ToolCallingChatOptions build() {
-			return this.options;
+		public B combineWith(ChatOptions.Builder<?> other) {
+			super.combineWith(other);
+			if (other instanceof Builder<?> options) {
+				if (options.toolCallbacks != null) {
```
sobychacko: `toolCallbacks` initializes to an `ArrayList` always. Why do we need this null check? This means that even when the other type's value is an empty collection, we overwrite the value for this, no?

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
```diff
@@ -184,133 +185,152 @@ public void setTopP(@Nullable Double topP) {
...
+					this.toolCallbacks = options.toolCallbacks;
+				}
+				if (options.toolNames != null) {
```
sobychacko: Same question as the `toolCallbacks` above.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
```diff
@@ -184,133 +185,152 @@ public void setTopP(@Nullable Double topP) {
...
+					this.toolNames = options.toolNames;
+				}
+				if (options.toolContext != null) {
```
sobychacko: `toolContext` always initialized to a `HashMap`.

spring-ai-model/src/main/java/org/springframework/ai/chat/prompt/ChatOptions.java
```diff
@@ -82,81 +82,102 @@ public interface ChatOptions extends ModelOptions {
...
-		Builder topP(Double topP);
+		B topP(@Nullable Double topP);
 
 		/**
 		 * Build the {@link ChatOptions}.
 		 * @return the Chat options.
 		 */
 		ChatOptions build();
 
+		B combineWith(ChatOptions.Builder<?> other);
```
sobychacko, line 184: We should add some javadoc explaining this method, since this is part of public API.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
```diff
@@ -184,133 +185,152 @@ public void setTopP(@Nullable Double topP) {
...
-			return this;
-		}
-
-		@Override
-		public ToolCallingChatOptions.Builder topP(@Nullable Double topP) {
-			this.options.setTopP(topP);
-			return this;
+		public ToolCallingChatOptions build() {
+			DefaultToolCallingChatOptions options = new DefaultToolCallingChatOptions();
+			if (this.toolCallbacks != null) {
```
sobychacko: Its already initialized to an `ArrayList` though.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
```diff
@@ -184,133 +185,152 @@ public void setTopP(@Nullable Double topP) {
...
+				options.setToolCallbacks(this.toolCallbacks);
+			}
+			if (this.toolNames != null) {
```
sobychacko: Already initialized, so why the check?

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
```diff
@@ -184,133 +185,152 @@ public void setTopP(@Nullable Double topP) {
...
+				options.setToolNames(this.toolNames);
+			}
+			if (this.toolContext != null) {
```
sobychacko: Already initialized.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
```diff
@@ -184,133 +185,152 @@ public void setTopP(@Nullable Double topP) {
...
-		public ToolCallingChatOptions.Builder topK(@Nullable Integer topK) {
-			this.options.setTopK(topK);
-			return this;
-		}
-
-		@Override
-		public ToolCallingChatOptions.Builder topP(@Nullable Double topP) {
-			this.options.setTopP(topP);
-			return this;
+		public ToolCallingChatOptions build() {
```
sobychacko, line 288: This `build()` method is not consistent with the `OllamaChatModel`'s. Is that intentional?

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaChatOptionsTests.java
```diff
@@ -34,7 +35,18 @@
...
+		extends AbstractChatOptionsTests<OllamaChatOptions, B> {
+
+	@Override
+	protected Class<OllamaChatOptions> getConcreteOptionsClass() {
+		return OllamaChatOptions.class;
+	}
+
+	@Override
+	protected B readyToBuildBuilder() {
+		return (B) OllamaChatOptions.builder();
```
sobychacko, line 48: unchecked cast.

sobychacko (changes requested):

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
ericbottard: Good point. I went back and forth on several of those semantics which lead to some inconsistent leftovers, I'll simplify.

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingChatOptions.java
ericbottard: Can you clarify what you mean?

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaChatOptionsTests.java
ericbottard: This is needed b/c of the self referencing generics trick.

ericbottard:
Force pushed changes to the first commit. Please merge as two separate commits

ericbottard:
Hi @sobychacko. I believe comments have been addressed and this PR is able to be merged, so that other work can build on it (incl. maybe ctor management if we find a good pattern).

ilayaperumalg:
Rebased and merged into main via 3db5c2bd4b1047df98e6c38a4d3f2a2930b903e8 and 5f09a8f9560b2134bb6e4156ffaff8cc55a03466

PR #5528, opened 2026-03-03
fix: track token count of first document in new batch in TokenCountBatchingStrategy

### LINKED ISSUE #5525: The currentSize accumulation in TokenCountBatchingStrategy is inaccurate
When I was examining the source code of TokenCountBatchingStrategy, I identified an issue with the batching and saving process in the batch() method. 

When creating a new instance of currentBatch, the currentSize is reset to 0, but the Document documents traversed during the current iteration will be added to the new instance of currentBatch 

This results in a situation where, during the loop, each time the currentSize is reset, one token length of the current iteration through the Document is not accumulated

coding node：
	for (Map.Entry<Document, Integer> entry : documentTokens.entrySet()) {
		Document document = entry.getKey();
		currentSize += entry.getValue();
		if (currentSize > this.maxInputTokenCount) {
			batches.add(currentBatch);
			currentBatch = new ArrayList<>();
			currentSize = 0;
		}
		currentBatch.add(document);
	}

I apologize for my poor English. This is the bug description I generated through machine translation

Image

weiguangli-io:
Closes #5525

## Summary

When a document's token count causes `currentSize` to exceed `maxInputTokenCount` in `TokenCountBatchingStrategy.batch()`, the current batch is saved and a new one is created. However, `currentSize` is reset to `0` while the triggering document is still added to the new batch — its token count is silently dropped from the running total.

This means each new batch can accept one extra document's worth of tokens beyond the configured limit before the next split is triggered.

### Fix

Change `currentSize = 0` to `currentSize = entry.getValue()` so the first document in each new batch is properly counted.

### Test

Added `batchShouldTrackTokenCountAcrossBatchBoundaries` test that creates multiple documents with a small token limit (10 tokens), verifying that batches are correctly split and no documents are lost.

> AI-assisted testing, AI-assisted review

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/embedding/TokenCountBatchingStrategy.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/embedding/TokenCountBatchingStrategy.java
@@ -160,7 +160,7 @@ public class TokenCountBatchingStrategy implements BatchingStrategy {
 			if (currentSize > this.maxInputTokenCount) {
 				batches.add(currentBatch);
 				currentBatch = new ArrayList<>();
-				currentSize = 0;
+				currentSize = entry.getValue();
 			}
 			currentBatch.add(document);
 		}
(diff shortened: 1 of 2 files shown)


sobychacko:
PR merged upstream to `main` and back-ported to `1.1.x`. Thanks for the contribution.

PR #5543, opened 2026-03-05
Implement notification logic for MCP server transports

chemicL:
Implemented the logic to support resource subscription from https://github.com/modelcontextprotocol/java-sdk/pull/839


chemicL:
As noted by @tzolov this can be integrated since java-sdk `1.1.0-SNAPSHOT` is ok for the current main branch.

Issue #5546, opened 2026-03-06
Add `@Nullable` to `value` parameter in `ChatClientResponse.Builder.context(String key, Object value)` for consistency


CokeLee777:
**Summary**

`ChatClientResponse.Builder` has a method `context(String key, Object value)` where the `value` parameter is not annotated with `@Nullable`. The same overload in `ChatClientRequest.Builder` uses `context(String key, @Nullable Object value)`, and the context map type is `Map<String, @Nullable Object>` in both cases, so null values are allowed. Adding `@Nullable` to the response builder would align the two APIs and better reflect the actual contract.

**Current behavior**

- `ChatClientRequest.Builder.context(String key, @Nullable Object value)` — value is `@Nullable`
- `ChatClientResponse.Builder.context(String key, Object value)` — value has no nullability annotation

**Proposed change**

In `spring-ai-client-chat` → `ChatClientResponse.java`, change:

```java
public Builder context(String key, Object value) {
```

to:

```java
public Builder context(String key, @Nullable Object value) {
```

**Rationale**

- Consistency with `ChatClientRequest.Builder.context(String, @Nullable Object)`.
- The stored context is `Map<String, @Nullable Object>`, so the parameter semantics should allow null.
- Improves static analysis / IDE null-safety and documents the API contract.

**Scope**

- Single line change (add `@Nullable` to the parameter). No behavior change; `Assert.notNull` is only applied to `key`.

CokeLee777:
Opened a PR for this: #5547

sdeleuze:
Closing in favor of the PR #5547.

PR #5547, opened 2026-03-06
Add `@Nullable` to `ChatClientResponse.Builder#context` value parameter

CokeLee777:
## Summary

Adds `@Nullable` to the `value` parameter in `ChatClientResponse.Builder.context(String key, Object value)` for consistency with `ChatClientRequest.Builder.context(String key, @Nullable Object value)` and with the context map type `Map<String, @Nullable Object>`.

## Changes

* **ChatClientResponse.java**: `context(String key, @Nullable Object value)` — add `@Nullable` to `value`
* **ChatClientResponseTests.java**: Add `whenBuilderContextWithNullValueThenCreateSuccessfully()` to verify that the builder accepts null as context value

Fixes GH-5546


sdeleuze:
Merged, thanks!

PR #5551, opened 2026-03-06
Use new builder pattern for `MistralAiChatOptions`

sdeleuze:


models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
```diff
@@ -163,29 +165,57 @@ public class MistralAiChatOptions implements ToolCallingChatOptions, StructuredO
...
+			@Nullable Integer maxTokens, @Nullable Boolean safePrompt, @Nullable Integer randomSeed,
+			@Nullable ResponseFormat responseFormat, @Nullable List<String> stop, @Nullable Double frequencyPenalty,
+			@Nullable Double presencePenalty, @Nullable Integer n, @Nullable List<FunctionTool> tools,
+			@Nullable ToolChoice toolChoice, @Nullable List<ToolCallback> toolCallbacks,
+			@Nullable Set<String> toolNames, @Nullable Boolean internalToolExecutionEnabled,
+			@Nullable Map<String, Object> toolContext) {
+
+		this.model = model;
+		this.temperature = temperature;
+		if (topP != null) {
```
ericbottard, line 182: All of those null checks in the ctor are not needed, are they?

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
```diff
@@ -386,9 +416,32 @@ public void setOutputSchema(String outputSchema) {
...
+	public MistralAiChatOptions.Builder<?> mutate() {
+		return builder()
+			// ChatOptions
+			.model(this.model)
+			.frequencyPenalty(this.frequencyPenalty)
+			.maxTokens(this.maxTokens)
+			.presencePenalty(this.presencePenalty)
+			.stop(this.stop == null ? null : new ArrayList<>(this.stop))
+			.temperature(this.temperature)
+			.topP(this.topP)
```
ericbottard, line 432: Missing topK.
If this is not supported, I still add the line with a comment to clarify. See https://github.com/spring-projects/spring-ai/pull/5552/changes#diff-36c3bf74085463ed9c9dbc5b38f199d63939dde406c7c4deaf30e6d9c6c57b1bR371

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
```diff
@@ -163,29 +165,57 @@ public class MistralAiChatOptions implements ToolCallingChatOptions, StructuredO
 	@JsonIgnore
 	private Map<String, Object> toolContext = new HashMap<>();
 
-	public static Builder builder() {
-		return new Builder();
+	// Temporary constructor to maintain compat with ModelOptionUtils
+	public MistralAiChatOptions() {
```
ericbottard, line 169: We should see if we can make this protected/package-pted and still have Jackson work in the interim. This would force more use of the builder

ericbottard (approved):
LGTM with small polish. Feel free to merge yourself when done

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
sdeleuze: I think they are to not override the default non null value on related fields.

ericbottard: Ok. We shall harmonize the handling of callections later anyway.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
sdeleuze: I was trying to stay consistent with what has been done in `AzureOpenAiChatOptions` and limit the breaking changes for users, especially pre M3.

That said, if we want to force the builder, and keep Jackson working as expected, we should do it by:
 - Removing this public constructor
 - Remove setters
 - Annotating the builder with `@JsonPOJOBuilder`

I will commit as it is and we can tackle this potential refinement after discussing/deciding with the team.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
sdeleuze: I was not sure about that one, but yeah your proposal makes sense to me, I will refine.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
```diff
@@ -163,29 +165,57 @@ public class MistralAiChatOptions implements ToolCallingChatOptions, StructuredO
...
+	}
+
+	protected MistralAiChatOptions(String model, @Nullable Double temperature, @Nullable Double topP,
```
tzolov, line 172: I presume that those @Nullable will allow creating a `delta` options at runtime with the builder?

PR #5554, opened 2026-03-06
Use new builder pattern for GoogleGenAiChatOptions.

ericbottard:


models/spring-ai-google-genai/src/main/java/org/springframework/ai/google/genai/GoogleGenAiChatOptions.java
```diff
@@ -549,7 +564,42 @@ public String toString() {
...
+			.topK(this.topK)
+			.topP(this.topP)
+			// ToolCallingChatOptions
+			.toolCallbacks(this.getToolCallbacks())
+			.toolNames(this.getToolNames())
+			.toolContext(this.getToolContext())
+			.internalToolExecutionEnabled(this.getInternalToolExecutionEnabled())
+			// StructuredOutputChatOptions
+			.outputSchema(this.getOutputSchema())
+			.responseMimeType(this.responseMimeType)
```
sdeleuze, line 588: You should probably put `.responseMimeType(this.responseMimeType)` first due to `.outputSchema(this.getOutputSchema())` conditional processing based on `this.responseMimeType` value.

ericbottard: That's an interesting point. I did push the change, but to me, users should be able to use builder methods in any order. So there is something slightly broken in the existing behavior

sdeleuze: Thanks for updating the PR, from my perspective, not sure something is broken, the logic seems correct to me as `outputSchema` side-effect is by design but let's discuss that later outside of this PR. I will review again and likely merge the PR.

PR #5566, opened 2026-03-09
Add User-Agent header for Anthropic API calls

mikelambert:
Passes User-Agent: spring-ai to the Anthropic API, matching what's already done for the OpenAI connector.


sdeleuze:
@sobychacko Should we decline this or at least apply it only in 1.1.x, as we are going to replace the current implementation by the SDK one via #5366?

sobychacko:
@sdeleuze Ya, I think we may want to apply that only to the `1.1.x` branch. In the meantime, I can see how we can add that in the SDK based module.

mikelambert:
Ah thanks! And let me know if there's another branch you'd like me to re apply this too as well!

sdeleuze:
> @sdeleuze Ya, I think we may want to apply that only to the `1.1.x` branch. In the meantime, I can see how we can add that in the SDK based module.

@sobychacko Not sure we will be able to be in control of that with the Anthropic SDK, but let see.

sdeleuze:
Spring AI 1.x is out of OSS support so closing that issue PR unmerged.

PR #5567, opened 2026-03-09
feat(mcp): add mcp-annotations module

### LINKED ISSUE #5457: Migrate mcp-annotations into Spring AI project
Migrate the existing [mcp-annotations](https://github.com/spring-ai-community/mcp-annotations) project into Spring AI project. 

While the `package names` and artifacts `groupID` will be updated, the core annotation modules are designed to remain framework-agnostic — they can be used with the default java-sdk without any Spring dependencies.

The existing annotations provide comprehensive coverage of MCP Client and Server capabilities, including:

- Server-side: `@McpTool`, `@McpResource`, `@McpPrompt`, `@McpComplete`
- Client-side: `@McpLogging`, `@McpSampling`, `@McpElicitation`, `@McpProgress`, `@McpToolListChanged`, `@McpResourceListChanged`, `@McpPromptListChanged`

tzolov:
Migrate the framework-agnostic mcp-annotations core from spring-ai-community/mcp-annotations into spring-ai as part of the mcp/mcp-annotations-spring module.

- Includes annotations (@McpTool, @McpPrompt, @McpResource, etc.), method callbacks, providers (tool, prompt, resource, sampling, progress, completion, elicitation, logging), request context, and adapters — covering both sync/async and stateless variants.
- Align the ClassUtills and JsonParser with existing spring-ai utils.
- Registers the module in root pom.xml and spring-ai-bom.
- Update MCP annotations docs
- Add migration instructions and open-rewrite receipe

Resolves #5457

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/utils/McpJsonParser.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/utils/McpJsonParser.java
@@ -26,16 +26,17 @@ import org.springframework.util.Assert;
 
 /**
  * Additional utilities for JSON parsing operations specific to MCP annotations and tools.
- * Reuses the underlying JsonMapper from {@link JsonParser} but provides convenience methods
- * for converting between Maps and Java objects, which is a common pattern in MCP tool interactions.
+ * Reuses the underlying JsonMapper from {@link JsonParser} but provides convenience
+ * methods for converting between Maps and Java objects, which is a common pattern in MCP
+ * tool interactions.
  */
 public final class McpJsonParser {
 
-	private McpJsonParser() {
-	}
-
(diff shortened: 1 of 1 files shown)


spring-ai-model/src/main/java/org/springframework/ai/util/json/JsonParser.java
```diff
@@ -189,4 +194,19 @@ else if (javaType.isEnum()) {
 		return result;
 	}
 
+	public static Map<String, Object> toMap(Object object) {
```
tzolov: Those additional methods are used by the mcp-annotations. 
I guess we can use the more generic `toTypedObject` but it will enforce some additional unnecessary checks, conversions and castings that might harm the performance.

sdeleuze: Would be nice if we could avoid introducing those methods to `JsonParser` and keep something more internal before we replace it by something better.

PR #5582, opened 2026-03-10
Use generic builder pattern for ZhiPuAiChatOptions.

ericbottard:


models/spring-ai-zhipuai/src/main/java/org/springframework/ai/zhipuai/ZhiPuAiChatOptions.java
```diff
@@ -396,181 +405,125 @@ public String toString() {
...
-			return this;
-		}
-
-		public Builder toolContext(Map<String, Object> toolContext) {
-			if (this.options.toolContext == null) {
-				this.options.toolContext = toolContext;
-			}
-			else {
-				this.options.toolContext.putAll(toolContext);
+		public B combineWith(ChatOptions.Builder<?> other) {
```
sdeleuze, line 494: Could we add `@Override` here?

ericbottard: done. (and rebased)

PR #5589, opened 2026-03-11
Move Huggingface model and autoconfiguration out of the main repo

ericbottard:
This PR moves the hugging-face model out of the main repository, pending a shift to the community org, based on feedback.


sdeleuze (changes requested):
There are still a few "huggingface" occurences like in `org.springframework.ai.model.SpringAIModels#HUGGINGFACE` or some `*.xml` files, could you please make a pass to remove the ones not required?

ericbottard:
Hi @sdeleuze, I force pushed a new version that should be good now.
Every remaining occurence is either about HF as a general repository of models and AI resources (in docs), or refers to the embeding model (which is unrelated to the removal of HF as a model)

PR #5599, opened 2026-03-13
Use the new builder pattern for `OpenAiSdkChatOptions`

sdeleuze:


models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkChatOptions.java
```diff
@@ -716,421 +843,429 @@ public StreamOptions build() {
...
+					this.maxRetries = that.maxRetries;
+				}
+			}
+			return self();
 		}
 
+		@Override
 		public OpenAiSdkChatOptions build() {
-			return this.options;
+			OpenAiSdkChatOptions options = new OpenAiSdkChatOptions();
```
ericbottard, line 1221: This is not using the all args ctor, is this intended?

ericbottard: Will merge nevertheless for now, we'll revisit anyway

ericbottard (approved):

models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkChatOptions.java
sdeleuze: Good catch, refined via 45e1882b87ae48c4d7c35ae15aa8bc039717541a.

PR #5601, opened 2026-03-15
refactor: improve time complexity and reduce memory allocations

chabinhwang:
## Summary

- **DefaultToolCallingManager**: Pre-compute `HashSet` of existing tool names to reduce `resolveToolDefinitions` lookup from O(N×M) to O(N+M)
- **ModelOptionsUtils**: Change `interfaceNames` from `List` to `Set` for O(1) `contains()` instead of O(N), use `Character.toUpperCase()` to avoid unnecessary `String` allocation in `toGetName()`
- **Vector stores (16 files)**: Replace `documents.indexOf(document)` O(N) lookup inside loops with index-based access O(1), eliminating O(N²) overall complexity in `doAdd` methods

### DefaultToolCallingManager (`resolveToolDefinitions`)
Previously, for each tool name in `chatOptions.getToolNames()`, the code streamed through all `toolCallbacks` with `.anyMatch()` — resulting in O(N×M) complexity. Now a `HashSet<String>` of existing tool callback names is pre-computed once, reducing the check to O(1) per iteration.

### ModelOptionsUtils (`mergeBeans` / `toGetName`)
- `interfaceNames` was collected as a `List<String>` but used with `.contains()` which is O(N) per call. Changed to `Set<String>` for O(1) lookups.
- `toGetName()` used `name.substring(0, 1).toUpperCase()` which creates an intermediate `String` object. Changed to `Character.toUpperCase(name.charAt(0))` to avoid unnecessary allocation.

### Vector Stores (16 files)
All vector store `doAdd` methods used `documents.indexOf(document)` inside a for-each loop to correlate documents with their embeddings. Since `indexOf` is O(N), this made the overall loop O(N²). Replaced with indexed for-loops using `embeddings.get(i)` for O(1) access.

Affected stores: Cassandra, Chroma, Couchbase, GemFire, MariaDB, Milvus, MongoDB Atlas, Neo4j, OpenSearch, Oracle, Pinecone, Qdrant, Redis, Typesense, Azure, CosmosDB

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/ModelOptionsUtils.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/ModelOptionsUtils.java
@@ -18,12 +18,14 @@ package org.springframework.ai.model;
 
 import java.beans.PropertyDescriptor;
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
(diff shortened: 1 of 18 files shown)


ericbottard (approved):

ericbottard:
Hi @chabinhwang,

thanks for this very good contribution!

PR #5602, opened 2026-03-15
[refactor] use slf4j placeholders instead of concat

viacheslav-dobrynin:


ericbottard:
Hi, thanks for your contribution @viacheslav-dobrynin

PR #5605, opened 2026-03-16
MCP WebMvc transports: fix header usage

Kehrlann:
`ServerTransportSecurityIT` was broken, because Spring FW 7 `ServerRequest` 's HTTP headers don't support map operations anymore.
This introduces a function to collect headers in a map.

It reintroduces some `@Disabled` integration tests.

It is not very efficient, but I have an issue opened to improve this: https://github.com/modelcontextprotocol/java-sdk/issues/870


chemicL:
@Kehrlann FYI https://github.com/spring-projects/spring-framework/issues/36354 might help in framework 7.1 to simplify this, perhaps worth sharing this use case.

PR #5628, opened 2026-03-18
Fix AzureOpenAiChatOptions defaulting stop to empty list instead of null

### LINKED ISSUE #5627: AzureOpenAiChatOptions initializes stop to empty list instead of null, breaking models that don't support the parameter
**Bug description**
Since Spring AI 2.0.0-M3, Azure OpenAI requests include "stop": [] in the request body, even when no stop sequences are configured. Models that don't support the stop parameter (e.g. gpt-5-mini) reject the request with "Unsupported parameter: 'stop' is not supported with this model". I think it might be caused by this commit https://github.com/spring-projects/spring-ai/commit/e6ec123574c263094dd3bc866d67c1390f190f82

**Environment**
- Spring AI: 2.0.0-M3
- Spring Boot: 4.0.3
- Java: 25
- AI provider: Azure OpenAI (spring-ai-starter-model-azure-openai)
- Model: gpt-5-mini (Azure deployment)

**Steps to reproduce**
  1. Configure spring-ai-starter-model-azure-openai with a model that does not support the stop parameter (e.g. gpt-5-mini):
  ```yaml
spring:
    ai:
      azure:
        openai:
          endpoint: https://<your-endpoint>.cognitiveservices.azure.com/
          api-key: <your-key>
          chat:
            options:
              deployment-name: gpt-5-mini
```
  2. Send any chat completion request via ChatClient — no stop sequences configured anywhere.
  3. The request fails with:
  ```
{
    "error": {
      "message": "Unsupported parameter: 'stop' is not supported with this model.",
      "type": "invalid_request_error",
      "param": "stop",
      "code": "unsupported_parameter"
    }
  }
```

**Expected behavior**
When no stop sequences are configured, the stop parameter should be omitted from the request body.

I've created a small reproducer here: https://github.com/maxhov/spring-ai-stop-bug

maxhov:
The stop field was initialized to new ArrayList<>() causing "stop": [] to be serialized in every request, even when no stop sequences were configured. Models that don't support the stop parameter (e.g. gpt-5-mini) reject requests with this empty list present.

Changed the default to null so @JsonInclude(NON_NULL) correctly omits it from serialization.

Fixes #5627

### DIFF
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatOptions.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatOptions.java
@@ -169,7 +169,7 @@ public class AzureOpenAiChatOptions implements ToolCallingChatOptions {
 	 * A collection of textual sequences that will end completions generation.
 	 */
 	@JsonProperty("stop")
-	private List<String> stop = new ArrayList<>();
+	private List<String> stop;
 
 	/**
 	 * A value that influences the probability of generated tokens appearing based on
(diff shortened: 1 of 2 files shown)


sdeleuze:
`AzureOpenAiChatOptions` has been removed since Azure support comes via OpenAI SDK support now.

PR #5635, opened 2026-03-19
Fix CosmosDBVectorStore metadata isolation in search results

### LINKED ISSUE #5632: CosmosDBVectorStore: Shared metadata map causes all documents to have identical merged metadata
https://github.com/spring-projects/spring-ai/blob/0ad60decf02dab62cf4a09003ac07aefd9064e90/vector-stores/spring-ai-azure-cosmos-db-store/src/main/java/org/springframework/ai/vectorstore/cosmosdb/CosmosDBVectorStore.java#L409-L437

In the document retrieval path, metadata is collected into a single docFields map across all documents, then that same map reference is assigned to every Document object. This has two consequences:

Every returned Document contains metadata merged from all documents, not just its own.
If multiple documents share a metadata key with different values, only the last-written value survives — silently losing data.

Reproduction:
Store two documents with distinct metadata (e.g., {"author": "Alice"} and {"author": "Bob"}). Retrieve both — every returned document will have {"author": "Bob"}.

Suggested fix:
Move metadata extraction inside the per-document mapping:

```java
return documents.stream()
    .map(doc -> {
        Map<String, Object> metadata = new HashMap<>();
        JsonNode metadataNode = doc.get("metadata");
        if (metadataNode != null) {
            metadataNode.propertyNames().forEach(property -> {
                JsonNode value = metadataNode.get(property);
                Object parsedValue = value.isTextual() ? value.asText()
                        : value.isNumber() ? value.numberValue()
                        : value.isBoolean() ? value.booleanValue()
                        : value.toString();
                metadata.put(property, parsedValue);
            });
        }
        return Document.builder()
            .id(doc.get("id").asText())
            .text(doc.get("content").asText())
            .metadata(metadata)
            .build();
    })
    .collect(Collectors.toList());
```

This builds a fresh metadata map per document and also adds a null-check on the metadata node for robustness.

Yaklede:
Fixes #5632

## Summary

`CosmosDBVectorStore.doSimilaritySearch()` was accumulating metadata from all matched documents into a single shared map and then assigning that merged map to every returned `Document`.

This change keeps the fix minimal and localized to the document conversion path:
- build a fresh metadata map for each returned document
- keep the existing value conversion behavior for string, number, boolean, and fallback string conversion
- handle a missing `metadata` node defensively
- add regression coverage for the broken behavior

## Root Cause

The similarity-search result mapping logic created one metadata map outside the per-document conversion loop, populated it while iterating over all result documents, and then reused that accumulated map for every returned `Document`.

That caused:
- metadata from multiple results to be merged together
- same-key collisions to become last-write-wins
- returned documents to no longer reflect their own stored metadata

## Changes

- Refactored the similarity-search document conversion so each result document builds its own metadata map from its own `metadata` node.
- Added a null check for the `metadata` node and return an empty metadata map when it is absent.
- Kept the existing conversion behavior:
  - string -> `String`
  - number -> `Number`
  - boolean -> `Boolean`
  - everything else -> `toString()`

## Testing

Added unit tests in:

- `vector-stores/spring-ai-azure-cosmos-db-store/src/test/java/org/springframework/ai/vectorstore/cosmosdb/CosmosDBVectorStoreTests.java`

What the tests verify:

- `mapSimilaritySearchDocumentsShouldKeepMetadataScopedToEachDocument`
  - creates two similarity-search result documents with different metadata values
  - verifies that each returned `Document` contains only its own metadata
  - verifies that metadata does not leak across results
  - verifies that string, number, boolean, and fallback string conversion behavior is preserved

- `mapSimilaritySearchDocumentsShouldHandleMissingMetadataNode`
  - verifies that a result without a `metadata` node is handled safely
  - verifies that the returned `Document` receives an empty metadata map instead of failing

## Why this fix is safe

This change only affects the read-side conversion of similarity-search results into Spring AI `Document` instances.

It does not change:
- query generation
- filtering behavior
- persistence format
- add/delete behavior
- metadata value conversion semantics

So the behavioral change is limited to correcting the returned metadata for search results.

### DIFF
--- a/vector-stores/spring-ai-azure-cosmos-db-store/src/main/java/org/springframework/ai/vectorstore/cosmosdb/CosmosDBVectorStore.java
+++ b/vector-stores/spring-ai-azure-cosmos-db-store/src/main/java/org/springframework/ai/vectorstore/cosmosdb/CosmosDBVectorStore.java
@@ -78,6 +78,7 @@ import org.springframework.util.Assert;
  * @author Theo van Kraay
  * @author Soby Chacko
  * @author Thomas Vitale
+ * @author Yaklede
  * @since 1.0.0
  */
 public class CosmosDBVectorStore extends AbstractObservationVectorStore implements AutoCloseable {
@@ -330,7 +331,7 @@ public class CosmosDBVectorStore extends AbstractObservationVectorStore implemen
 								+ "' not found in metadata for document with id: " + id);
 					}
 
-					partitionKeyValue = metadataNode.get(metadataKey).asText();
+					partitionKeyValue = metadataNode.get(metadataKey).asString();
 				}
 				else {
 					throw new IllegalArgumentException("Unsupported partition key path: " + this.partitionKeyPath);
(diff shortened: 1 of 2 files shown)


sdeleuze:
CosmosDB support has been moved to community side, see https://azurecosmosdb.github.io/spring-ai/docs/index.html.

PR #5655, opened 2026-03-22
GH-5654: fix: clear entire conversation history in RedisChatMemoryRepository

### LINKED ISSUE #5654: RedisChatMemoryRepository clear() only deletes first 10 messages in a conversation
`RedisChatMemoryRepository.clear(conversationId)` appears to delete only the first 10 matching messages when a conversation has more than 10 entries.

Cause: FT.SEARCH is called without pagination, so Redis Search default limit (0 10) is applied.

> limits the results to the offset and number of results given. Note that the offset is zero-indexed. 
The default is 0 10, which returns 10 items starting from the first result. You can use LIMIT 0 0 to count the number of documents in the result set without actually returning them.

Ref: https://redis.io/docs/latest/commands/ft.search/

jbl428:
resolves #5654

### DIFF
--- a/memory/repository/spring-ai-model-chat-memory-repository-redis/src/main/java/org/springframework/ai/chat/memory/repository/redis/RedisChatMemoryRepository.java
+++ b/memory/repository/spring-ai-model-chat-memory-repository-redis/src/main/java/org/springframework/ai/chat/memory/repository/redis/RedisChatMemoryRepository.java
@@ -485,12 +485,19 @@ public final class RedisChatMemoryRepository implements ChatMemoryRepository, Ad
 		// Use QueryBuilders to create a tag field query
 		QueryNode queryNode = QueryBuilders.intersect("conversation_id",
 				Values.tags(RediSearchUtil.escape(conversationId)));
-		Query query = new Query(queryNode.toString());
-		SearchResult result = jedis.ftSearch(config.getIndexName(), query);
+		int batchSize = Math.max(10, config.getMaxMessagesPerConversation());
 
-		try (Pipeline pipeline = jedis.pipelined()) {
-			result.getDocuments().forEach(doc -> pipeline.del(doc.getId()));
-			pipeline.sync();
+		while (true) {
+			Query query = new Query(queryNode.toString()).limit(0, batchSize);
+			SearchResult result = jedis.ftSearch(config.getIndexName(), query);
+			if (result.getDocuments().isEmpty()) {
+				break;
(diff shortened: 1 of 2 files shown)


memory/repository/spring-ai-model-chat-memory-repository-redis/src/main/java/org/springframework/ai/chat/memory/repository/redis/RedisChatMemoryRepository.java
```diff
@@ -485,12 +485,19 @@ public void clear(String conversationId) {
...
-			pipeline.sync();
+		while (true) {
+			Query query = new Query(queryNode.toString()).limit(0, batchSize);
+			SearchResult result = jedis.ftSearch(config.getIndexName(), query);
+			if (result.getDocuments().isEmpty()) {
+				break;
+			}
+
+			try (Pipeline pipeline = jedis.pipelined()) {
+				result.getDocuments().forEach(doc -> pipeline.del(doc.getId()));
```
jbl428, line 498: Would it make sense to use UNLINK instead of DEL here for conversation cleanup?
It might reduce blocking during large deletes while keeping the same user-facing behavior.
If this sounds reasonable, I can either include it in this PR or open a separate issue for follow-up.

memory/repository/spring-ai-model-chat-memory-repository-redis/src/main/java/org/springframework/ai/chat/memory/repository/redis/RedisChatMemoryRepository.java
```diff
@@ -485,12 +485,19 @@ public void clear(String conversationId) {
 		// Use QueryBuilders to create a tag field query
 		QueryNode queryNode = QueryBuilders.intersect("conversation_id",
 				Values.tags(RediSearchUtil.escape(conversationId)));
-		Query query = new Query(queryNode.toString());
```
JaiVora, line 488: Why not just:
```java
Query query = new Query(queryNode.toString())
        .limit(0, config.getMaxMessagesPerConversation());
```

janik-unite:
I'm also affected by #5654. `RedisChatMemoryRepository.clear()` leaves messages behind when a conversation contains more than 10 entries, so stale chat history remains in Redis.

#5655 looks like it addresses the problem. Is there anything currently blocking review/merge, or anything I can provide to help move this forward?

PR #5657, opened 2026-03-22
Add unified cache usage metrics to Usage interface

sobychacko:
Prompt caching can save up to 90% on input token costs, but today reading cache metrics requires provider-specific code — casting to native SDK types, navigating nested records, or reading from metadata maps differently for each provider.

Add getCacheReadInputTokens() and getCacheWriteInputTokens() as default methods on the Usage interface, returning null for providers that don't support caching. Update DefaultUsage with new fields, constructor, and JSON serialization. Update UsageCalculator to accumulate cache metrics across streaming chunks and tool-calling loops.

Wire up cache metrics extraction in four providers:

 - Anthropic: cacheReadInputTokens and cacheCreationInputTokens (sync and streaming)
 - Bedrock: cacheReadInputTokens and cacheWriteInputTokens (sync and streaming)
 - OpenAI SDK: cachedTokens from PromptTokensDetails (read only)
 - Google GenAI: cachedContentTokenCount override (read only)

Add unit tests for DefaultUsage cache fields including serialization round-trips. Add integration test assertions to Anthropic and Bedrock caching tests verifying Usage interface values match native SDK values.

Update usage-handling reference documentation with cache metrics table.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/metadata/Usage.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/metadata/Usage.java
@@ -67,4 +67,28 @@ public interface Usage {
 	 */
 	@Nullable Object getNativeUsage();
 
+	/**
+	 * Returns the number of input tokens read from the prompt cache, if the provider
+	 * supports prompt caching. Cached tokens are tokens that were previously processed
+	 * and stored by the provider, reducing cost and latency for repeated prompt prefixes.
+	 * @return the number of cached input tokens read, or {@code null} if the provider
+	 * does not support prompt caching or no cache hit occurred.
+	 * @since 2.0.0
+	 */
+	default @Nullable Long getCacheReadInputTokens() {
+		return null;
+	}
+
+	/**
+	 * Returns the number of input tokens written to the prompt cache, if the provider
+	 * supports prompt caching. Cache writes occur when new prompt content is cached for
+	 * the first time.
+	 * @return the number of input tokens written to cache, or {@code null} if the
+	 * provider does not support prompt caching or no cache write occurred.
+	 * @since 2.0.0
+	 */
+	default @Nullable Long getCacheWriteInputTokens() {
+		return null;
+	}
+
 }
--- a/spring-ai-model/src/main/java/org/springframework/ai/support/UsageCalculator.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/support/UsageCalculator.java
@@ -61,7 +61,24 @@ public final class UsageCalculator {
 			promptTokens += usageFromPreviousChatResponse.getPromptTokens();
 			generationTokens += usageFromPreviousChatResponse.getCompletionTokens();
 			totalTokens += usageFromPreviousChatResponse.getTotalTokens();
-			return new DefaultUsage(promptTokens, generationTokens, totalTokens);
+			// Accumulate cache metrics, preserving null when neither side reports them.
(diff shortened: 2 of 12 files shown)


tzolov:
@sobychacko after rebasing the testSystemOnlyPromptCaching() is failing

spring-ai-model/src/main/java/org/springframework/ai/chat/metadata/Usage.java
```diff
@@ -67,4 +67,28 @@ default Integer getTotalTokens() {
...
 
+	/**
+	 * Returns the number of input tokens read from the prompt cache, if the provider
+	 * supports prompt caching. Cached tokens are tokens that were previously processed
+	 * and stored by the provider, reducing cost and latency for repeated prompt prefixes.
+	 * @return the number of cached input tokens read, or {@code null} if the provider
+	 * does not support prompt caching or no cache hit occurred.
+	 * @since 2.0.0
+	 */
+	default @Nullable Long getCacheReadInputTokens() {
```
tzolov, line 78: I wonder if `null` is the right approach to express no-tokens. 
I can see that in `getTotalTokens` we default to `0`. Is it too bad to use 0 if there are no CachedReadInputTokens()?  I was thinking of negative numbers (e.g. NO_OPTS = -123) but this could lead to wrong token subtraction if we miss to validate.

spring-ai-model/src/main/java/org/springframework/ai/chat/metadata/Usage.java
```diff
@@ -67,4 +67,28 @@ default Integer getTotalTokens() {
...
+
+	/**
+	 * Returns the number of input tokens written to the prompt cache, if the provider
+	 * supports prompt caching. Cache writes occur when new prompt content is cached for
+	 * the first time.
+	 * @return the number of input tokens written to cache, or {@code null} if the
+	 * provider does not support prompt caching or no cache write occurred.
+	 * @since 2.0.0
+	 */
+	default @Nullable Long getCacheWriteInputTokens() {
```
tzolov, line 90: same like above

spring-ai-model/src/main/java/org/springframework/ai/chat/metadata/Usage.java
tzolov: On second read, null has special and useful semantics in Json serialization (e.g. you can filter out null value fields)

spring-ai-model/src/main/java/org/springframework/ai/support/UsageCalculator.java
```diff
@@ -61,7 +61,24 @@ public static Usage getCumulativeUsage(final Usage currentUsage,
 			promptTokens += usageFromPreviousChatResponse.getPromptTokens();
 			generationTokens += usageFromPreviousChatResponse.getCompletionTokens();
 			totalTokens += usageFromPreviousChatResponse.getTotalTokens();
-			return new DefaultUsage(promptTokens, generationTokens, totalTokens);
+			// Accumulate cache metrics, preserving null when neither side reports them.
+			Long cacheRead = null;
```
tzolov, line 65: this might change if the null semantics change

sobychacko:
@tzolov Thanks for the review. I will go ahead and merge it.

sobychacko:
Merged via https://github.com/spring-projects/spring-ai/commit/7cfad4559be4dbb4e4c60afe081cc3d169e631bb.

PR #5659, opened 2026-03-22
Allow custom StructuredOutputConverter(s) to participate in Native Structured Output

filiphr:
This builds a bit on top of #5412.

The purpose of this is the fact that we extensively use the structured output with our own `StructuredOutputConverter`(s), which do not extend from `BeanOutputConverter`.  Unfortunately, we now need to do something like

```
if (StringUtils.isNotEmpty(outputSchema)) {
    clientRequestBuilder.context(ChatClientAttributes.STRUCTURED_OUTPUT_SCHEMA.getKey(), outputSchema);
    clientRequestBuilder.context(ChatClientAttributes.STRUCTURED_OUTPUT_NATIVE.getKey(), true);
}
```

instead of just passing our output converter to the `ChatClient`. I came up with the workaround because I read the code and how everything works under the hood in Spring AI.

From what I could see the `BeanOutputConverter` is not really needed for this to work, since any `StructuredOutputConverter` should be able to provide a JSON Schema if they want to.

Would appreciated your feedback on this.

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelCallAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelCallAdvisor.java
@@ -16,7 +16,7 @@
 
 package org.springframework.ai.chat.client.advisor;
 
-import java.util.Map;
+import java.util.HashMap;
 
 import org.jspecify.annotations.Nullable;
 
@@ -59,7 +59,7 @@ public final class ChatModelCallAdvisor implements CallAdvisor {
 
 		return ChatClientResponse.builder()
 			.chatResponse(chatResponse)
-			.context(Map.copyOf(formattedChatClientRequest.context()))
+			.context(new HashMap<>(formattedChatClientRequest.context()))
 			.build();
 	}
 
@@ -88,10 +88,7 @@ public final class ChatModelCallAdvisor implements CallAdvisor {
 				.text(userMessage.getText() + System.lineSeparator() + outputFormat)
 				.build());
 
-		return ChatClientRequest.builder()
-			.prompt(augmentedPrompt)
-			.context(Map.copyOf(chatClientRequest.context()))
-			.build();
+		return chatClientRequest.mutate().prompt(augmentedPrompt).build();
 	}
 
 	@Override
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelCallAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelCallAdvisor.java
@@ -16,7 +16,7 @@
 
 package org.springframework.ai.chat.client.advisor;
 
-import java.util.HashMap;
+import java.util.Map;
 
 import org.jspecify.annotations.Nullable;
 
@@ -59,7 +59,7 @@ public final class ChatModelCallAdvisor implements CallAdvisor {
 
 		return ChatClientResponse.builder()
 			.chatResponse(chatResponse)
-			.context(new HashMap<>(formattedChatClientRequest.context()))
+			.context(Map.copyOf(formattedChatClientRequest.context()))
 			.build();
 	}
 
@@ -88,7 +88,10 @@ public final class ChatModelCallAdvisor implements CallAdvisor {
 				.text(userMessage.getText() + System.lineSeparator() + outputFormat)
 				.build());
 
-		return chatClientRequest.mutate().prompt(augmentedPrompt).build();
+		return ChatClientRequest.builder()
+			.prompt(augmentedPrompt)
+			.context(Map.copyOf(chatClientRequest.context()))
+			.build();
 	}
 
 	@Override
--- a/spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/ChatClientNativeStructuredResponseTests.java
+++ b/spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/ChatClientNativeStructuredResponseTests.java
@@ -16,18 +16,20 @@
 
 package org.springframework.ai.chat.client;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 import net.javacrumbs.jsonunit.assertj.JsonAssertions;
 import net.javacrumbs.jsonunit.core.Option;
+import org.jspecify.annotations.NonNull;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import tools.jackson.databind.JsonNode;
 
 import org.springframework.ai.chat.client.advisor.api.CallAdvisor;
 import org.springframework.ai.chat.client.advisor.api.CallAdvisorChain;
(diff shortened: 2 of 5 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelCallAdvisor.java
```diff
@@ -59,7 +59,7 @@ public ChatClientResponse adviseCall(ChatClientRequest chatClientRequest, CallAd
 
 		return ChatClientResponse.builder()
 			.chatResponse(chatResponse)
-			.context(Map.copyOf(formattedChatClientRequest.context()))
+			.context(new HashMap<>(formattedChatClientRequest.context()))
```
filiphr: This is needed due to the fact that `Map.copyOf` does not allow `null` values, whereas the contract of the context is `Map<String, @Nullable Object>`.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelCallAdvisor.java
```diff
@@ -88,10 +88,7 @@ private static ChatClientRequest augmentWithFormatInstructions(ChatClientRequest
 				.text(userMessage.getText() + System.lineSeparator() + outputFormat)
 				.build());
 
-		return ChatClientRequest.builder()
-			.prompt(augmentedPrompt)
-			.context(Map.copyOf(chatClientRequest.context()))
-			.build();
+		return chatClientRequest.mutate().prompt(augmentedPrompt).build();
```
filiphr: Same reason as the change of `Map.copyOf`

spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/ChatClientNativeStructuredResponseTests.java
```diff
@@ -325,12 +329,199 @@ public void dynamicDisableNativeEntityTest() {
...
+		verify(this.structuredOutputChatOptions, never()).setOutputSchema(anyString());
+	}
+
 	record UserEntity(String name, int age) {
 	}
 
 	private static class ContextCatcherCallAdvisor implements CallAdvisor {
 
-		private Map<String, Object> context = new ConcurrentHashMap<>();
+		private Map<String, Object> context = new HashMap<>();
```
filiphr: Same reason as the `Map.copyOf`. `ConcurrentHashMap` does not allow `null` values.

filiphr (commented):
Adding some explanation for some changes that are not entirely related to the initial request. In theory this could lead to some problems outside of the proposed changes here since the contract between the context and the underlying implementations differs.

If you prefer I can split this PR into 2 to handle the contract difference.

filiphr:
Hey @ilayaperumalg, @tzolov, sorry for the ping, but I wanted to check if this PR is something that you would consider for Spring AI. I personally think that it's a good improvement of the API.

Thanks for looking into it.

filiphr:
Hey @ilayaperumalg, have you had the chance to have a look at this one perhaps? If this something that is interesting for Spring AI. Currently, we have to fiddle with some of the lower level flags to trigger the structured output.

ilayaperumalg:
@filiphr Sorry about the delay in response. The changes look good to me. We'll target to get this in for M5.

ilayaperumalg (approved):

filiphr:
Thanks for the review @ilayaperumalg. Looking forward to M5.

spring-ai-model/src/main/java/org/springframework/ai/converter/StructuredOutputConverter.java
```diff
@@ -26,7 +28,17 @@
...
+ * @author Filip Hrisafov
  */
 public interface StructuredOutputConverter<T> extends Converter<String, T>, FormatProvider {
 
+	/**
+	 * Returns the JSON schema for the structured output of an LLM call.
+	 * @return the JSON schema or {@code null} if not available
+	 * @since 2.0.0
+	 */
+	@Nullable default String getJsonSchema() {
```
tzolov, line 40: Using `null` as no-op is undesired and can lead to side effects.
Consider defining an explicit  `NO_JSON_SCHEMA = ""` directly on `StructuredOutputConverter`, change the default method to return it (non-null), and let the existing `StringUtils.hasText()` guard in `ChatModelCallAdvisor` handle it — which, btw, already treats "" and null identically.

tzolov: Then you can remove the @Nullable.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelCallAdvisor.java
tzolov: With NO_JSON_SCHEMA this change is not needed

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelCallAdvisor.java
tzolov: With NO_JSON_SCHEMA this change is not needed

spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/ChatClientNativeStructuredResponseTests.java
tzolov: With NO_JSON_SCHEMA this change is not needed

tzolov (changes requested):
Thank you @filiphr. Very useful improvement. Check my comments around the No-Ops approach. It should simplify the PR.

Also could you please add some documentation about using custom structured-output-convertors in the https://docs.spring.io/spring-ai/reference/2.0-SNAPSHOT/api/structured-output-converter.html#_native_structured_output 
(This is optional we can always do it afterwards)

filiphr:
Thanks for the review @tzolov 

I'll do the adjustment for the No-Op, it's cleaner.

However, the contract of `ChatClientRequest#context` is still `Map<String, @Nullable Object>`

https://github.com/spring-projects/spring-ai/blob/main/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClientRequest.java#L36C15-L36C32

I think the change in the production code for that is still needed. However, for that I'll write a separate test case and PR for it.

filiphr:
@tzolov I've applied your entire feedback in this PR. Let me know ift there is something else left that I should do from my side.

Btw, I force pushed, but that was only my "Apply review feedback" that I pushed too early.

filiphr:
I'm not sure why the MCP Integration tests are failing. I've run them locally and they are green for me.

tzolov (commented):
LGTM @filiphr! One thing — we should use an explicit `NO_JSON_SCHEMA` check instead of comparing against an empty string, otherwise a value change would break the logic. We could address this in the merge or a follow-up PR, perhaps by adding a default `hasJsonSchema()` method to `StructuredOutputConverter`. Thoughts?

filiphr:
>One thing — we should use an explicit `NO_JSON_SCHEMA` check instead of comparing against an empty string, otherwise a value change would break the logic

I am not sure that I am following this. Check this where? In the `DefaultChatClient` before putting it in the request context or in the `ChatModelCallAdvisor`?

In 

https://github.com/spring-projects/spring-ai/blob/ed2c7c7c058e1cc3910a5c7f5a5e98b47ef1b169/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ChatModelCallAdvisor.java#L70-L75

There is already a `StringUtils.hasText(outputSchema)`. If we change the logic in the `ChatModelCallAdvisor` and we change the value of `NO_JSON_SCHEMA` then it is possible that people that used the `hatClientAttributes.STRUCTURED_OUTPUT_SCHEMA` key directly (we use it) to have a behaviour change as well.

>perhaps by adding a `default hasJsonSchema()` method to `StructuredOutputConverter`. Thoughts?

From a developer experience point of view, I find this a bit too verbose. People would need to override 2 methods to provide their logic.

PR #5660, opened 2026-03-23
Add portable PromptCacheOptions for provider-agnostic cache configura…

sobychacko:
…tion

Today, configuring prompt caching requires learning provider-specific APIs — AnthropicCacheOptions with AnthropicCacheStrategy for Anthropic, BedrockCacheOptions with BedrockCacheStrategy for Bedrock. Users who switch providers must rewrite their cache configuration, and the conceptually identical strategies (SYSTEM_ONLY, TOOLS_ONLY, etc.) live in separate, unrelated enums.

Introduce PromptCacheStrategy and PromptCacheOptions as core types on the ChatOptions interface, allowing users to configure prompt caching once using portable types that work across providers. Each supporting provider translates these to its native cache mechanism — Anthropic maps strategy, TTL, and minContentLength to AnthropicCacheOptions; Bedrock maps strategy to BedrockCacheOptions (logging that TTL is ignored since Bedrock uses a fixed 5-minute TTL).

Provider-specific cache options take precedence when both are set, preserving full backward compatibility and access to provider-specific features like Anthropic's multi-block system caching or per-message-type TTL configuration.


tzolov (commented):
Looks great. The options abstraction nicely unifies caching capabilities, and the documentation clearly explaining explicit (Anthropic), implicit (OpenAI), and dedicated-service (Gemini) types is helpful — I wasn't aware the others don't expose direct caching control.
@sobychacko, before we merge, though: is this generalization actually warranted? I initially supported it assuming multiple providers would implement caching differently — similar to structured output — but if it'll only ever apply to Anthropic, do we really need the abstraction layer?

sobychacko:
@tzolov Thats a fair point, something I also struggled a bit. However, it is applied to both Anthropic and Bedrock. They both provide similar API's for prompt caching and we have duplication across those two modules. Bedrock provides prompt caching for both nova and claude models. So, this is up to you. I am ok if we decide not to merge; we just need to live with the redundancy in those two modules, which may not be an unreasonable thing to do in this case.

sdeleuze:
As discussed with @sobychacko, the proposals in this PR are going toward a better design, but multiple reasons make me think 2.0 is probably too early to introduce that. As pointed out by @tzolov, current abstraction is mainly designed against Anthropic prompt caching support (directly or via Bedrock), 2.1 features like Responses API / Interactions API may impact the design, same for spring-ai-session and some providers support other kind of caching than prompt one.

We could potentially include a shortcut to `AnthropicCacheStrategy` or similar without forcing to use the builder but that's something we could include later as a non breaking change.

So as agreed with Soby, I will close this PR unmerged.

PR #5661, opened 2026-03-23
Fix `AzureOpenAiChatOptions#stop` field initializer regression

### LINKED ISSUE #5656: AzureOpenAiChatOptions.stop field initializer regression in M3 causes "stop": [] to be sent to Azure, breaking reasoning models
In M3, `AzureOpenAiChatOptions` changed `private List<String> stop;` to `private List<String> stop = new ArrayList<>()`. This causes `"stop": []` to be sent to Azure on every request, even when stop
sequences are never set.

`ModelOptionsUtils.merge()` (called by `buildRequestPrompt()`) serializes options to a map, merges, then deserializes via Jackson into a new `AzureOpenAiChatOptions`. The no-arg constructor runs the
field initializer (`stop = new ArrayList<>()`), and since `stop` isn't in the JSON, Jackson never overwrites it. The merge logic sees `getStop() != null` → true → sends `"stop": []`.

Reasoning models (o1, gpt-5.1) reject this: `Unsupported parameter: 'stop' is not supported with this model.`

Setting `stop` to null doesn't help — `buildRequestPrompt()` recreates the object via Jackson, re-triggering the field initializer. We had to roll back to M2.

PR #5619 (merged for M4) fixes the builder path but not the `ModelOptionsUtils.merge()` path, which bypasses the builder.

**Environment**

- Spring AI: 2.0.0-M3 (works in M2)
- Java: 21
- Azure OpenAI SDK: 1.0.0-beta.16

**Steps to reproduce**

```java
var model = AzureOpenAiChatModel.builder()
        .openAIClientBuilder(clientBuilder)
        .defaultOptions(AzureOpenAiChatOptions.builder().deploymentName("gpt-5.1").build())
        .build();

model.call(new Prompt(new UserMessage("hello")));
// Fails: "Unsupported parameter: 'stop' is not supported with this model."
```

**Expected behavior**

stop should stay null through all internal merge paths when never explicitly set.

```java
@Test
void stopFieldShouldBeNullAfterJacksonRoundtrip() {
    var mapper = JsonMapper.builder().build();
    var options = AzureOpenAiChatOptions.builder().deploymentName("gpt-5.1").build();

    String json = mapper.writeValueAsString(options);
    var deserialized = mapper.readValue(json, AzureOpenAiChatOptions.class);

    // Fails in M3: stop is [] instead of null
    assertThat(deserialized.getStop()).isNull();
}
```

ZhangDT-sky:
This PR fixes a regression where the `stop` field in `AzureOpenAiChatOptions` was initialized to an empty `ArrayList`. This caused the JSON payload to always include `"stop": []`, which is rejected by Azure OpenAI reasoning models (e.g., o1, o3).

### Changes:
- Removed the default initialization of the `stop` field.
- Added a regression test in `AzureOpenAiChatOptionsTests` to ensure the `stop` field remains `null` after options merging and Jackson roundtrips.

**Fixes #5656**

### DIFF
--- a/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatOptions.java
+++ b/models/spring-ai-azure-openai/src/main/java/org/springframework/ai/azure/openai/AzureOpenAiChatOptions.java
@@ -169,7 +169,7 @@ public class AzureOpenAiChatOptions implements ToolCallingChatOptions {
 	 * A collection of textual sequences that will end completions generation.
 	 */
 	@JsonProperty("stop")
-	private List<String> stop = new ArrayList<>();
+	private List<String> stop;
 
 	/**
 	 * A value that influences the probability of generated tokens appearing based on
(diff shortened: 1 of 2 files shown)


sdeleuze:
Polished (checkstyle error) and merged, thanks.

PR #5667, opened 2026-03-24
Add Anthropic service tier support for priority capacity selection

sobychacko:
Anthropic offers priority and standard capacity tiers for API requests. This adds service tier configuration so users can control whether requests opportunistically use priority capacity (lower latency) or always use standard capacity.

Introduces AnthropicServiceTier as a Java enum rather than exposing the SDK's Kotlin sealed class directly, enabling Spring Boot property binding via spring.ai.anthropic.chat.options.service-tier.


tzolov:
Rebased and merged at b1c6375acfe073c18710a6a5fa9df2e6c548bb26

PR #5683, opened 2026-03-26
Fix CosmosDB NIN filter operator from !IN to NOT IN

### LINKED ISSUE #5665: CosmosDB incorrect filter conversion
https://github.com/spring-projects/spring-ai/blob/0ad60decf02dab62cf4a09003ac07aefd9064e90/vector-stores/spring-ai-azure-cosmos-db-store/src/main/java/org/springframework/ai/vectorstore/cosmosdb/CosmosDBFilterExpressionConverter.java#L128

This is incorrect and CosmosDB will not recognize it! Try it out in the CosmosDB Data Explorer UI, the correct formulation is NOT IN instead of !IN

chanani:
## Summary

Fixes #5665

## Problem

`CosmosDBFilterExpressionConverter` was generating `!IN` for `NIN` 
filter expressions, but CosmosDB NoSQL API does not recognize `!IN` 
as a valid operator.

## Root Cause

The `NIN` case in `CosmosDBFilterExpressionConverter` was producing 
the wrong operator string:
```java
// Before
case NIN -> " !IN ";

// After
case NIN -> " NOT IN ";
```

## Verification

The correct syntax can be verified in the 
[CosmosDB Data Explorer](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query/keywords#not-in).
```sql
-- Invalid (before)
SELECT * FROM c WHERE c.category !IN ("tech", "science")

-- Valid (after)
SELECT * FROM c WHERE c.category NOT IN ("tech", "science")
```

### DIFF
--- a/vector-stores/spring-ai-azure-cosmos-db-store/src/main/java/org/springframework/ai/vectorstore/cosmosdb/CosmosDBFilterExpressionConverter.java
+++ b/vector-stores/spring-ai-azure-cosmos-db-store/src/main/java/org/springframework/ai/vectorstore/cosmosdb/CosmosDBFilterExpressionConverter.java
@@ -125,7 +125,7 @@ class CosmosDBFilterExpressionConverter extends AbstractFilterExpressionConverte
 			case GT -> " > ";
 			case GTE -> " >= ";
 			case IN -> " IN ";
-			case NIN -> " !IN ";
+			case NIN -> " NOT IN ";
 			default -> throw new RuntimeException("Not supported expression type:" + exp.type());
 		};
 	}


chanani:
Hi @tzolov ,

Could a maintainer please review this PR? It fixes a bug where `CosmosDB NIN` filter expressions were incorrectly generating `!IN` instead of the valid `NOT IN` operator.

The fix is a one-line change, well-scoped, and resolves #5665. Let me know if anything needs to be updated. Thanks!

sobychacko:
@chanani Hi, we will have an update on a few PR's and issues around CosmosDB very soon (in a few days or so). Thanks for the patience.

sdeleuze:
The CosmosDB now lives in https://azurecosmosdb.github.io/spring-ai/docs/index.html.

PR #5684, opened 2026-03-26
Remove SpringAiTestAutoConfigurations

ericbottard:
This commit removes the `SpringAiTestAutoConfigurations` class and its corresponding tests. This class wrongfully assumed that `@AutoConfiguration.after` was a way to express dependencies.

A lot of `@AutoConfiguration.after` were removed because they were not necessary. Such order on `@AutoConfiguration B` on `A` (via `after = A.class`) only makes sense if `B` _e.g._ defines beans that are `@ConditionalOnMissingBean` AND `A` also defines such beans, and we want to make sure that `A`'s beans are defined before `B`'s.

Everytime the attribute was legit, a comment was added to explain why.

Also, some `B.afterName = A` were replaced by `A.before = B`, which is typesafe.

All tests that used `SpringAiTestAutoConfigurations` have been changed to explicity create a list of auto configurations based on the previous value of the after attribute. This list may be a superset of what is actually needed, but for the sake of simplicity, this is what this commit does. Also note that `SpringAiTestAutoConfigurations` did not compute a transitive closure of the after dependencies (only the 1st level), so it could not be "corect" anyway.


sdeleuze (approved):
I won't claim I was able to review everything, but the principle makes sense to me.

Issue #2535, opened 2025-03-20
DefaultUsage#nativeUsage can crash SimpleLoggingAdvisor


mands:
The `nativeUsage` object within `DefaultUsage` may not be JSON serialisable and hence causes a crash when using `SimpleLoggingAdvisor`.

I've hit this when using VertexAI, where the `nativeUsage` instance, which is cast as an `Object`, is a protobuf of type `GenerateContentRepsonse$UsageMetadata`.

This can't be serialised for pretty printing within the SimpleLoggerAdvisor repsonse code, which calls `ModelOptionsUtils#toJsonStringPrettyPrinter` and thus throws an Exception.

```java
	public static String toJsonStringPrettyPrinter(Object object) {
		try {
			return OBJECT_MAPPER.writerWithDefaultPrettyPrinter().writeValueAsString(object);
		}
		catch (JsonProcessingException e) {
			throw new RuntimeException(e);
		}
	}
```	

I think simplest solution would be to add a `@JsonIgnore` to `DefaultUsage#getNativeUsage` as it's not always obvious if it can be logged.

PR #5690, opened 2026-03-27
Fix Bedrock Converse nativeUsage JSON serialization

### LINKED ISSUE #5682: No serializer found for class software.amazon.awssdk.services.bedrockruntime.model.TokenUsage
**Bug description**
After upgrade from Spring AI 1.1.2 to 1.1.3 following Exception is thrown:
```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: No serializer found for class software.amazon.awssdk.services.bedrockruntime.model.TokenUsage and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) (through reference chain: org.springframework.ai.chat.model.ChatResponse["metadata"]->org.springframework.ai.chat.metadata.ChatResponseMetadata["usage"]->org.springframework.ai.chat.metadata.DefaultUsage["nativeUsage"])
	at com.fasterxml.jackson.databind.exc.InvalidDefinitionException.from(InvalidDefinitionException.java:77) ``` 


in this code:

```java
  @GetMapping("/hello-response")
  public ChatResponse helloResponse(@RequestParam String prompt) {
    String promptFromDb = "Plan me a {days}-day trip to {city}";
    return chatClient
        .prompt()
        .user(u -> u.text(promptFromDb).param("days", 3).param("city", prompt))
        .call()
        .chatResponse();
  }

```  

**Environment**
Java 25
Spring Boot 3.5
Spring AI 1.1.3
Amazon Bedrock Converse

**Steps to reproduce**
Execute above code. 

**Expected behavior**
No exception is thrown like in Spring AI 1.1.2

ZhangDT-sky:
## Summary

Fixes #5682 by making Bedrock Converse `nativeUsage` JSON-serializable when `ChatResponse` is returned directly from a Spring MVC endpoint.

Previously, `BedrockProxyChatModel` stored the AWS SDK `TokenUsage` object in `DefaultUsage.nativeUsage`. Since that type is not Jackson-serializable by default, returning `ChatResponse` from a controller could fail with `InvalidDefinitionException`.

This change stores a JSON-safe `Map<String, Object>` in `nativeUsage` instead of introducing Bedrock-specific Jackson customization, while preserving token and cache usage aggregation across tool-calling rounds and streaming responses.

## Changes

- Replace Bedrock Converse `nativeUsage` from AWS SDK `TokenUsage` with a JSON-safe `Map<String, Object>`
- Preserve aggregation for both legacy `TokenUsage` values and the new `Map` representation
- Apply the same behavior to streaming responses
- Update Bedrock Converse usage aggregation tests

## Why

This keeps `ChatResponse` safe to serialize out of the box in common Spring Boot MVC usage without requiring users to disable `FAIL_ON_EMPTY_BEANS` or register custom Jackson serializers.

## Testing

```bash
mvn -pl models/spring-ai-bedrock-converse test -DskipITs
```


sobychacko:
Thanks for the PR and the analysis. I submitted another [PR](https://github.com/spring-projects/spring-ai/pull/6919) that addresses this issue at the core level. The serialization problem isn't specific to Bedrock, the same thing was reported for Vertex in #2535, so handling it in `DefaultUsage` takes care of every provider at once. It also keeps the AWS `TokenUsage` object in `nativeUsage` as is, since users rely on casting to that type.

Closing this one in favor of #6919. Appreciate you digging into it.

PR #5714, opened 2026-03-30
Vertex AI removal for non-embedding use cases

sdeleuze:
The PR removes non-embedding support for Vertex AI, as Google GenAI should now be used instead.


spring-ai-docs/src/main/antora/modules/ROOT/pages/api/embeddings/google-genai-embeddings-text.adoc
```diff
@@ -286,23 +286,12 @@ GoogleGenAiTextEmbeddingOptions options = GoogleGenAiTextEmbeddingOptions.builde
...
-
-**Use Google GenAI Embeddings when:**
-- You want quick prototyping with API keys
-- You need the latest embedding features from the Developer API
-- You want flexibility to switch between API key and Vertex AI modes
-- You're already using Google GenAI for chat
-
-**Use Vertex AI Text Embeddings when:**
-- You have existing Vertex AI infrastructure
-- You need multimodal embeddings (currently only available in Vertex AI)
```
ericbottard, line 307: Dont we want to keep this diff bullet list (and this bullet in particular)?

spring-ai-docs/src/main/antora/modules/ROOT/pages/api/vectordbs.adoc
```diff
@@ -259,38 +259,24 @@ TokenCountBatchingStrategy strategy = new TokenCountBatchingStrategy(
...
 
-Here's an example configuration for using Vertex AI with auto-truncation and custom `BatchingStrategy` and then using them in the PgVectorStore:
+Here's an example configuration for using an embedding model with auto-truncation and custom `BatchingStrategy` and then using them in the PgVectorStore:
 
 [source,java]
 ----
 @Configuration
 public class AutoTruncationEmbeddingConfig {
 
-    @Bean
```
ericbottard, line 277: Seems like this example makes no sense now without this. I think everything related to embedding should be kept in docs, no?

spring-ai-docs/src/main/antora/modules/ROOT/pages/api/embeddings/google-genai-embeddings-text.adoc
sdeleuze: This section is specifically about text embedding which is supported on GenAI side, so I don't think we need to argue about GenAI versus Vertex in that context anymore.

spring-ai-docs/src/main/antora/modules/ROOT/pages/api/vectordbs.adoc
sdeleuze: Indeed, I will revert.

spring-ai-docs/src/main/antora/modules/ROOT/pages/api/embeddings/google-genai-embeddings-text.adoc
ericbottard: 👍

ericbottard (approved):

PR #5715, opened 2026-03-30
Fix Javadoc documentation and configuration

sdeleuze:
Before this PR, the documented javadoc generation command was broken.

This PR updates it, and refines the javadoc configuration to just skip the javadoc generation in `spring-ai-spring-boot-testcontainers` (which contains no public class) instead of disabling `failOnError`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ To update the year on license headers using the [license-maven-plugin](https://o
 
 To check javadocs using the [javadoc:javadoc](https://maven.apache.org/plugins/maven-javadoc-plugin/)
 ```shell
-./mvnw javadoc:javadoc -Pjavadoc
+./mvnw javadoc:javadoc
 ```
 
 #### Source Code Style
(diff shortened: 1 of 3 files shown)


ericbottard:
Wouldn't the same issue apply to the starters modules?

ericbottard:
Otherwise, this way of doing seems good to me!

ericbottard (approved):

sdeleuze:
@ericbottard I don't see another `<failOnError>false</failOnError>` occurence and the build works fine as it is, so I am not sure we need to refine something else. Maybe the Javadoc plugin is clever enough to skip modules without `src/main` directory.

ericbottard:
> @ericbottard I don't see another `<failOnError>false</failOnError>` occurence and the build works fine as it is, so I am not sure we need to refine something else. Maybe the Javadoc plugin is clever enough to skip modules without `src/main` directory.

Must be something like that. Confirmed this works as is.

PR #5719, opened 2026-03-30
Add OpenAI SDK based AudioTranscription model support

ilayaperumalg:
- Introduces audio transcription model implementation using the official OpenAI Java SDK (com.openai:openai-java) as an alternative to the existing Spring RestClient-based implementation.

### DIFF
--- a/models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkAudioTranscriptionOptions.java
+++ b/models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkAudioTranscriptionOptions.java
@@ -22,8 +22,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.openai.azure.AzureOpenAIServiceVersion;
 import com.openai.credential.Credential;
 import com.openai.models.audio.AudioModel;
@@ -42,7 +40,6 @@ import org.springframework.ai.audio.transcription.AudioTranscriptionOptions;
  * @author Piotr Olaszewski
  * @since 2.0.0
  */
-@JsonInclude(JsonInclude.Include.NON_NULL)
 public class OpenAiSdkAudioTranscriptionOptions extends AbstractOpenAiSdkOptions implements AudioTranscriptionOptions {
 
 	/**
@@ -55,22 +52,16 @@ public class OpenAiSdkAudioTranscriptionOptions extends AbstractOpenAiSdkOptions
 	 */
 	public static final AudioResponseFormat DEFAULT_RESPONSE_FORMAT = AudioResponseFormat.TEXT;
 
-	@JsonProperty("model")
 	private @Nullable String model;
 
-	@JsonProperty("response_format")
 	private AudioResponseFormat responseFormat = DEFAULT_RESPONSE_FORMAT;
 
-	@JsonProperty("prompt")
 	private @Nullable String prompt;
 
-	@JsonProperty("language")
 	private @Nullable String language;
 
-	@JsonProperty("temperature")
 	private @Nullable Float temperature;
 
-	@JsonProperty("timestamp_granularities")
 	private @Nullable List<TranscriptionCreateParams.TimestampGranularity> timestampGranularities;
 
 	public static Builder builder() {


models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkAudioTranscriptionOptions.java
```diff
@@ -0,0 +1,362 @@
...
+	 */
+	public static final String DEFAULT_TRANSCRIPTION_MODEL = AudioModel.WHISPER_1.asString();
+
+	/**
+	 * Default response format.
+	 */
+	public static final AudioResponseFormat DEFAULT_RESPONSE_FORMAT = AudioResponseFormat.TEXT;
+
+	@JsonProperty("model")
+	private @Nullable String model;
```
sdeleuze, line 55: Can we initialize it to `DEFAULT_TRANSCRIPTION_MODEL` and make it non-nullable?

models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkAudioTranscriptionOptions.java
```diff
@@ -0,0 +1,362 @@
...
+
+	@JsonProperty("response_format")
+	private @Nullable AudioResponseFormat responseFormat;
```
sdeleuze: Can we initialize it to `DEFAULT_RESPONSE_FORMAT` and make it non-nullable?

models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkAudioTranscriptionOptions.java
ilayaperumalg: Fixed this for AudioResponseFormat. This needs to be addressed across the OpenAI SDK models. Can we address this separately after this one and the moderation model PR are merged?

auto-configurations/models/spring-ai-autoconfigure-model-openai-sdk/src/test/java/org/springframework/ai/model/openaisdk/autoconfigure/OpenAiSdkAudioTranscriptionAutoConfigurationIT.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Integration tests for {@link OpenAiSdkAudioTranscriptionAutoConfiguration}.
+ *
+ * @author Ilayaperumal Gopinathan
+ * @author Michael Lavelle
+ * @author Christian Tzolov
+ * @author Thomas Vitale
+ * @since 2.0.0
+ */
+@EnabledIfEnvironmentVariable(named = "OPENAI_API_KEY", matches = ".*")
```
ericbottard:
```suggestion
@EnabledIfEnvironmentVariable(named = "OPENAI_API_KEY", matches = ".+")
```

auto-configurations/models/spring-ai-autoconfigure-model-openai-sdk/src/test/java/org/springframework/ai/model/openaisdk/autoconfigure/OpenAiSdkAudioTranscriptionAutoConfigurationIT.java
```diff
@@ -0,0 +1,106 @@
...
+public class OpenAiSdkAudioTranscriptionAutoConfigurationIT {
+
+	private static final Log logger = LogFactory.getLog(OpenAiSdkAudioTranscriptionAutoConfigurationIT.class);
+
+	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner().withPropertyValues(
+			"spring.ai.openai-sdk.api-key=" + System.getenv("OPENAI_API_KEY"),
```
ericbottard, line 48: Does this test really require a real API key, and thus, is it really an Integration test?

What if we put "bogus-key" here? Will it still startup and allow us testing the values based on ConfigurationProperties.

Seems like only `transcribe()` actually does real processing. Can we split this class in two?

auto-configurations/models/spring-ai-autoconfigure-model-openai-sdk/src/test/java/org/springframework/ai/model/openaisdk/autoconfigure/OpenAiSdkAudioTranscriptionPropertiesTests.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Unit tests for {@link OpenAiSdkAudioTranscriptionProperties}.
+ *
+ * @author Ilayaperumal Gopinathan
+ * @author Michael Lavelle
+ * @author Christian Tzolov
+ * @author Piotr Olaszewski
+ */
+class OpenAiSdkAudioTranscriptionPropertiesTests {
```
ericbottard, line 35: Actually, seems like there already is a class testing the effect of `@ConfigurationProperties` here. There may be some redundancy

models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkAudioTranscriptionModel.java
```diff
@@ -0,0 +1,231 @@
...
+			this.options = options;
+			return this;
+		}
+
+		/**
+		 * Builds a new {@link OpenAiSdkAudioTranscriptionModel} instance.
+		 * @return the configured transcription model
+		 */
+		public OpenAiSdkAudioTranscriptionModel build() {
+			return new OpenAiSdkAudioTranscriptionModel(this);
```
ericbottard, line 226: Can we use a more traditional builder pattern where the thing being built does not know about the builder, but vice-versa

models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkAudioTranscriptionOptions.java
```diff
@@ -0,0 +1,362 @@
...
+/**
+ * OpenAI SDK Audio Transcription Options.
+ *
+ * @author Ilayaperumal Gopinathan
+ * @author Michael Lavelle
+ * @author Christian Tzolov
+ * @author Piotr Olaszewski
+ * @since 2.0.0
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
ericbottard: Are those "options" being serialized on the wire because the SDK uses them (I doubt that), or are they "options" in the spring-ai sense?

If the latter, they shouldn't be Jackson annotated (at least eventually).

models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkAudioTranscriptionOptions.java
```diff
@@ -0,0 +1,362 @@
...
+	@Override
+	public String toString() {
+		return "OpenAiSdkAudioTranscriptionOptions{" + "model='" + this.model + '\'' + ", responseFormat="
+				+ this.responseFormat + ", prompt='" + this.prompt + '\'' + ", language='" + this.language + '\''
+				+ ", temperature=" + this.temperature + ", timestampGranularities=" + this.timestampGranularities + '}';
+	}
+
+	public static final class Builder {
+
+		private final OpenAiSdkAudioTranscriptionOptions options;
```
ericbottard: The way this builder pattern is implemented is dangerous.

A builder should pass tests similar in spirit to [those](https://github.com/spring-projects/spring-ai/blob/cafecdf0141fb4e86f31f87f0a606a3d374b249e/spring-ai-test/src/main/java/org/springframework/ai/test/options/AbstractChatOptionsTests.java#L35-L63).

ilayaperumalg:
Squashed, rebased and merged via 4aaca6df1

PR #5725, opened 2026-03-31
Move options merging to ChatClient level via builder combineWith() method

ericbottard:
This commit gets rid of the Jackson oriented ModelOptionsUtils.merge() usage in the buildRequestPrompt() method of models.

Instead, "options merging" is now implemented at the ChatClient level, and uses the builder combineWith() method to merge options.

This means that ChatModel.call() now expects the prompt it receives to already have the options merged OR to have a prompt with null options (in which case the model default options will be used).

Lots of tests using mocks fixed by adding when(mockModel.getDefaultOptions)).thenReturn(someOptions))

All models in the repo have been changed except AzureOpenAiChatModel

ITs pass on OllamaChatModel, need to see if others pass too.

### DIFF
--- a/spring-ai-integration-tests/pom.xml
+++ b/spring-ai-integration-tests/pom.xml
@@ -80,7 +80,7 @@
 
 		<dependency>
 			<groupId>org.springframework.ai</groupId>
-			<artifactId>spring-ai-starter-model-openai</artifactId>
+			<artifactId>spring-ai-starter-model-openai-sdk</artifactId>
 			<version>${project.parent.version}</version>
 			<scope>test</scope>
 		</dependency>
--- a/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/client/advisor/QuestionAnswerAdvisorIT.java
+++ b/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/client/advisor/QuestionAnswerAdvisorIT.java
@@ -33,7 +33,7 @@ import org.springframework.ai.document.DocumentReader;
 import org.springframework.ai.evaluation.EvaluationRequest;
 import org.springframework.ai.evaluation.EvaluationResponse;
 import org.springframework.ai.integration.tests.TestApplication;
-import org.springframework.ai.openai.OpenAiChatModel;
+import org.springframework.ai.openaisdk.OpenAiSdkChatModel;
 import org.springframework.ai.reader.markdown.MarkdownDocumentReader;
 import org.springframework.ai.reader.markdown.config.MarkdownDocumentReaderConfig;
 import org.springframework.ai.template.st.StTemplateRenderer;
@@ -57,7 +57,7 @@ public class QuestionAnswerAdvisorIT {
 	private List<Document> knowledgeBaseDocuments;
 
 	@Autowired
-	OpenAiChatModel openAiChatModel;
+	OpenAiSdkChatModel openAiChatModel;
 
 	@Autowired
 	PgVectorStore pgVectorStore;
--- a/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/client/advisor/QuestionAnswerAdvisorStreamIT.java
+++ b/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/client/advisor/QuestionAnswerAdvisorStreamIT.java
@@ -30,8 +30,8 @@ import org.springframework.ai.chat.client.advisor.vectorstore.QuestionAnswerAdvi
 import org.springframework.ai.document.Document;
 import org.springframework.ai.document.DocumentReader;
 import org.springframework.ai.integration.tests.TestApplication;
-import org.springframework.ai.openai.OpenAiChatModel;
-import org.springframework.ai.openai.OpenAiChatOptions;
+import org.springframework.ai.openaisdk.OpenAiSdkChatModel;
+import org.springframework.ai.openaisdk.OpenAiSdkChatOptions;
 import org.springframework.ai.reader.markdown.MarkdownDocumentReader;
 import org.springframework.ai.reader.markdown.config.MarkdownDocumentReaderConfig;
 import org.springframework.ai.vectorstore.pgvector.PgVectorStore;
@@ -53,7 +53,7 @@ public class QuestionAnswerAdvisorStreamIT {
 	private List<Document> knowledgeBaseDocuments;
 
 	@Autowired
-	OpenAiChatModel openAiChatModel;
+	OpenAiSdkChatModel openAiChatModel;
 
 	@Autowired
 	PgVectorStore pgVectorStore;
@@ -86,7 +86,7 @@ public class QuestionAnswerAdvisorStreamIT {
 			.build()
 			.prompt(question)
 			.advisors(qaAdvisor)
-			.options(OpenAiChatOptions.builder().streamUsage(true).build())
+			.options(OpenAiSdkChatOptions.builder().streamUsage(true))
 			.stream()
 			.content();
 
--- a/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/client/advisor/RetrievalAugmentationAdvisorIT.java
+++ b/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/client/advisor/RetrievalAugmentationAdvisorIT.java
@@ -34,7 +34,7 @@ import org.springframework.ai.document.DocumentReader;
 import org.springframework.ai.evaluation.EvaluationRequest;
 import org.springframework.ai.evaluation.EvaluationResponse;
 import org.springframework.ai.integration.tests.TestApplication;
-import org.springframework.ai.openai.OpenAiChatModel;
+import org.springframework.ai.openaisdk.OpenAiSdkChatModel;
 import org.springframework.ai.rag.advisor.RetrievalAugmentationAdvisor;
 import org.springframework.ai.rag.preretrieval.query.expansion.MultiQueryExpander;
 import org.springframework.ai.rag.preretrieval.query.transformation.CompressionQueryTransformer;
@@ -63,7 +63,7 @@ class RetrievalAugmentationAdvisorIT {
 	private List<Document> knowledgeBaseDocuments;
 
 	@Autowired
-	OpenAiChatModel openAiChatModel;
+	OpenAiSdkChatModel openAiChatModel;
 
 	@Autowired
 	PgVectorStore pgVectorStore;
--- a/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/rag/generation/augmentation/ContextualQueryAugmenterIT.java
+++ b/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/rag/generation/augmentation/ContextualQueryAugmenterIT.java
@@ -23,7 +23,7 @@ import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
 
 import org.springframework.ai.document.Document;
 import org.springframework.ai.integration.tests.TestApplication;
-import org.springframework.ai.openai.OpenAiChatModel;
+import org.springframework.ai.openaisdk.OpenAiSdkChatModel;
 import org.springframework.ai.rag.Query;
 import org.springframework.ai.rag.generation.augmentation.ContextualQueryAugmenter;
 import org.springframework.ai.rag.generation.augmentation.QueryAugmenter;
@@ -42,7 +42,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 class ContextualQueryAugmenterIT {
 
 	@Autowired
-	OpenAiChatModel openAiChatModel;
+	OpenAiSdkChatModel openAiChatModel;
 
 	@Test
 	void whenContextIsProvided() {
--- a/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/rag/preretrieval/query/expansion/MultiQueryExpanderIT.java
+++ b/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/rag/preretrieval/query/expansion/MultiQueryExpanderIT.java
@@ -23,7 +23,7 @@ import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
 
 import org.springframework.ai.chat.client.ChatClient;
 import org.springframework.ai.integration.tests.TestApplication;
-import org.springframework.ai.openai.OpenAiChatModel;
+import org.springframework.ai.openaisdk.OpenAiSdkChatModel;
 import org.springframework.ai.rag.Query;
 import org.springframework.ai.rag.preretrieval.query.expansion.MultiQueryExpander;
 import org.springframework.ai.rag.preretrieval.query.expansion.QueryExpander;
@@ -42,7 +42,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 class MultiQueryExpanderIT {
 
 	@Autowired
-	OpenAiChatModel openAiChatModel;
+	OpenAiSdkChatModel openAiChatModel;
 
 	@Test
 	void whenExpanderWithDefaults() {
--- a/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/rag/preretrieval/query/transformation/CompressionQueryTransformerIT.java
+++ b/spring-ai-integration-tests/src/test/java/org/springframework/ai/integration/tests/rag/preretrieval/query/transformation/CompressionQueryTransformerIT.java
@@ -23,7 +23,7 @@ import org.springframework.ai.chat.client.ChatClient;
 import org.springframework.ai.chat.messages.AssistantMessage;
 import org.springframework.ai.chat.messages.UserMessage;
 import org.springframework.ai.integration.tests.TestApplication;
-import org.springframework.ai.openai.OpenAiChatModel;
+import org.springframework.ai.openaisdk.OpenAiSdkChatModel;
 import org.springframework.ai.rag.Query;
 import org.springframework.ai.rag.preretrieval.query.transformation.CompressionQueryTransformer;
 import org.springframework.ai.rag.preretrieval.query.transformation.QueryTransformer;
@@ -42,7 +42,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 class CompressionQueryTransformerIT {
 
 	@Autowired
-	OpenAiChatModel openAiChatModel;
+	OpenAiSdkChatModel openAiChatModel;
 
 	@Test
 	void whenTransformerWithDefaults() {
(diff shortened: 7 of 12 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClientUtils.java
```diff
@@ -105,57 +100,38 @@ static ChatClientRequest toChatClientRequest(DefaultChatClient.DefaultChatClient
...
-			if (processedChatOptions == null) {
-				processedChatOptions = new DefaultToolCallingChatOptions();
-			}
-			else if (processedChatOptions instanceof DefaultChatOptions defaultChatOptions) {
-				processedChatOptions = ModelOptionsUtils.copyToTarget(defaultChatOptions, ChatOptions.class,
-						DefaultToolCallingChatOptions.class);
-			}
+		ChatOptions.Builder<?> builder = inputRequest.getChatModel().getDefaultOptions().mutate();
+		if (inputRequest.getOptionsCustomizer() != null) {
+			builder = builder.combineWith(inputRequest.getOptionsCustomizer());
```
sobychacko, line 105: Right now this only checks `getOptionsCustomizer()`, but the deprecated `options(T)` still writes to `chatOptions` and nothing reads it back, correct?. Users can still call `.options(SomeOptions.builder().temperature(0.9).build()) `through `ChatClient,` which hits the deprecated path and stores into `chatOptions`. Even if we're pulling the deprecated method before GA, there's a window between M5 and removal where both APIs exist, but only one actually does anything. Feels like we should fall back to `chatOptions.mutate()` here if optionsCustomizer is null, so the old
path isn't quietly broken while it's still available, so that `chatOptions.mutate()` acts as the customizer? Maybe not a big deal, since we are going to remove them soon, but just wanted to bring it your attention. Thanks!

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -579,67 +575,13 @@ public ChatResponse internalCall(Prompt prompt, @Nullable ChatResponse previousC
...
-				if (!originalAnthropicOptions.getHttpHeaders().isEmpty()) {
-					requestOptions.setHttpHeaders(new HashMap<>(originalAnthropicOptions.getHttpHeaders()));
-				}
-				if (originalAnthropicOptions.getSkillContainer() != null) {
-					requestOptions.setSkillContainer(originalAnthropicOptions.getSkillContainer());
-				}
-			}
-		}
+		var requestOptions = (AnthropicChatOptions) prompt.getOptions();
+		requestOptions = requestOptions == null ? this.options : requestOptions;
```
sobychacko, line 602: I have a few questions around this - maybe I am missing some other code paths. But imagine a scenario where someone accesses the `ChatModel` directly without the `ChatClient`, as shown below.
```
AnthropicChatModel model = AnthropicChatModel.builder()
      ...
          .model("claude-sonnet-4-6")
          .maxTokens(4096)
          .temperature(0.7)
          .toolCallbacks(myDefaultTool)
          .build())
      .build();
```
Now, if I do something as below, won't it cause a CCE from line 579 above? (since we are strictly expecting `AnthropicChatOptions` and we are providing a `ChatOptions` (same is true for `ToolCallingChatOptions`, I guess)

```
ChatOptions options = ChatOptions.builder().temperature(0.9).build();
 model.call(new Prompt("hello", options));
```

Another potential issue:

Consider the following code - 

```
AnthropicChatOptions options = AnthropicChatOptions.builder()
      .temperature(0.9)
      .build();
  model.call(new Prompt("hello", options));
```
Before we were merging on top of the defaults (model - claude-sonnet-4-6, maxTokesn - 4096), and then overriding the temperature to 0.9. But with this change on line 579, I think we are replaing the defaults entirely - we may end up with model and maxTokens as `null`. Or maybe not - there might be some other code path i am missing. But can you look into that scenario to see if it is a valid one?

Same kind of concern with custom tool callbacks. Imagine this scenario:

```
// Add a one-off tool for this request, expecting default tools to still be there
  AnthropicChatOptions options = AnthropicChatOptions.builder()
      .toolCallbacks(myOneOffTool)
      .build();
  model.call(new Prompt("hello", options));
```

We were using `mergeToolCallbacks()` before to combine the default and custom toolcallbacks. But with this change, I can only see the new custom one and the default is gone since we are over-wrting. no?

I believe these concerns apply to all `buildRequestPrompt` methods in other models as well.

spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/DefaultChatClientTests.java
```diff
@@ -2286,9 +2292,10 @@ void whenToolCallbackProviderThenNotEagerlyEvaluated() {
 		verify(provider, never()).getToolCallbacks();
 	}
 
+	@Disabled("TODO: check this test does not make sense anymore")
```
sobychacko: Why are we disabling these tests? Were they testing behaviors we exposed to the users? I feel like we should get to the bottom of why those tests were broken due to the changes. If it is a legitimate reason, I would remove the tests and document the reasoning. What do you think?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClientUtils.java
```diff
@@ -105,57 +100,38 @@ static ChatClientRequest toChatClientRequest(DefaultChatClient.DefaultChatClient
...
 		}
 
-		if (processedChatOptions instanceof ToolCallingChatOptions toolCallingChatOptions) {
+		if (builder instanceof ToolCallingChatOptions.Builder<?> tbuilder) {
```
sobychacko, line 108: This assumes `getDefaultOptions().mutate()` always returns a `ToolCallingChatOptions.Builder`. That's true for all current models since they all implement `ToolCallingChatOptions,` but if someone has a custom `ChatModel` with plain `ChatOptions` defaults, any tools set via `ChatClient` would silently fall through. Might be worth a comment or a fallback, but not a big deal.

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaChatOptionsTests.java
```diff
@@ -50,6 +50,10 @@ protected Builder readyToBuildBuilder() {
 
 	@Test
 	void testBasicOptions() {
+		var b1 = OllamaChatOptions.builder().model("model").mainGPU(12);
+
+		var b = OllamaChatOptions.builder().mainGPU(12).model("model");
+
```
sobychacko: Looks like they are leftover vars?

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatModelIT.java
```diff
@@ -459,7 +450,7 @@ OllamaApi ollamaApi() {
 		OllamaChatModel ollamaChat(OllamaApi ollamaApi) {
 			return OllamaChatModel.builder()
 				.ollamaApi(ollamaApi)
-				.defaultOptions(OllamaChatOptions.builder().model(MODEL).temperature(0.9).build())
+				.defaultOptions(OllamaChatOptions.builder().model(MODEL).temperature(0.0).build())
```
sobychacko, line 453: Why did the `temperature` value change? Was there a particular reason?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClientUtils.java
ericbottard: That is true. I dumped this PR here as a stepping stone, knowing that we'd remove the `options(T)` soon, but maybe we can make sure it stays compatible (or assess the usage of the old method and remove it now if ready). Thanks for the catch!

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
ericbottard: > Now, if I do something as below, won't it cause a CCE 

Yes and this is now by design. If you use `call()` (ie model without ChatClient), you need to provide either `null` or options **of the correct concrete type** and "**full**". We'll make it clearer in the docs and release notes

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClientUtils.java
ericbottard: The equivalent was already true with the old approach, so I don't see what needs attention here?

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatModelIT.java
ericbottard: This is unrelated to the core of this PR, but was an attempt at having more predictable output from this test that tends to be flaky

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
sobychacko: Thanks @ericbottard. Aren't we limiting ourselves with this restriction imposed on users, though? Because they now have to think about it in such a way that they always have to make sure they are providing the concrete option type, and not the core abstraction. FWIW, if someone is using a chat model directly, it is reasonable for us to expect them to use the corresponding optons type, but our API's still allow one to use the core type, and that could run into an issue. If there is a compelling reason to do this, let's proceed. I see this as potentially an issue and want to flag it.

What about the merging issues? Do you see a problem there?

Thanks!

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClientUtils.java
sobychacko: Ya, not a big concern, but I thought this might be an opportunity for us to clean up.

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
ericbottard: Yes, this is a tradeoff we discussed at length with @sdeleuze and @tzolov. The way the types are currently designed, there is not much we can do to force users to use the correct type. But we think the overall benefits outweigh this small edge case.
Overall, I think people wanting to benefit from "the abstraction value" would better use ChatClient, esp if we move tool calling as an advisor, etc

sobychacko: Ok. That sounds good then. Although I don't think this is an edge case, since chat options as a core abstraction are one of the strengths of Spring AI, people may want to stick with that for portability reasons. But since you said the benefits outweigh this limitation, let it be.

sobychacko: Do you think the concerns I raised about merging are potential issues? Thanks!

spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/DefaultChatClientTests.java
ericbottard: Re-enabled in my latest push

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaChatOptionsTests.java
ericbottard: Fixed

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClientUtils.java
ericbottard: I don't really see what fallback we can use here. If tools are not supported b/c of the concrete type of the options, we simply don't deal with tools

sobychacko: Ok, let's not worry too much about it then, since we are not breaking any existing behavior. Sorry for the noise on that.

spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
```diff
@@ -1,6 +1,19 @@
 [[upgrade-notes]]
 = Upgrade Notes
 
+[[upgrading-to-2-0-0-M5]]
+== Upgrading to 2.0.0-M5
+
+=== ChatOptions Handling
+Clients can interact with ChatModels in two ways:
```
tzolov: "Users" instead of "Clients"

spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
```diff
@@ -1,6 +1,19 @@
...
+
+ * either directly by calling xref:api/chatmodel.adoc[ ChatModel.call(Prompt) / ChatModel.stream(Prompt)`],
+ * or by using the xref:api/chatclient.adoc[ChatClient] API.
+
+Starting with Spring AI 2.0.0.M5, when using the `ChatModel` approach, options set in the prompt need to be of the concrete type used by the concrete model (_e.g._ `AnthropicChatOptions` when using `AnthropicChatModel`) and need to be either "full" (_i.e._ all relevant fields set) or `null` (in which case the default options set in the model will be used).
```
tzolov, line 13: Perhaps we should explain how to merge the default options with new custom options manually before calling the chat model using the new customizer API?

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
sdeleuze: @ericbottard Just wondering if we could support the use case where not the right type of option is passed via a fallback on `.mutate()` + apply it on the default option type of the model. Do you see what I mean?

spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chatclient.adoc
```diff
@@ -645,6 +645,8 @@ java.util.function.Function<I, O> function)`
 
 * `advisors(Consumer<AdvisorSpec> advisorSpecConsumer)`
 
+image::chat-client-options-merging.png[align="center"]
```
tzolov, line 648: The chat-client-options-merging.png is incorrectly copied.

Also, the Chat Option Merging deserves a dedicated chapter (e.g. == ChatOptions Management) with explantation how the chat options lifecycle works at ChatClient level. E.g. the Default/Start-up vs run-time Options. The new ChatOptions.Builder/customizer mechanism. 
And we can add a warning that starting with M5+ the chat-options is not performed automatically at ChatModel level and provide a link to the work around showing how to do the options merging manually outside the ChatModel.

spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
```diff
@@ -1,6 +1,63 @@
...
+ChatOptions opts = AnthropicChatOptions.builder()
+    .maxTokens(100)
+    .temperature(0.7)
+    .disableParallelToolUse(true)
+    .build();
+String response = client.prompt("Tell me a joke")
+    .options(opts)
+    .call().content();
+
+// New
```
tzolov, line 31: IMO, we still should provide a option for the users to do manual merging before calling the ChatModel. 
E.g. if they didn't use the ChatClient they still can re-work their existing code doing the merging manually using the new ChatOption.Builder?

spring-ai-docs/src/main/antora/modules/ROOT/images/chat-client-options-merging.png
tzolov, line 1: this image is wrong

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
tzolov: @sobychacko wouldn't the new documentation how to do options merging manually (below) be enough to mitigate this?

models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/AnthropicChatModelTests.java
```diff
@@ -206,25 +206,6 @@ void getDefaultOptionsReturnsCopy() {
 		assertThat(defaultOptions1.getModel()).isEqualTo(defaultOptions2.getModel());
 	}
 
-	@Test
-	void buildRequestPromptMergesOptions() {
```
tzolov, line 210: Maybe this should test the manual merging mechanism instead?

models/spring-ai-mistral-ai/src/test/java/org/springframework/ai/mistralai/MistralAiChatCompletionRequestTests.java
```diff
@@ -89,40 +83,6 @@ void chatCompletionRequestWithOptionsTest() {
 		assertThat(request.stream()).isTrue();
 	}
 
-	@Test
-	void whenToolRuntimeOptionsThenMergeWithDefaults() {
```
tzolov, line 93: Perhaps convert this into manual chatoption merging test

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatRequestTests.java
```diff
@@ -53,39 +51,6 @@ class OllamaChatRequestTests {
 		.retryTemplate(RetryUtils.DEFAULT_RETRY_TEMPLATE)
 		.build();
 
-	@Test
-	void whenToolRuntimeOptionsThenMergeWithDefaults() {
```
tzolov, line 57: Perhaps convert this into manual chatoption merging test

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClientUtils.java
tzolov: I'm not sure i can find the temp backward compatability support for the deprecated request chatOptions?

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
```diff
@@ -208,8 +208,11 @@ interface ChatClientRequestSpec {
 
 		ChatClientRequestSpec messages(List<Message> messages);
 
+		@Deprecated
 		<T extends ChatOptions> ChatClientRequestSpec options(T options);
 
+		<B extends ChatOptions.Builder<?>> ChatClientRequestSpec options(B customizer);
```
ilayaperumalg, line 211: Since `options(B customizer)` accepts the builder, should we changes method to `optionsBuilder` instead of just `options` as it is confusing when we pass the builder to the options() method.

sdeleuze: Right now, I am in favor of keeping `options` name as it helps for migration (just remove `.build()` and your are done), communicate the right semantics and is consistent as the type will vary based on the context.

If we had to change, it would be more something like `optionsCustomizer` but that's pretty long and I am not convinced this is better.

PR #5726, opened 2026-03-31
Upgrade GH actions to latest versions

ericbottard:
Github runners will soon only support Node 24, so
actions need to be upgraded to versions that support it.


ilayaperumalg:
Rebased and merged via de8ad9c17

PR #5730, opened 2026-04-01
Add OpenAI SDK Moderation Model implementation

ilayaperumalg:
- Implement OpenAiSdkModerationModel and OpenAiSdkModerationOptions.
- Add auto-configuration for the moderation model.
- Add unit tests for the moderation model.

### DIFF
--- a/models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkModerationOptions.java
+++ b/models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkModerationOptions.java
@@ -21,8 +21,6 @@ import java.time.Duration;
 import java.util.Map;
 import java.util.Objects;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.openai.azure.AzureOpenAIServiceVersion;
 import com.openai.credential.Credential;
 import org.jspecify.annotations.Nullable;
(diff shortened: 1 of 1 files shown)


models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkModerationOptions.java
```diff
@@ -0,0 +1,209 @@
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class OpenAiSdkModerationOptions extends AbstractOpenAiSdkOptions implements ModerationOptions {
+
+	/**
+	 * Default moderation model.
+	 */
+	public static final String DEFAULT_MODERATION_MODEL = "omni-moderation-latest";
+
+	@JsonProperty("model")
+	private @Nullable String model;
```
sdeleuze, line 44: Could we set it to `DEFAULT_MODERATION_MODEL` by default and make it non-null?

ilayaperumalg: That's a good idea. Looks like this needs to be addressed across the OpenAI SDK models. Can we address this separately after this one and the transcription model PR are merged?

models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkModerationOptions.java
```diff
@@ -0,0 +1,285 @@
...
+import org.springframework.ai.moderation.ModerationOptions;
+
+/**
+ * OpenAI SDK Moderation Options.
+ *
+ * @author Ilayaperumal Gopinathan
+ * @author Ahmed Yousri
+ * @since 2.0.0
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
ericbottard: Seems like "merging" is implemented via https://github.com/spring-projects/spring-ai/pull/5730/changes#diff-f70df6089d6b9eff28d67795cee26977042a8489eebfb458a3fd33198b0702deR164

If so, this class should not be Jackson annotated.

The same comment applies to the other PR about OpenAI SDK

PR #5734, opened 2026-04-02
Add extraBody support for OpenAI SDK models

ilayaperumalg:
Fix extraBody serialization for OpenAI SDK and compatible proxies

- The `extra_body` parameter is designed to allow users to pass custom or proprietary parameters (like `top_k`, `num_ctx`, or `max_tokens`) to OpenAI-compatible endpoints without being blocked by the official SDK's strictly typed parameter list.
- OpenAI SDK Integration: `extraBody` properties are now properly intercepted and mapped directly to `additionalBodyProperties()`. This flattens the custom parameters into the root of the generated JSON request.
- Official OpenAI API Behavior: The literal `extra_body` wrapper is stripped out. If an unsupported parameter is provided, the official OpenAI API will correctly reject it as an "unknown parameter" (validated via `OpenAiSdkChatModelIT`).
- Compatible Providers (Ollama, DeepSeek, Groq, etc.): Custom parameters defined in `extraBody` are successfully flattened and transmitted to the proxy APIs, allowing provider-specific features to work seamlessly (validated via `max_tokens` truncation assertions across all proxy integration tests).

### DIFF
--- a/models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkChatOptions.java
+++ b/models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkChatOptions.java
@@ -591,7 +591,7 @@ public class OpenAiSdkChatOptions extends AbstractOpenAiSdkOptions
 	}
 
 	@JsonAnySetter
-	public void setExtraBodyProperty(String key, Object value) {
+	public void addExtraBodyProperty(String key, Object value) {
 		if (this.extraBody == null) {
 			this.extraBody = new HashMap<>();
 		}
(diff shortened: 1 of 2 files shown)


models/spring-ai-openai-sdk/src/main/java/org/springframework/ai/openaisdk/OpenAiSdkChatOptions.java
```diff
@@ -565,6 +581,23 @@ public void setServiceTier(@Nullable String serviceTier) {
...
+	public @Nullable Map<String, Object> getExtraBody() {
+		return this.extraBody;
+	}
+
+	public void setExtraBody(@Nullable Map<String, Object> extraBody) {
+		this.extraBody = extraBody;
+	}
+
+	@JsonAnySetter
+	public void setExtraBodyProperty(String key, Object value) {
```
sdeleuze: I know it was done like that in `OpenAiApi`, but can we rename this method to `addExtraBodyProperty` ?

Issue #5748, opened 2026-04-03
antlr4-runtime conflicts when used with other libraries that use Antlr4.


nielsbasjes:
Please do a quick search on GitHub issues first, there might be already a duplicate issue for the one you are about to create.
If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

**Bug description**
Several modules use ANTLR4 for parsing and thus the antlr4-runtime dependency.

In the generated ANTLR4 code there is a very strict check to ensure the runtime version and the code generator version are an exact match. If not then the system will not start.

The effect is that if this library is used in conjuction with another library that has or needs ANTLR4 the combined system will no longer work if not all components have the exact same version.

In my own software I have chosen to shade the antlr-runtime into the library to ensure no conflicts can occur.
See this as a reference: https://github.com/nielsbasjes/yauaa/blob/main/analyzer/pom.xml

**Environment**
Any

**Steps to reproduce**
Try to combine this library with a different library which has a slightly different version of the ANTLR4 runtime.

**Expected behavior**
No conflicts and working software.

**Minimal Complete Reproducible example**
This is a bit hard. I'll see what I can prepare.

nielsbasjes:
I found more problems related to Antlr dependencies ... fixing it.

nielsbasjes:
I provided a pull request with fixes for the issues I found: https://github.com/spring-projects/spring-ai/pull/5752

PR #5752, opened 2026-04-03
Fixing antlr dependencies

nielsbasjes:
Changes in this pull request:
- **Problem to fix**: The spring-ai-template-st module references the Token class from Antlr to access something in the template library. The template library uses Antlr 3 and the added dependency is the Antlr 4  runtime. **It is a miracle that actually worked.**
- **Solution**: Shade the already provided antlr-runtime (which is V3) and relocate it to a package that is chosen to avoid conflicts with any downstream consuming application.

- **Problem to fix**: https://github.com/spring-projects/spring-ai/issues/5748
When using Antlr in a project the antlr runtime needs to be an exact matched version of the plugin used to generate the code. If you have a downstream project that has a different dependency that also uses Antlr then there is a high probability there is a conflict in the used runtime and a least one of the components will not work.
- **Solution**: Shade the applicable antlr4-runtime and relocate it to a package that is chosen to avoid conflicts with any downstream consuming application. Also: The generated Anltr4 code was committed to the repo which is not needed at all.

- Chore: Updating to the latest Antlr4 version.

NOTE: I used this pattern to relocate the antlr runtime to. This can be changed to `anything` as long as the new package avoids conflicts with any downstream users.
`shaded.${project.groupId}.${project.artifactId}.library.org.antlr`


nielsbasjes:
Note: The ST4 must be shaded in also because that is the main library that needs the shaded antlr-runtime also.

tzolov:
@nielsbasjes , thanks for raising this issue and the related PR. 

We will triage it soon

nielsbasjes:
@tzolov I rebased and fixed the conflicts. Have you had the time to consider this? Is there anything you need me to do?

PR #5753, opened 2026-04-03
GH-5724: Add purgeEmbeddings method to ChromaVectorStore

### LINKED ISSUE #5724: [Feature Request] Add deleteAll() or clear() method to VectorStore interface
**Expected Behavior**
Suggest adding a default void deleteAll() method to the VectorStore interface to provide a unified collection clearing method for various vector store implementations.

**Current Behavior**
Currently, the VectorStore interface only provides the delete(List<String> idList) method.  There is no unified high-level clearing method, forcing developers to rely on specific implementations (such as directly operating ChromaApi), which breaks the encapsulation effect of the abstraction layer.

**Context**
I am a developer currently using Spring AI 1.1.2 and Chroma to develop an AI intent recognition platform. During integration testing and model performance evaluation, I frequently need to reset the vector data to ensure that each test run starts from a clean state.

**Willingness to contribute**
I very much hope to contribute code to Spring AI, and I would like to ask the maintainers for confirmation: Is it allowed to add deleteAll() in VectorStore and implement the ChromaVectorStore version? If this proposal is consistent, I will be very happy to complete this Pull Request after receiving your confirmation.

suwe12:
Fixes #5724

This pull request adds the `purgeEmbeddings()` method to the `ChromaVectorStore` class. 
This method allows users to delete all embeddings within a specific collection while keeping the collection structure intact.

Key changes:
* Implemented `purgeEmbeddings()` in `ChromaVectorStore.java` using batch deletion.
*Added integration test `purgeEmbeddingsCollection()` in `ChromaVectorStoreIT.java` to verify the functionality.
* Ensured all commits are signed-off per DCO requirements.

### DIFF
--- a/vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/chroma/vectorstore/ChromaVectorStore.java
+++ b/vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/chroma/vectorstore/ChromaVectorStore.java
@@ -63,6 +63,8 @@ import org.springframework.util.CollectionUtils;
  */
 public class ChromaVectorStore extends AbstractObservationVectorStore implements InitializingBean {
 
+	private static final int DEFAULT_PURGE_BATCH_SIZE = 1000;
+
 	private final ChromaApi chromaApi;
 
 	private final String tenantName;
@@ -173,6 +175,44 @@ public class ChromaVectorStore extends AbstractObservationVectorStore implements
 		Assert.notNull(idList, "Document id list must not be null");
 		this.chromaApi.deleteEmbeddings(this.tenantName, this.databaseName, this.requireCollectionId(),
 				new DeleteEmbeddingsRequest(idList));
+	}
+
+	/**
+	 * Deletes all embeddings from the current collection while keeping the collection
+	 * itself intact. Embeddings are deleted in batches by repeatedly fetching the first
+	 * page of ids until the collection is empty.
+	 * <p>
+	 * Note: This method does not support concurrency control. Concurrent writes or
+	 * deletions may result in incomplete deletions or duplicate deletion attempts.
+	 * @return the number of embeddings requested for deletion
+	 * @since 2.0.0-SNAPSHOT
+	 */
+	public int purgeEmbeddings() {
+		int deleteCount = 0;
+		String collectionId = this.requireCollectionId();
+		try {
+			while (true) {
+				var response = this.chromaApi.getEmbeddings(this.tenantName, this.databaseName, collectionId,
+						new ChromaApi.GetEmbeddingsRequest(List.of(), Map.of(), DEFAULT_PURGE_BATCH_SIZE, 0,
+								List.of()));
+
+				if (response == null || CollectionUtils.isEmpty(response.ids())) {
+					return deleteCount;
+				}
+				List<String> ids = response.ids();
+				this.chromaApi.deleteEmbeddings(this.tenantName, this.databaseName, collectionId,
+						new DeleteEmbeddingsRequest(ids));
+				deleteCount += ids.size();
+
+				if (ids.size() < DEFAULT_PURGE_BATCH_SIZE) {
+					return deleteCount;
+				}
+			}
+		}
+		catch (Exception e) {
+			logger.error("Purge failed: {}", e.getMessage());
+			throw new IllegalStateException("Failed to purge chroma collection", e);
+		}
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/chroma/vectorstore/ChromaVectorStore.java
```diff
@@ -175,6 +177,44 @@ public void doDelete(List<String> idList) {
...
+				this.chromaApi.deleteEmbeddings(this.tenantName, this.databaseName, collectionId,
+						new DeleteEmbeddingsRequest(ids));
+				deleteCount += ids.size();
+
+				if (ids.size() < DEFAULT_PURGE_BATCH_SIZE) {
+					return deleteCount;
+				}
+			}
+		}
+		catch (Exception e) {
```
AlecHFerguson, line 212: Generic try/catch is not recommended in Java. Are you sure you want to handle, say, a network exception the same way as a syntax exception? Also, why throw IllegalStateException? Do ALL instances of thrown `Exception` represent this `a method has been invoked at an illegal or inappropriate time`?
https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/IllegalStateException.html

vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/chroma/vectorstore/ChromaVectorStore.java
```diff
@@ -175,6 +177,44 @@ public void doDelete(List<String> idList) {
...
+	 * Deletes all embeddings from the current collection while keeping the collection
+	 * itself intact. Embeddings are deleted in batches by repeatedly fetching the first
+	 * page of ids until the collection is empty.
+	 * <p>
+	 * Note: This method does not support concurrency control. Concurrent writes or
+	 * deletions may result in incomplete deletions or duplicate deletion attempts.
+	 * @return the number of embeddings requested for deletion
+	 * @since 2.0.0-SNAPSHOT
+	 */
+	public int purgeEmbeddings() {
```
AlecHFerguson, line 190: Out of curiosity, how would this actually be called from a standard RAG application? Standard tutorials from Spring AI recommend dependency injection of `VectorStore` alone, with concrete implementation detected from configuration. If you only have `VectorStore`, would you need to call `getNativeClient()`? Or do you recommend creating a custom bean which instantiates `ChromaVectorStore` directly? How do you recommend using this method?

AlecHFerguson:
Nice work tackling this. I probably wouldn't add to Spring AI until `VectorStore` defines a method; usage without this is a bit cumbersome. I commented in #5724; I'm not sure a `deleteAll()` method is appropriate inside Spring specifically. I would certainly wait for community feedback.

Note that the same `while` loop can be implemented in your application for testing purposes. Or for VectoreStore more generally -- this worked for me with PGVector.

```java
@Service
public class VectorDbCleanupService {
    private final VectorStore vectorStore;

    public VectorDbCleanupService(VectorStore vectorStore) {
        this.vectorStore = vectorStore;
    }

    public void truncate() {
        var documentsFound = true;
        while (documentsFound) {
            final var searchRequest =
                    SearchRequest
                            .builder()
                            .query("abcdefghijklmnopqrstuvwxyz")
                            .similarityThreshold(SIMILARITY_THRESHOLD_ACCEPT_ALL)
                            .topK(1000)
                            .build();
            final var documentsToDelete = vectorStore.similaritySearch(searchRequest);
            documentsFound = !documentsToDelete.isEmpty();
            final var idsToDelete = documentsToDelete.stream().map(Document::getId).toList();
            vectorStore.delete(idsToDelete);
        }
    }
}
```

suwe12:
@AlecHFerguson Thanks for sharing your thoughts on #5724. I agree that a general API for this operation needs broader discussion.

I implemented this on `ChromaVectorStore` rather than `VectorStore` because `VectorStore` is a shared abstraction used by different vector-store implementations. Adding this operation there would first require clarifying what it should mean across providers—for example, whether it deletes only embeddings, clears an index, or drops and recreates a collection, as well as its concurrency and error behavior.

It would also require reviewing and potentially adapting every concrete vector-store implementation. I could not find existing Spring AI guidance for this operation, so I did not want this PR to decide its behavior unilaterally.

My intention with this PR is to provide a concrete, tested reference for that discussion. If the maintainers decide that this should become a general `VectorStore` capability, this implementation could help inform the broader design and the Chroma adaptation.

vector-stores/spring-ai-chroma-store/src/main/java/org/springframework/ai/chroma/vectorstore/ChromaVectorStore.java
suwe12: If the application explicitly uses ChromaDB, I think it can inject the auto-configured `ChromaVectorStore` directly. As I understand it, adding this method to `VectorStore` would require agreeing on common semantics and updating all implementations. I’ve explained my reasoning in more detail in [this comment](https://github.com/spring-projects/spring-ai/pull/5753#issuecomment-5691696466).
I’d welcome your thoughts on this approach.

Issue #1611, opened 2024-10-28
Health indicators for vector stores (and any other external APIs)


habuma:
Given that vector stores are databases that are external to the application, which could go down or otherwise become unavailable, and as such contribute to the health (or lack thereof) of a Spring AI application...

It would be great if Spring AI provided health indicators for the vector stores that contribute to Actuator's health endpoint.

One question concerning this is whether such health indicators would be needed for vector stores that are also databases for which Spring already has health indicators. For example, Postgres and Mongo probably are already covered by health indicators. But things like Chroma and Milvus probably aren't.

eddumelendez:
I started this some time ago and shared it with @markpollack and @tzolov at Spring IO but never raised the PR. Currently, it is missing some tests. I can start raising a draft PR to start discussing about it.

markpollack:
@eddumelendez just checking in after a long time on this - it is a decent suggestion, just low on the priority list now for the team.

celikfatih:
Hi, is anyone currently working on this issue? If not, I’d be happy to help and work on an initial implementation. Please let me know if that would be welcome.

DDINGJOO:
I would like to work on this issue.

If no one is actively implementing it right now, I plan to start with a small design proposal and then submit an initial PR with tests.

Initial scope proposal:
1. Introduce a common health-indicator pattern for vector stores.
2. Implement it for one or two stores first.
3. Extend to additional stores in follow-up PRs.

DDINGJOO:
Opened a PR with an initial implementation for vector store health indicators (Qdrant + Weaviate), including conditional auto-configuration and unit tests.

PR: https://github.com/spring-projects/spring-ai/pull/5553

I can extend the same pattern to additional vector stores in follow-up PRs.

suryateja-g13:
I've submitted an implementation for this in PR #6030.

PR #5764, opened 2026-04-07
Add health indicator for PgVector vector store

iaJingda:
## Summary

This PR adds a Spring Boot Actuator `HealthIndicator` for the PgVector vector store, addressing #1611.

## Why

Spring AI vector stores can become unavailable independently from the application (e.g., database connection lost, the `vector` PostgreSQL extension uninstalled). Production deployments need observability into the health of these external dependencies, following Spring Boot's standard `HealthIndicator` pattern.

## Changes

- Added `PgVectorStoreHealthIndicator` that verifies the `vector` PostgreSQL extension is installed and reachable
- Added `PgVectorStoreHealthAutoConfiguration` for conditional auto-configuration
- Registered the new auto-configuration in `AutoConfiguration.imports`
- Added optional `spring-boot-health` dependency in `pom.xml`
- Added unit tests covering 6 scenarios

## Behavior

The health indicator:
- Calls `SELECT extversion FROM pg_extension WHERE extname = 'vector'`
- Reports `UP` with details (`vectorExtensionVersion`, `database`) when the extension is installed
- Reports `DOWN` with `reason` when the extension is missing
- Reports `DOWN` when the query throws an exception
- Can be disabled via `management.health.pgvector.enabled=false`
- Backs off when the user provides a custom `pgvectorHealthIndicator` bean

## Tests

Unit tests cover:
- Health indicator registered when `PgVectorStore` bean is present
- `UP` status when vector extension is installed (with version details)
- `DOWN` status when vector extension is missing
- `DOWN` status on JDBC exception
- Not registered when `management.health.pgvector.enabled=false`
- Backs off when a custom `pgvectorHealthIndicator` bean is provided

All 6 tests pass locally.

## Notes

This is intentionally a focused, minimal-scope PR for a single vector store. If the approach is approved, I'm happy to follow up with similar health indicators for other vector stores (Redis, Elasticsearch, etc.) in separate PRs.

Related to the earlier work in #5553 (closed), which proposed a similar pattern for Qdrant and Weaviate.

### DIFF
--- a/auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-pgvector/pom.xml
+++ b/auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-pgvector/pom.xml
@@ -96,5 +96,11 @@
 			<version>${project.parent.version}</version>
 			<scope>test</scope>
 		</dependency>
+
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-health</artifactId>
+			<optional>true</optional>
+		</dependency>
 	</dependencies>
 </project>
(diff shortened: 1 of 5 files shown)


iaJingda:
Friendly ping — I understand merging a health indicator for just one vector store might create an inconsistency with the other stores. If the maintainers are open to it, I'd be happy to follow up with PRs for Redis, Elasticsearch, Milvus, Qdrant, and Chroma vector stores, reusing the same pattern established in this PR. That way we can address all vector stores as a series of focused, minimal-scope PRs rather than blocking on a single large change. Just let me know if this approach works for the project's roadmap.

PR #5770, opened 2026-04-07
fix: parse large integer constants as Long in FilterExpressionTextParser

### LINKED ISSUE #4705: FilterExpressionTextParser treats all number literals as Integer, failing on Long values

anuragg-saxenaa:
## Summary

Fixes #4705.

`FilterExpressionTextParser.visitIntegerConstant` used `Integer.valueOf()` which throws `NumberFormatException` for values exceeding `Integer.MAX_VALUE` (2147483647). This causes filter expressions like `id == 9223372036854775807` to fail with a parse error even though the value fits in a `long`.

**Change:** Fall back to `Long.parseLong()` when the value is out of Integer range.

```java
// before

// after
try {
} catch (NumberFormatException e) {
}
```

The existing `L`-suffixed long syntax (`id == 9223372036854775807L`) is unaffected — it routes through `visitLongConstant`. Values within Integer range continue to produce `Integer` values (no behavioral change for existing usage).

## Testing

Added `testLargeIntegerFallsBackToLong()` to `FilterExpressionTextParserTests` covering:
- `Long.MAX_VALUE` without `L` suffix → parsed as `Long`
- Any value > `Integer.MAX_VALUE` → parsed as `Long`  
- Values within Integer range → still parsed as `Integer`

All existing tests pass.

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
@@ -230,7 +230,13 @@ public class FilterExpressionTextParser {
 
 		@Override
 		public Filter.Operand visitIntegerConstant(FiltersParser.IntegerConstantContext ctx) {
-			return new Filter.Value(Integer.valueOf(ctx.getText()));
+			String text = ctx.getText();
+			try {
+				return new Filter.Value(Integer.parseInt(text));
+			}
+			catch (NumberFormatException e) {
+				return new Filter.Value(Long.parseLong(text));
+			}
 		}
 
 		@Override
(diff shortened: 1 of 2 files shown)


ericbottard:
Hi @anuragg-saxenaa,

thanks for your contribution. 

Sadly, I will close this PR as a WONTFIX as we don't want to magically convert ints to longs. When the user is entering a literal, they do know whether they're dealing with an int or a long. Just use the `L` syntax.

Issue #4691, opened 2025-10-21
Do not swallow unhandled exceptions


nkavian:
**Bug description & Expected Behavior**

When using `@MCPTool` on a method that throws an exception, the exception is swallowed by the Spring AI library and after it is output to the caller.

It's expected that the unhandled exception handler provided by the Bugsnag library should receive it and perform its own logic. Developers will not have visibility into the errors that are happening. Local logs and toolings such as Bugsnag are being skipped.

**Steps to reproduce**

1. Create a REST controller with a callable REST endpoint.
2. Annotate the endpoint with `@MPCTool`.
3. Call the endpoint using Postman and notice that Bugsnag and other tooling can execute.
4. Call the MCP method and notice that the MCP caller gets an error but that Bugsnag will not be triggered.

**Minimal Complete Reproducible example**

```
    @RequireAuthentication
    @GetMapping("{currency}")
    @Operation(summary = GET_BALANCE_TITLE, description = GET_BALANCE_DESCRIPTION)
    @McpTool(title = GET_BALANCE_TITLE, description = GET_BALANCE_DESCRIPTION, generateOutputSchema = true)
    public BalancePayload getBalance(@PathVariable final Currency currency) {
        throw new RuntimeException();
    }
```

Issue #4488, opened 2025-09-27
Ability to react to exceptions thrown by @McpTool methods


rnett:
**Expected Behavior**

There is a way for me to react to `@McpTool` methods throwing exceptions, e.g. to log them.

**Current Behavior**

There is no way to react to exceptions, they are turned into an MCP error response (great!) and then discard, losing all stacktrace and cause information (not great!).

**Context**

I'm trying to diagnose errors in an MCP tool and not having the errors logged makes it quite painful.

rnett:
Part of the reason for this is that only exceptions that are of the class returned by `AbstractMcpToolProvider.doGetToolCallException` are turned into responses, but it returns `java.lang.Exception`, so everything is.

PR #5772, opened 2026-04-07
Add pluggable exception handler for MCP tool method invocations

eocantu:
Introduces `McpToolCallExceptionHandler`, an interface for customizing how exceptions thrown during MCP tool method invocations are converted into `CallToolResult` error responses. The MCP annotation framework silently swallows all exceptions from tool method calls and returns a generic error result. There's no hook to intercept those exceptions to customize or control the error message format.

This feature should help with:
* https://github.com/spring-projects/spring-ai/issues/4488
* https://github.com/spring-projects/spring-ai/issues/4691

**Summary of the changes:**
* `McpToolCallExceptionHandler` - New functional interface with a `defaultHandler()` factory method
* `AbstractMcpToolProvider.setExceptionHandler()` to configure the handler at the provider level
* `SyncMcpAnnotationProviders` / `AsyncMcpAnnotationProviders` utility overloads accepting a handler
* Auto-configuration registers a `@ConditionalOnMissingBean` `McpToolCallExceptionHandler` bean for Spring Boot override support

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
@@ -54,10 +54,15 @@ public abstract class AbstractMcpToolMethodCallback<T, RC extends McpRequestCont
 
 	protected final ReturnMode returnMode;
 
+	protected final String mcpToolName;
+
 	protected AbstractMcpToolMethodCallback(ReturnMode returnMode, Method toolMethod, Object toolObject) {
 		this.toolMethod = toolMethod;
 		this.toolObject = toolObject;
 		this.returnMode = returnMode;
+		McpTool annotation = toolMethod.getAnnotation(McpTool.class);
+		this.mcpToolName = (annotation != null && annotation.name() != null && !annotation.name().isEmpty())
+				? annotation.name() : toolMethod.getName();
 	}
 
 	/**
@@ -78,7 +83,14 @@ public abstract class AbstractMcpToolMethodCallback<T, RC extends McpRequestCont
 			throw new RuntimeException("Failed to access tool method", ex);
 		}
 		catch (InvocationTargetException ex) {
-			throw new RuntimeException("Error invoking method: " + this.toolMethod.getName(), ex.getCause());
+			Throwable cause = ex.getCause();
+			if (cause instanceof RuntimeException re) {
+				throw re;
+			}
+			if (cause instanceof Error error) {
+				throw error;
+			}
+			throw new RuntimeException(cause);
 		}
 		return result;
 	}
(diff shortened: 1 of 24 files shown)


mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
```diff
@@ -78,7 +83,14 @@ protected Object callMethod(Object[] methodArguments) {
...
 		catch (InvocationTargetException ex) {
-			throw new RuntimeException("Error invoking method: " + this.toolMethod.getName(), ex.getCause());
+			Throwable cause = ex.getCause();
+			if (cause instanceof RuntimeException re) {
+				throw re;
+			}
+			if (cause instanceof Error error) {
+				throw error;
+			}
+			throw new RuntimeException(cause);
```
eocantu, line 93: This `callMethod` always re-wrapped exceptions in a `RuntimeException("Error invoking method...)`. This destroyed the original exception type before it reached the `toolCallExceptionClass` matching logic, silently breaking the feature for any specific exception subclass.

Additionally, error messages used Java internals (the Java method name), rather than the MCP tool name the client actually knows.

This change here necessitated some updates to tests to match the update to the error string.

mcp/mcp-annotations/src/test/java/org/springframework/ai/mcp/annotation/method/tool/SyncMcpToolMethodCallbackExceptionHandlingTests.java
```diff
@@ -97,10 +97,10 @@ public void testNonMatchingExceptionClass_ThrowsException() throws Exception {
 		McpSyncServerExchange exchange = mock(McpSyncServerExchange.class);
 		CallToolRequest request = new CallToolRequest("runtime-exception-tool", Map.of("input", "test"));
 
-		// The RuntimeException from callMethod should NOT be caught (not an
-		// IllegalArgumentException)
+		// The RuntimeException from the tool method should NOT be caught (not an
+		// IllegalArgumentException) and propagates with its original message
 		assertThatThrownBy(() -> callback.apply(exchange, request)).isInstanceOf(RuntimeException.class)
-			.hasMessageContaining("Error invoking method");
+			.hasMessageContaining("Runtime error: test");
```
eocantu, line 103: Previously `callMethod` always re-threw as new `RuntimeException("Error invoking method: ")`. The original exception now propagates unchanged, so the assertion must match its actual message.

mcp/mcp-annotations/src/test/java/org/springframework/ai/mcp/annotation/method/tool/SyncMcpToolMethodCallbackExceptionHandlingTests.java
```diff
@@ -138,9 +138,10 @@ public void testCheckedExceptionHandling_WithSpecificClass() throws Exception {
 		McpSyncServerExchange exchange = mock(McpSyncServerExchange.class);
 		CallToolRequest request = new CallToolRequest("checked-exception-tool", Map.of("input", "test"));
 
-		// The RuntimeException wrapper should NOT be caught
+		// The checked exception wrapped in RuntimeException should NOT be caught and
+		// propagates with the original cause accessible
 		assertThatThrownBy(() -> callback.apply(exchange, request)).isInstanceOf(RuntimeException.class)
-			.hasMessageContaining("Error invoking method")
+			.hasMessageContaining("Business error: test")
```
eocantu, line 144: Same root change: checked exceptions are now re-wrapped as new RuntimeException(cause) (no custom message), so the wrapper's message becomes cause.toString() which includes "Business error: test".

eocantu:
Hi @ericbottard ! This error handler would be really helpful to us. Do you think we could get some eyes on this PR? Any thoughts?

eocantu:
Hi @tzolov !

We have a downstream SDK that needs structured error classification from `@McpTool` methods, and this PR adds a `McpToolCallExceptionHandler` extension point to enable that, rather than the current hardcoded conversion behavior. Happy to address any feedback or adjust the approach if that helps the review. Thanks for your time!

PR #5783, opened 2026-04-09
Fix Streamable HTTP WebClient transport to accept SSE frames without `event` field

### LINKED ISSUE #5780: WebClientStreamableHttpTransport drops valid SSE frames when `event:` is omitted (should default to `message`)

## Description

```text
````

## Current Behavior

```java
    // parse JSON-RPC payload
}
else {
    // ignore frame
}
```

## Expected Behavior

* `event == null`
* `event == ""`
* `event == "message"`

## Why this matters

In practice, this can cause:

* request timeouts such as:

## Steps to Reproduce

```text
```

## Actual Behavior

## Environment

* Spring AI: `2.0.0-M3`
* Spring Boot: `4.0.4`
* MCP Java SDK: `1.1.0`
* Java: `25.0.2`
* Reactor Netty: `1.3.4`
* OS: macOS

## Related Issues

## Additional Context

ZhangDT-sky:
## Summary

This PR fixes SSE parsing in `WebClientStreamableHttpTransport` for Streamable HTTP.

Per SSE semantics, when `event:` is omitted, the event type defaults to `message`.  
Previously, the transport only accepted frames where `event == "message"` explicitly, so valid `data:`-only frames could be dropped.

This could cause JSON-RPC responses (including `initialize`) to be ignored and lead to startup/timeout failures.

## Root Cause

`WebClientStreamableHttpTransport.parse(...)` only parsed SSE frames when:

- `event.event().equals("message")`

If `event.event()` was `null` (or empty), the frame was ignored.

## Changes

- Updated message-event detection in `WebClientStreamableHttpTransport` to treat all of the following as message events:
1. `event == null`
2. `event == ""`
3. `event == "message"`

- Added integration test:
`WebClientStreamableHttpTransportSseParsingIT#shouldParseSseEventWithoutEventFieldAsMessage`

The test serves `text/event-stream` with only `data:` (no `event:`) and verifies the JSON-RPC response is parsed and delivered.

## Why This Is Safe

- Scope is limited to SSE event-type filtering in Streamable HTTP WebClient transport.
- Existing explicit `event: message` behavior is unchanged.
- Fix aligns behavior with standard SSE semantics and improves interoperability with spec-compliant servers.

## Verification

- Added dedicated regression test for `data:`-only SSE frames.
- Local targeted test passed for the new scenario.

Fixes #5780

### DIFF
--- a/mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransport.java
+++ b/mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransport.java
@@ -503,16 +503,21 @@ public final class WebClientStreamableHttpTransport implements McpClientTranspor
 
 	private Tuple2<Optional<String>, Iterable<McpSchema.JSONRPCMessage>> parse(ServerSentEvent<String> event) {
 		if (isMessageEvent(event.event())) {
+			String data = event.data();
+			if (data == null || data.isEmpty()) {
+				logger.debug("Ignoring SSE message event with empty data: {}", event);
+				return Tuples.of(Optional.empty(), List.of());
+			}
 			try {
 				// We don't support batching ATM and probably won't since the next version
 				// considers removing it.
-				McpSchema.JSONRPCMessage message = McpSchema.deserializeJsonRpcMessage(this.jsonMapper, event.data());
+				McpSchema.JSONRPCMessage message = McpSchema.deserializeJsonRpcMessage(this.jsonMapper, data);
 				String eventId = event.id();
 				Optional<String> idOpt = (eventId != null) ? Optional.of(eventId) : Optional.empty();
 				return Tuples.of(idOpt, List.of(message));
 			}
 			catch (IOException ioException) {
-				throw new McpTransportException("Error parsing JSON-RPC message: " + event.data(), ioException);
+				throw new McpTransportException("Error parsing JSON-RPC message: " + data, ioException);
 			}
 		}
 		else {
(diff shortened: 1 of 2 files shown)


Kehrlann:
For reference, here's the Java SDK equivalent https://github.com/modelcontextprotocol/java-sdk/pull/913

mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransport.java
```diff
@@ -502,17 +502,22 @@ public <T> T unmarshalFrom(Object data, TypeRef<T> typeRef) {
 	}
 
 	private Tuple2<Optional<String>, Iterable<McpSchema.JSONRPCMessage>> parse(ServerSentEvent<String> event) {
-		if (MESSAGE_EVENT_TYPE.equals(event.event())) {
+		if (isMessageEvent(event.event())) {
+			String data = event.data();
+			if (data == null || data.isEmpty()) {
+				logger.debug("Ignoring SSE message event with empty data: {}", event);
+				return Tuples.of(Optional.empty(), List.of());
+			}
```
Kehrlann, line 510: This is untested; what's the use-case?

mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransport.java
```diff
@@ -521,6 +526,11 @@ private Tuple2<Optional<String>, Iterable<McpSchema.JSONRPCMessage>> parse(Serve
 		}
 	}
 
+	private static boolean isMessageEvent(@Nullable String eventType) {
+		// Per SSE semantics, missing/blank event type defaults to "message".
+		return eventType == null || eventType.isEmpty() || MESSAGE_EVENT_TYPE.equals(eventType);
```
Kehrlann, line 531:
```suggestion
		return !StringUtils.hasLength(eventType) || MESSAGE_EVENT_TYPE.equals(eventType);
```

Kehrlann (commented):
Thank you for your contribution.

I have one open question before we merge this.

Kehrlann:
Merged as part of db7544a. Thank you for your contribution!

PR #913, opened 2024-06-20
Add customizable logger advisor

tzolov:
Resolves #912

PR #5788, opened 2026-04-10
fix: treat omitted SSE event type as default 'message' in WebClientStreamableHttpTransport

### LINKED ISSUE #5780: WebClientStreamableHttpTransport drops valid SSE frames when `event:` is omitted (should default to `message`)

anuragg-saxenaa:
## Problem

`WebClientStreamableHttpTransport.parse()` only processes SSE frames where `event.event()` equals `"message"`. Per the [SSE specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), omitting the `event:` field means the event type defaults to `"message"`. When Spring/WebFlux decodes such a frame, `ServerSentEvent.event()` returns `null` — so the frame was silently dropped.

This caused MCP clients to miss valid JSON-RPC responses from servers that emit bare `data:`-only SSE frames, resulting in initialization timeouts like:
- `Client failed to initialize by explicit API call`
- `Did not observe any item or terminal signal within ...`

This is the Spring AI counterpart of `modelcontextprotocol/java-sdk#885` / PR `#913`, which fixed the `HttpClient` variant but explicitly noted the `WebClient` variant needed a separate fix here.

## Fix

Accept all three equivalent event types for SSE message frames:
- `event == null` (field omitted)
- `event == ""` (field present but empty)
- `event == "message"` (explicit)

```java
String eventType = event.event();
if (eventType == null || eventType.isEmpty() || MESSAGE_EVENT_TYPE.equals(eventType)) {
```

## Testing

Added `testSseFrameWithoutEventTypeIsAccepted()` to `WebClientStreamableHttpTransportErrorHandlingIT` — spins up a local HTTP server that returns an SSE stream with only a `data:` line (no `event:` line) and verifies the message handler receives the parsed JSON-RPC message.

Closes #5780

### DIFF
--- a/mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransport.java
+++ b/mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransport.java
@@ -502,7 +502,8 @@ public final class WebClientStreamableHttpTransport implements McpClientTranspor
 	}
 
 	private Tuple2<Optional<String>, Iterable<McpSchema.JSONRPCMessage>> parse(ServerSentEvent<String> event) {
-		if (MESSAGE_EVENT_TYPE.equals(event.event())) {
+		String eventType = event.event();
+		if (eventType == null || eventType.isEmpty() || MESSAGE_EVENT_TYPE.equals(eventType)) {
 			try {
 				// We don't support batching ATM and probably won't since the next version
 				// considers removing it.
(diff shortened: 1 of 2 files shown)


redinside-dev:
Hi maintainers 👋 This PR has been waiting for review for 13 days. Happy to make any adjustments — just let me know! Targeting Sprint 1 close by Apr 25.

anuragg-saxenaa:

PR #5789, opened 2026-04-10
Remove ModelOptionsUtils.merge calls

ericbottard:
This PR removes remaining usages of MOU.merge() and removes Jackson annotations from options classes.


sdeleuze (approved):

sdeleuze:
@ericbottard LGTM, but I don't have the time to test it locally or test it with Ollama ITs, so I will let you merge it in case more check are needed.

Issue #5754, opened 2026-04-03
ToolCallAdvisor causes infinite loop when a tool returns long descriptive text on invalid input (empty string parameter)


visoeclipse:
When a tool is called with an empty string for a required parameter (e.g., `inputData = ""`), and the tool implementation returns a long descriptive “task prompt” instead of a concise error message, the `ToolCallAdvisor` enters an infinite loop. The model keeps calling the same tool with the same empty argument, consuming excessive tokens and never terminating.

**Expected behavior**  
The `ToolCallAdvisor` should detect repeated identical tool calls (same name + same arguments) and break the loop after a configurable number of attempts, or at least allow the model to receive a clear error that stops the cycle.

**Actual behavior**  
The advisor continues indefinitely, because the tool’s output (a long text that looks like a valid response) is interpreted as a normal tool result, and the model decides to call the same tool again with the same empty input.

**Steps to Reproduce**  
1. Define a tool that requires a non‑empty `inputData` parameter:
   ```java
   @Tool(description = "Analyzes data")
   public String analyze(@ToolParam(required = true) String inputData) {
       if (inputData == null || inputData.trim().isEmpty()) {
           // ❌ Returning a long descriptive text
           return "## Role\nYou are an expert...\n## Task\nPlease provide data...";
       }
       return "analysis result";
   }
   ```
2. Configure `ChatClient` with `ToolCallAdvisor`:
   ```java
   var advisor = ToolCallAdvisor.builder()
       .toolCallingManager(toolCallingManager)
       .build();
   var chatClient = ChatClient.builder(chatModel)
       .defaultAdvisors(advisor)
       .defaultTools(analyzeTool)
       .build();
   ```
3. Ask a question that leads the model to call the tool with an empty `inputData` (e.g., the model hallucinates the required data).
4. Observe infinite tool‑call loop in logs.

**Environment**  
- Spring AI version: `2.0.0-M2`
- Java version: `17`
- LLM provider: Doubao / OpenAI (any)
- Spring Boot version: `3.2.x`

**Logs (excerpt)**  
```
2026-04-03 17:34:52.476 ... "tool_calls":[{"function":{"arguments":" {\"inputData\": \"\", ...", "name":"audioChronicleInsight"}}]
2026-04-03 17:34:52.600 ... Successful execution of tool: audioChronicleInsight
... (repeats many times with same empty inputData)
```

**Possible workarounds**  
[...]

Issue #4464, opened 2025-09-24
ToolCalling loop with tool_choice: function


TLoud:
Setting an OpenAi's tool_choice option to "function" triggers an infinite loop of llm trying to call this function. 
Looks like when spring-ai calls llm with a tool execution result it copies chatOptions with tool_choice set to function, which forces model to call the same tool again. Ive tried to reset the tool_choice option in debugger before options are copied for sending a tool call result (at the bottom of org.springframework.ai.openai.OpenAiChatModel#createRequest) - looks like it helped llm to exit loop. But i didnt find an intended way to get around this problem. So maybe im doing something wrong.

Maybe disbling internal toolExecution will help - so i would be able to create OpenAiChatOptions with tool_choice - auto while construncting a message with tool execution result. But it would be nice to continue use internalToolExecution.

I've also encontered similar behaviour with DeepSeek's toolChoice. Didn't try to reset the tool_choice though.

public class ToolChoiceTest {

	public void test() {
		try {
			String openaiApiKey = System.getenv("OPENAI_API_KEY");
			var openAiApi = OpenAiApi.builder().apiKey(openaiApiKey).build();
			Object toolChoice = OpenAiApi.ChatCompletionRequest.ToolChoiceBuilder.FUNCTION("getMarineYetiDescription");
			var openAiChatOptions = OpenAiChatOptions.builder().model("gpt-5").toolChoice(toolChoice).build();
			var chatModel = OpenAiChatModel.builder().openAiApi(openAiApi).defaultOptions(openAiChatOptions).build();
			var chatClient = ChatClient.builder(chatModel).defaultTools(this).build();
			var prompt = new Prompt("What is the color of marine yeti?");
			String response = chatClient.prompt(prompt).call().content();
			System.out.println(response);
		} catch (Exception exception) {
			System.err.println(exception);
		}
	}

	@Tool(description = "Get description of marine yeti")
	public String getMarineYetiDescription() {
		return "Marine yeti is orange";
	}

	public static void main(String[] args) {
		new ToolChoiceTest().test();
	}

}

TLoud:
Looking at OpenAI's Agents SDK. They are using a reset_tool_choice param with true default value

<img width="778" height="185" alt="Image" src="https://github.com/user-attachments/assets/78b96ca4-4aaa-4aa4-860f-4f5baccf1cc0" />

https://openai.github.io/openai-agents-python/ref/agent/#agents.agent.Agent

PR #5790, opened 2026-04-10
Fix silent null-arg tool dispatch causing runaway tool-call loops

laran:
Related: https://github.com/spring-projects/spring-ai/issues/5754
Related: https://github.com/spring-projects/spring-ai/issues/3333
Related: https://github.com/spring-projects/spring-ai/issues/2383
Related: https://github.com/spring-projects/spring-ai/issues/4464
Related: https://github.com/spring-projects/spring-ai/issues/4617

The "Handle the possible null parameter situation in streaming mode" logic added in b059cdf9 silently replaced null or empty tool-call arguments with "{}". When a tool has required parameters, the downstream MethodToolCallback then called the method with null for every required argument and silently returned whatever the tool produced. Many tool implementations return a valid-looking but empty result in that case, which the model interprets as a transient failure and retries — often with the identical call. Combined with the absence of any iteration limit on Spring AI's tool-call recursion (#3333), this can produce multi-million-token runaway loops in a single turn.

Fix:

1. DefaultToolCallingManager: when tool arguments are null or empty, raise a ToolExecutionException and route it through the standard ToolExecutionExceptionProcessor so the resulting error becomes a proper tool response. The model can then adjust its approach rather than retry blindly. Well-formed tool calls in the same batch still execute normally.

2. MethodToolCallback.buildMethodArguments: when a required parameter (default: @ToolParam.required = true) is missing from the tool input map, throw a ToolExecutionException with a clear "Missing required parameter" message. Previously, buildTypedArgument silently returned null for missing values, allowing method invocation to proceed with null arguments. Optional parameters marked with @ToolParam(required = false) still pass null through unchanged, and zero-parameter tools are unaffected.

3. Sync/AsyncMcpToolCallback: apply the same fix as (1) for MCP callbacks that had the identical silent-"{}" fallback.

Tests:

- DefaultToolCallingManagerTest: the two tests that previously asserted the buggy behavior (shouldHandleNullArgumentsInStreamMode, shouldHandleEmptyArgumentsInStreamMode) are rewritten to assert that the tool callback is NOT invoked and that the conversation history contains a tool response with the error from the exception processor. A new test (shouldExecuteValidToolsWhileReturningErrorForMalformedTool) verifies that a batch containing a malformed call and a valid call processes both independently.

- MethodToolCallbackExceptionHandlingTest: new tests verify that (a) missing required parameters throw ToolExecutionException and the underlying method is never invoked, (b) optional parameters are still allowed to be null, and (c) zero-param tools remain callable with "{}".

- SyncMcpToolCallbackTests / AsyncMcpToolCallbackTest: rewritten null/empty input tests to assert the new error path and verify that the MCP client is never invoked.

- DefaultToolCallingManagerTests#whenMixedMethodToolCallsInChatResponse ThenExecute was implicitly relying on the silent-null behavior by calling TestGenericClass.call(String) with an empty args map. Updated to supply a value for the required parameter.

Loop safety of the new tests: every new test invokes its callback exactly once and asserts on the single result. There is no recursion, retry loop, or chat model involved, so the tests cannot themselves reproduce the runaway loop they guard against.

Thank you upfront!

### DIFF
--- a/mcp/common/src/main/java/org/springframework/ai/mcp/AsyncMcpToolCallback.java
+++ b/mcp/common/src/main/java/org/springframework/ai/mcp/AsyncMcpToolCallback.java
@@ -107,11 +107,19 @@ public class AsyncMcpToolCallback implements ToolCallback {
 	@Override
 	public String call(String toolCallInput, @Nullable ToolContext toolContext) {
 
-		// Handle the possible null parameter situation in streaming mode.
+		// Reject null/empty tool call input. This typically indicates a streaming tool
+		// call aggregation failure upstream. Previously, it was silently replaced with
+		// "{}" which let the MCP call proceed with no arguments, producing silent
+		// failures that caused the model to retry indefinitely. Throwing a
+		// ToolExecutionException ensures the standard error processor surfaces a clear
+		// message back to the model.
 		if (!StringUtils.hasText(toolCallInput)) {
-			logger.warn("Tool call arguments are null or empty for MCP tool: {}. Using empty JSON object as default.",
+			logger.warn("Tool call arguments are null or empty for MCP tool: {}. Surfacing as a tool error response.",
 					this.tool.name());
-			toolCallInput = "{}";
+			throw new ToolExecutionException(this.getToolDefinition(),
+					new IllegalArgumentException("Tool call arguments were null or empty; the MCP tool '"
+							+ this.tool.name() + "' was not invoked. This usually indicates a malformed or "
+							+ "incomplete tool invocation from the model."));
 		}
 
 		Map<String, Object> arguments = ModelOptionsUtils.jsonToMap(toolCallInput);
--- a/mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallback.java
+++ b/mcp/common/src/main/java/org/springframework/ai/mcp/SyncMcpToolCallback.java
@@ -108,11 +108,19 @@ public class SyncMcpToolCallback implements ToolCallback {
 	@Override
 	public String call(String toolCallInput, @Nullable ToolContext toolContext) {
 
-		// Handle the possible null parameter situation in streaming mode.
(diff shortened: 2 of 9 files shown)


0xbrainkid:
The fix is correct and well-scoped — routing null/empty args through `ToolExecutionException` rather than silently passing `{}` closes both the runaway loop risk and the silent null-invocation behavior.

One trust dimension worth adding to the exception handling: **agent identity context on the exception.** When a tool execution fails due to null/missing arguments, the exception currently contains the tool name and error type. In multi-agent Spring AI deployments where tool calls originate from different agents, the exception should also carry the calling agent identity — this is essential for diagnosing whether null-arg failures are concentrated in specific agents (suggesting a model-side bug or prompt issue) vs. distributed across all callers (suggesting a tool registration problem).

A minimal addition to `ToolExecutionException`:

```java
public class ToolExecutionException extends RuntimeException {
    private final String toolName;
    private final String callerAgentId;  // which agent made this call
    // ...
}
```

This also matters for behavioral trust: an agent that consistently produces null-arg tool calls has a measurable behavioral signature — high null-arg error rate across sessions is an anomaly signal. If `ToolExecutionException` carries caller identity, it can feed directly into observability tools that track per-agent tool call quality, without requiring a separate identity propagation layer.

laran:
Good callout. I updated to include a callerAgentId.

laran:
@tzolov I'd like to encourage this fix be prioritized. I encountered this issue when running locally, using Opus 4.6, in an application that uses tools heavily. Because of the bug this PR fixes I ended up experiencing an infinite loop when calling the Anthropic API that cost me $125 in input tokens alone in a matter of 10 minutes. I've added several safeguards to protect myself since. But ultimately, fixing this bug is what's needed. Until it's fixed I'm just running a locally patched version.

Please prioritize this ASAP. The bug has the potential to cause catastrophic API cost.

PR #5807, opened 2026-04-16
Fix #5470: exclude duplicate query from history in CompressionQueryTransformer

### LINKED ISSUE #5470: CompressionQueryTransformer is broken
**Bug description**
`CompressionQueryTransformer` adds the current user query in the history section (and of course in the follow-up query section) of its `promptTemplate` - which does not make much sense.

**Environment**
spring-ai 1.1.2 (same behaviour in v2.0.0-M2)

**Steps to reproduce**
see attached unit test `CompressionQueryTransformerTests.java`

**Expected behavior**
`CompressionQueryTransformer` should not add the current user query in the history section. If the history section would be empty, than there is nothing to transform (in this case `transform(...)` should return the provided query as is).

see attached unit test `CompressionQueryTransformerTests.java`

Example output of the unit test:
```
Given the following conversation history and a follow-up query, your task is to synthesize
a concise, standalone query that incorporates the context from the history.
Ensure the standalone query is clear, specific, and maintains the user's intent.

Conversation history:
USER: first question
ASSISTANT: first answer
USER: 5ccc3a5b-e8e3-4168-a44d-ef4d6308ab0f

Follow-up query:
5ccc3a5b-e8e3-4168-a44d-ef4d6308ab0f

Standalone query:
```

[CompressionQueryTransformerTests.java](https://github.com/user-attachments/files/25373362/CompressionQueryTransformerTests.java)

anuragg-saxenaa:
Fix #5470: CompressionQueryTransformer was adding the current query to both history and follow-up sections. Added logic to exclude duplicate USER message from history when it matches current query text. Also added early return when history is empty.

### DIFF
--- a/spring-ai-rag/src/main/java/org/springframework/ai/rag/preretrieval/query/transformation/CompressionQueryTransformer.java
+++ b/spring-ai-rag/src/main/java/org/springframework/ai/rag/preretrieval/query/transformation/CompressionQueryTransformer.java
@@ -77,11 +77,29 @@ public class CompressionQueryTransformer implements QueryTransformer {
 	public Query transform(Query query) {
 		Assert.notNull(query, "query cannot be null");
 
+		if (query.history().isEmpty()) {
+			logger.debug("No conversation history to compress. Returning the original query.");
+			return query;
+		}
+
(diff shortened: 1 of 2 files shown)


anuragg-saxenaa:

Issue #2903, opened 2025-04-26
spring-ai-pgvector-store: Add support for 'halfvec' for vector embeddings and distance functions in PgVectorStore


bjconlan:
**Expected Behavior**

When using the PgVector vector database (PgVectorStore) functionality I was hoping to use the `halfvec` type due to requirements needed from my local model.

**Current Behavior**

Currently only the `vector` type and distance functions supporting said type are supported

**Context**

Using the newer [gemma models](https://huggingface.co/bartowski/google_gemma-3-4b-it-GGUF?show_file_info=google_gemma-3-4b-it-Q6_K.gguf) their embedded vector lengths are upward of 2000 (which is the [limit of the `vector` column type](https://github.com/pgvector/pgvector/issues/461) which is used thoughout the `org.springframework.ai.vectorstore.pgvector.PgVectorStore` and related documentation. 

I can work around this by explicitly creating the schema via:
```sql
CREATE EXTENSION IF NOT EXISTS vector;
CREATE EXTENSION IF NOT EXISTS hstore;
CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
CREATE TABLE IF NOT EXISTS vector_store (
  id uuid DEFAULT uuid_generate_v4() PRIMARY KEY,
  content text,
  metadata json,
  embedding halfvec(2560)
);
CREATE INDEX ON vector_store USING HNSW (embedding halfvec_l2_ops);
```
But cannot use the database as the `PgVectorStore.PgDistanceType` enumeration doesn't provide the appropriate distance type functions for these specific column types.

Ive explicitly modified the code for my case to include the appropriate enum
```java
public static enum PgDistanceType {
    HALFVEC_EUCLIDEAN_DISTANCE("<->", "halfvec_l2_ops", "SELECT *, embedding <-> ? AS distance FROM %s WHERE embedding <-> ? < ? %s ORDER BY distance LIMIT ? "),
...
```
Which allowed me to overcome the issue but perhaps should be added to the enum; perhaps along with [other types that may need to be supported here too](https://github.com/pgvector/pgvector?tab=readme-ov-file#hnsw):

PR #5813, opened 2026-04-17
PgVectorStore : support halfvec, bit and sparsevec column type

R3gardless:
Related to #2903 

Extends `PgVectorStore` to support pgvector's three non-default column types - `halfvec`, `bit` and `sparsevec` in addition to the existing `vector`, Also adds the bit-only `HAMMING_DISTANCE` / `JACCARD_DISTANCE` metrics.
Existing users are unaffected: the default remains `vector` + `cosine` + `HNSW`.

- **pgvector 0.7+ required** for `halfvec`, `bit`, and `sparsevec`. Currently not validated at init, so older version fails with the error `halfvec does not exist`. 
- **bit quantization** the Spring AI internally sign-bit quantizes to bind PGvector `PGbit`. 
 
Reference 
- https://github.com/pgvector/pgvector-java
- https://github.com/pgvector/pgvector


R3gardless:
Could I backport this feature to 1.1.X version as well?

PR #5818, opened 2026-04-18
Fix: toolCallbacks empty when DefaultChatOptions provided via options()

anuragg-saxenaa:
## Summary

When `options()` is called with `DefaultChatOptions` (instead of `ToolCallingChatOptions`), toolCallbacks specified via `.toolCallbacks()` were not propagated to the model.

**Root cause:** `ModelOptionsUtils.copyToTarget()` doesn't copy fields that don't exist in the source type. Since `DefaultChatOptions` has no `toolCallbacks` field, the callbacks from the request spec were lost.

**Fix:** After `copyToTarget` creates `DefaultToolCallingChatOptions`, explicitly propagate toolCallbacks (and toolCallbackProviders) from the request spec.

## Changes

- `DefaultChatClientUtils.java`: After `copyToTarget` from `DefaultChatOptions`, set toolCallbacks from request spec
- `DefaultChatClientUtilsTests.java`: Added regression test

## Testing

- All existing `DefaultChatClientUtilsTests` pass
- New regression test added

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ To clone it you have to either:
 The project targets and build artifacts compatible with Java 17+, but requires JDK 21
 to build. This is enforced by the maven enforcer plugin.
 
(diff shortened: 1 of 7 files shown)


anuragg-saxenaa:
Upstream `DefaultChatClientUtils` was refactored to a builder-based approach after this branch was created — the `toChatClientRequest` method now uses `getDefaultOptions().mutate()` instead of `getChatOptions()` directly.

The original fix logic needs to be re-implemented for the new builder API: specifically, when tools are configured but the builder is not a `ToolCallingChatOptions.Builder` (e.g. because the model returns `DefaultChatOptions` as its defaults), the builder should be upgraded before tool callbacks are applied.

Will update this PR with the re-implemented fix shortly.

PR #5824, opened 2026-04-18
Add methods to handle output schema resource in StructuredOutputChatOptions

nicolaskrier:


spring-ai-model/src/main/java/org/springframework/ai/model/tool/StructuredOutputChatOptions.java
```diff
@@ -16,26 +16,62 @@
...
 	void setOutputSchema(String outputSchema);
 
+	/**
+	 * <p>
+	 * Set output schema resource. It can be useful to define the output schema in the
+	 * application properties, for example:
+	 * {@code spring.ai.ollama.chat.options.output-schema-resource=classpath:schemas/country-json-schema.json}.
+	 * </p>
+	 * <p>
+	 * Note: A default method can't be defined here because {@code JavaBeanBinder} doesn't
```
nicolaskrier, line 47: I am not satisfied with my current implementation. I would prefer to define a default method in this interface but it is not possible at the moment as explained in this note.

I would appreciate feedback on this. Also, defining a default setter in an interface that calls another setter might be considered bad practice.

Issue #5820, opened 2026-04-18
Introduce events in ChatClient streaming API


cijaaimee:
The advisor chain is a very powerful tool. However, I feel like there is a piece missing to the puzzle.

Whilst you can alter a `ChatClientRequest` or a `ChatClientResponse` in any way you'd like through an advisor, there is no clean way of signaling that something happened. The workaround I've found here is to inject data into the response's context and then looking for it after calling the ChatClient API, but to me that's a very hacky bandaid fix.

A practical example of this limitation is #5792, where the current API does not offer a clean way of propagating the tool responses to the user calling the ChatClient API.

To solve this, I'd like to propose introducing what I call a `ChatClientEvent`. This would be an interface used to signal anything that happens throughout the entire API call. Here's what it could look like:
```kotlin
interface ChatClientEvent {
    data class ModelGeneration(val generation: Generation) : ChatClientEvent
    data class ToolCalled(val id: String, val name: String, val responseData: String)  : ChatClientEvent
    //  any other events: InputValidationFailed, etc.
}
```
This interface could also be extended by users to define their own custom events.

The `StreamAdvisor` interface would be modified to return a `Flux<ChatClientEvent>`, like so:
```java
public interface StreamAdvisor extends Advisor {
	Flux<ChatClientEvent> adviseStream(ChatClientRequest chatClientRequest, StreamAdvisorChain streamAdvisorChain);
}
```

Then, users could act on events just like they do with `#chatClientResponse()`, `#chatResponse()` or `#content()`:
```kotlin
chatClient.prompt("Hello World!")
    .stream()
    .events()
    .doOnNext { event ->
        when(event) {
            is ModelGeneration -> logger.info("AI response: ${event.generation.output.text}, tool calls: ${event.generation.output.toolCalls}")
            is ToolCalled -> logger.info("Tool execution completed: id=${event.id}, tool=${event.name}, responseData=${event.responseData}")
        }
    }
```

The blocking API could also be updated to keep track of all events, adding an `events` field to the `ChatClientResponse`:
```java
public record ChatClientResponse(@Nullable ChatResponse chatResponse, Map<String, @Nullable Object> context, List<ChatClientEvent> events)
```

[...]

PR #5827, opened 2026-04-19
Emit tool execution result in streaming Flux when streamToolCallResponses is enabled

### LINKED ISSUE #5792: The toolCalls of ChatResponse are always empty, with no tool call information whatsoever
When SpringAI ChatClient calls a model and the model invokes a tool during execution, the toolCalls field of all elements in the returned Flux ChatResponse is empty, meaning no tool call information is returned. However, when called via SpringAI-Alibaba's ReactAgent.streamMessages, tool call information is returned.

Is this intentional design in SpringAI? If I want to retrieve tool call information from ChatResponse, what should I do?

Version: 1.1.2

bkoragan:
**Summary**
- Closes the emission gap in `ToolCallAdvisor`'s streaming path so that, when `streamToolCallResponses(true)` is set, the downstream `Flux<ChatClientResponse>` now includes a synthetic chunk carrying the tool execution result between the model's tool-call request and its follow-up answer.
- Opt-in and additive: default behavior (`streamToolCallResponses=false`) is unchanged, and no existing advisor implementation is touched. The fix is confined to `ToolCallAdvisor.handleToolCallRecursion`.

**Context**
Reported in #5792: consumers of `ChatClient ... .stream()` observe that `ChatResponse.toolCalls` is "always empty". The underlying cause is the default filter in `ToolCallAdvisor` that drops any chunk where `hasToolCalls()` is `true`. Setting `streamToolCallResponses(true)` restores those chunks — but even with the flag on, the tool *execution result* was never emitted into the `Flux` for the non-`returnDirect` path, so the full round-trip was not observable downstream.

This PR is intentionally scoped as **short-term**: a narrow, zero-breakage fix that reuses the existing flag to make the round-trip fully visible. #5820(@cijaaimee) remains the long-term direction as a follow-up.

**Behavior**
With `streamToolCallResponses(true)`, per tool-call iteration the `Flux` now emits, in order:
1. Streamed assistant chunks that carry the model's tool-call request (`ChatResponse.hasToolCalls() == true`).
2. **New:** a synthetic `ChatClientResponse` whose `ChatResponse` is built via `ToolExecutionResult.buildGenerations(...)` — same shape the `returnDirect` path already uses. Each `Generation`:
   - has the tool's raw output as the `AssistantMessage` content,
   - has `finishReason == ToolExecutionResult.FINISH_REASON` (`"returnDirect"`),
   - carries `ToolExecutionResult.METADATA_TOOL_ID` and `METADATA_TOOL_NAME` in `ChatGenerationMetadata`.
3. Streamed assistant chunks of the model's final answer.

With `streamToolCallResponses(false)` (default) the synthetic chunk is **not** emitted and the existing filter continues to suppress the assistant tool-call chunk, so downstream consumers on the default path observe only the final answer — no behavior change.

**What this does not do**
- It does not flip the `streamToolCallResponses` default. Consumers who want visibility still need to opt in explicitly.
- It does not introduce a typed advisor-chain event model — that is #5820.

closes : #5792

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/TOOLCALLADVISOR_STREAMING_DESIGN.md
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/TOOLCALLADVISOR_STREAMING_DESIGN.md
@@ -131,10 +131,23 @@ ToolCallAdvisor advisor = ToolCallAdvisor.builder()
     .build();
 ```
 
-| Configuration | Intermediate Tool Calls | Final Answer |
-|--------------|------------------------|--------------|
-| `streamToolCallResponses(false)` (default) | Filtered out | Streamed |
-| `streamToolCallResponses(true)` | Streamed | Streamed |
+| Configuration | Assistant Tool-Call Chunk | Tool Execution Result Chunk | Final Answer |
+|--------------|---------------------------|-----------------------------|--------------|
+| `streamToolCallResponses(false)` (default) | Filtered out | Not emitted | Streamed |
+| `streamToolCallResponses(true)` | Streamed | Synthetic chunk emitted between iterations | Streamed |
+
+When enabled, the flow for a single tool-call round-trip is:
+
+```
+[assistant chunks + tool_calls] → [synthetic tool-response chunk] → [assistant final-answer chunks]
+```
+
+The synthetic tool-response chunk is built from the `ToolExecutionResult` via
+`ToolExecutionResult.buildGenerations(...)`, matching the shape used by the
+`returnDirect` path. Each generation carries the tool result text as the
+`AssistantMessage` content and the tool's id/name plus a `returnDirect` finish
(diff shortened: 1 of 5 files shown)


bkoragan:
@cijaaimee @anxinyh @maintainers, Please review the changes for the issue proposal # https://github.com/spring-projects/spring-ai/issues/5792 and share feedback. Thanks!

cijaaimee:
Looks good @bkoragan!
My only concern here is using ToolExecutionResult#buildGeneration as-is, here's why:
* The finish reason is set to ToolExecutionResult#FINISH_REASON, which is is `returnDirect`
* Should a finish reason be set at all?

anxinyh:
Thank you very much for this PR.

Currently, I have customized a NewToolCallAdvisor in version 1.1.3, which inherits the original ToolCallAdvisor and overrides the adviseStream method, applying your code.

Later, I thought that ToolResponseMessage could be placed in ChatResponse by adding a toolResponses field, making it listed alongside generations.

For example, like the NodeOutput in SpringAi Alibaba.

Issue #5829, opened 2026-04-19
OutOfMemoryError in ForkPDFLayoutTextStripper on parsing offset whitespace


asw12:
**Bug description**
ForkPDFLayoutTextStripper has this code:
https://github.com/spring-projects/spring-ai/blob/061d9a15313e32379c0bfe683df0d2e7e7ea96dc/document-readers/pdf-reader/src/main/java/org/springframework/ai/reader/pdf/layout/ForkPDFLayoutTextStripper.java#L169-L177

When parsing a document where there are two consecutive TextPositions offset along the y-axis by more than `5.5` points and representing whitespace characters (such that `height == 0.0`), there is a division by zero. As the resultant `Double.POSITIVE_INFINITY` is cast to an `int`, `numberOfLines` ends up being set to `2147483646`. As each new line creates a char[] buffer based on the number of characters that can fit in each line, this can easily end up being more than 400 Gi of memory.

**Environment**
spring-ai 1.1.4 (currently 1.1.x), Corretto OpenJDK 21

**Steps to reproduce**
This issue was encountered in the wild with the (open access) PDF found at: https://www.clinical-lung-cancer.com/article/S1525-7304(22)00115-2/fulltext ([direct link](https://www.clinical-lung-cancer.com/action/showPdf?pii=S1525-7304%2822%2900115-2)).

When saved, you can just run
```java
new PagePdfDocumentReader("<path_to_pdf>/PIIS1525730422001152.pdf").get()
```
and it'll either consume ~416 Gi of memory, or more likely, fail with OutOfMemoryError.

<img width="1280" height="1564" alt="Image" src="https://github.com/user-attachments/assets/1fddafd2-6695-4eec-9ae8-93436beb2017" />
<img width="2560" height="1564" alt="Image" src="https://github.com/user-attachments/assets/cec35194-c78f-4145-bad6-574264b14ce4" />

The quirk about this document specifically is that on the authors section of the research paper:
* there is a space set on the superscript [1] level, right after the first author's name
* there is a space at the baseline level, before the next author's name
* the different in y-positions exceeds 5.5 points.

**Expected behavior**
Parsing consecutive whitespace text that is offset along the Y-axis should not create an exorbitant amount of character buffers, and there should never be a division by zero type of error when determining the number of newlines to add.

**Minimal Complete Reproducible example**

Here is a (mostly Gemini-vibecoded) test example I came up with to avoid copyright burdens:
```java
[...]

PR #5830, opened 2026-04-19
Prevent OutOfMemoryError in ForkPDFLayoutTextStripper for zero-height whitespace

bkoragan:
**Fixes**: #5829

**Summary**

- `ForkPDFLayoutTextStripper.getNumberOfNewLinesFromPreviousTextPosition` now guards against non-positive and non-finite `TextPosition.getHeight()`, and caps the resulting line count at 500. Removes a division-by-zero path that produced `Double.POSITIVE_INFINITY → Integer.MAX_VALUE → ~400 GiB` of `TextLine` allocations.
- Narrow in scope: one method in one file + a regression test. Default behavior on well-formed input is unchanged.

**Root cause**
Two consecutive whitespace `TextPosition`s with `height == 0.0`, separated by more than 5.5 points along the Y axis, hit the division-by-zero branch in `getNumberOfNewLinesFromPreviousTextPosition`:
1. `Math.floor(delta) / 0.0` → `Double.POSITIVE_INFINITY`
2. `(int) POSITIVE_INFINITY` → `Integer.MAX_VALUE`
3. `createNewEmptyNewLines(~2.1B)` allocates a `TextLine` per iteration, each with a `char[]` sized by the page width — hundreds of GB of attempted allocation → heap OOM.

The pattern occurs naturally in published PDFs (the reporter hit it on the authors section of an open-access journal article). That makes this reachable from untrusted PDF input — DoS-flavored..

**Fix**
- Primary guard: if `!Double.isFinite(height) || height <= 0.0`, return `1` (matches the existing `Math.max(1, ...)` floor — a Y-delta > 5.5 points still warrants at least one line break; we just can't compute an exact count without a valid height).
- Defensive cap: clamp `numberOfLines` at `MAX_NEW_LINES_PER_POSITION_GAP = 500`, covering legitimate-but-tiny positive heights (e.g. 0.001) that would otherwise produce thousands of spurious blank lines.. Any real PDF layout is well below 500.

No API change, no autoconfig change. Well-formed input path is unchanged.

**Validation**

Added `PagePdfDocumentReaderOomTests` with a deterministic PDF reproducer (adapted from the reporter's minimal example).
**Before the fix** on `main`, with a 512 MB heap cap:
java.lang.OutOfMemoryError: Java heap space
at TextLine.<init>(TextLine.java:41)
at ForkPDFLayoutTextStripper.addNewLine(ForkPDFLayoutTextStripper.java:185)
at ForkPDFLayoutTextStripper.createNewEmptyNewLines(ForkPDFLayoutTextStripper.java:157)
BUILD FAILURE (21.9 s)

**After the fix**, normal heap:
Tests run: 1, Failures: 0, Errors: 0 — 0.194 s

**Test plan**
- [x] `./mvnw -pl document-readers/pdf-reader -Dtest=PagePdfDocumentReaderOomTests test` — passes in 194 ms; reproduces OOM on `main` under `-Xmx512m`.
- [x] `./mvnw -pl document-readers/pdf-reader -am verify` — spring-javaformat:validate, checkstyle clean..
- [ ] Cross-check against the reporter's original PDF (https://www.clinical-lung-cancer.com/action/showPdf?pii=S1525-7304%2822%2900115-2) — not run locally; the synthesized reproducer test exercises the same code path.

**Backport candidates**
Happy to open backport PRs as required for previous versions - once this lands on `main`.

### DIFF
--- a/document-readers/pdf-reader/src/main/java/org/springframework/ai/reader/pdf/layout/ForkPDFLayoutTextStripper.java
+++ b/document-readers/pdf-reader/src/main/java/org/springframework/ai/reader/pdf/layout/ForkPDFLayoutTextStripper.java
@@ -48,10 +48,13 @@ public class ForkPDFLayoutTextStripper extends PDFTextStripper {
 
 	/**
 	 * Defensive upper bound on the number of empty lines that may be inserted between two
-	 * consecutive {@link TextPosition}s. Any real PDF layout produces a value well below
-	 * this; anything higher indicates a malformed document (see gh-5829).
+	 * consecutive {@link TextPosition}s. Chosen well below the threshold where
+	 * {@code ExtractedTextFormatter.trimAdjacentBlankLines} becomes unstable on long
+	 * blank-line runs (gh-2247), so the stripper's output can always be safely fed
+	 * through the default reader pipeline. Covers any realistic paragraph or section
+	 * break; anything higher indicates a malformed document (see gh-5829).
 	 */
-	static final int MAX_NEW_LINES_PER_POSITION_GAP = 500;
+	static final int MAX_NEW_LINES_PER_POSITION_GAP = 20;
 
 	private double currentPageWidth;
 
--- a/document-readers/pdf-reader/src/main/java/org/springframework/ai/reader/pdf/layout/ForkPDFLayoutTextStripper.java
+++ b/document-readers/pdf-reader/src/main/java/org/springframework/ai/reader/pdf/layout/ForkPDFLayoutTextStripper.java
@@ -46,6 +46,13 @@ public class ForkPDFLayoutTextStripper extends PDFTextStripper {
 
 	public static final int OUTPUT_SPACE_CHARACTER_WIDTH_IN_PT = 4;
 
+	/**
+	 * Defensive upper bound on the number of empty lines that may be inserted between two
+	 * consecutive {@link TextPosition}s. Any real PDF layout produces a value well below
+	 * this; anything higher indicates a malformed document (see gh-5829).
+	 */
+	static final int MAX_NEW_LINES_PER_POSITION_GAP = 500;
+
 	private double currentPageWidth;
 
 	private @Nullable TextPosition previousTextPosition;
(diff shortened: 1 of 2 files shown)


bkoragan:
@maintainers Please review this PR changes, and let me know if any feedback. This fixes #5829. Thanks!

document-readers/pdf-reader/src/main/java/org/springframework/ai/reader/pdf/layout/ForkPDFLayoutTextStripper.java
```diff
@@ -46,6 +46,13 @@ public class ForkPDFLayoutTextStripper extends PDFTextStripper {
 
 	public static final int OUTPUT_SPACE_CHARACTER_WIDTH_IN_PT = 4;
 
+	/**
+	 * Defensive upper bound on the number of empty lines that may be inserted between two
+	 * consecutive {@link TextPosition}s. Any real PDF layout produces a value well below
+	 * this; anything higher indicates a malformed document (see gh-5829).
+	 */
+	static final int MAX_NEW_LINES_PER_POSITION_GAP = 500;
```
asw12: Thanks for the prompt pull request! It does solve my problem; my only comment is on the use of 500 here. In actuality, wouldn't 1 or 2 new lines suffice?

If the main purpose of `ForkPDFLayoutTextStripper` is to feed text into the `*PdfDocumentReader` classes (and of course, there's nothing that guarantees these classes will be tightly coupled together in perpetuity), then in fact one of the very next things we do is feed the extracted text into this `ExtractedTextFormatter` regex (catastrophically backtracking! https://github.com/spring-projects/spring-ai/issues/2247) that strips out consecutive newlines anyway:
https://github.com/spring-projects/spring-ai/blob/c66a3c55cd19048a63fc2f34c93e46a9809771b1/spring-ai-commons/src/main/java/org/springframework/ai/reader/ExtractedTextFormatter.java#L87

So 500 new lines still seems like a waste of ~100 kB to me, iff we are certain that the `ExtractedTextFormatter` is guaranteed to be set upon its output.

bkoragan: @asw12  Good catch on #2247 — I just looked into it, the regex is a textbook catastrophic-backtracking case ((^$[\r\n]+?^)+ overlaps on itself), with the added wrinkle that the JDK matcher's recursion blows the stack before the exponential blow-up even kicks in.. That is actually a useful constraint for this cap: the stripper should never emit more blank lines than the downstream formatter can safely collapse..

Using 20 as the cap keeps us (1) order of magnitude below the #2247 threshold while still covering any realistic paragraph/section break and bounding per-gap allocation to ~8 KB. I'll update the PR with 20 shortly. Hope that works with safe cap.

#2247 seems a separate fix — either rewrite the collapse with an atomic group / possessive quantifier, or just replace the regex with a two-pass loop. Happy to tackle it as a follow-up PR once this one lands..

bkoragan: Updated the PR with cap tightened to 20, with the Javadoc now calling out the #2247 stack-overflow constraint so the rationale survives in code.. Per-gap allocation drops from ~100 kB to ~8 kB. Happy to tackle #2247 itself in a separate PR. Please review again. Thanks!

asw12 (approved):

PR #5834, opened 2026-04-20
JsonParser: handle raw strings for jsr310 types

Kehrlann:
MCP tool objects don't work with JSR310 types. E.g.

```java
@McpTool(name = "local-date-tool", description = "Tool with date input")
public String dateTool(@McpToolParam LocalDate date) {
	return "Date: " + date.toString();
}
```

when called with `{ "date": "2026-04-20" }` will produce the following error:

```
Conversion from JSON to java.time.LocalDate failed
Unexpected character ('-' (code 45)): Expected space separating root-level values
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); byte offset: #UNKNOWN]
```

This is a hotfix rather than a deep investigation into the intricacies of our JSON parsing setup.

### DIFF
--- a/mcp/mcp-annotations/src/test/java/org/springframework/ai/mcp/annotation/method/tool/SyncMcpToolMethodCallbackTests.java
+++ b/mcp/mcp-annotations/src/test/java/org/springframework/ai/mcp/annotation/method/tool/SyncMcpToolMethodCallbackTests.java
@@ -17,6 +17,7 @@
 package org.springframework.ai.mcp.annotation.method.tool;
 
 import java.lang.reflect.Method;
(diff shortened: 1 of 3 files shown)


tzolov (approved):
Looks good as temp patch. Go ahead

sdeleuze (approved):
This is fixing a regression introduced by the migration to Jackson 3 where I changed the `catch (Exception e)` to `catch (JacksonException e)`, this PR fix it so indeed looks like a reasonable workaround.

We should later do a related refactoring, as this `JsonParser` should be implemented differently.

PR #5842, opened 2026-04-21
Fix #5780: Accept SSE frames without event: field in WebClientStreamableHttpTransport

### LINKED ISSUE #5780: WebClientStreamableHttpTransport drops valid SSE frames when `event:` is omitted (should default to `message`)

anuragg-saxenaa:
## Description

Per SSE spec, omitting `event:` defaults to `message`. The `parse()` method in `WebClientStreamableHttpTransport` now accepts null/empty/"message" event types as valid message events.

Previously, SSE frames sent without an explicit `event:` field were silently dropped because `event.event()` returned null, causing MCP client initialization failures and request timeouts.

Example dropped frame:
```
```

`MESSAGE_EVENT_TYPE.equals(null)` returned false — frame ignored.

## Fix

**File:** `mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransport.java`

**Change:**
```java
// Before:

// After:
String eventType = event.event();
// Per SSE spec, omitting event: defaults to "message". Accept null/empty/"message" as valid.
if (MESSAGE_EVENT_TYPE.equals(eventType) || eventType == null || eventType.isEmpty()) {
```

## Regression Test

Added `WebClientStreamableHttpTransportSseEventTest.java` covering:
- **null event** — SSE frame with no `event:` field (regression scenario for #5780)
- **empty event** — SSE frame with `event:` (empty value)
- **explicit message event** — SSE frame with `event: message` (backward compatibility)

## Related

- MCP Java SDK PR #913 fixed the same issue for the `HttpClient` variant
- `modelcontextprotocol/java-sdk#885` — original upstream discussion
- Fix follows the same pattern as the SDK fix

### DIFF
--- a/mcp/transport/mcp-spring-webflux/src/test/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransportSseEventTest.java
+++ b/mcp/transport/mcp-spring-webflux/src/test/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransportSseEventTest.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.mcp.client.webflux.transport;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+
+import com.sun.net.httpserver.HttpServer;
+import io.modelcontextprotocol.spec.HttpHeaders;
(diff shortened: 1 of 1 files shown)


anuragg-saxenaa:
LGTM — clean SSE spec compliance fix.

The core change is correct: parse() now treats null/empty event type as equivalent to "message", which aligns with the SSE spec (omitting event: defaults to "message"). The regression tests are thorough — covering null event, empty event, explicit "message" event, and unknown event types. The bug description and test comments make the rationale clear.

No concerns.

anuragg-saxenaa:
Hi team — this PR is ready for review (fixes #5780). CI is waiting for first-contributor workflow approval. Could a maintainer approve the workflow run when convenient? Thank you!

anuragg-saxenaa:
Hi team — this PR is ready for review (fixes #5780). CI is waiting for first-contributor workflow approval. Could a maintainer approve the workflow run when convenient? Thank you!

anuragg-saxenaa:

PR #5843, opened 2026-04-21
Enable MCP IT tool with date

Kehrlann:
Follow up on #5834, the snapshots in CI should now handle JSR310 dates in integration tests.


anuragg-saxenaa:
LGTM — removing `@Disabled` to re-enable the `testToolWithDateParameter` test in CI now that the spring-ai-model SNAPSHOT is published. Clean, targeted fix.

redinside-dev (approved):
LGTM — removing the Disabled annotation now that the snapshot is published. Clean change, test logic is sound.

PR #5854, opened 2026-04-22
GH-5806: Merge streaming tool call chunks in MessageAggregator

### LINKED ISSUE #5806: Streaming tool calls are incorrectly merged, causing empty arguments and IllegalArgumentException
## Bug description
When using streaming mode (stream=true) with tool/function calling in Spring AI, tool execution may fail with:
`java.lang.IllegalArgumentException: toolInput cannot be null or empty`
Root cause appears to be incorrect merging of streaming tool call chunks in NodeExecutor.mergeToolCalls() (or equivalent internal logic).

In streaming responses, the model emits tool calls in multiple chunks. For example:
`ToolCall 1: { id: "call_xxx", name: "queryCourse", arguments: "" }/
ToolCall 2: { id: "", name: "", arguments: "{\"query\":\"edu\"" }
ToolCall 3: { id: "", name: "", arguments: ",\"page\":4" }
ToolCall 4: { id: "", name: "", arguments: "}" }`

However, the current merge logic:
- Deduplicates only by `id`
- Does **not properly concatenate `arguments`**
- Does **not propagate `name` correctly**

As a result, the merged `ToolCall` may end up with:
`arguments = ""`

which causes the exception during tool execution.

---

## Environment

- Spring AI version: `1.0.0-M6`
- Spring Boot: `3.x`
- Java: `17`
- Model: OpenAI-compatible (streaming enabled)
- Mode: `stream=true`
- Tool calling: enabled (FunctionCallback / ToolCallback)
- Vector store: N/A

---

## Steps to reproduce

1. Enable streaming `mode:`
`chatClient.prompt()
    .user("search courses about edu")
    .tools(new QueryCourseTool())
    .stream()
    .content();`
2. Ensure the model triggers a tool call with arguments
3. Debug inside:
`DefaultToolCallingManager.executeToolCall(...)`
4. Inspect:
`assistantMessage.getToolCalls()`
5. Observe multiple partial tool calls:
`arguments: ""
arguments: "{\"query\":\"edu\""
arguments: ",\"page\":4"
arguments: "}"`
6. Execution reaches:
`toolCallback.call(toolInputArguments, toolContext);`
7.Exception is thrown:
`IllegalArgumentException: toolInput cannot be null or empty`

## Expected behavior
Streaming tool call chunks should be merged into a complete valid tool call:
`
{
  id: "call_xxx",
  name: "queryCourse",
  arguments: "{\"query\":\"edu\",\"page\":4}"
}
`
Specifically:

 - arguments should be concatenated in order
 - name should be preserved from the first valid chunk
 - Empty chunks should not overwrite valid values
## Minimal Complete Reproducible example
```@Bean
public FunctionCallback queryCourse() {
    return FunctionCallback.builder()
        .function("queryCourse", (Map<String, Object> input) -> {
            return "ok";
        })
        .build();
}
```
```
chatClient.prompt()
    .user("帮我查一下edu相关课程，第4页")
    .tools(queryCourse())
    .stream()
    .content()
    .subscribe(System.out::println);
```
## Debug Observation
```
ToolCall 0:
  name = "queryCourse"
  arguments = ""

ToolCall 1:
  name = null
  arguments = "{\"query\":\"edu\""

ToolCall 2:
  name = null
  arguments = ",\"page\":4"

ToolCall 3:
  name = null
  arguments = "}"
```
After merge:

`arguments = ""   <-- incorrect`

## Suggested fix

In mergeToolCalls():

1. Merge by id
2. Concatenate arguments instead of overwriting:
`existing.arguments += incoming.arguments;`
3. Preserve first non-null name:
```
if (existing.name == null) {
    existing.name = incoming.name;
 }
```
4. Ignore empty argument chunks if necessary
## Additional context

This issue only occurs in streaming mode.
Non-streaming responses return fully assembled tool calls and work correctly.

suryateja-g13:
## Summary

- Streaming tool calls arrive in multiple chunks: first chunk carries `id` and `name`, subsequent chunks carry partial `arguments` with an empty `id`
- `MessageAggregator` was calling `addAll()` on every chunk, producing a list of incomplete `ToolCall` entries with empty `arguments`, ultimately causing `IllegalArgumentException: toolInput cannot be null or empty` during execution
- Added `mergeToolCallChunk()` helper: chunks with a non-empty `id` start a new entry; chunks with an empty `id` append their `arguments` to the most recently started call
- Handles single tool calls split across N chunks and multiple sequential tool calls streamed one after another

## How to test locally

```bash
./mvnw test -pl spring-ai-model -Dtest=MessageAggregatorTests
```

- `should_merge_tool_call_chunks_into_single_complete_tool_call` — replicates the exact streaming pattern from the issue report
- `should_pass_through_complete_tool_call_unchanged` — single complete tool call (no merging needed) still works
- `should_merge_multiple_parallel_tool_calls` — two sequential tool calls each split across two chunks
- `should_handle_chunk_with_null_arguments` — null arguments handled without NPE

Fixes GH-5806 (https://github.com/spring-projects/spring-ai/issues/5806)

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/model/MessageAggregator.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/model/MessageAggregator.java
@@ -116,7 +116,9 @@ public class MessageAggregator {
 				}
 				AssistantMessage outputMessage = chatResponse.getResult().getOutput();
 				if (!CollectionUtils.isEmpty(outputMessage.getToolCalls())) {
-					toolCallsRef.get().addAll(outputMessage.getToolCalls());
+					for (ToolCall chunk : outputMessage.getToolCalls()) {
+						mergeToolCallChunk(toolCallsRef.get(), chunk);
+					}
 				}
 
 			}
@@ -205,6 +207,23 @@ public class MessageAggregator {
 			metadataRateLimitRef.set(new EmptyRateLimit());
 
 		}).doOnError(e -> logger.error("Aggregation Error", e));
+	}
+
+	/**
+	 * Merges a streaming tool call chunk into the accumulated tool call list. If the
+	 * chunk has a non-empty {@code id}, it starts a new tool call entry. Otherwise, its
+	 * {@code arguments} are appended to the most recently added tool call.
+	 */
+	private static void mergeToolCallChunk(List<ToolCall> toolCalls, ToolCall chunk) {
+		if (StringUtils.hasText(chunk.id())) {
+			toolCalls.add(chunk);
+		}
+		else if (!toolCalls.isEmpty()) {
+			ToolCall last = toolCalls.remove(toolCalls.size() - 1);
+			String merged = (last.arguments() != null ? last.arguments() : "")
+					+ (chunk.arguments() != null ? chunk.arguments() : "");
+			toolCalls.add(new ToolCall(last.id(), last.type(), last.name(), merged));
+		}
 	}
 
 	public record DefaultUsage(Integer promptTokens, Integer completionTokens, Integer totalTokens) implements Usage {
--- a/spring-ai-model/src/test/java/org/springframework/ai/chat/model/MessageAggregatorTests.java
+++ b/spring-ai-model/src/test/java/org/springframework/ai/chat/model/MessageAggregatorTests.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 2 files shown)


anuragg-saxenaa (approved):
LGTM. The fix changes `addAll` to a `mergeToolCallChunk()` loop that correctly handles partial tool call chunks: chunks with a non-empty `id` start a new entry, while empty-id chunks append arguments to the last entry. The logic correctly reconstructs streamed tool calls across multiple chunks.

The new `MessageAggregatorTests` has good coverage: merging sequential chunks, passing through complete tool calls, handling multiple parallel tool calls, and null arguments. Edge case handling looks solid.

One minor observation: the code uses `StringUtils.hasText(chunk.id())` for the empty-id check. Since `id` is a String record component, this is safe — empty string or null both return false from `hasText`. ✅

Issue #5600, opened 2026-03-14
isText() at Document ignores MimeType of Media field.


CarvalhoLeonardo:
The test for Text types at [Document.java](https://github.com/spring-projects/spring-ai/blob/b0ce25aed319d4fb0c80535375a81d8ce57bdb80/spring-ai-commons/src/main/java/org/springframework/ai/document/Document.java#L197) seems off.

If I add Media types "text/plain" or "text/xml", it should test it before throwing the exception at [AbstractObservationVectorStore](https://github.com/spring-projects/spring-ai/blob/b0ce25aed319d4fb0c80535375a81d8ce57bdb80/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/observation/AbstractObservationVectorStore.java#L92), since they are text.

On the other hand, setting the text and setting the media at the same time is not allowed, so, using the media seems impossible at this time.

sdeleuze:
The current design enforces that either `String text` or `Media media` are set, so this kind of hybrid use case where `Media` is used to represent a text is not supported. Could you please share more details on your use case?

CarvalhoLeonardo:
I am creating a RAG pipeline with Web sources, and was trying to save the documents following the Mime types (all being XML, JSON or Text), and found this behavior.

I may be misinterpreting the API, but seems odd to have the text field being the unique validation factor on this scenario.

sdeleuze:
I see, I will discuss this with the team and provide a feedback.

karansahani78:
Hi! I’ve been looking into this issue and would like to contribute if it aligns with the intended design.

From my understanding, `Document` currently considers only the `text` field as textual, while `media` is ignored even when its MIME type represents textual content (e.g. `text/plain`, `text/xml`, `application/json`).

In use cases like RAG pipelines, documents are often stored as structured formats (JSON/XML) via `media`, but still need to be treated as text for downstream processing. This seems to lead to the `IllegalArgumentException` in `AbstractObservationVectorStore`.

One possible direction could be:

* Extending `isText()` to also consider MIME types that represent textual content, or
* Introducing a clearer distinction/helper for “textual media” vs binary media

I initially considered checking for `text/*`, but I’m unsure if the scope should also include formats like `application/json` or `application/xml`.

Happy to open a PR once the preferred direction is clarified. Would love your guidance on this.

CarvalhoLeonardo:
I think

_Introducing a clearer distinction/helper for “textual media” vs binary media_

[...]

PR #5892, opened 2026-04-27
Fix Document.isText() media type handling and add test coverage

karansahani78:
Improve handling of media types in Document.isText()

Previously, certain text-based media types were not correctly identified as text. 
This change improves the detection logic to properly handle such cases.

Additionally, a new test class (DocumentIsTextMediaTest) has been added to 
verify correct behavior across different media types.

- Fix media type detection in Document.isText()
- Add test coverage for text and non-text media types

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/document/Document.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/document/Document.java
@@ -190,12 +190,34 @@ public class Document {
 
 	/**
 	 * Determines whether this document contains text or media content.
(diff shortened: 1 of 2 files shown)


karansahani78:
This PR is related to #5600.

Based on the ongoing discussion there, it seems introducing a clearer distinction for “textual media” may be preferred over modifying `isText()` directly.

I’m keeping this PR as a reference for the current approach, but I’m happy to refactor it (e.g. introduce a helper like `isTextualMedia()` or adjust toward a formatter-based solution) once the direction is confirmed.

PR #5895, opened 2026-04-28
Wire mcpServerJsonMapper into stateless MCP server builders

### LINKED ISSUE #5894: MCP server: McpServerStatelessAutoConfiguration does not pass mcpServerObjectMapper into McpServer.sync().jsonMapper()
## Bug description

`McpServerStatelessAutoConfiguration#mcpStatelessSyncServer` (and the async equivalent) builds the `McpServer` like this:

```java
StatelessSyncSpecification serverBuilder = McpServer.sync(statelessTransport).serverInfo(serverInfo);
// … configures tools, capabilities, instructions, requestTimeout …
return serverBuilder.build();
```

It never calls `.jsonMapper(...)`. So when `serverBuilder.build()` runs, it falls back to `McpJsonMapper.getDefault()` — a strict vanilla Jackson `ObjectMapper` with `FAIL_ON_UNKNOWN_PROPERTIES=true`. That default is then used inside `McpStatelessAsyncServer#asyncInitializeRequestHandler` to do the second-stage `convertValue(req, InitializeRequest.class)`.

Meanwhile, the lenient `mcpServerObjectMapper` bean produced by `McpServerObjectMapperAutoConfiguration` is wired into the transport layer (`WebMvcStatelessServerTransport`) for the first-stage envelope deserialization. So:

- **Stage 1 (transport)** — JSON-RPC envelope → `JSONRPCRequest`: uses the lenient mapper ✅
- **Stage 2 (handler)** — `params` → `InitializeRequest`: uses the strict default ❌

## Impact

Any MCP client implementing the [2025-11-25 spec](https://modelcontextprotocol.io/specification/2025-11-25) sends `capabilities.elicitation.form: {}` (or `url: {}`) on the initialize request. The MCP Java SDK's `ClientCapabilities.Elicitation` record (≤ 0.17.x) is empty, so the strict default mapper rejects the unknown fields and the handshake fails with:

```
-32603: Unrecognized field "form" (class io.modelcontextprotocol.spec.McpSchema$ClientCapabilities$Elicitation),
not marked as ignorable (0 known properties: ])
 ... through reference chain: InitializeRequest["capabilities"]->ClientCapabilities["elicitation"]->Elicitation["form"]
```

Codex (OpenAI's CLI), and any other client on spec 2025-11-25, can't connect to a Spring AI MCP server because of this.

## Fix

`McpServer.StatelessSyncSpecification` (and `StatelessAsyncSpecification`) already expose a `jsonMapper(McpJsonMapper)` builder method. The autoconfigure should call it:

```java
StatelessSyncSpecification serverBuilder = McpServer.sync(statelessTransport)
    .serverInfo(serverInfo)
    .jsonMapper(new JacksonMcpJsonMapper(mcpServerObjectMapper));   // <-- add this
```

The same wiring already exists for stdio transport in `McpServerAutoConfiguration#stdioServerTransport`. The stateless path was missed.

## Reproducer

Spring AI 1.1.2 + `spring-ai-starter-mcp-server-webmvc`, configured with `spring.ai.mcp.server.protocol: STATELESS`. POST to the MCP endpoint:

```json
{
  "jsonrpc": "2.0", "id": 1, "method": "initialize",
  "params": {
    "protocolVersion": "2025-11-25",
    "capabilities": { "elicitation": { "form": {}, "url": {} } },
    "clientInfo": { "name": "codex", "version": "0.x" }
  }
}
```

Returns the strict-mode error above. Same payload with the workaround below succeeds.

## Workaround

Register a `McpJsonMapperSupplier` via SPI (`META-INF/services/io.modelcontextprotocol.json.McpJsonMapperSupplier`) that returns a lenient `JacksonMcpJsonMapper`. Spring Boot's classloader picks user resources before transitive deps, so this overrides `McpJsonMapper.getDefault()`. Works but obviously the proper fix is to plumb `mcpServerObjectMapper` through.

## Related

#5178 reports a similar symptom in a different code path (tool call argument deserialization using a static `OBJECT_MAPPER`). Same class of bug — incomplete plumbing of `mcpServerObjectMapper` through the MCP stack — different location.

## Environment

- Spring AI 1.1.2 (also reproducible on 1.1.5 via the same pinned MCP SDK 0.17.0)
- Spring Boot 4.0.3 / Java 25
- MCP Java SDK 0.17.0 (transitive)

PVPankaj:
`McpServerStatelessAutoConfiguration` builds the stateless sync and async `McpServer` without calling `.jsonMapper(...)`. This causes the handler stage (e.g. `convertValue(req, InitializeRequest.class)`) to fall back to `McpJsonMapper.getDefault()` - a strict vanilla ObjectMapper with `FAIL_ON_UNKNOWN_PROPERTIES=true`.

Meanwhile, the transport layer correctly uses the lenient `mcpServerJsonMapper` bean. So we end up with two different mappers at different stages:

- **Stage 1 (transport)** - JSON-RPC envelope -> `JSONRPCRequest`: lenient mapper ✅
- **Stage 2 (handler)** - `params` -> `InitializeRequest`: strict default ❌

Any MCP client implementing the 2025-11-25 spec (e.g. Codex) sends fields like `capabilities.elicitation.form` that the current SDK records don't model yet, so the strict mapper rejects them and the initialize handshake fails with:

```
Unrecognized field "form" (class ...ClientCapabilities$Elicitation),
not marked as ignorable
```

This change adds `@Qualifier("mcpServerJsonMapper") JsonMapper` to both `mcpStatelessSyncServer` and `mcpStatelessAsyncServer` bean methods and wires it via `.jsonMapper(new JacksonMcpJsonMapper(...))` on the builder - the same pattern already used by `McpServerAutoConfiguration` for the stdio transport.

Closes gh-5894

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerStatelessAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerStatelessAutoConfiguration.java
@@ -19,6 +19,7 @@ package org.springframework.ai.mcp.server.common.autoconfigure;
 import java.util.ArrayList;
 import java.util.List;
 
+import io.modelcontextprotocol.json.jackson3.JacksonMcpJsonMapper;
 import io.modelcontextprotocol.server.McpServer;
 import io.modelcontextprotocol.server.McpServer.StatelessAsyncSpecification;
 import io.modelcontextprotocol.server.McpServer.StatelessSyncSpecification;
@@ -37,9 +38,11 @@ import io.modelcontextprotocol.server.McpStatelessSyncServer;
 import io.modelcontextprotocol.spec.McpSchema;
 import io.modelcontextprotocol.spec.McpSchema.Implementation;
 import io.modelcontextprotocol.spec.McpStatelessServerTransport;
+import tools.jackson.databind.json.JsonMapper;
 
 import org.springframework.ai.mcp.server.common.autoconfigure.properties.McpServerProperties;
 import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.AllNestedConditions;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
@@ -75,6 +78,7 @@ public class McpServerStatelessAutoConfiguration {
 	@ConditionalOnProperty(prefix = McpServerProperties.CONFIG_PREFIX, name = "type", havingValue = "SYNC",
 			matchIfMissing = true)
 	public McpStatelessSyncServer mcpStatelessSyncServer(McpStatelessServerTransport statelessTransport,
+			@Qualifier("mcpServerJsonMapper") JsonMapper mcpServerJsonMapper,
 			McpSchema.ServerCapabilities.Builder capabilitiesBuilder, McpServerProperties serverProperties,
 			ObjectProvider<List<SyncToolSpecification>> tools,
 			ObjectProvider<List<SyncResourceSpecification>> resources,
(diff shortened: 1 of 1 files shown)


sobychacko:
Closing in favor of #6031, which makes the same fix and also updates the integration test setup. Without that, most of the tests in that suite fail because the new dependency isn't available in the test context.

The production change here is correct; it's only the test harness that needed the extra line.

Issue #5797, opened 2026-04-13
RESPECT_JSONPROPERTY_ORDER not enabled in JsonSchemaGenerator


clement-s-bridge:
Hello,

**Bug description**
I've noticed a difference between the `JacksonSchemaModule` configuration in `BeanOutputConverter.java` and `JsonSchemaGenerator.java`.

In `BeanOutputConverter.java` there's `JacksonOption.RESPECT_JSONPROPERTY_ORDER` but not in `JsonSchemaGenerator.java`.

This setting is useful to set the `reflexion` property first for MistralAI and have a more accurate response.

IMO we can add the missing settings in `JsonSchemaGenerator.java` it shouldn't cause any harm.

**Environment**
Java 25, Spring Boot 4.0.5, SpringAI 2.0.0 M4

**Steps to reproduce**
Create a json schema with `ResponseFormat.jsonSchema(MyExpectedOutput.class)`. With `MyExpectedOutput` containing  `@JsonPropertyOrder(...)`.

**Expected behavior**
The json schema should respect the `@JsonPropertyOrder(...)` annotation but it doesn't.

suryateja-g13:
Submitted a fix in PR #5847: https://github.com/spring-projects/spring-ai/pull/5847

Issue #5341, opened 2026-01-27
BeanOutputConverter .entity() ignores @Schema annotations despite documentation claiming support - forces code duplication with @JsonPropertyDescription


mustafagoksever:
The `BeanOutputConverter` (used internally by `.entity()`) ignores Swagger/OpenAPI `@Schema` annotations for field descriptions, despite the official API documentation explicitly listing `@Schema` as a supported annotation. This forces developers to duplicate field descriptions using `@JsonPropertyDescription`, creating unnecessary code duplication and maintenance overhead.

**Environment**
- Spring AI version: 1.1.2
- Java version: Java 21
- AI Provider:  OpenAI

---

**Documentation vs. Reality**

According to the [JsonSchemaGenerator API documentation](https://docs.spring.io/spring-ai/docs/current/api/org/springframework/ai/util/json/schema/JsonSchemaGenerator.html), the following annotations are **officially supported**:

- `@ToolParam(required = ..., description = ...)`
- `@JsonProperty(required = ...)`
- `@JsonClassDescription(...)`
- `@JsonPropertyDescription(...)`
- **`@Schema(required = ..., description = ...)` ✓**
- `@Nullable`

However, in practice with `.entity()` / `BeanOutputConverter`, **only `@JsonPropertyDescription` actually works**.

---

**Steps to Reproduce**

1. Create a POJO with `@Schema` annotations:

```java
import io.swagger.v3.oas.annotations.media.Schema;

public class Product {
    @Schema(description = "Unique product identifier")
    private Long id;
    
    @Schema(description = "Product name as displayed in catalog")
    private String name;
    
    @Schema(description = "Product price in USD")
    private Double price;
    
    // getters and setters
}
```

2. Use structured output with `.entity()`:

```java
Product result = chatClient.prompt()
    .user("Extract product information: iPhone 15 Pro, $999")
    .call()
    .entity(Product.class);
```

3. Debug the generated schema:

```java
BeanOutputConverter<Product> converter = new BeanOutputConverter<>(Product.class);
System.out.println(converter.getFormat());
```

---

**Expected Behavior**

The generated JSON schema should include `description` fields from `@Schema` annotations:

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "object",
  "properties": {
    "id": {
      "type": "integer",
      "description": "Unique product identifier"
    },
    "name": {
      "type": "string",
      "description": "Product name as displayed in catalog"
    },
    "price": {
[...]

PR #5897, opened 2026-04-28
Reuse `JsonSchemaGenerator` in `BeanOutputConverter`

### LINKED ISSUE #5884: Inconsistent JSON Schema support between tool calling and structured outputs
**Expected Behavior**

The JSON Schema logic used for tool calling support (encapsulated in the `JsonSchemaGenerator` class) should be re-used also in `BeanOutputConverter` to ensure consistent and coherent support.

**Current Behavior**

Currently, `BeanOutputConverter` uses internal logic for handling the JSON Schema generation. It should be updated to re-use the `JsonSchemaGenerator` class where we improved the JSON Schema support when we introduced the declarative tool calling features. Unfortunately, we never got around to update the `BeanOutputConverter` after that.

ChoMinGi:
Fixes https://github.com/spring-projects/spring-ai/issues/5884

Replace `BeanOutputConverter`'s internal victools schema generation with `JsonSchemaGenerator.generateForType()` delegation to ensure consistent JSON Schema behavior between tool calling and structured outputs.

- Add `RESPECT_JSONPROPERTY_ORDER` and conditional `KotlinModule` to `JsonSchemaGenerator`
- Remove duplicate victools setup from `BeanOutputConverter`
- Remove `postProcessSchema(JsonNode)` extension point
- Replace it with a `protected generateSchema()` method as the new extension point — subclasses can post-process by calling `super.generateSchema()` or replace the generation logic entirely
- Schema formatting is now delegated to `JsonSchemaGenerator` (Jackson's default pretty printer)

**Behavior changes:**
- `@JsonProperty` without explicit `required = true` no longer forced into `required` array
- Kotlin nullable fields (`String?`) no longer forced into `required`
- Unannotated fields remain required by default (unchanged)

This also addresses the root cause behind #5797 and #5341, as both stem from `BeanOutputConverter` maintaining a separate schema generation setup.

**Open question(resolved)**
- The `required` behavior changes for fields annotated with `@JsonProperty` (without explicit `required = true`). Is this acceptable, or should backward compatibility be preserved?
- **Resolved: accepted as bug fix.** Documented in 2.0.0-M6 upgrade notes.


sdeleuze:
@ThomasVitale Could your please review this PR?

ThomasVitale:
I'll have a look later today, thanks

spring-ai-model/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
```diff
@@ -187,23 +180,15 @@ private static ResponseTextCleaner createDefaultTextCleaner() {
...
-			.with(jacksonModule)
-			.with(Option.FORBIDDEN_ADDITIONAL_PROPERTIES_BY_DEFAULT);
-
-		configBuilder.forFields().withRequiredCheck(f -> true);
-
-		if (KotlinDetector.isKotlinReflectPresent()) {
-			configBuilder.with(new KotlinModule());
+		String schemaString = JsonSchemaGenerator.generateForType(this.type);
+		JsonNode jsonNode;
+		try {
```
ThomasVitale: I'm a bit unsure about this part. I understand it's meant to maintain support for `postProcessSchema()`. I'm thinking if we should take this opportunity for the 2.0.0 release to remove the `postProcessSchema()` entirely. Until now it was necessary because the entire JSON Schema generation logic was included in this class and we needed a way to make it customizable that wouldn't require re-inventing the whole logic. And I know it's been used to work around some of the issues in the logic (e.g. the cumbersome handling of required fields or missing support for annotations) which this PR solved once and for all.

Now the whole logic is one call to `JsonSchemaGenerator.generateForType(this.type)`. We could remove the `postProcessSchema()` method, make the `BeanOutputConverter.generateSchema()` method `protected`, and identify that one as the extension point to customize the JSON Schema generation logic. That would make this even more flexible since developers would still have the option to apply post-processing logic, but they would also have the option to completely change the generation logic if they wanted to.

My main concern with the suggested solution is the extra Jackson deserialisation+serialization operations that are always performed, even though there's no post-processing logic defined (which is probably most of the times). Jackson operations are not cheap. With the additional risk of exceptions being thrown performing an operation that is often not needed.

Thoughts? @ChoMinGi @sdeleuze

ThomasVitale:
@ChoMinGi thanks so much for the PR. About your open question.

The `JsonSchemaGenerator` was configured from the start to treat all properties as required by default (to ensure correct behaviour across different model providers), unless a property is annotated explicitly to be optional via any of these annotations:

- `@ToolParam(required = false)`
- `@JsonProperty(required = false)`
- `@Schema(required = false)`
- `@Nullable`

The logic handling that is in the `SpringAiSchemaModule` class used internally by the `JsonSchemaGenerator`.

I think it's correct to apply this same logic to structured output conversion. 

- I see as a bug the fact that Kotlin nullable fields were treated as required, so it's great that's now fixed by doing this consolidation.
- I also see as a bug that if I had a field annotated with `@JsonProperty` (whose `required` value is `false` by default if unspecified) was actually treated as required. Now, we are respecting the annotation value, and also reducing the element of surprise.

So I think that's an acceptable change as it's actually about fixing existing bugs in the `BeanOutputConverter` implementation. @sdeleuze what do you think?

Still, we're changing the behaviour, so it's important to make people aware of it. In the context of this PR, I would suggest mentioning the two behaviour differences in the upgrades notes for the upcoming release [here](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc) in a new "Upgrading to 2.0.0-M6" section.

And then I would perhaps suggest creating a separate task for reviewing and consolidating the documentation about JSON Schema generation between Tool Calling and Structured Output Conversion.

spring-ai-model/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
ChoMinGi: @ThomasVitale 
Thanks for raising this. I agree with that direction.

I kept the deserialize/post-process/serialize flow only to preserve the existing `postProcessSchema()` extension point.
But since this is for 2.0.0 and schema generation is now delegated to `JsonSchemaGenerator.generateForType(this.type)`, removing `postProcessSchema()` and making `generateSchema()` the protected extension point seems cleaner and more flexible.

That keeps the same construction-time extension lifecycle as today, while avoiding the extra Jackson roundtrip in the default path. Subclasses can still post-process the generated schema by calling `super.generateSchema()`, or replace the generation logic entirely.

I'll update the PR accordingly and add the migration note under a new Upgrading to 2.0.0-M6 section in `upgrade-notes.adoc`.

ChoMinGi: Updated in the latest push.

`postProcessSchema()` has been removed, `generateSchema()` is now protected, and `BeanOutputConverter` now uses the `JsonSchemaGenerator.generateForType(this.type)` output directly without the Jackson deserialize/serialize roundtrip.

I also updated the affected schema format assertions, removed the obsolete line separator tests, and added the 2.0.0-M6 upgrade notes for the required-property behavior changes and the migration from `postProcessSchema()` to `generateSchema()`.

Tests: `BeanOutputConverterTest`, `JsonSchemaGeneratorTests`, and `KotlinBeanOutputConverterTests` pass locally.

ChoMinGi:
@ThomasVitale 
Thanks for the clarification on the required-property semantics.
Agreed it's effectively a bug fix, and good to have structured output conversion aligned with JsonSchemaGenerator.
The two behavior differences are documented in the new "Upgrading to 2.0.0-M6"section.

I agree the JSON Schema documentation consolidation between Tool Calling and Structured Output Conversion is better handled as a follow-up. I can open a  separate issue for that.

ChoMinGi:
Follow-up issue created: #5915

KoreaNirsa:
This appears to introduce a behavioral change in how `additionalProperties: false` is applied

Previously, `BeanOutputConverter` used `Option.FORBIDDEN_ADDITIONAL_PROPERTIES_BY_DEFAULT`, which enforced `additionalProperties: false` across all generated object schemas, including nested objects and array items

With the current delegation to `JsonSchemaGenerator.generateForType()`, `additionalProperties: false` seems to be applied only to the root schema node, leaving nested object schemas more permissive

This is visible in the updated test expectations where `items` schemas no longer include `additionalProperties: false`

I think this may effectively relax schema strictness for nested objects, which could impact
- structured output prompting
- provider-side strict schema handling
- consumers relying on stricter nested JSON Schema output

If this change is intentional, it would be helpful to document it in the upgrade notes

mustafagoksever:
I ran into the same issue described here and previously discussed it in https://github.com/spring-projects/spring-ai/issues/5341.

In my case, the previous behavior of `BeanOutputConverter` was indeed problematic, especially around how `@JsonProperty` and nullable fields were treated as required.

This change aligns much better with expected Jackson semantics and fixes the inconsistency between structured output and tool schema generation.

+1

ChoMinGi:
Thanks @KoreaNirsa — you're right, this is worth addressing.

**Root cause**
`JsonSchemaGenerator.processSchemaOptions()` applies `additionalProperties: false` only on the root node ([`560baaae5`](https://github.com/spring-projects/spring-ai/commit/560baaae5)).
`BeanOutputConverter` previously used victools' `FORBIDDEN_ADDITIONAL_PROPERTIES_BY_DEFAULT` which applied it recursively. This PR's consolidation surfaces that gap.

**Impact**
This wasn't visible for tool calling (parameters are typically flat), but it breaks structured output with nested types — `OpenAiChatModelResponseFormatIT.jsonSchemaBeanConverter()` will fail since it asserts nested `additionalProperties: false` and sends the schema to OpenAI with `strict(true)`.

**Two options**
- (A) Fix in this PR — Add recursive `forbidAdditionalProperties()` in `JsonSchemaGenerator` (same pattern as `convertTypeValuesToUpperCase()`), with a `!node.has("additionalProperties")` guard to preserve
`Map<K,V>` schemas. 
Restore nested assertions in tests.

- (B) Follow-up PR— Document the behavior change in upgrade notes, open a separate PR for the recursive fix.

I lean toward (A) — this PR's goal is schema consolidation, and leaving a strict-mode regression feels like unfinished consolidation rather than a follow-up.

@ThomasVitale Happy to push the fix here as a separate commit, or split it out if you prefer to keep this PR's scope focused.

ThomasVitale:
@ChoMinGi I also like option A better, and I agree we should make it consistent. Thanks!

ChoMinGi:
@ThomasVitale Pushed option A in a separate commit (591ede4) — recursive traversal with a Map-preservation guard.

Changes
- `forbidAdditionalProperties()` recursively applies `additionalProperties: false` to all object schemas with `properties`, mirroring the traversal pattern in  `convertTypeValuesToUpperCase()`.
- `!node.has("additionalProperties")` guard preserves Map schemas (which set `additionalProperties` to a schema, not `false`).
- Restored the nested assertions in the existing tests that were dropped in the previous revision.
- Added coverage for the ALLOW option with nested objects and Map field preservation.

(Thanks @KoreaNirsa for catching the Map regression.)

Verified `OpenAiChatModelResponseFormatIT.jsonSchemaBeanConverter()` passes locally.
Happy to adjust if you'd prefer a different shape.

sdeleuze:
@ChoMinGi Please rebase this PR on `main` and make sure the tests introduced by #5978 are still green.

ChoMinGi:
@sdeleuze Rebased. The three tests from #5978 stay green.

Two test alignments followed from your `checkRequired` refinement (e1cc9ebe7):
- `KotlinBeanOutputConverterTests`: default-value Kotlin field assertion flipped to `doesNotContain` (matches your `non-null property with default value should not be required`).
- `ChatClientNativeStructuredResponseTests`: fixture updated for the `format: "int32"` hint now visible through `BeanOutputConverter` after consolidation.

Upgrade notes moved into a new `[[upgrading-to-2-0-0-M7]]` section (M6 has shipped).

tzolov:
LGTM, thanks @ChoMinGi

ilayaperumalg (approved):

ilayaperumalg:
@ChoMinGi Thanks for the PR and the follow up! Rebased, resolved the conflicts, and merged via e548ecbb9

ChoMinGi:
@sdeleuze @tzolov @ilayaperumalg 
Thank you! It was great working on this. I look forward to contributing to spring-ai again when possible.

ilayaperumalg:
@ChoMinGi Thank you once again for the contribution and support! Looking forward to your contributions!

PR #5902, opened 2026-04-29
Restore OpenAI embedding encoding format option

### LINKED ISSUE #5901: Spring AI 2.0.0-M5: Embedding fails with OpenAI-compatible APIs (Base64 string vs float[])
## Bug description

Spring AI `2.0.0-M5` fails when using OpenAI-compatible embedding APIs (e.g. IONOS bge-m3) due to a mismatch in embedding response format.

The error indicates that the embedding is returned as a **string (likely Base64)** instead of a `float[]`, which causes deserialization to fail:
json: cannot unmarshal string into Go struct field Embedding.data.embedding of type []float32

This worked correctly in `2.0.0-M4`.

---

## Environment

- Spring AI: `2.0.0-M5`
- Java: 21
- Spring Boot: 4.0.6
- Vector Store: Qdrant
- Dependencies:
  - `spring-ai-starter-model-openai`
  - `spring-ai-qdrant-store`
  - `spring-ai-rag`
- Provider: IONOS OpenAI-compatible endpoint
- Model: `bge-m3`

---

## Steps to reproduce

1. Configure Spring AI with an OpenAI-compatible endpoint (e.g. IONOS)
2. Use `spring-ai-starter-model-openai`
3. Call embeddings

KoreaNirsa:
## Summary

This restores the OpenAI embedding `encodingFormat` option and passes it through to `EmbeddingCreateParams`

In Spring AI 2.0.0-M5, the OpenAI model implementation migrated to the official OpenAI Java SDK. During that migration, the previous `encodingFormat` option from `OpenAiEmbeddingOptions` was dropped. The OpenAI Java SDK 4.28.0 defaults embedding requests to `base64`

As a result, even when users configure `spring.ai.openai.embedding.options.encoding-format=float`, the actual request is sent with `encoding_format: base64`. This breaks some OpenAI-compatible embedding APIs, such as IONOS bge-m3, where float array embeddings are expected

This change
- Restores `encodingFormat` on `OpenAiEmbeddingOptions`
- Defaults it to `float`
- Passes it to `EmbeddingCreateParams`
- Preserves it across builder `from(...)`, `merge(...)`, and `from(EmbeddingCreateParams)`
- Adds unit coverage for the default value, explicit `base64`, and copy/merge behavior

Fixes gh-5901

## Regression Cause

In `v2.0.0-M4`, `OpenAiEmbeddingOptions` exposed `encodingFormat` and mapped it to the `encoding_format` request field

Evidence
`OpenAiEmbeddingOptions` in `v2.0.0-M4` contains

```java
private @JsonProperty("encoding_format") String encodingFormat;

public String getEncodingFormat()
public void setEncodingFormat(String encodingFormat)

public Builder encodingFormat(String encodingFormat)
```

Source
https://raw.githubusercontent.com/spring-projects/spring-ai/v2.0.0-M4/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiEmbeddingOptions.java

In `v2.0.0-M5`, after the migration to the official OpenAI Java SDK, that option is no longer present in `OpenAiEmbeddingOptions`, and `toOpenAiCreateParams(...)` does not call `builder.encodingFormat(...)`

Evidence
`OpenAiEmbeddingOptions` in `v2.0.0-M5` contains `user` and `dimensions`, but no `encodingFormat`, `getEncodingFormat`, `setEncodingFormat`, or builder `encodingFormat(...)`

Source
https://raw.githubusercontent.com/spring-projects/spring-ai/v2.0.0-M5/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiEmbeddingOptions.java

Therefore, when Spring AI does not explicitly set the value, the SDK default `base64` is used. That is the regression introduced in M5

### Internal Unit Test

```text
Command
.\mvnw.cmd -pl models/spring-ai-openai -Dtest=OpenAiEmbeddingOptionsTests test

Result
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

### Spring Boot Reproducer - Before Fix, 2.0.0-M5

```text
Dependency
org.springframework.ai:spring-ai-starter-model-openai:2.0.0-M5

Application property
spring.ai.openai.api-key=test-key
spring.ai.openai.embedding.options.model=bge-m3
spring.ai.openai.embedding.options.encoding-format=float

Result
AssertionError because the request body did not contain "encoding_format":"float".
```

image

---

### Spring Boot Verification - After Fix, 2.0.0-SNAPSHOT

```text
Dependency
org.springframework.ai:spring-ai-starter-model-openai:2.0.0-SNAPSHOT

Application property
spring.ai.openai.api-key=test-key
spring.ai.openai.embedding.options.model=bge-m3
spring.ai.openai.embedding.options.encoding-format=float
```

image

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiEmbeddingOptions.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiEmbeddingOptions.java
@@ -36,11 +36,18 @@ public class OpenAiEmbeddingOptions extends AbstractOpenAiOptions implements Emb
 
 	public static final String DEFAULT_EMBEDDING_MODEL = EmbeddingModel.TEXT_EMBEDDING_ADA_002.asString();
 
+	public static final String DEFAULT_ENCODING_FORMAT = EmbeddingCreateParams.EncodingFormat.FLOAT.asString();
+
 	/**
 	 * An identifier for the caller or end user of the operation. This may be used for
 	 * tracking or rate-limiting purposes.
 	 */
 	private @Nullable String user;
+
+	/**
+	 * The format to return the embeddings in. Can be either float or base64.
+	 */
+	private @Nullable String encodingFormat = DEFAULT_ENCODING_FORMAT;
 
 	/*
 	 * The number of dimensions the resulting output embeddings should have. Only
@@ -60,6 +67,14 @@ public class OpenAiEmbeddingOptions extends AbstractOpenAiOptions implements Emb
 		this.user = user;
 	}
 
+	public @Nullable String getEncodingFormat() {
+		return this.encodingFormat;
+	}
+
+	public void setEncodingFormat(@Nullable String encodingFormat) {
+		this.encodingFormat = encodingFormat;
+	}
+
 	@Override
 	public @Nullable Integer getDimensions() {
 		return this.dimensions;
@@ -72,7 +87,8 @@ public class OpenAiEmbeddingOptions extends AbstractOpenAiOptions implements Emb
 	@Override
 	public String toString() {
 		return "OpenAiEmbeddingOptions{" + "user='" + this.user + '\'' + ", model='" + this.getModel() + '\''
-				+ ", deploymentName='" + this.getDeploymentName() + '\'' + ", dimensions=" + this.dimensions + '}';
+				+ ", deploymentName='" + this.getDeploymentName() + '\'' + ", encodingFormat='" + this.encodingFormat
+				+ '\'' + ", dimensions=" + this.dimensions + '}';
 	}
 
 	public EmbeddingCreateParams toOpenAiCreateParams(List<String> instructions) {
@@ -93,6 +109,9 @@ public class OpenAiEmbeddingOptions extends AbstractOpenAiOptions implements Emb
 		}
 		if (this.getUser() != null) {
 			builder.user(this.getUser());
+		}
+		if (this.getEncodingFormat() != null) {
+			builder.encodingFormat(EmbeddingCreateParams.EncodingFormat.of(this.getEncodingFormat()));
 		}
 		if (this.getDimensions() != null) {
 			builder.dimensions(this.getDimensions());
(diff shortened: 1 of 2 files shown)


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiEmbeddingOptions.java
```diff
@@ -36,12 +36,21 @@ public class OpenAiEmbeddingOptions extends AbstractOpenAiOptions implements Emb
...
 	/**
 	 * An identifier for the caller or end user of the operation. This may be used for
 	 * tracking or rate-limiting purposes.
 	 */
 	private @Nullable String user;
 
+	/**
+	 * The format to return the embeddings in. Can be either float or base64.
+	 */
+	private @Nullable String encodingFormat = DEFAULT_ENCODING_FORMAT;
```
sdeleuze: I think I would suggest to align on OpenAI SDK default by not setting some defaults on Spring AI side, and document this in the upgrade notes. If that's the case, we can remove the `encodingFormatSet` field.

@tzolov Thoughts?

sdeleuze:
Thanks for this PR, I think we are going to keep the default to the OpenAI SDK one (Base64) and document that in the upgrade notes.

Also I am tempted to expose this option with the strongly typed `EmbeddingCreateParams.EncodingFormat` enum provided by OpenAI SDK rather than one we artificially introduced at Spring AI level.

See my refined implementation at https://github.com/sdeleuze/spring-ai/tree/openai-embedding-encoding-format.

KoreaNirsa:
@sdeleuze Thanks for the feedback and for sharing the refined implementation.

I hadn’t considered that approach, but I agree it is more appropriate. Aligning with the OpenAI SDK default and using `EmbeddingCreateParams.EncodingFormat` directly makes the implementation cleaner.

I’ve applied the suggested changes. Could you please take another look when you have a chance?

Thanks!

sdeleuze:
Refined and merged (using my original commit). After discussing with the team, we chose to use a Spring AI enum instead.

Next time @KoreaNirsa, probably better to cherry-pick additional commits keeping the author info, etc. Thanks for your contribution.

PR #5923, opened 2026-05-03
Remove setters from `OllamaChatOptions`

sdeleuze:
- Remove `@NestedConfigurationProperty` from `OllamaChatProperties`
- Expose options directly under the `spring.ai.ollama.chat` prefix instead of `spring.ai.ollama.chat.options`
- Implement Spring Boot deprecation mechanism for the legacy `options` nested properties to gracefully guide users during upgrades


ericbottard:
This looks good to me.
In addition to removing the setters in options (the long term goal), it also removes the `.options.` segment which was here for the wrong reasons in the first place.
The migration path for users is seamless. Of course, this will need to be applied to all models.

ericbottard (approved):

PR #5927, opened 2026-05-03
Improve tool calling observations

### LINKED ISSUE #5926: Improve tool calling observations
When executing tool calls via the `ToolCallingManager`, observations are created based on the `ToolCallingObservationContext` and `ToolCallingObservationConvention`.

I'd like to suggest some improvements:

- Introduce a dedicated "operation type" for tool calling: "execute_tool". Currently, such operations are generally marked as "framework" operations, which is not very specific.
- Add two new high cardinality attributes to the tool calling observations to capture the tool call id (important for tracking the tool call origin and mapping it to the right tool among the ones provided to the model) and the tool type (by default it would be "function", but it could vary when integrating MCP and agentic workflows).
- Fix a bug when tool call arguments are an empty string. In that case, the empty JSON object string is used by the `DefaultToolCallingManager`. However, in the observation, we capture an empty string instead of  “{}”.

ThomasVitale:
- Introduce “execute_tool” operation type, which is now used in tool calling observations instead of “framework”. As a consequence, tool calling spans are now named “execute_tool <tool-name>” and the “gen_ai.operation.name” attribute is set to “execute_tool” instead of “framework”. Before, we were using the SpringAIKind in the observation name instead of AiOperationType, which is what all the other observations do. The reason was due to the lack of dedicated operation type value. That is now fixed, too.
- Add new attributes in tool calling observations to capture tool call type (low cardinality, “spring.ai.tool.type”) and tool call id (high cardinality, “spring.ai.tool.call.id”)
- Fix bug when tool call arguments are an empty string. Before, it would use the empty string. Now, it correctly sets it to an empty JSON object string: “{}”, which is consistent with how the arguments are actually parsed in the DefaultToolCallingManager.

Fixes gh-5926

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/observation/DefaultToolCallingObservationConvention.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/observation/DefaultToolCallingObservationConvention.java
@@ -19,6 +19,7 @@ package org.springframework.ai.tool.observation;
 import io.micrometer.common.KeyValue;
 import io.micrometer.common.KeyValues;
 
+import org.springframework.ai.observation.conventions.AiOperationType;
 import org.springframework.ai.observation.conventions.SpringAiKind;
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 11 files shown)


sdeleuze:
@tzolov Could you please qualify this PR?

spring-ai-model/src/main/java/org/springframework/ai/tool/observation/DefaultToolCallingObservationConvention.java
```diff
@@ -51,15 +52,19 @@ public String getName() {
 	public String getContextualName(ToolCallingObservationContext context) {
 		Assert.notNull(context, "context cannot be null");
 		String toolName = context.getToolDefinition().name();
-		return "%s %s".formatted(SpringAiKind.TOOL_CALL.value(), toolName);
+		return "%s %s".formatted(AiOperationType.EXECUTE_TOOL.value(), toolName);
```
tzolov, line 55: This change if expected, breaks the DefaultToolCallingManagerIT#observationForToolCall() (line 94).

tzolov (changes requested):
Thanks @ThomasVitale. Useful update. 
- Please fix the DefaultToolCallingManagerIT test failures. 

Also can you please fix the `spring.ai.tool.call.result` description in observability/index.adoc. It is an unrelated to copy/paste error i believe. The right description should be something like "The result returned from the tool call execution. (Only when enabled)". Otherwise i will fix it with a followup commit

spring-ai-model/src/main/java/org/springframework/ai/tool/observation/DefaultToolCallingObservationConvention.java
ThomasVitale: Ooh let me fix that

ThomasVitale: Fixed

ThomasVitale:
@tzolov I fixed the test and updated the docs for `spring.ai.tool.call.result`.

tzolov (approved):
LGTM

tzolov:
rebased and merged at 5845155bf10cbe64e0724faadee1a18d79e51f5a

PR #5929, opened 2026-05-04
Preserve OpenAI response metadata

### LINKED ISSUE #5922: Preserve unmapped root-level provider fields in ChatResponseMetadata.map
**Bug description**
Spring AI drops several root-level provider-specific fields when mapping a ChatCompletion response into `ChatResponse`.

For OpenAI-compatible providers (e.g., OpenRouter), fields such as `provider`, `system_fingerprint`, and other top-level attributes present in the raw response are not propagated into `ChatResponseMetadata`.

Although `ChatResponseMetadata` already provides a flexible `map` (via `AbstractResponseMetadata`) for arbitrary metadata, these root-level fields are currently not stored there, resulting in loss of useful provider information.

---

**Environment**

* Spring AI version: 1.0.x / 1.1.x
* Java version: 17 / 21
* Provider: OpenRouter (OpenAI-compatible API)
* Model: `moonshotai/kimi-k2.5-0127`
* Integration: `ChatModel.call()`
* No vector store used

---

**Steps to reproduce**

1. Configure Spring AI with an OpenAI-compatible endpoint (e.g., OpenRouter).

2. Execute:

```java id="t7xk2p"
ChatResponse response = chatModel.call(
    new Prompt("hi")
);
```

3. Inspect metadata:

```java id="q92ldk"
ChatResponseMetadata metadata = response.getMetadata();

System.out.println(metadata.get("provider"));           // null
System.out.println(metadata.get("system_fingerprint")); // null
```

4. Compare with actual provider response (below)

---

**Provider response (actual)**

```json id="r3k9d1"
{
  "id": "gen-1888888888-XYZabc123NewId",
  "object": "chat.completion",
  "created": 1777799928,
  "model": "moonshotai/kimi-k2.5-0127",
  "provider": "Together",
  "system_fingerprint": null
}
```

---

**Expected behavior**
All additional **root-level fields** from the provider response should be preserved and made accessible via `ChatResponseMetadata.map`.

Example:

```java id="l92kd1"
metadata.get("provider");
metadata.get("system_fingerprint");
metadata.get("created");
```

Currently only created field is present and others root-level fields are silently dropped during mapping.

---

**Proposed improvement**
Since `ChatResponseMetadata` already extends `AbstractResponseMetadata` and includes:

```java id="z8x1qp"
protected final Map<String, Object> map = new ConcurrentHashMap<>();
```

Spring AI can:

* Move all **unmapped root-level fields** into `metadata.map`
* Keep existing structured fields (`id`, `model`, `usage`, etc.) unchanged
* Avoid breaking changes while improving observability and extensibility

---

```java id="w91kdl"
@SpringBootTest
class ChatResponseMetadataTest {

    @Autowired
    ChatModel chatModel;

    @Test
    void shouldExposeRootLevelFields() {
        ChatResponse response = chatModel.call(
                new Prompt("hi")
        );

        ChatResponseMetadata metadata = response.getMetadata();

        assertNotNull(metadata);

        // Expected but currently missing
        assertTrue(metadata.containsKey("provider"));
        assertTrue(metadata.containsKey("system_fingerprint"));
    }
}
```

yaner-here:
Fixes https://github.com/spring-projects/spring-ai/issues/5922. Preserves unmapped root-level OpenAI chat completion response fields in `ChatResponseMetadata`.

The OpenAI SDK exposes provider-specific root fields through `ChatCompletion._additionalProperties()`. These fields were previously dropped when Spring AI mapped `ChatCompletion` into `ChatResponseMetadata`. This change copies those additional properties into the metadata map.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -64,6 +64,7 @@ import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.scheduler.Schedulers;
 import tools.jackson.databind.JsonNode;
+import tools.jackson.databind.ObjectMapper;
 
 import org.springframework.ai.chat.messages.AssistantMessage;
 import org.springframework.ai.chat.messages.MessageType;
@@ -116,6 +117,8 @@ public final class OpenAiChatModel implements ChatModel {
 	private static final ChatModelObservationConvention DEFAULT_OBSERVATION_CONVENTION = new DefaultChatModelObservationConvention();
 
 	private static final ToolCallingManager DEFAULT_TOOL_CALLING_MANAGER = ToolCallingManager.builder().build();
+
+	private static final ObjectMapper objectMapper = new ObjectMapper();
 
 	private final Logger logger = LoggerFactory.getLogger(OpenAiChatModel.class);
 
@@ -552,21 +555,37 @@ public final class OpenAiChatModel implements ChatModel {
 		Assert.notNull(result, "OpenAI ChatCompletion must not be null");
 		result.model();
 		result.id();
-		return ChatResponseMetadata.builder()
+		ChatResponseMetadata.Builder metadataBuilder = ChatResponseMetadata.builder()
 			.id(result.id())
 			.usage(usage)
 			.model(result.model())
-			.keyValue("created", result.created())
-			.build();
+			.keyValue("created", result.created());
+
+		result._additionalProperties().forEach((key, jsonValue) -> {
+			try {
+				String jsonValueString = objectMapper.writeValueAsString(jsonValue);
+				Object value = ModelOptionsUtils.JSON_MAPPER.readValue(jsonValueString, Object.class);
+				metadataBuilder.keyValue(key, value);
+			}
+			catch (Exception e) {
+				logger.error("Error parsing JSON value for key '{}': {}", key, jsonValue, e);
+				metadataBuilder.keyValue(key, jsonValue);
+			}
+		});
+
+		return metadataBuilder.build();
 	}
 
 	private ChatResponseMetadata from(ChatResponseMetadata chatResponseMetadata, Usage usage) {
 		Assert.notNull(chatResponseMetadata, "OpenAI ChatResponseMetadata must not be null");
-		return ChatResponseMetadata.builder()
+		ChatResponseMetadata.Builder builder = ChatResponseMetadata.builder()
 			.id(chatResponseMetadata.getId())
 			.usage(usage)
-			.model(chatResponseMetadata.getModel())
-			.build();
+			.model(chatResponseMetadata.getModel());
+
+		chatResponseMetadata.entrySet().forEach(e -> builder.keyValue(e.getKey(), e.getValue()));
+
+		return builder.build();
 	}
 
 	/**
@@ -618,6 +637,7 @@ public final class OpenAiChatModel implements ChatModel {
 			.model(chunk.model())
 			.usage(chunk.usage()
 				.orElse(CompletionUsage.builder().promptTokens(0).completionTokens(0).totalTokens(0).build()))
+			.putAllAdditionalProperties(chunk._additionalProperties())
 			.build();
 	}
 
--- a/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java
+++ b/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java
@@ -16,18 +16,34 @@
 
 package org.springframework.ai.openai;
 
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
 import com.openai.client.OpenAIClient;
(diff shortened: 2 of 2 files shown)


ilayaperumalg (commented):
Thanks for the PR! It's a great catch to preserve these additional properties.

However, there is one significant issue with leaking SDK-specific types into Spring AI's generic abstractions, and one minor refactoring suggestion:

1. **Leaking `com.openai.core.JsonValue`:**
By doing `result._additionalProperties().forEach(metadataBuilder::keyValue);`, the map is populated with OpenAI SDK's `JsonValue` objects. As shown in the added test (`metadata.<JsonValue>get(\non_null_field\)`), this forces users to cast values to an OpenAI-specific class. This breaks Spring AI's portability—if a user switches to another provider, their code expecting `JsonValue` will break. 

**Recommendation:** Unwrap the `JsonValue` into standard Java types (String, Number, Boolean, List, Map) before adding them. You can use Jackson for this:
```java
result._additionalProperties().forEach((key, jsonValue) -> {
    try {
        Object unmarshalled = ModelOptionsUtils.JSON_MAPPER.readValue(jsonValue.toString(), Object.class);
        metadataBuilder.keyValue(key, unmarshalled);
    } catch (Exception e) {
        metadataBuilder.keyValue(key, jsonValue.toString());
    }
});
```

2. **Minor Map Copying Simplification:**
In `from(ChatResponseMetadata chatResponseMetadata, Usage usage)`, creating a temporary map via Streams is fine, but you could simplify it by iterating directly over the entries:
```java
ChatResponseMetadata.Builder builder = ChatResponseMetadata.builder()
    .id(chatResponseMetadata.getId())
    .usage(usage)
    .model(chatResponseMetadata.getModel());

chatResponseMetadata.entrySet().forEach(e -> builder.keyValue(e.getKey(), e.getValue()));

return builder.build();
```

yaner-here:
Thanks for the review! All suggestions have been addressed.

---

### Slight implementation difference

There's a slight difference. OpenAI's `JsonValue.toString()` returns non-standard JSON string. For example, `JsonValue.from(Map.of("key1", 1, "key2", 2)).toString()` actually returns `{key2=2, key1=1}`, which is expected to be `{"key2": 2, "key1": 1}`, thus can't be handled by Jackson. 

```
tools.jackson.core.exc.StreamReadException: Unrecognized token 'key2': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); byte offset: #UNKNOWN]
	at tools.jackson.core.JsonParser._constructReadException(JsonParser.java:1856)
	at tools.jackson.core.json.ReaderBasedJsonParser._reportInvalidToken(ReaderBasedJsonParser.java:3130)
	at tools.jackson.core.json.ReaderBasedJsonParser._handleOddValue(ReaderBasedJsonParser.java:2075)
	at tools.jackson.core.json.ReaderBasedJsonParser.nextToken(ReaderBasedJsonParser.java:809)
	at tools.jackson.databind.deser.jdk.UntypedObjectDeserializerNR._deserializeNR(UntypedObjectDeserializerNR.java:261)
	at tools.jackson.databind.deser.jdk.UntypedObjectDeserializerNR.deserialize(UntypedObjectDeserializerNR.java:65)
	at tools.jackson.databind.deser.DeserializationContextExt.readRootValue(DeserializationContextExt.java:266)
	at tools.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:2639)
```

So I use `ObjectMapper` to get the valid JSON string.

---

### Another better implementation?

BTW, `JsonValue.convert(Object.class)` could **perfectly** converting itself to Java types. For example, `JsonValue.from(List.of("abc", 123, Map.of("key", "value"))).convert(Object.class)` returns the expected nested instance.
image

Would `JsonValue.convert(Object.class)` be better? I would appreciate your thoughts on this.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -117,6 +118,8 @@ public final class OpenAiChatModel implements ChatModel {
 
 	private static final ToolCallingManager DEFAULT_TOOL_CALLING_MANAGER = ToolCallingManager.builder().build();
 
+	private static final ObjectMapper objectMapper = new ObjectMapper();
```
ilayaperumalg, line 121: Can we reuse the JSON_MAPPER from ModelOptionsUtils?

ilayaperumalg (approved):

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
ilayaperumalg: @yaner-here The changes look good except the ObjectMapper usage. I can update when merging.

yaner-here: @ilayaperumalg OK, thanks a lot ❤️

ilayaperumalg:
@yaner-here Thank you for the PR and the follow up updates! Rebased and merged via 145b250f49ce0d16d5e0447018fb12fabc796cb4. Added a change to replace the ObjectMapper with `ModelOptionsUtils.JSON_MAPPER.convertValue` via 4ace53c194625881bbaf6b0fa848953eeb4e979c

PR #5935, opened 2026-05-04
Switch to JDK 17.0.19 to build the project

sdeleuze:
With the recent inclusion of `-XDaddTypeAnnotationsToSymbol` in JDK 17, it is now possible to use JDK 17.0.19 to compile the project and get NullAway checks working as expected.

We use this opportunity to switch to Liberica distribution which is the preferred distribution for all Spring projects.

Build and CI configuration are updated accordingly.

This commit also upgrades NullAway to 0.13.4 and updates the documentation in README.md to explain those requirements and how to use SDKMAN! to install and configure the right JDK.

Once merged, I will update the `1.1.x` and `1.0.x` branches to use the same Liberica JDK 17.0.19 version.

### DIFF
--- a/.github/workflows/artifactory-milestone-release.yml
+++ b/.github/workflows/artifactory-milestone-release.yml
@@ -18,8 +18,8 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v5
         with:
-          java-version: '21'
-          distribution: 'temurin'
+          java-version: '17'
+          distribution: 'liberica'
           cache: 'maven'
 
       - name: Capture release version
--- a/pom.xml
+++ b/pom.xml
@@ -342,9 +342,8 @@
 		<antlr.version>4.13.1</antlr.version>
 		<maven-compiler-plugin.version>3.14.1</maven-compiler-plugin.version>
 		<maven-enforcer-plugin.version>3.6.2</maven-enforcer-plugin.version>
-		<!-- The version (range) of the JDK we want to use to compile.
-			This is different from ${java.version} which we use as -release target. -->
-		<compiler.jdk.version>[21.0.8,)</compiler.jdk.version>
+		<!-- The version (range) of the JDK we want to use to compile -->
+		<compiler.jdk.version>[17.0.19,)</compiler.jdk.version>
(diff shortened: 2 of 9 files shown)


pom.xml
```diff
@@ -365,8 +364,8 @@
 		<spring-javaformat-checkstyle.version>0.0.47</spring-javaformat-checkstyle.version>
 		<maven-gpg-plugin.version>3.2.8</maven-gpg-plugin.version>
 
-		<error-prone.version>2.46.0</error-prone.version>
-		<nullaway.version>0.13.0</nullaway.version>
+		<error-prone.version>2.42.0</error-prone.version>
```
sdeleuze, line 367: 2.42.0 is the latest Error Prone version supporting JDK 17.

.github/workflows/artifactory-milestone-release.yml
```diff
@@ -18,8 +18,8 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v5
         with:
-          java-version: '21'
```
ericbottard, line 21: There are other CI workflows that need to be updated to align with jdk 17.
And we need to make sure that liberica 17 on github actually gives us 17.0.19+ (I guess we'll see but there is no reason it doesn't).

In particular, pr-check.yaml needs editing and it applies to all branches, not just main that you're editing here. 

Otherwise LGTM on paper.

sdeleuze: Good catch, I have updated remaining ones and plan to apply those changes on all branches once review accepted (so please don't merge, I will do it).

sdeleuze:
I have tested locally that build is green after `killall java` to kill any daemon and after removing the compilation cache.

ericbottard (approved):
LGTM let's proceed and see what happens

ilayaperumalg (approved):

ilayaperumalg:
Rebased and merged as 935d79d6b

sdeleuze:
Branches `1.1.x` and  `1.0.x` have been updated to use Liberica JDK 17.0.19 has well.

PR #5936, opened 2026-05-05
Document HTTP-layer observability gap for Anthropic and OpenAI chat models

### LINKED ISSUE #5916: Document Anthropic and OpenAI don't provide HTTP metrics/traces nor context propagation anymore
After the switch to the Anthropic SDK in 2.0.0-M2 and the OpenAI SDK in 2.0.0-M5, the application will not provide any metrics or traces for the HTTP interactions happening under the hood between the `AnthropicChatModel`/`OpenAiChatModel` and the inference service. They are also not propagating the `traceId` any longer, so it's not possible to correlate spans all the way down to the inference services.

I would suggest highlighting that in the documentation and release notes for 2.0.0, since it's a breaking change for applications running in production and relying on that telemetry. It can be especially problematic for the Spring AI OpenAI module when used in enterprise settings relying on AI Gateways and/or using OpenAI API-compatible inference systems, for which it won't be possible to trace requests through any longer, due to the lack of context propagation.

The Anthropic SDK doesn't provide the possibility to customize the underlying OkHttpClient. @salaboy submitted a PR (https://github.com/anthropics/anthropic-sdk-java/pull/315) to address that, but it hasn't received any answer yet. The OpenAI SDK seems to have a similar issue.

Micrometer ships with instrumentation for OkHttp. I hope that Spring AI can re-use that in order to re-introduce out-of-the-box HTTP instrumentation for both Anthropic and OpenAI, once the two SDKs make the OkHttpClient configurable.

Tip for developers reading this and in need for HTTP observability: consider bringing your own instrumented HTTP client instead of the built-in OkHttp module. See also: https://github.com/spring-projects/spring-ai/issues/5838

sobychacko:
Fixes: https://github.com/spring-projects/spring-ai/issues/5916

- Vendor SDKs build their own OkHttpClient and don't expose an interceptor hook, so HTTP metrics, spans, and traceparent propagation are missing
- Add a note to both Anthropic and OpenAI chat docs
- Cross-reference from the observability index

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/anthropic-chat.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/anthropic-chat.adoc
@@ -1589,8 +1589,28 @@ spring.ai.anthropic.chat.options.web-search-tool.user-location.country=US
 
 == Observability
 
-The Anthropic SDK implementation supports Spring AI's observability features through Micrometer.
-All chat model operations are instrumented for monitoring and tracing.
+Spring AI emits Micrometer metrics and tracing spans at the chat model layer for every Anthropic call.
+Each span wraps the SDK invocation and carries request parameters, response metadata, and token usage.
+See the xref:observability/index.adoc[observability reference] for the full set.
+
+[IMPORTANT]
+====
+**HTTP layer is not instrumented**
+
+Since `2.0.0-M3`, Anthropic chat goes through the official `anthropic-java` SDK.
+The SDK builds its own `OkHttpClient` and doesn't expose an interceptor hook on it, so Spring AI has no way to attach the Micrometer or OpenTelemetry interceptors that would instrument the HTTP request.
+You won't see:
+
+* HTTP client metrics (per-endpoint latency, status codes, retries)
+* A span for the HTTP call itself, between Spring AI and the Anthropic API
+* Outbound `traceparent` propagation to the API or any intermediary (proxy, AI gateway, OpenAI-compatible inference service)
+
+Before `2.0.0-M3` the `RestClient` based code picked up HTTP instrumentation from Spring's HTTP stack for free.
+The switch to the SDK removed that side effect.
+
+If you need HTTP-level observability today, the https://github.com/open-telemetry/opentelemetry-java-instrumentation[OpenTelemetry Java agent] is the simplest path.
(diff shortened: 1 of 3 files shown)


spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chat/anthropic-chat.adoc
```diff
@@ -1589,8 +1589,28 @@ spring.ai.anthropic.chat.options.web-search-tool.user-location.country=US
...
+You won't see:
+
+* HTTP client metrics (per-endpoint latency, status codes, retries)
+* A span for the HTTP call itself, between Spring AI and the Anthropic API
+* Outbound `traceparent` propagation to the API or any intermediary (proxy, AI gateway, OpenAI-compatible inference service)
+
+Before `2.0.0-M3` the `RestClient` based code picked up HTTP instrumentation from Spring's HTTP stack for free.
+The switch to the SDK removed that side effect.
+
+If you need HTTP-level observability today, the https://github.com/open-telemetry/opentelemetry-java-instrumentation[OpenTelemetry Java agent] is the simplest path.
```
ThomasVitale: I'm afraid this won't work as the OpenTelemetry Java Agent is typically mutually exclusive with the Spring Observability stack. So you either use the Micrometer-based approach or the Agent.

It might be possible to turn off all features in the Agent except the OkHttp instrumentation, but even if that worked, it would only be for traces (and maybe context propagation, but it would need to be tested) as Spring doesn't support the OpenTelemetry Metrics API, so no metrics would be exported. 

If the plan is to address this issue in 2.0.0-M1, I would consider removing this suggestion. Or else I would probably suggest referring to the SDK docs on how to implement a custom http client module and then suggest to customize the spring ai integration to use that custom module.

sdeleuze: I would remove this suggestion in all cases, this is a current limitation not easy to fix.

sobychacko: @ThomasVitale @sdeleuze @ilayaperumalg Thanks for the review. I updated the docs; please take a look.

ilayaperumalg (approved):

ilayaperumalg:
Squashed, Rebased and merged via db95798ad

PR #5960, opened 2026-05-06
Remove OpenAI AI setters from options

sdeleuze:
- Remove `@NestedConfigurationProperty` from `OpenAiChatOptions`
- Expose options directly under the `spring.ai.openai.chat` prefix instead of `spring.ai.openai.chat.options`
- Implement Spring Boot deprecation mechanism for the legacy `options` nested properties to gracefully guide users during upgrades


ilayaperumalg:
The class `AbstractOpenAiOptions` still has some setter methods.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatOptions.java
```diff
@@ -651,8 +441,10 @@ public void setToolContext(Map<String, Object> toolContext) {
 	@Override
 	public void setOutputSchema(@Nullable String outputSchema) {
```
ilayaperumalg, line 428: I think this setter needs to be addressed (removed) via StructuredOutputOptions?

ericbottard: Indeed this one needs to stay until https://github.com/spring-projects/spring-ai/pull/5957 is merged.

We'll address all those once these PRs Sebastien is working on are merged

auto-configurations/models/spring-ai-autoconfigure-model-openai/src/main/java/org/springframework/ai/model/openai/autoconfigure/OpenAiChatProperties.java
```diff
@@ -16,15 +16,24 @@
...
+import org.springframework.boot.context.properties.DeprecatedConfigurationProperty;
 
 /**
  * OpenAI SDK Chat autoconfiguration properties.
  *
  * @author Christian Tzolov
+ * @author Sebastien Deleuze
  */
 @ConfigurationProperties(OpenAiChatProperties.CONFIG_PREFIX)
 public class OpenAiChatProperties extends AbstractOpenAiOptions {
```
ericbottard: We should consider breaking that inheritance relation at some later point (not a breaking change from user POV)

auto-configurations/models/spring-ai-autoconfigure-model-openai/src/main/java/org/springframework/ai/model/openai/autoconfigure/OpenAiChatProperties.java
```diff
@@ -33,11 +42,638 @@ public class OpenAiChatProperties extends AbstractOpenAiOptions {
...
+	}
+
+	public void setInternalToolExecutionEnabled(@Nullable Boolean internalToolExecutionEnabled) {
+		this.internalToolExecutionEnabled = internalToolExecutionEnabled;
+	}
+
+	public OpenAiChatOptions toOptions() {
+		OpenAiChatOptions.Builder builder = OpenAiChatOptions.builder();
+		builder.model(this.getModel());
+		if (this.frequencyPenalty != null) {
```
ericbottard, line 321: Given this is a fresh builder, we can actually skip all those null checks

auto-configurations/models/spring-ai-autoconfigure-model-openai/src/main/java/org/springframework/ai/model/openai/autoconfigure/OpenAiChatProperties.java
sdeleuze: I agree, and tried since I did it for Anthropic, but it was too involved so I chose to defer that to a later PR.

sdeleuze: After discussing with @ilayaperumalg I will give it another try.

auto-configurations/models/spring-ai-autoconfigure-model-openai/src/main/java/org/springframework/ai/model/openai/autoconfigure/OpenAiChatProperties.java
```diff
@@ -16,28 +16,669 @@
...
+	@Deprecated(since = "2.0.0", forRemoval = true)
+	public Options getOptions() {
 		return this.options;
 	}
 
+	public void setOptions(Options options) {
+		this.options = options;
+	}
+
+	public class Options {
```
ilayaperumalg, line 414: These options don't include the properties from AbstractOpenAiOptions (which are now getting deprecated). For instance, the property `spring.ai.openai.chat.base-url` is no longer supported/deprecated.

ilayaperumalg:
OpenAiBindingsPropertiesProcessor and TanzuBindingsPropertiesProcessor need to be fixed with the changes.

ilayaperumalg:
Can we add a note at the upgrade-notes on the changes introduced as part of these ongoing changes with the deprecations across the models?

spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
```diff
@@ -4,6 +4,13 @@
 [[upgrading-to-2-0-0-M6]]
 == Upgrading to 2.0.0-M6
 
+=== Options and Properties
+
+The configuration properties do not use anymore `.option`, so for example the former `spring.ai.openai.chat.options.model` is now `spring.ai.openai.chat.model`.
```
ilayaperumalg: This needs to be`.options`. Will fix when merging.

sdeleuze: Good catch, fixed.

ericbottard: Also, the wording should be "do not use `.options` anymore", instead of "do not use anymore `.options`"

ilayaperumalg (approved):

ilayaperumalg:
Rebased and merged without squashing the commits to make it easier to go through the changes.

PR #5975, opened 2026-05-08
feat: allow custom modules in BeanOutputConverter JSON schema generation

### LINKED ISSUE #794: Generate complex Json schemas with BeanOutputConverter
**Expected Behavior**

Be able to generate more complex schemas with the `BeanOutputConverter` class, for instance with constrains definitions like max elements in array and max or min length for strings. This can be handled adding support to the definition of what modules should be loaded for _jsonschema-generator_ library. Could be allowing override the private method `generateSchema` or adding it as a constructor parameter.

**Current Behavior**

Right now the `BeanOutputConverter` is adding just the _jackson-module_ to the configuration of _jsonschema-generator_ library, this is a great starting point to have a schema definition but does not allow the definition of more complex scenarios. The code to generate the value for jsonSchema field is private so we cannot override it to adding more modules to the libray.

**Context**

Under some circumstances we can ask to the AI model to generate content with specific rules, for instance if we want to generate specific number of answers for one question. In that case we can specify it with the prompt but probably could be a good idea add it as a constraint to the expected schema just to be sure the AI model respect it. This can be handled with the maven library _jsonschema-module-jakarta-validation_. I tested locally and just add it to the  _jsonschema-generator_ library configuration it's enough to add constraints to the generated schema.

EvanYao826:
## Summary

Allow users to register additional victools `jsonschema-generator` modules in `BeanOutputConverter` for custom JSON schema generation.

### Problem

`BeanOutputConverter` currently hardcodes only the `JacksonSchemaModule` and `Swagger2Module` during schema generation. The `generateSchema()` method is `private` with no way to extend it. This prevents users from adding modules like `JakartaValidationModule` to derive schema constraints (`@Size`, `@Min`, `@Max`, `@NotNull`, etc.) from Jakarta Validation annotations.

### Solution

- Added a `Builder` pattern (`BeanOutputConverter.builder(Class)`) with a `withModule(Module)` method for registering additional schema modules
- Added constructor overloads accepting `Module...` varargs for simpler use cases
- All existing constructors remain unchanged — fully backward compatible
- No new dependencies required in `spring-ai-model`

### Usage Example

```java
// Using Builder (recommended for multiple modules)
BeanOutputConverter<MyBean> converter = BeanOutputConverter.builder(MyBean.class)
    .withModule(new JakartaValidationModule())
    .build();

// Using constructor varargs (simple single-module case)
BeanOutputConverter<MyBean> converter = new BeanOutputConverter<>(MyBean.class, new JakartaValidationModule());
```

Closes #794

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
@@ -17,9 +17,13 @@
 package org.springframework.ai.converter;
 
 import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
+import com.github.victools.jsonschema.generator.Module;
 import com.github.victools.jsonschema.generator.Option;
 import com.github.victools.jsonschema.generator.SchemaGenerator;
 import com.github.victools.jsonschema.generator.SchemaGeneratorConfig;
@@ -80,12 +84,26 @@ public class BeanOutputConverter<T> implements StructuredOutputConverter<T> {
 	/** The text cleaner used to preprocess LLM responses before parsing. */
 	private final ResponseTextCleaner textCleaner;
 
+	/** Additional jsonschema-generator modules to apply during schema generation. */
+	private final List<Module> extraModules;
+
 	/**
 	 * Constructor to initialize with the target type's class.
 	 * @param clazz The target type's class.
 	 */
 	public BeanOutputConverter(Class<T> clazz) {
-		this(clazz, null, null);
+		this(clazz, null, null, Collections.emptyList());
+	}
+
+	/**
+	 * Constructor to initialize with the target type's class and extra modules for JSON
+	 * schema generation.
+	 * @param clazz The target type's class.
+	 * @param extraModules Additional {@link Module} instances (e.g.
+	 * {@code JakartaValidationModule}) to apply during schema generation.
+	 */
+	public BeanOutputConverter(Class<T> clazz, Module... extraModules) {
+		this(clazz, null, null, List.of(extraModules));
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


ThomasVitale:
The upcoming changes in https://github.com/spring-projects/spring-ai/pull/5897 will make it possible to customize the JSON Schema generation logic in BeanOutputConverter by overriding the `generateSchema()` method. Would that help with your use case?

EvanYao826:
Thanks for the pointer! Yes, #5897's approach of making `generateSchema()` protected would allow achieving similar functionality through subclassing. However, I believe these two approaches are complementary:

1. **Ergonomics**: Our `Builder.withModule()` approach allows one-line customization without subclassing:
   ```java
   BeanOutputConverter.builder(MyBean.class)
       .withModule(new JakartaValidationModule())
       .build();
   ```
   vs. the subclass approach which requires creating a full custom class.

2. **Composability**: Multiple modules can be stacked easily via `withModule()` calls.

3. **Compatibility**: If #5897 merges first, we'd be happy to rebase and adapt the `withModule()` API to work with the new `JsonSchemaGenerator` delegation — the Builder can inject modules into whichever schema generation mechanism is used internally.

That said, if the maintainers feel #5897 alone is sufficient, we're happy to close this PR. Happy to adapt either way!

EvanYao826:
Hi @ThomasVitale, thanks again for the feedback on #5897. I've updated the PR description to clarify how the Builder approach complements the protected `generateSchema()` method. Would love to hear your thoughts on whether both approaches could coexist, or if you'd prefer I close this in favor of #5897. Happy to adjust either way!

EvanYao826:
Hi @ThomasVitale, following up on this PR. I noticed the codebase has been significantly refactored since this was opened — specifically, `generateSchema()` was extracted to `JsonSchemaGenerator.generateForType()`, making the inline module approach in this PR incompatible with the current API.

Would you prefer I:
1. Rework this PR to extend `JsonSchemaGenerator` with module support, or
2. Close this PR and open a new one with the updated approach?

Happy to adapt to whatever direction makes sense for the project.

PR #5976, opened 2026-05-08
OllamaChatOptions getOutputSchema() throws exception

ThomasVitale:
The OllamaChatOptions implementation breaks the nullability contract in the parent StructuredOutputOptions interface, throwing an exception instead of returning null if there's no format specified.


ThomasVitale:
I noticed that the native structured output integration tests for Ollama are currently failing on the main branch (so, not related to the changes in this PR). I opened a separate issue to look into that: https://github.com/spring-projects/spring-ai/issues/5977

tzolov (approved):
Nice catch. Thanks @ThomasVitale

Issue #1818, opened 2024-11-25
OpenAiAutoConfiguration does not recognize the API Key configured via property key: 'spring.ai.openai.image.api-key'


hardikSinghBehl:
**Bug description**
When only an `ImageModel` bean is needed, while configuring the OpenAI API key via the below property:

```yaml
spring:
  ai:
    openai:
      image:
        api-key: ${OPENAI_API_KEY}
```
I got the below error:

```bash
Caused by: java.lang.IllegalArgumentException: OpenAI API key must be set. Use the connection property: spring.ai.openai.api-key or spring.ai.openai.chat.api-key property.
```

The error persisted when I tried disabling the chat client autoconfiguration using:

```yaml
spring:
  ai:
    chat:
      client:
        enabled: false
```
Ended up using the `spring.ai.openai.api-key` property to resolve the startup exception but the former key should've worked as well.

**Environment**
Using the Spring AI version : 1.0.0-M4 with Java 23

**Expected behavior**
The expectation of having the API key recognized when configured through the property `spring.ai.openai.image.api-key` was not met.

**Steps to reproduce and Minimal Complete Reproducible example**
TODO

emredmrcan:
@markpollack can I check and contribute to this issue?

hungrytech:
@emredmrcan  Is it still unresolved? If so, can I contribute to solving it?

emredmrcan:
> [@markpollack](https://github.com/markpollack) [@emredmrcan](https://github.com/emredmrcan) Is it still unresolved? If so, can I contribute to solving it?

Hello @hungrytech , I haven't received a reply yet so If they approve your contribution request, please feel free to go

thelightway24:
Hello @emredmrcan , @hungrytech 
It seems there hasn’t been an update on this issue for a while.  
Has this been resolved?
If not and no one’s working on it, I'd be happy to work on it.

sdeleuze:
Superseded by #5989.

PR #5990, opened 2026-05-09
Fix OpenAI auto-config when only sub-prefix `api-key` is set

### LINKED ISSUE #5989: OpenAI auto-config startup fails when only a sub-prefix credential is set
## Background

While digging into #1818 (`spring.ai.openai.image.api-key` alone fails the context), I traced the failure to how OpenAI auto-configuration gates its model beans, not to anything image-specific. The same shape of bug applies to chat, embedding, audio.speech, audio.transcription, and moderation — `image` is just the first one that surfaced because that's the path the
original reporter took.

I'm filing this as a separate issue (with a follow-up PR) because the fix touches all six sub-configurations and introduces a small new `Condition`, which feels closer to a refactor than a single-bean bug fix. The change closes #1818 as a consequence. Happy to fold it back into #1818 if you'd prefer that flow.

## Reproduction

```yaml
# application.yml — startup fails with only this set
spring:
  ai:
    openai:
      image:
        api-key: sk-xxx
```

```
java.lang.IllegalStateException: At least one credential source must be specified:
    credential (apiKey), workloadIdentity, or adminApiKey
    at com.openai.core.ClientOptions$Builder.effectiveCredential(ClientOptions.kt:503)
    ... (wrapped by Spring as BeanCreationException during context startup)
```

Substitute `chat`, `embedding`, `audio.speech`, `audio.transcription`, or `moderation` for `image` above and the failure is identical — only the failing sub-configuration changes. Programmatic credentials, the `OPENAI_API_KEY` env var, and Microsoft Foundry passwordless setups exhibit the same shape: one valid credential source, five other
sub-configurations still try to register and fail fast.

## Current behavior

The six OpenAI auto-configurations gate registration only on `spring.ai.openai.api-key` (the common prefix), even though Spring AI already accepts several other valid credential sources at the SDK level:

- `spring.ai.openai.<chat|image|embedding|audio.speech|audio.transcription|moderation>.api-key`
- a programmatic `Credential` set on `OpenAiCommonProperties` or any of the sub-`*Properties`
- `OPENAI_API_KEY` env var (default OpenAI endpoint)
- Microsoft Foundry endpoints, where the api-key may legitimately be absent because passwordless auth is used. Detected by any of:
    - `spring.ai.openai.microsoft-foundry=true`
    - `base-url` ending in `openai.azure.com` or `cognitiveservices.azure.com`
    - `microsoft-deployment-name`, `deployment-name`, or `microsoft-foundry-service-version` set

In all of these cases, all six `@Configuration` classes still register; `OpenAiSetup` then fails fast on the sub-configurations that have no usable credential, and the failure takes the whole application context down before the working sub-configuration ever boots.

## Why a custom Condition

The first thing I tried was widening the existing `@ConditionalOnProperty` — the same shape that's been attempted before:

```java
@ConditionalOnProperty(
    prefix = "spring.ai.openai",
    name = { "api-key", "image.api-key" },
    matchIfMissing = false
)
```

That doesn't work: Spring Boot's `name = {a, b}` semantics is AND, not OR — it matches only when both keys are set, which is the opposite of what we need. The condition needs OR-of-credential-sources logic, which I don't see a way to express declaratively without a custom `SpringBootCondition`.

## Proposed change

A single `OpenAiConnectionCondition` utility with one inner condition per sub-configuration (`Chat`, `Image`, `Embedding`, `AudioSpeech`, `AudioTranscription`, `Moderation`). Each one independently checks the credential sources listed above with short-circuit OR semantics. If none match, the corresponding `@Bean` does not register, and the rest of the context boots normally.

The conditions sit on the `@Bean` method, not the `@Configuration` class, so:

- `@EnableConfigurationProperties(OpenAi*Properties)` still runs — the properties bean stays available for other code that injects it.
- The existing model-selector `@ConditionalOnProperty(spring.ai.model.<kind>)` keeps working unchanged.

A working branch with this change and 13 tests covering each credential source per sub-configuration is ready (all green on the `spring-ai-autoconfigure-model-openai` module). I'll open the PR right after this issue if the direction looks reasonable.

## Scope / non-goals

- No public API changes, no new dependencies.
- No behavior change for users who already set `spring.ai.openai.api-key` — that branch matches first, exactly as today.
- Closes #1818 as a side effect.

jewoodev:
Fix OpenAI auto-configuration so each model bean is created only when a usable credential source exists for that bean. The #1818 image case is one example of the same gating issue across chat, embedding, image, audio speech, audio transcription, and moderation: a sub-prefix key, programmatic `Credential`, `OPENAI_API_KEY`, GitHub Models token, or Foundry passwordless setup could still let unrelated OpenAI beans start and fail in `OpenAiSetup`.

This adds a single `OnAvailableOpenAiConnection` condition on the model `@Bean` methods. It preserves the existing `spring.ai.model.<kind>` selector and property binding behavior, while resolving the sub-prefix from the bean method's `AbstractOpenAiProperties` parameter.

Regression tests cover common and per-model credentials, no-credential startup, selector opt-out, user-defined chat model backoff, env-var fallback, and Foundry passwordless auth. Scope is limited to the OpenAI auto-configurations.

Closes #5989


auto-configurations/models/spring-ai-autoconfigure-model-openai/pom.xml
```diff
@@ -128,6 +128,13 @@
...
 			<scope>test</scope>
 		</dependency>
+
+		<dependency>
+			<groupId>uk.org.webcompere</groupId>
+			<artifactId>system-stubs-jupiter</artifactId>
+			<version>2.1.7</version>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
```
jewoodev, line 126: After re-checking the new test dependency, I think it is justified.

`OpenAiSetup` lives in `models/spring-ai-openai`, has no Spring imports, and is also used by plain Java builders, so its env-var fallback necessarily goes through `System.getenv()`. The auto-config condition should mirror that; if it read `OPENAI_API_KEY` through Spring `Environment`, a property-only value could make the condition match while `OpenAiSetup` still fails.

`system-stubs-jupiter` is only used to exercise that direct env-var branch in `OpenAiConnectionConditionAutoConfigurationTests`.

If a new test dependency is a blocker, two acceptable fallbacks are:
- use `junit-pioneer` instead
- drop the condition-level env-var test and rely on existing `OpenAiSetupTests` coverage

jewoodev (commented):
I'll check this new dependency is necessary one more time. ➝ The detailed check is complete!

jewoodev:
Superseded by the update below: I initially asked whether to keep six explicit condition wrappers or try the lighter single-condition shape. I pushed the single-condition variant in `7564522` so it can be reviewed directly.

jewoodev:
Pushed the single-condition variant in `7564522` for direct comparison. It replaces the six wrapper conditions with `OnAvailableOpenAiConnection`, resolves the sub-prefix from the bean method's `AbstractOpenAiProperties` parameter, and moves credential checks to `AbstractOpenAiProperties`.

Trade-off: the sub-prefix is inferred instead of named at the `@Conditional` call site. Adding a new OpenAI sub-bean now only needs `@Conditional(OnAvailableOpenAiConnection.class)`. Feedback on which shape fits Spring AI better would be helpful.

PR #5993, opened 2026-05-10
GH-5987: fix OpenAI chat streaming latency without tools

### LINKED ISSUE #5987: `OpenAiChatModel#stream` buffers the entire response before emitting chunks
**Bug description**

    if (list.isEmpty()) {
        return Flux.empty();
    }

        .filter(Objects::nonNull)

    if (!hasToolCalls) {
    }

    // tool call handling...
});`
**Environment**
Spring AI version:
2.0.0-M5
2.0.0-M6

Spring Boot 4.0.6

**Steps to reproduce**

1. Create an OpenAiChatModel.
Example:
        .model("gpt-4o-mini")
        .build();

        .options(options)
        .build();

        .doOnNext(response -> {
        })
        .blockLast();
`
**Expected behavior**

yaner-here:
Fix GH-5987.

OpenAI chat streaming responses were being collected with `collectList()` before being emitted downstream, so even normal non-tool streaming requests only delivered chunks after the upstream SSE stream completed. This happened because the same aggregation path was used to detect and merge streaming tool calls.

This PR lets requests without configured tools stream responses immediately while still aggregating the final response for observation metadata.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -77,6 +77,7 @@ import org.springframework.ai.chat.metadata.Usage;
 import org.springframework.ai.chat.model.ChatModel;
 import org.springframework.ai.chat.model.ChatResponse;
 import org.springframework.ai.chat.model.Generation;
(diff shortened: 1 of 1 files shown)


sdeleuze:
Does not look like a proper fix to me, will reopen if I am wrong.

PR #5996, opened 2026-05-10
fix: parse large integer constants as Long in FilterExpressionTextParser

### LINKED ISSUE #4705: FilterExpressionTextParser treats all number literals as Integer, failing on Long values

EvanYao826:
Fixes #4705

## Summary

`FilterExpressionTextParser.visitIntegerConstant` uses `Integer.valueOf()` which throws `NumberFormatException` for values exceeding `Integer.MAX_VALUE` (2147483647). This prevents using filter expressions with `long` metadata IDs.

## Change

Fall back to `Long.parseLong()` when the integer literal overflows the 32-bit range:

```java
// before

// after
String text = ctx.getText();
try {
} catch (NumberFormatException ex) {
}
```

- Values within `Integer` range continue to produce `Integer` (backward compatible)
- The existing `L`-suffixed long syntax is unaffected

## Testing

Added `testLargeIntegerFallsBackToLong()` covering:
- `Long.MAX_VALUE` without L suffix parsed as `Long`
- `Long.MIN_VALUE` without L suffix parsed as `Long`
- Values within Integer range still parsed as `Integer`

All 16 tests pass (including 15 existing + 1 new).

## Notes

Two prior PRs (#5431 and #5770) were closed without merge for the same issue.

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
@@ -230,7 +230,13 @@ public class FilterExpressionTextParser {
 
 		@Override
 		public Filter.Operand visitIntegerConstant(FiltersParser.IntegerConstantContext ctx) {
-			return new Filter.Value(Integer.valueOf(ctx.getText()));
+			String text = ctx.getText();
+			try {
+				return new Filter.Value(Integer.parseInt(text));
+			}
+			catch (NumberFormatException ex) {
+				return new Filter.Value(Long.parseLong(text));
+			}
 		}
 
 		@Override
--- a/spring-ai-vector-store/src/test/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParserTests.java
+++ b/spring-ai-vector-store/src/test/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParserTests.java
@@ -204,6 +204,23 @@ public class FilterExpressionTextParserTests {
 	}
 
 	@Test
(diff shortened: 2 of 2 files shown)


EvanYao826:
Hi! Gentle ping on this fix. I noticed two prior PRs (#5431, #5770) were closed for the same issue (large integer overflow in FilterExpressionTextParser). This fix uses a try-catch fallback approach that's fully backward compatible — values within Integer range still produce Integer. Would love to get your feedback. Thanks!

EvanYao826:
Hi maintainers! Gentle follow-up ping on this PR. It's been open for about a month with no review feedback yet.

This is a small, focused fix that addresses a real issue. Happy to rebase or adjust if needed. Could someone take a look when convenient? Thanks!

EvanYao826:
Hi, gentle ping on this PR — it fixes a bug where large integer constants (exceeding `Integer.MAX_VALUE`) cause a `NumberFormatException` in `FilterExpressionTextParser`, preventing the use of long metadata IDs in filter expressions.

This issue has had two prior PR attempts (#5431, #5770) that were closed without merge, yet the underlying bug persists. This version falls back to `Long.parseLong()` while remaining fully backward-compatible with values in the Integer range.

Would appreciate a review. Thank you!

Issue #5978, opened 2026-05-08
Tool input schema marks Kotlin nullable fields as required in `JsonSchemaGenerator`


jeanchristophecascade:
### Bug description

Every `val ... = null` property of a Kotlin data class used inside a `@McpTool` parameter (or as the input/output of structured-output conversion) is emitted with `required: true` in the generated JSON Schema, regardless of nullability or default value.

The two schema generators involved are:
- `org.springframework.ai.util.json.schema.JsonSchemaGenerator` (spring-ai-model)
- `org.springframework.ai.mcp.annotation.method.tool.utils.McpJsonSchemaGenerator` (spring-ai-mcp-annotations)

Both initialise their victools `SchemaGenerator` with `JacksonSchemaModule + Swagger2Module + SpringAiSchemaModule` only — `org.springframework.ai.model.KotlinModule` (shipped via PR #1766 to resolve #1667 for `ModelOptionsUtils`) does not appear to be registered in either of these newer paths.

This is consistent with the fragmentation flagged in #5915 and the sibling cases in #4487 / #5341.

### Environment

Spring AI 2.0.0-M5, Kotlin 2.3, JDK 25, `spring-ai-starter-mcp-server-webmvc`.

### Steps to reproduce

```kotlin
data class Filter(val name: String? = null, val ids: List<String>? = null)

@McpTool(description = "...")
fun search(@McpToolParam(required = false) filter: Filter? = null): String = ""
```

`tools/list` returns:

```json
{
  "inputSchema": {
    "type": "object",
    "properties": {
      "filter": {
        "type": "object",
        "properties": { "ids": { }, "name": { } },
        "required": ["ids", "name"]
      }
    },
    "required": []
  }
}
```

The top-level `required: []` is correct (param-level honours `@McpToolParam(required = false)` and `@org.jspecify.Nullable`). The nested `required: ["ids", "name"]` is the issue.

### Expected behavior

`required: []` (or absent) for the nested object — both fields are Kotlin-nullable constructor parameters with default values. This matches what `KotlinModule#isRequired` already computes:

```java
boolean isNonNullable = !returnType.isMarkedNullable();
boolean isInConstructor = constructorParamsWithoutDefault.contains(name);
return isNonNullable && isInConstructor;
```

### One possible approach

Registering `KotlinModule` in each generator's static initialiser when Kotlin is on the classpath:

```java
SchemaGeneratorConfigBuilder builder = new SchemaGeneratorConfigBuilder(...)
[...]

PR #5997, opened 2026-05-11
Fix Kotlin nullable fields in MCP tool input schema

qixiangme:
## Summary
This PR is related to #5978 and fixes the MCP Tools schema generation path for Kotlin nullable fields. 

PR #5897 addresses the standard `@Tool` schema path by registering `org.springframework.ai.model.KotlinModule` in `JsonSchemaGenerator`. MCP tools use `McpJsonSchemaGenerator` for method input schemas, so Kotlin nullable properties inside MCP tool parameter data classes could still be emitted as required in nested JSON Schema objects.

This change applies the same conditional `KotlinModule` registration pattern to `McpJsonSchemaGenerator`, keeping the fix narrowly scoped to MCP tool input schema generation.

## Tests

The regression test verifies both sides of Kotlin requiredness behavior:

- nullable Kotlin constructor properties with default null values, such as `String? = null` and `List<String>? = null`, are not included in nested `required` arrays
- non-null Kotlin constructor properties without defaults remain required

Commands run:

```bash
./mvnw -pl mcp/mcp-annotations '-Denforcer.skip=true' 'kotlin:test-compile@test-compile'
./mvnw -pl mcp/mcp-annotations -Dtest=McpJsonSchemaGeneratorKotlinTests '-Denforcer.skip=true' surefire:test
./mvnw -pl mcp/mcp-annotations -Dtest=McpJsonSchemaGeneratorKotlinTests '-Denforcer.skip=true' package
```

Related issue: Addresses the MCP Tools portion of #5978.

### DIFF
--- a/mcp/mcp-annotations/pom.xml
+++ b/mcp/mcp-annotations/pom.xml
@@ -67,6 +67,18 @@
 			<scope>test</scope>
 		</dependency>
 
+		<dependency>
+			<groupId>org.jetbrains.kotlin</groupId>
+			<artifactId>kotlin-stdlib</artifactId>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>org.jetbrains.kotlin</groupId>
+			<artifactId>kotlin-reflect</artifactId>
+			<scope>test</scope>
+		</dependency>
+
 	</dependencies>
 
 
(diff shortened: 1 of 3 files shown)


sdeleuze:
FYI The error is caused by a `javac` behavior described in https://github.com/openjdk/jdk/pull/28018, I will refine to fix it.

qixiangme:
Thanks for the clarification. I'll keep an eye on the updated fix.

PR #6013, opened 2026-05-12
Make ToolCallAdvisor the default tool execution mechanism

### LINKED ISSUE #5459: Make ToolCallAdvisor the default took call management option
Currently, the tool-call lifecycle is managed internally by each concrete ChatModel implementation. 

The ToolCallAdvisor allows you to manage this lifecycle as a part of the Advisor chains. The latter provides clear visibility for the tool request and response messages, allowing manipulation and storing of those messages in Memory. This is a requirement for many AI and Agentic use cases.

When the ToolCallAdvisor is added to the chain (e.g. ChatClient.advisors(ToolCallAdsvior.builder()...), it disables the low-level ChatMemory tool handling implementations. The proposal is to auto-register the ToolCallAdvisor by default when there are registered ChatClient Tools or tool callbacks.

Also a the toolcall advisor should disable its internal conversation memory support if there are ChatMemory advisors registered after it in the chain.

tzolov:
Resolves #5459  by making the ToolCallAdvisor the default tool execution mechanism

- **Auto-register `ToolCallAdvisor`**: When tools are configured on a `ChatClient` call, a `ToolCallAdvisor` is automatically added to the advisor chain — no manual wiring needed. Auto-registration is suppressed when the chain already contains a `ToolCallHandlingAdvisor`.
- **Propagate `ToolCallingManager`**: `ChatClient.builder()` and `DefaultChatClientBuilder` now accept and propagate a `ToolCallingManager` so the auto-registered advisor uses the same manager injected into the application context.
- **Deprecate internal `ChatModel` tool execution**: `internalToolExecutionEnabled` flag handling, builder setters (`toolCallingManager`, `toolExecutionEligibilityPredicate`), and the `ToolExecutionEligibilityPredicate` interface are all deprecated since 2.0.0 for removal in 3.0.0 across OpenAI, Anthropic, Google GenAI, Ollama, Bedrock Converse, Mistral AI, MiniMax, and DeepSeek models. A one-time runtime warning is logged when the deprecated path is triggered.
- **Docs**: Updated `tools.adoc`, `chatclient.adoc`, and `advisors.adoc` to document the new auto-registration behaviour and deprecations.

The auto-registration logic is implemented in the https://github.com/spring-projects/spring-ai/pull/6013/changes#diff-1c346f07e7f4cfdf523ec42a8cade77dcb481def7d045f03e1ff91db6ebf0fee

#### Tests

- `ToolCallAdvisorAutoRegistrationTests` — auto-registration logic, suppression via `ToolCallHandlingAdvisor`, `AdvisorParams` helpers
- `DefaultChatClientTests` — `ToolCallingManager` preservation through `mutate()`
- `DefaultChatClientBuilderTests` — builder propagation
- `ChatClientAutoConfigurationTests` — auto-config wires the injected `ToolCallingManager`


tzolov:
TODO - perhaps we should pass a default a ToolCallHandlingAdvisor to the DefaultChatClient (instead of ToolCallingManger) to use as a default ToolCallAdvisor instance to auto-register. 
This will allow use to register custom `ToolCallHandlingAdvisor` such as ToolSearchToolCallAdvisor as spring beans.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
```diff
@@ -1035,6 +1070,69 @@ private BaseAdvisorChain buildAdvisorChain() {
...
+		 */
+		private void autoRegisterToolCallAdvisor(boolean streaming) {
+
+			boolean autoRegisterDisabled = Boolean.FALSE
+				.equals(this.advisorParams.get(ChatClientAttributes.TOOL_CALL_ADVISOR_AUTO_REGISTER.getKey()));
+			if (autoRegisterDisabled) {
+				return;
+			}
+
+			boolean hasTools = !this.toolCallbacks.isEmpty() || !this.toolCallbackProviders.isEmpty()
```
ilayaperumalg, line 1091: This currently takes only the ChatClient request properties, but not the tool options set via ToolCallingChatOptions (toolNames and toolCallbacks).

ilayaperumalg:
> perhaps we should pass a default a ToolCallHandlingAdvisor to the DefaultChatClient (instead of ToolCallingManger) to use as a default ToolCallAdvisor instance to auto-register.

I think it's a better approach. This way, ChatClient doesn't need to know about the ToolCallingManager. ChatClient only manages the AdvisorChain and the advisor handles the tools - would be a nice separation of concerns.

ilayaperumalg (approved):

ilayaperumalg:
Rebased, squashed and merged via 2bbd955969ee22e003f61cc03a08044da2bf7c44

PR #6022, opened 2026-05-13
Fix Google GenAI Text Embedding missing `taskType`

### LINKED ISSUE #5966: Google GenAI Text Embedding does not pass task type to the API
**Bug description**

`GoogleGenAiTextEmbeddingModel` accepts `task-type` via YAML configuration and `GoogleGenAiTextEmbeddingOptions`, but never passes it to the underlying Google GenAI SDK when
building the `EmbedContentConfig`.

In `GoogleGenAiTextEmbeddingModel.java`, the `call()` method builds an `EmbedContentConfig` and sets `outputDimensionality` when specified, but the task type is left as a TODO
comment:

```java
EmbedContentConfig.Builder configBuilder = EmbedContentConfig.builder();

if (options.getDimensions() != null) {
    configBuilder.outputDimensionality(options.getDimensions());
}

// Set task type if specified - this might need to be handled differently
// as the new SDK might not have a direct taskType field
// We'll need to check the SDK documentation for this

EmbedContentConfig config = configBuilder.build();
```

The Google GenAI SDK (com.google.genai:google-genai) does support task type on EmbedContentConfig.Builder:

configBuilder.taskType("RETRIEVAL_QUERY");

**Environment**

- Spring AI: 1.1.5 (also present in 1.1.4 and on main)
- Google GenAI SDK: 1.51.0
- EmbedContentConfig.Builder.taskType(String) is available in the SDK

**Steps to reproduce**
Generate embeddings for any text with different `spring.ai.google.genai.embedding.text.options.task-type` set and verify they are the same vectors.

**Expected behavior**

When `spring.ai.google.genai.embedding.text.options.task-type` is configured (or set via `GoogleGenAiTextEmbeddingOptions.builder().taskType(...)`), the value should be passed to `EmbedContentConfig.Builder.taskType()`.

Suggested fix:
After the dimensions block, add:

```java
if (options.getTaskType() != null) {
    configBuilder.taskType(options.getTaskType().name());
}
```

jewoodev:
### Summary

Forward the configured `taskType` into `EmbedContentConfig` so `GoogleGenAiTextEmbeddingModel#call` stops silently falling back to the SDK default.

Closes #5966.

- **Why**: `buildEmbeddingRequest` merges `taskType` into the runtime options correctly, but `call` never passes it to `EmbedContentConfig.Builder`, so every request goes out with the SDK's default task type regardless of what the user set.

- **Fix**: Pass `options.getTaskType().name()` to `EmbedContentConfig.Builder#taskType(String)` when a task type is set, mirroring the existing `outputDimensionality` block right above it.

### Tests

Added regression coverage in `GoogleGenAiTextEmbeddingModelTests` for:
- an explicit per-request `taskType`
- the `RETRIEVAL_DOCUMENT` default from `initializeDefaults()`

```bash
./mvnw -pl models/spring-ai-google-genai-embedding test \
    -Dtest='GoogleGenAiTextEmbeddingModelTests'
```

### DIFF
--- a/models/spring-ai-google-genai-embedding/src/main/java/org/springframework/ai/google/genai/text/GoogleGenAiTextEmbeddingModel.java
+++ b/models/spring-ai-google-genai-embedding/src/main/java/org/springframework/ai/google/genai/text/GoogleGenAiTextEmbeddingModel.java
@@ -60,6 +60,7 @@ import org.springframework.util.StringUtils;
  * @author Soby Chacko
  * @author Dan Dobrin
  * @author Sebastien Deleuze
+ * @author Jewoo Shin
  * @since 1.0.0
  */
 public class GoogleGenAiTextEmbeddingModel extends AbstractEmbeddingModel {
@@ -151,9 +152,10 @@ public class GoogleGenAiTextEmbeddingModel extends AbstractEmbeddingModel {
 					configBuilder.outputDimensionality(options.getDimensions());
 				}
 
-				// Set task type if specified - this might need to be handled differently
-				// as the new SDK might not have a direct taskType field
-				// We'll need to check the SDK documentation for this
+				// Set task type if specified
+				if (options.getTaskType() != null) {
+					configBuilder.taskType(options.getTaskType().name());
+				}
 
 				EmbedContentConfig config = configBuilder.build();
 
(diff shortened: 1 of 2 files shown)


jewoodev:
Gentle ping on this PR.

This is still mergeable and all checks are green. The scope is intentionally small:

- forwards the already-merged `taskType` option into `EmbedContentConfig`
- adds unit coverage for an explicit `taskType`
- adds unit coverage for the default `RETRIEVAL_DOCUMENT` task type

Would a maintainer be able to triage/review this when convenient?

PR #6027, opened 2026-05-13
GH-6023: Fix IndexOutOfBoundException when choice.index() != list position

### LINKED ISSUE #6023: IndexOutOfBoundsException in OpenAiChatModel.internalStream when choice.index() != list position
**Bug description**
`OpenAiChatModel.internalStream()` throws `IndexOutOfBoundsException` when a streaming chunk contains a choice whose `index` field value does not match its position in the `choices` array.

At [OpenAiChatModel.java:322](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java#L322), the code uses the API-level `choice.index()` field to index into the raw `chunk.choices()` list:

```java
"chunkChoice", chunk.choices().get((int) choice.index())
```

The `choice.index()` value is the **semantic index** from the OpenAI API response (e.g., `0` or `1`), not the positional index within the `choices` array. When a chunk contains a single choice with `index: 1`, `chunk.choices()` has length 1, but `chunk.choices().get(1)` is called — causing the `IndexOutOfBoundsException`.

**Environment**
- **Spring AI version**: 2.0.0-M6
- **Java version**: 25
- **LLM provider**: OpenAI-compatible API gateway

**Steps to reproduce**
1. Configure `OpenAiChatModel` with `stream_options.include_usage: true` (the default for streaming)
2. Use an OpenAI-compatible API gateway that proxies non-OpenAI models
3. Send a streaming chat completion request
4. The gateway sends chunks where a choice's `index` field may not match its position in the `choices` array

**Expected behavior**
Streaming should complete successfully regardless of the `index` field values in choices.

## Actual Behavior

```
java.lang.IndexOutOfBoundsException: Index 1 out of bounds for length 1
	at java.base/java.util.ArrayList.get(ArrayList.java:428)
	at org.springframework.ai.openai.OpenAiChatModel.lambda$internalStream$6(OpenAiChatModel.java:322)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:214)
	at java.base/java.util.Collections$2.tryAdvance(Collections.java:5182)
	at java.base/java.util.Collections$2.forEachRemaining(Collections.java:5190)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:570)
	...
```

## Root Cause Analysis

In `OpenAiChatModel.internalStream()`, when building generation metadata:

```java
List<Generation> generations = chatCompletion.choices().stream().map(choice -> {
    // ...
    Map<String, Object> metadata = Map.of(
        // ...
        "chunkChoice", chunk.choices().get((int) choice.index())  // BUG: uses API index as list position
    );
    return buildGeneration(choice, metadata, request);
}).toList();
```

`chunkToChatCompletion(chunk)` converts chunk choices to `ChatCompletion.Choice` objects, preserving the original `index` field from the API response. The code then iterates over `chatCompletion.choices()` and uses `choice.index()` to look up the original chunk choice from `chunk.choices()`.

This breaks when `choice.index()` (API field) ≠ list position, which can happen when:
- The gateway sends chunks with non-zero-based or non-contiguous choice indices
- A usage-only final chunk has fewer choices than expected
- The gateway sends choices in non-sequential order

## Suggested Fix

Use the stream iteration index (list position) instead of the API `choice.index()` field to look up the corresponding chunk choice:

```java
List<ChatCompletion.Choice> choiceList = chatCompletion.choices();
List<Generation> generations = IntStream.range(0, choiceList.size()).mapToObj(i -> {
    ChatCompletion.Choice choice = choiceList.get(i);
    // ...
    Map<String, Object> metadata = Map.of(
        // ...
        "chunkChoice", chunk.choices().get(i)  // use list position, not API index
    );
    return buildGeneration(choice, metadata, request);
}).toList();
```

Or alternatively, build a lookup map from `index` → chunk choice:

```java
Map<Long, ChatCompletionChunk.Choice> chunkChoicesByIndex = chunk.choices().stream()
    .collect(Collectors.toMap(ChatCompletionChunk.Choice::index, Function.identity()));

// Then in the mapping:
"chunkChoice", chunkChoicesByIndex.get(choice.index())
```

shounak-26:
Fixes #6023 

## Problem
- OpenAiChatModel.java:322 internalStream() used choice.index() as a positional index into chunk.choices() list
- When a gateway sends a chunk, where choice.index() == 1, but the choices list has only 1 element, which caused IOBE

## Fix
- Build a Hashmap inside chunk.choices() is gettting stored based on there index & corresponding object
- So when look was happening from hasmap the correct object based on the index ID value was getting mapped

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import com.openai.client.OpenAIClient;
@@ -310,6 +311,7 @@ public final class OpenAiChatModel implements ChatModel {
 					try {
 						ChatCompletion chatCompletion = chunkToChatCompletion(chunk);
 						String id = chatCompletion.id();
+						Map<Long, ChatCompletionChunk.Choice> chunkChoiceByIndex = chunk.choices().stream().collect(Collectors.toMap(ChatCompletionChunk.Choice::index, Function.identity()));
 						List<Generation> generations = chatCompletion.choices().stream().map(choice -> {
 							roleMap.putIfAbsent(id, choice.message()._role().asString().isPresent()
 									? choice.message()._role().asStringOrThrow() : "");
@@ -319,7 +321,7 @@ public final class OpenAiChatModel implements ChatModel {
 									choice.message().refusal().isPresent() ? choice.message().refusal() : "",
 									"annotations", choice.message().annotations().isPresent()
 											? choice.message().annotations() : List.of(),
-									"chunkChoice", chunk.choices().get((int) choice.index()));
+									"chunkChoice", chunkChoiceByIndex.get(choice.index()));
 
 							return buildGeneration(choice, metadata, request);
 						}).toList();


shounak-26:
Hi please review this @tzolov, @sdeleuze

sobychacko:
Thanks for this fix, @shounak-26. This targets the same `IndexOutOfBoundsException` in `OpenAiChatModel.internalStream()` for issue #6023 as PR #6221, which also changes the same line so that `choice.index()` is no longer used as a positional list lookup. We are going ahead with #6221 since it includes a regression test and uses a positional lookup that aligns with the ordering already guaranteed by `chunkToChatCompletion()`. One neutral note on the approach here: collecting into a `Map` keyed by `choice.index()` via `Collectors.toMap` would raise an error if two choices ever shared the same index. We really appreciate you digging into this, and we hope you will consider contributing again on future issues.

PR #6029, opened 2026-05-13
Fix MessageWindowChatMemory to drop orphaned ToolResponseMessages after window eviction

### LINKED ISSUE #5940: MessageWindowChatMemory truncation breaks tool_use/tool_result pairs (Anthropic 400)
## Summary

`MessageWindowChatMemory` performs naïve count-based head truncation when the message count exceeds `maxMessages`. The cut point is not aware of `AssistantMessage` (with `tool_use` blocks) ↔ `ToolResponseMessage` (with `tool_result` blocks) pairs. When the cut falls between such a pair, the assistant message is dropped but the tool result survives — and the next request to the provider fails because the conversation now starts with an orphan `tool_result`.

## Reproducer

Anthropic's API enforces this strictly:

> 400 invalid_request_error
> messages.0.content.0: unexpected `tool_use_id` found in `tool_result` blocks: toolu_bdrk_…. Each `tool_result` block must have a corresponding `tool_use` block in the previous message.

In our environment (Spring AI 2.0.0-M5, Anthropic Claude Sonnet, `MessageWindowChatMemory` with `maxMessages=100`, tool-heavy ReAct conversations) we hit this every few hours on long-lived threads. A single user turn often produces 4–10 tool calls, so 100 messages is reached after only ~15–20 user turns. Once the window starts sliding, any cut that lands inside a tool_use→tool_result pair corrupts the conversation.

A self-contained reproducer with `MessageWindowChatMemory.builder().maxMessages(4).build()`:

```java
// Pretend the LLM made a tool call in turn 1 and we kept appending.
chatMemory.add(conversationId, List.of(
    new UserMessage("u1"),
    AssistantMessage.builder().content("").toolCalls(List.of(toolCall("t1"))).build(),
    new ToolResponseMessage(List.of(new ToolResponse("t1", "tool_a", "result_a"))),
    AssistantMessage.builder().content("done").build(),
    new UserMessage("u2")));   // window slides; messages[0] is now ToolResponseMessage(tool_use_id=t1)
                               // but the AssistantMessage that contained t1 was dropped.
```

`chatMemory.get(conversationId)` returns a list whose head is a `ToolResponseMessage` referencing a tool_use that no longer exists in the history.

## Why the bot's recovery isn't enough

Anthropic returns the 400 *every* request until the corrupt state is cleared. Our app catches `BadRequestException` and resets the entire conversation — which is annoying and loses context the user expects to be there. We can write our own pair-aware wrapper, but a fix at the framework layer would benefit every user.

## Suggested fix

`MessageWindowChatMemory` (or a pair-aware sibling) should, after computing the window cut, advance the cut forward while `messages[0] instanceof ToolResponseMessage` (or while `messages[0]` is a `tool_result`-bearing message whose matching `tool_use` is no longer in the kept window). Equivalent forms:

- Keep cutting from the head while the first kept message is a `ToolResponseMessage`.
- Or: track tool_use_id → tool_use position and never cut between a `tool_use` and its matching `tool_result`.

Either approach prevents the orphan-tool_result corruption without changing the public API. The same idea applies to OpenAI's `tool` messages — same constraint, same wire-format requirement.

## Related but distinct issues

- #2101 — tool calls not persisted at all (different bug)
- #2529 — tool calls duplicated in memory (different bug)
- #3366 — streaming mode loses `toolCalls` on `AssistantMessage` (different bug)
- #4315 — tool messages not stored when using `.stream.chatClientResponse()` (different bug)

This issue is specifically about *truncation* in the persisted store breaking pairs.

## Environment

- Spring AI 2.0.0-M5
- `MessageWindowChatMemory` + `RedisChatMemoryRepository` (Redis JSON)
- Anthropic Claude Sonnet 4.6
- Java 25, Spring Boot 4.0

suryateja-g13:
## Problem

\`MessageWindowChatMemory\` truncates the message list from the head when the window size is exceeded. However, it does not account for the structural constraint that every \`ToolResponseMessage\` must be preceded by an \`AssistantMessage\` containing the matching tool-use call.

When the window evicts exactly enough messages to remove an \`AssistantMessage\` that contained a tool-use block, the corresponding \`ToolResponseMessage\` is left at the start of the retained list — as an orphan with no preceding tool call. Sending this to an LLM causes an API error (e.g. Anthropic returns \`400: tool_use block must be followed by tool_result\`).

### Reproduction scenario

\`\`\`
Window limit = 3, per-add saves (as done by MessageChatMemoryAdvisor):
  add([UserMsg1])             → [UserMsg1]
  add([AssistantMsg(tool)])   → [UserMsg1, AssistantMsg(tool)]
  add([ToolResponseMsg])      → [UserMsg1, AssistantMsg(tool), ToolResponseMsg]   ← at limit
  add([AssistantMsg_final])   → trim 1 → [AssistantMsg(tool), ToolResponseMsg, AssistantMsg_final]

Next turn:
  add([UserMsg2])             → trim 1 → [ToolResponseMsg, AssistantMsg_final, UserMsg2]
                                           ^^^ orphaned — no preceding tool_use → Anthropic 400
\`\`\`

## Solution

After the standard head-truncation, scan the retained list for any \`ToolResponseMessage\` entries that appear before the first non-system, non-tool-response message. Drop all of them — they are orphans whose \`AssistantMessage(tool_use)\` partner was evicted.

\`\`\`java
private static List<Message> dropOrphanedToolResponses(List<Message> messages) {
    // skip leading SystemMessages (always preserved)
    // drop all consecutive ToolResponseMessages at the head
    // (their AssistantMessage partner was evicted by the window trim)
}
\`\`\`

## Changes

- \`MessageWindowChatMemory.java\`: call \`dropOrphanedToolResponses()\` after head-truncation
- \`MessageWindowChatMemoryTests.java\`: four new tests

## Tests

All existing 18 tests continue to pass. Four new tests added:

- \`orphanedToolResponseIsDroppedWhenToolUseEvicted\` — single orphan
- \`multipleConsecutiveOrphanedToolResponsesAreAllDropped\` — two back-to-back orphans from a multi-tool-call turn
- \`toolPairPreservedWhenBothFitInWindow\` — complete pair survives eviction
- \`systemMessagePreservedWhenOrphanedToolResponseDropped\` — system message stays at position 0
- \`loneOrphanedToolResponseWithMaxMessagesOneResultsInEmptyHistory\` — edge case: single orphan leaves empty history

## Known limitation

\`dropOrphanedToolResponses\` only catches orphans created by **FIFO head-first eviction** — the only eviction strategy currently implemented. If a future strategy evicts an \`AssistantMessage(tool_call)\` from a non-leading position, its \`ToolResponseMessage\` would not be detected.

After orphan drop, the conversation may start with an \`AssistantMessage\` (clean text, no tool calls) if the user message that preceded it was also evicted in the same cycle. This is a separate concern — the specific error reported in \#5940 (\`tool_result\` with no matching \`tool_use\`) is fully addressed here.

Closes #5940

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
@@ -23,6 +23,7 @@ import java.util.Set;
 
 import org.springframework.ai.chat.messages.Message;
 import org.springframework.ai.chat.messages.SystemMessage;
+import org.springframework.ai.chat.messages.ToolResponseMessage;
 import org.springframework.util.Assert;
 
 /**
@@ -108,7 +109,36 @@ public final class MessageWindowChatMemory implements ChatMemory {
 			}
 		}
 
-		return trimmedMessages;
+		return dropOrphanedToolResponses(trimmedMessages);
+	}
+
+	/**
+	 * Drop any {@link ToolResponseMessage} instances at the head of the non-system
+	 * portion of the list whose paired {@code AssistantMessage} tool-use was evicted by
+	 * the window trim. Leaving such messages causes providers (e.g. Anthropic) to reject
+	 * the request with a 400 because the {@code tool_result} has no matching
+	 * {@code tool_use} in the preceding turn.
+	 */
+	private static List<Message> dropOrphanedToolResponses(List<Message> messages) {
+		// Find where non-system messages begin
+		int firstNonSystem = 0;
+		while (firstNonSystem < messages.size() && messages.get(firstNonSystem) instanceof SystemMessage) {
+			firstNonSystem++;
+		}
+
+		// Advance past any ToolResponseMessages at the head (their tool_use was evicted)
+		int cutFrom = firstNonSystem;
+		while (cutFrom < messages.size() && messages.get(cutFrom) instanceof ToolResponseMessage) {
+			cutFrom++;
+		}
+
+		if (cutFrom == firstNonSystem) {
+			return messages;
+		}
+
+		List<Message> result = new ArrayList<>(messages.subList(0, firstNonSystem));
+		result.addAll(messages.subList(cutFrom, messages.size()));
+		return result;
 	}
 
 	public static Builder builder() {
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
@@ -17,9 +17,9 @@
 package org.springframework.ai.chat.memory;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import org.springframework.ai.chat.messages.Message;
 import org.springframework.ai.chat.messages.SystemMessage;
@@ -81,17 +81,10 @@ public final class MessageWindowChatMemory implements ChatMemory {
 	private List<Message> process(List<Message> memoryMessages, List<Message> newMessages) {
 		List<Message> processedMessages = new ArrayList<>();
 
-		// Compare by text only — AbstractMessage.equals() includes metadata, which
-		// persistence stores (Cassandra, JDBC, MongoDB) may enrich on save, causing a
-		// false "new system message" detection that silently wipes existing system
-		// context.
-		Set<String> memorySystemTexts = memoryMessages.stream()
-			.filter(SystemMessage.class::isInstance)
-			.map(Message::getText)
-			.collect(Collectors.toSet());
+		Set<Message> memoryMessagesSet = new HashSet<>(memoryMessages);
 		boolean hasNewSystemMessage = newMessages.stream()
 			.filter(SystemMessage.class::isInstance)
-			.anyMatch(message -> !memorySystemTexts.contains(message.getText()));
+			.anyMatch(message -> !memoryMessagesSet.contains(message));
 
 		memoryMessages.stream()
 			.filter(message -> !(hasNewSystemMessage && message instanceof SystemMessage))
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/memory/MessageWindowChatMemory.java
@@ -17,9 +17,9 @@
 package org.springframework.ai.chat.memory;
 
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.springframework.ai.chat.messages.Message;
 import org.springframework.ai.chat.messages.SystemMessage;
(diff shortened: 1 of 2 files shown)


suryateja-g13:
Flagging for maintainer attention — the original issue (#5940) has had active discussion recently that reinforces why this fix is needed.

Key findings from the thread:
- The recommended Spring AI pattern (`disableInternalConversationHistory()` + `MessageChatMemoryAdvisor` at default order) **does** store tool pairs in memory on every tool iteration — this is by design, confirmed by `AbstractToolCallAdvisorIT`
- There is **no workaround** available to users short of placing `MessageChatMemoryAdvisor` outside the tool loop (a non-obvious deviation from the recommended pattern)
- The orphan issue is reproducible in production with any tool-heavy conversation once the window starts sliding — the original author hits it every few hours with `maxMessages=100`

This fix is the correct place to handle it — in `MessageWindowChatMemory` itself, transparently, without requiring users to change their advisor configuration.

anuragg-saxenaa:
@suryateja-g13 — independent local validation + a gap in the test matrix.

## Local validation (clean)

Built and tested your branch locally against `spring-ai-model` on Java 25 / Maven 3.9.13:

```
mvn -pl spring-ai-model -am test -Dtest=MessageWindowChatMemoryTests
→ Tests run: 22, Failures: 0, Errors: 0, Skipped: 0  (BUILD SUCCESS)
```

All 22 of your existing tests pass. The fix is clean and the multi-orphan test (`multipleConsecutiveOrphanedToolResponsesAreAllDropped`) is a good case.

## Gap in the test matrix

One case your fix does not cover: a single `AssistantMessage` with TWO tool calls whose `tool_response`s are *interleaved* with other messages in the retained window. The head-only scan in `dropOrphanedToolResponses` correctly drops leading orphans, but it leaves orphans that are *sandwiched* between non-`ToolResponseMessage` messages.

### Concrete scenario (window=5)

Input:
```
[system, u1, assistant{t1,t2}, toolResponse{t1}, user2, toolResponse{t2}, u3, nextTurn]
```

After head-trim (keep last 5):
```
[toolResponse{t1}, user2, toolResponse{t2}, u3, nextTurn]
```

`dropOrphanedToolResponses` sees `toolResponse{t1}` at the head, advances past it, stops at `user2`, returns `[user2, toolResponse{t2}, u3, nextTurn]`. **But `toolResponse{t2}` is also an orphan** — its `tool_use` was inside the evicted `assistant{t1,t2}`. The current fix leaves it in the result. Sending this to Anthropic causes `400: unexpected tool_use_id found in tool_result blocks`.

### Failing test (run locally, fails against the current fix)

```java
@Test
void orphanToolResponseInsideRetainedConversationIsDropped() {
    int limit = 5;
    MessageWindowChatMemory mem = MessageWindowChatMemory.builder().maxMessages(limit).build();
    String id = UUID.randomUUID().toString();

    AssistantMessage assistantWithTwoToolCalls = AssistantMessage.builder()
        .toolCalls(List.of(new AssistantMessage.ToolCall("t1", "function", "tool_a", "{}"),
                new AssistantMessage.ToolCall("t2", "function", "tool_b", "{}")))
        .build();
    ToolResponseMessage toolResponse1 = ToolResponseMessage.builder()
        .responses(List.of(new ToolResponseMessage.ToolResponse("t1", "tool_a", "result_a")))
        .build();
    UserMessage user2 = new UserMessage("user2");
    ToolResponseMessage toolResponse2 = ToolResponseMessage.builder()
        .responses(List.of(new ToolResponseMessage.ToolResponse("t2", "tool_b", "result_b")))
        .build();
    UserMessage u3 = new UserMessage("u3");
    AssistantMessage nextTurn = new AssistantMessage("next turn reply");

    mem.add(id, List.of(new UserMessage("u1"), assistantWithTwoToolCalls, toolResponse1, user2, toolResponse2, u3,
            nextTurn));

    List<Message> result = mem.get(id);

    // Both tool_results are orphans because their tool_use was inside
    // assistantWithTwoToolCalls which was evicted by the window trim.
    assertThat(result.stream().filter(ToolResponseMessage.class::isInstance).count()).isZero();
    assertThat(result).doesNotContain(toolResponse1, toolResponse2);
    // Sanity: non-orphan messages survive.
    assertThat(result).contains(user2, u3, nextTurn);
}
```

Result against the current fix: `Tests run: 23, Failures: 1` — `expected: 0L but was: 1L` (one orphan `ToolResponseMessage` survives in the result).

## Possible fix shape (for the maintainer's review)

`dropOrphanedToolResponses` would need a second pass: after the head-only scan, walk the retained non-system list and remove any `ToolResponseMessage` whose `tool_call_id`s cannot be matched to a `tool_call` on any retained `AssistantMessage` with `toolCalls`. This is a tool_use-id map scan, not a position-based scan, so it correctly catches orphans that are no longer at the head.

I'd be happy to drive this to merge if a maintainer is willing to reassign — but I'd want to credit you as the original author of the orphan-drop logic and the multi-orphan test. The added work (the gap test + the id-based fix) would land as additional commits on top of your three.

If you'd rather finish the gap-fix yourself, I'm also fine with that — just flagging it because the issue is real and the existing fix is in the right place to handle it.

— @anuragg-saxenaa (independent reviewer, RedOS context, no conflicts of interest)

PR #6030, opened 2026-05-13
Add health indicators for VectorStore beans via Spring Boot Actuator

### LINKED ISSUE #1611: Health indicators for vector stores (and any other external APIs)

suryateja-g13:
## Problem

Spring AI's `VectorStore` implementations have no out-of-the-box integration with Spring Boot Actuator's health endpoint. Users cannot tell whether their vector store is reachable simply by checking `/actuator/health`.

Closes #1611

## Solution

Add `VectorStoreHealthIndicator` and `VectorStoreHealthContributorAutoConfiguration` to the `spring-ai-autoconfigure-vector-store-observation` module.

### Behaviour

- **Single store** → a plain `HealthIndicator` named `vectorStoreHealthContributor` is registered; visible at `/actuator/health/vectorStore`.
- **Multiple stores** → a `CompositeHealthContributor` is registered, keyed by bean name; each store is individually visible at `/actuator/health/vectorStore/<beanName>`.
- **Health check** → performs a lightweight `similaritySearch(query="health", topK=1)`. Reports `UP` with the store name as a detail; reports `DOWN` with the exception message on failure.
- **Toggle** → respects `management.health.vectorStore.enabled=false` (defaults to `true`).

### Example output

```json
{
  "status": "UP",
  "components": {
    "vectorStore": {
      "status": "UP",
      "details": { "vectorStore": "SimpleVectorStore" }
    }
  }
}
```

## Changes

| File | Change |
|------|--------|
| `pom.xml` | Add optional `spring-boot-health` dependency (Spring Boot 4.x module) |
| `VectorStoreHealthIndicator.java` | New — `AbstractHealthIndicator` implementation |
| `VectorStoreHealthContributorAutoConfiguration.java` | New — auto-configuration with `@ConditionalOnEnabledHealthIndicator("vectorStore")` |
| `AutoConfiguration.imports` | Register new auto-configuration |
| `VectorStoreHealthContributorAutoConfigurationTests.java` | New — 4 tests via `ApplicationContextRunner` |

## Tests

- `healthIndicatorRegisteredForSingleVectorStore` — single store → plain `HealthIndicator`, status `UP`
- `compositeHealthContributorRegisteredForMultipleVectorStores` — two stores → `CompositeHealthContributor` with 2 entries
- `healthIndicatorCanBeDisabled` — `management.health.vectorStore.enabled=false` → no bean registered
- `healthIndicatorReportsDownWhenStoreThrows` — store throws on search → status `DOWN`

All 4 tests pass.

### DIFF
--- a/auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-observation/pom.xml
+++ b/auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-observation/pom.xml
@@ -50,6 +50,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-health</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter</artifactId>
 		</dependency>
 		<dependency>
--- a/auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-observation/src/main/java/org/springframework/ai/vectorstore/observation/autoconfigure/VectorStoreHealthContributorAutoConfiguration.java
+++ b/auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-observation/src/main/java/org/springframework/ai/vectorstore/observation/autoconfigure/VectorStoreHealthContributorAutoConfiguration.java
@@ -21,6 +21,7 @@ import java.util.Map;
 
 import org.springframework.ai.vectorstore.VectorStore;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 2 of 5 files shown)


StantonMatt (commented):
I took a pass over this branch because it overlaps the earlier store-specific attempts in #5553/#5764 and the generic contributor shape here seems like the main design question.

I validated the focused module in a JDK 21 Maven container:

`mvn -ntp -B -Dmaven.repo.local=/m2 -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -pl auto-configurations/vector-stores/spring-ai-autoconfigure-vector-store-observation -am -Dtest=VectorStoreHealthContributorAutoConfigurationTests test`

Result: reactor success, and all 6 `VectorStoreHealthContributorAutoConfigurationTests` tests passed.

The remaining tradeoff I see is the probe semantics: `similaritySearch("health", topK=1)` is the only common `VectorStore` API, but for most stores it still runs through the embedding pipeline. That means the indicator can report `DOWN` for embedding-model outages/rate limits even when the vector store is reachable. The branch documents that behavior clearly, so if maintainers are comfortable with that app-health interpretation, I do not see a separate implementation issue from this pass.

sobychacko:
Thanks for the PR!

Closing in favor of #5764, which targets the same issue (#1611) and was opened earlier. Two technical points drove it:

1. `VectorStoreHealthContributorAutoConfiguration` has `@ConditionalOnBean(VectorStore.class)` but is only ordered with `@AutoConfigureAfter(VectorStoreObservationAutoConfiguration.class)`. The real `VectorStore` beans come from the per-store auto-configurations, which that hint does not cover. Boot sorts auto-configurations by class name before applying after/before hints, so this one evaluates before the pgvector, qdrant, redis, oracle, opensearch, pinecone, s3, typesense and weaviate ones, sees no bean, and silently never registers. `VectorStoreHealthContributorAutoConfigurationTests` does not catch it because `withUserConfiguration(...)` registers the store as a user bean, which is always in place first.

2. `similaritySearch(SearchRequest.builder().query("health").topK(1).build())` goes through the embedding pipeline, so every Actuator poll is a billed remote call and an embedding outage reports the store DOWN.

The branch is also conflicting with `main`. The generic approach is worth discussing on #1611: moving `@ConditionalOnBean` onto the `@Bean` method or injecting `ObjectProvider<Map<String, VectorStore>>` both sidestep the ordering problem.

Issue #3868, opened 2025-07-21
Spring AI MCP Server does not properly work when using spring.main.lazy-initialization=true


switschel:
**Bug description**
I'm using spring-ai-starter-mcp-server-webmvc in my project to expose tools. I recognized that it gives me a 500 error when I set the property `spring.main.lazy-initialization=false` and I try to access the `/sse` endpoint.

**Error**
```
Caused by: jakarta.servlet.ServletException: java.lang.NullPointerException: Cannot invoke "io.modelcontextprotocol.spec.McpServerSession$Factory.create(io.modelcontextprotocol.spec.McpServerTransport)" because "this.sessionFactory" is null
	at org.springframework.web.servlet.function.ErrorHandlingServerResponse.handleError(ErrorHandlingServerResponse.java:70)
	at org.springframework.web.servlet.function.AbstractServerResponse.writeTo(AbstractServerResponse.java:105)
	at org.springframework.web.servlet.function.support.HandlerFunctionAdapter.handle(HandlerFunctionAdapter.java:112)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1089)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:564)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:658)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:195)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:140)
	at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:51)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:164)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:140)
	at org.springframework.web.filter.CompositeFilter$VirtualFilterChain.doFilter(CompositeFilter.java:108)
[...]

PR #6033, opened 2026-05-14
Fix stateless MCP "handler not configured" with lazy init

### LINKED ISSUE #5964: McpStatelessSyncServer not initialized under spring.main.lazy-initialization=true, causing "MCP handler not configured" on every request
## Summary

When `spring.main.lazy-initialization=true` is active, `McpStatelessSyncServer` is never eagerly constructed. Because `setMcpHandler` is only called from the server's constructor/build phase, `WebMvcStatelessServerTransport.mcpHandler` stays `null` permanently. Every incoming MCP request then hits the null-check at `WebMvcStatelessServerTransport.handlePost` (line ~146) and returns HTTP 500 with JSON-RPC error `-32603`:

```json
{
  "jsonRpcError": {
    "code": -32603,
    "message": "MCP handler not configured"
  }
}
```

The transport bean and its `RouterFunction` _are_ created eagerly (because Spring MVC's handler mapping requires them at context refresh), so the endpoint is reachable — but the handler wire is never made.

## Related issue

#3868 describes the same root cause on the SSE transport (`WebMvcSseServerTransportProvider`). This issue covers the stateless transport path, which is a separate code path introduced later.

## Environment

- Spring AI: `2.0.0-M5`
- `spring-ai-starter-mcp-server-webmvc`
- `spring.ai.mcp.server.protocol=STATELESS`, `spring.ai.mcp.server.type=SYNC`
- `spring.main.lazy-initialization=true`
- Spring Boot 4.1.0-RC1 / Tomcat (WebMVC)

## Stack trace (trimmed)

```
io.modelcontextprotocol.spec.McpError$Builder.build(McpError.java:72)
org.springframework.ai.mcp.server.webmvc.transport.WebMvcStatelessServerTransport.handlePost(WebMvcStatelessServerTransport.java:150)
org.springframework.web.servlet.function.support.HandlerFunctionAdapter.handle(HandlerFunctionAdapter.java:106)
org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:963)
...
```

## Root cause

`McpServerStatelessAutoConfiguration.mcpStatelessSyncServer()` has no `@Lazy(false)` guard. Under global lazy init, the bean is deferred until first programmatic access, which never happens via the normal request path. The transport's `setMcpHandler` is therefore never called.

## Suggested fix

Add `@Lazy(false)` to the `mcpStatelessSyncServer` bean definition in `McpServerStatelessAutoConfiguration`:

```java
@Bean
@Lazy(false)
@ConditionalOnProperty(prefix = McpServerProperties.CONFIG_PREFIX, name = "type", havingValue = "SYNC",
        matchIfMissing = true)
public McpStatelessSyncServer mcpStatelessSyncServer(McpStatelessServerTransport statelessTransport, ...) {
    ...
}
```

This mirrors the fix that should also apply to the SSE transport variant in #3868.

jewoodev:
### Summary

Closes #5964.

When `spring.main.lazy-initialization=true` is enabled, the stateless MCP server beans are deferred during context refresh. The stateless transport endpoint is still registered, but the MCP handler is only wired when the server bean is built. That leaves incoming stateless requests failing with `MCP handler not configured`.

### Changes

- Add `@Lazy(false)` to `mcpStatelessSyncServer(...)`.
- Add `@Lazy(false)` to `mcpStatelessAsyncServer(...)`, which has the same lazy-init failure mode as the reported sync case.

### Tests

Added regression coverage in `McpStatelessServerAutoConfigurationIT` for:

- sync server eager creation under `LazyInitializationBeanFactoryPostProcessor`
- async server eager creation under the same post-processor

Each test asserts `beanFactory.containsSingleton(...)` after refresh, verifying eager instantiation without triggering the bean by lookup. This directly covers the eager-creation contract required for stateless transport handler wiring.

```bash
./mvnw -pl auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common -am test -Dtest='McpStatelessServerAutoConfigurationIT' -Dsurefire.failIfNoSpecifiedTests=false
```

Passes with 25 tests and 0 failures.

### Risk

Low. This only forces eager creation for stateless MCP server beans that must run their handler-wiring side effect during context refresh. It does not change behavior when global lazy initialization is disabled.

### Scope

This PR intentionally covers only the stateless path requested in #5964. The SSE / stateful path tracked in #3868 lives in different auto-configurations and is left out of this PR.

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerStatelessAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerStatelessAutoConfiguration.java
@@ -48,6 +48,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
+import org.springframework.context.annotation.Lazy;
 import org.springframework.core.env.Environment;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.util.CollectionUtils;
@@ -55,6 +56,7 @@ import org.springframework.web.context.support.StandardServletEnvironment;
 
 /**
  * @author Christian Tzolov
+ * @author Jewoo Shin
  */
 @AutoConfiguration
 @ConditionalOnClass(McpSchema.class)
@@ -72,6 +74,7 @@ public class McpServerStatelessAutoConfiguration {
 	}
 
 	@Bean
+	@Lazy(false)
 	@ConditionalOnProperty(prefix = McpServerProperties.CONFIG_PREFIX, name = "type", havingValue = "SYNC",
 			matchIfMissing = true)
 	public McpStatelessSyncServer mcpStatelessSyncServer(McpStatelessServerTransport statelessTransport,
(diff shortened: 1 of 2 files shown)


jewoodev:
For maintainers comparing this with #6041: the production fix is the same in both PRs (`@Lazy(false)` on the stateless sync/async server beans).

The only meaningful difference is the regression test. This PR asserts `beanFactory.containsSingleton(...)` after refresh, so it verifies eager instantiation without triggering the bean by lookup. That directly covers the handler-wiring side effect required by #5964.

Happy to align or close this PR if #6041 is preferred.

PR #6035, opened 2026-05-14
Fix recursive tool input schema by hoisting $defs to root

### LINKED ISSUE #5888: Tool inputSchema with recursive parameter types emits unresolvable `$ref` (nested `$defs`, root-relative `$ref`)
## Summary

`JsonSchemaGenerator.generateForMethodInput(Method)` produces a JSON Schema where `$defs` blocks are nested inside parameter property schemas, while `$ref` values use root-relative pointers (`#/$defs/<name>`). Per JSON Schema 2020-12, `#/...` resolves from the document root, so these refs are unresolvable. Strict resolvers (Anthropic Messages API, OpenAI tools, `langchain-mcp-adapters`, `jsonschema` Python lib) reject the schema with errors like:

```
Reference '#/$defs/FieldTermsRequestFilter' not found.
```

This breaks any `@Tool` method whose parameter type is — or transitively contains — a recursive Java type.

## Versions

- Spring AI: **1.1.4** (also reproduces on 1.1.2)
- Spring Boot: 3.5.x
- Module: `spring-ai-starter-mcp-server-webmvc` (but the bug is in `spring-ai-model`'s `JsonSchemaGenerator`, so it affects every tool-calling integration, not only MCP)
- Java: 21
- victools/jsonschema-generator: as transitively pulled by Spring AI 1.1.4

## Minimal reproducer

```java
public class Filter {
    public String field;
    public String operator;
    public List<Object> values;
    public List<Filter> filters; // ← recursive
}

public class SearchRequest {
    public List<Filter> filters;
    public int limit;
}

@Service
class SearchToolService {
    @Tool(description = "Search Books")
    public String searchBooks(@ToolParam(description = "Request") SearchRequest request) {
        return "...";
    }
}
```

Then:

```java
ToolCallback[] callbacks = MethodToolCallbackProvider.builder()
    .toolObjects(new SearchToolService())
    .build()
    .getToolCallbacks();

System.out.println(callbacks[0].getToolDefinition().inputSchema());
```

## Actual output (abridged, formatted)

```jsonc
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "object",
  "properties": {
    "request": {
      "$defs": {                                                    // ← $defs HERE
        "Filter": {
          "type": "object",
          "properties": {
            "filters": {
              "type": "array",
              "items": { "$ref": "#/$defs/Filter" } // ← root-relative
            },
            "field":    { "type": "string" },
            "operator": { "type": "string" },
            "values":   { "type": "array", "items": {} }
          },
          "required": ["field", "filters", "operator", "values"]
        }
      },
      "type": "object",
      "properties": {
        "filters": {
          "type": "array",
          "items": { "$ref": "#/$defs/Filter" }     // ← also broken
        },
        "limit": { "type": "integer", "format": "int32" }
      },
      "required": ["filters", "limit"]
    }
  },
  "required": ["request"],
  "additionalProperties": false
}
```

`$defs` lives at `#/properties/request/$defs/...`, but the refs assume `#/$defs/...`. Both refs are unresolvable.

## Expected output

`$defs` hoisted to the document root, leaving the existing root-relative refs valid:

```jsonc
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$defs": {
    "Filter": { /* … */ }
  },
  "type": "object",
  "properties": {
    "request": {
      "type": "object",
      "properties": {
        "filters": {
          "type": "array",
          "items": { "$ref": "#/$defs/Filter" }
        },
        "limit": { "type": "integer", "format": "int32" }
      },
      "required": ["filters", "limit"]
    }
  },
  "required": ["request"],
  "additionalProperties": false
}
```

## Root cause

In [`JsonSchemaGenerator.generateForMethodInput`](https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-model/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java) (path may differ across versions), each parameter's schema is generated independently:

```java
ObjectNode parameterNode = SUBTYPE_SCHEMA_GENERATOR.generateSchema(parameterType);
// ...
properties.set(parameterName, parameterNode);
```

`SUBTYPE_SCHEMA_GENERATOR` (victools) returns a **self-contained** schema rooted at the parameter type — including its own `$defs` block and `$ref` pointers that resolve against *that* root. The generator then slots the subtype schema directly under `properties.<paramName>` of a freshly built outer wrapper, without hoisting `$defs` to the document root or rewriting `$ref` paths.

For non-recursive parameter types, victools inlines everything and the bug is invisible. For recursive types, victools must use `$defs`/`$ref`, and the resulting tool schema is invalid.

## Impact

- Anthropic Messages API and OpenAI Tools both reject the schema (their tool-call path validates JSON Schema strictly).
- MCP servers built on `spring-ai-starter-mcp-server-*` expose broken `inputSchema` to **every** MCP client that validates schemas, including:
  - `langchain-mcp-adapters` (Python) — `Reference '...' not found.`
  - Anthropic's MCP integrations
  - Most `jsonschema`-based clients
- Affects every `@Tool` whose request DTO contains a self-referential field — common for filter / query DSL types.

## Workaround (in user code)

Wrap each `ToolCallback` with a decorator whose `getToolDefinition().inputSchema()` parses the JSON, hoists every nested `$defs` to the root, and returns the rewritten schema. Existing `#/$defs/...` refs then resolve. (Happy to share the snippet if helpful.)

## Suggested fix

After producing `parameterNode` in `generateForMethodInput`, lift any nested `$defs` to the outer schema:

```java
ObjectNode parameterNode = SUBTYPE_SCHEMA_GENERATOR.generateSchema(parameterType);
JsonNode nestedDefs = parameterNode.remove("$defs");
if (nestedDefs != null && nestedDefs.isObject()) {
    ObjectNode rootDefs = (ObjectNode) schema.withObject("/$defs"); // create or reuse
    nestedDefs.fields().forEachRemaining(e -> {
        if (rootDefs.has(e.getKey()) && !rootDefs.get(e.getKey()).equals(e.getValue())) {
            // log conflict, keep first
            return;
        }
        rootDefs.set(e.getKey(), e.getValue());
    });
}
```

Same treatment is needed in `generateForType` when wrapping multiple type schemas (e.g., for structured-output unions), and in any other code path that composes subtype schemas into a wrapper.

## Related

- #5640 — same root cause from the consumer side: `JsonSchemaConverter` rejects `$defs` from MCP server tool schemas (Google GenAI / Gemini path). PR #5211 forbids `$defs` rather than handling them; merging proper `$defs` support across both produce and consume paths would close both issues.
- modelcontextprotocol/java-sdk#749 (merged) — switched MCP `inputSchema` to a free-form `Map`, removing the SDK-side constraint. The remaining problem is purely Spring AI's schema generation.

Happy to open a PR if the maintainers agree on the approach.

jewoodev:
### Summary

Fixes #5888.

- **Why**: Spring AI currently inlines parameter schemas under `properties.<param>`, but their generated `$ref`s still point to root-level `$defs`.

- **Fix**: Move each parameter schema's `$defs` to the wrapper schema root before inlining the parameter. Reuse structurally equal definitions; rename conflicting simple names and rewrite refs within the hoisted schema.

### Tests
Added regression coverage in both generators for:
- transitive recursive parameter schemas
- duplicate equal definitions
- colliding simple-name definitions with ref rewrites

```bash
./mvnw -pl spring-ai-model,mcp/mcp-annotations test \
    -Dtest='JsonSchemaGeneratorTests,McpJsonSchemaGeneratorTests'
```

### Scope
This PR covers the transitive `$defs` failure reported in #5888 for both `JsonSchemaGenerator` and `McpJsonSchemaGenerator`. Direct self-recursive `$ref: "#"` handling is outside this PR.

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/utils/McpJsonSchemaGenerator.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/utils/McpJsonSchemaGenerator.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -44,6 +45,7 @@ import io.modelcontextprotocol.util.Assert;
 import io.modelcontextprotocol.util.Utils;
 import io.swagger.v3.oas.annotations.media.Schema;
 import org.jspecify.annotations.Nullable;
+import tools.jackson.databind.JsonNode;
 import tools.jackson.databind.node.ObjectNode;
 
 import org.springframework.ai.mcp.annotation.McpMeta;
@@ -179,6 +181,12 @@ public final class McpJsonSchemaGenerator {
 				required.add(parameterName);
 			}
 			ObjectNode parameterNode = SUBTYPE_SCHEMA_GENERATOR.generateSchema(parameterType);
+			// victools generates self-contained schemas where $defs and the $ref
+			// pointers into them are rooted at the sub-schema. Inlining the
+			// sub-schema under properties.<paramName> re-parents existing
+			// "#/$defs/<Name>" refs to the outer root, leaving them unresolvable.
+			// Hoist $defs to the outer root so those refs resolve again.
+			hoistDefsToRoot(schema, parameterNode);
 			String parameterDescription = getMethodParameterDescription(method, i);
 			if (Utils.hasText(parameterDescription)) {
 				parameterNode.put("description", parameterDescription);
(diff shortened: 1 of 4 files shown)


jewoodev:
Hi @danielvolovik96 and @jeanchristophecascade — apologies for going ahead without a check-in first. The issue had been quiet for ~18 days and I drafted this to keep it moving, but I want to be explicit: if either of you would prefer to lead the implementation, I'm happy to close this PR, hand the branch over, or co-author — just say the word. Same goes if maintainers prefer a different approach.

The diff covers the transitive `$defs` case described in #5888 (with rename + peer-ref rewrite for the same-simple-name corner case), applied to both `spring-ai-model` and `mcp-annotations` per @jeanchristophecascade's follow-up.

tzolov:
Thanks for addressing this issue @jewoodev !

tzolov:
Rebased, clean the code and duplications, squashed and merged at 0f9c2a5d7a19a219c9c0f5ac6903038eca3d3974

jeanchristophecascade:
Many thanks to both of you 😊

jewoodev:
Thank you @tzolov for the careful review, cleanup, and merge.

I really appreciate the refinements you made to the implementation. I learned from them and will keep that direction in mind for future contributions.

PR #6041, opened 2026-05-14
GH-5964: Add @Lazy(false) to stateless MCP server beans to fix lazy-init startup failure

### LINKED ISSUE #5964: McpStatelessSyncServer not initialized under spring.main.lazy-initialization=true, causing "MCP handler not configured" on every request

suryateja-g13:
Fixes #5964.

When `spring.main.lazy-initialization=true` is active, `McpStatelessSyncServer` and `McpStatelessAsyncServer` were never constructed at context refresh. The transport's `setMcpHandler` is called only during `build()`, so the handler stayed `null` permanently. Every incoming MCP request then hit the null-check in `WebMvcStatelessServerTransport.handlePost` and returned HTTP 500 with JSON-RPC `-32603 INTERNAL_ERROR: MCP handler not configured`.

`@Lazy(false)` is the standard Spring Boot pattern to override a global `spring.main.lazy-initialization=true` for beans that must be instantiated eagerly to wire side-effects. Without it, the transport endpoint is reachable but permanently misconfigured.

## Changes

- `McpServerStatelessAutoConfiguration` — added `@Lazy(false)` to `mcpStatelessSyncServer` and `mcpStatelessAsyncServer` bean methods
- `McpStatelessServerAutoConfigurationIT` — added two tests covering both SYNC and ASYNC variants under global lazy-init

## How to test locally

```bash
./mvnw test -pl auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common -Dtest=McpStatelessServerAutoConfigurationIT
```

## Risk

Low — adds `@Lazy(false)` to two bean methods with no other behavioural change. Has no effect when `spring.main.lazy-initialization` is not set.

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerStatelessAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerStatelessAutoConfiguration.java
@@ -48,6 +48,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
+import org.springframework.context.annotation.Lazy;
 import org.springframework.core.env.Environment;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.util.CollectionUtils;
@@ -72,6 +73,7 @@ public class McpServerStatelessAutoConfiguration {
 	}
 
 	@Bean
+	@Lazy(false)
 	@ConditionalOnProperty(prefix = McpServerProperties.CONFIG_PREFIX, name = "type", havingValue = "SYNC",
 			matchIfMissing = true)
 	public McpStatelessSyncServer mcpStatelessSyncServer(McpStatelessServerTransport statelessTransport,
@@ -163,6 +165,7 @@ public class McpServerStatelessAutoConfiguration {
 	}
 
 	@Bean
+	@Lazy(false)
 	@ConditionalOnProperty(prefix = McpServerProperties.CONFIG_PREFIX, name = "type", havingValue = "ASYNC")
 	public McpStatelessAsyncServer mcpStatelessAsyncServer(McpStatelessServerTransport statelessTransport,
 			McpSchema.ServerCapabilities.Builder capabilitiesBuilder, McpServerProperties serverProperties,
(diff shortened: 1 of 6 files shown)


sobychacko:
This looks like a duplicate of #6033, so closing it. Feel free to reopen if needed.

Note: this PR also bundled an unrelated Google GenAI `Locale.ROOT` fix (#5340). That one's worth keeping, please resubmit it as its own PR.

PR #6055, opened 2026-05-16
Fix GoogleGenAiTextEmbeddingModel ignoring taskType, title and autoTruncate options

### LINKED ISSUE #5966: Google GenAI Text Embedding does not pass task type to the API

suryateja-g13:
## Problem

`GoogleGenAiTextEmbeddingModel.embed()` builds an `EmbedContentConfig` but silently ignores three options that callers can set via `GoogleGenAiTextEmbeddingOptions`:
- `taskType` — determines the embedding space optimization (retrieval, classification, etc.)
- `title` — only relevant for `RETRIEVAL_DOCUMENT` task type
- `autoTruncate` — controls whether the model truncates inputs exceeding the token limit

A TODO comment in the code acknowledged this but left these options unwired.

Closes #5966

## Fix

Replace the TODO comment block with actual calls to `EmbedContentConfig.Builder`:
```java
if (options.getTaskType() != null) {
    configBuilder.taskType(options.getTaskType().name());
}
if (StringUtils.hasText(options.getTitle())) {
    configBuilder.title(options.getTitle());
}
if (options.getAutoTruncate() != null) {
    configBuilder.autoTruncate(options.getAutoTruncate());
}
```

`EmbedContentConfig.Builder.taskType()` accepts a `String`, and `GoogleGenAiTextEmbeddingOptions.TaskType` is an enum whose `.name()` produces the correct Google API string value.

## Test plan
- [ ] Verify `GoogleGenAiTextEmbeddingModel` compiles cleanly
- [ ] Manual test: create `GoogleGenAiTextEmbeddingOptions` with `taskType = RETRIEVAL_DOCUMENT`, verify the config is passed to the SDK

### DIFF
--- a/models/spring-ai-google-genai-embedding/src/main/java/org/springframework/ai/google/genai/text/GoogleGenAiTextEmbeddingModel.java
+++ b/models/spring-ai-google-genai-embedding/src/main/java/org/springframework/ai/google/genai/text/GoogleGenAiTextEmbeddingModel.java
@@ -52,7 +52,7 @@ import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 /**
- * A class representing a Vertex AI Text Embedding Model using the new Google Gen AI SDK.
+ * A class representing a Google Gen AI Text Embedding Model using the Google Gen AI SDK.
  *
  * @author Christian Tzolov
  * @author Mark Pollack
@@ -150,9 +150,15 @@ public class GoogleGenAiTextEmbeddingModel extends AbstractEmbeddingModel {
 					configBuilder.outputDimensionality(options.getDimensions());
 				}
 
-				// Set task type if specified - this might need to be handled differently
-				// as the new SDK might not have a direct taskType field
-				// We'll need to check the SDK documentation for this
+				if (options.getTaskType() != null) {
+					configBuilder.taskType(options.getTaskType().name());
+				}
+				if (StringUtils.hasText(options.getTitle())) {
+					configBuilder.title(options.getTitle());
+				}
+				if (options.getAutoTruncate() != null) {
+					configBuilder.autoTruncate(options.getAutoTruncate().booleanValue());
+				}
 
 				EmbedContentConfig config = configBuilder.build();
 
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks for this, @suryateja-g13. This overlaps with #6022, which was opened a few days earlier and makes the same change to the TODO block in `GoogleGenAiTextEmbeddingModel.call()`, so we are going to move forward there and close this one.

One thing worth calling out: your PR also wires `title` and `autoTruncate`, which are dropped today too. That part goes beyond what #5966 describes and we would like to keep it, so we will fold it into the surviving PR (with the option-forwarding assertions in a test class of their own rather than in `GoogleGenAiTextEmbeddingRetryTests`).

Thanks again for looking into this.

PR #6057, opened 2026-05-17
fix: add span attributes for input messages in ChatModelPromptContentObservationHandler

### LINKED ISSUE #6051: ChatModelPromptContentObservationHandler should tag span attributes for LLM Observability, not only log to SLF4J

**Bug description**
Spring AI ships ChatModelPromptContentObservationHandler to expose LLM prompt and completion content, but it only emits them via SLF4J logging. It does not attach them as OpenTelemetry span attributes (gen_ai.input.messages / gen_ai.output.messages).
As a result, observability backends that rely on OTEL span data — such as Datadog LLM Observability, Honeycomb, or any OTel-native platform — cannot display the actual prompt and completion content in traces. Users are forced to write and register custom ObservationHandler implementations just to bridge this gap (see workaround below).
The same gap exists for tool/function call inputs and outputs: there is no built-in handler that records these as span attributes.
Spring AI already defines the semantic conventions (gen_ai.*) and wires up the ObservationRegistry, so emitting this data as span tags should be supported out of the box.

**Environment**
Spring AI: 1.1.x (please confirm which version introduced ChatModelPromptContentObservationHandler)
Java: 21
Tracing: Micrometer Tracing + OpenTelemetry bridge (e.g. micrometer-tracing-bridge-otel)
Observability backend: Datadog (LLM Observability), but affects any OTEL-native backend

**Steps to reproduce**
Add Spring AI with a chat model (e.g. Anthropic, OpenAI).
Configure Micrometer Tracing with an OTEL exporter.
Enable prompt/completion content observation via spring.ai.chat.observations.include-prompt=true and spring.ai.chat.observations.include-completion=true.
Send a chat request and inspect the exported spans.

**Expected behavior**
The spans exported to an OTEL backend should include:

gen_ai.input.messages — serialized list of prompt messages with role and content
gen_ai.output.messages — serialized list of completion messages with role and content parts
Tool call inputs/outputs as span attributes when tools are invoked

This aligns with the [OpenTelemetry Generative AI semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/).

**Actual behavior**
ChatModelPromptContentObservationHandler logs prompt/completion content to SLF4J only. No span attributes are set. OTEL backends receive traces with no LLM-specific content.

Users currently have to write a custom handler like this to fill the gap:

public class ChatModelSpanContentObservationHandler 
        implements ObservationHandler<ChatModelObservationContext> {

    private final ObjectMapper objectMapper;

    @Override
    public void onStop(ChatModelObservationContext context) {
        TracingObservationHandler.TracingContext tracingContext =
                context.get(TracingObservationHandler.TracingContext.class);
        if (tracingContext == null) return;
        Span span = tracingContext.getSpan();
        if (span == null) return;
        tagPrompt(span, context);
        tagCompletion(span, context);
    }

    private void tagPrompt(Span span, ChatModelObservationContext context) {
        var instructions = context.getRequest().getInstructions();
        if (instructions == null || instructions.isEmpty()) return;
        try {
            List<Map<String, String>> messages = instructions.stream()
                .map(msg -> Map.of("role", msg.getMessageType().getValue(), 
                                   "content", msg.getText()))
                .collect(Collectors.toList());
            span.tag("gen_ai.input.messages", objectMapper.writeValueAsString(messages));
        } catch (Exception e) {
            logger.info("Failed to serialize prompt for span tagging", e);
        }
    }

    // ... tagCompletion similarly

    @Override
    public boolean supportsContext(Observation.Context context) {
        return context instanceof ChatModelObservationContext;
    }
}
This should not be necessary. Spring AI should either extend ChatModelPromptContentObservationHandler to also tag spans, or provide a dedicated ChatModelSpanContentObservationHandler that is auto-configured when tracing is on the classpath.

**Suggested Fix**
Extend ChatModelPromptContentObservationHandler#onStop to also call span.tag(...) when a TracingContext is present, similar to the workaround above.
Provide equivalent coverage for tool call observation contexts.
Auto-configure these handlers when micrometer-tracing is on the classpath, guarded by the existing spring.ai.chat.observations.include-prompt / include-completion flags.

EvanYao826:
Fixes #6051

---

## Problem

`ChatModelPromptContentObservationHandler.onStop()` emits prompt content to logs but does not set span attributes. Tracing systems have no visibility into the actual prompt content sent to the chat model.

## Fix

Add `gen_ai.input.messages` span tag in `onStop()` following OpenTelemetry GenAI semantic conventions. The tag is only set when:
- A `TracingContext` with a non-null `Span` is present
- The prompt contains at least one message

## Changes

- `ChatModelPromptContentObservationHandler.java`: Added `sampleSpanTag()` method that serializes messages as JSON array and sets span tag
- `ChatModelPromptContentObservationHandlerTests.java`: Added 5 tests covering: span tag set with tracing context, no tag without tracing context, no tag with empty prompt, no tag with null span, single message tag

---
*AI-agent involvement: This contribution was assisted by an AI coding agent.*

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/observation/conventions/AiObservationAttributes.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/observation/conventions/AiObservationAttributes.java
@@ -47,11 +47,6 @@ public enum AiObservationAttributes {
 	 * OpenTelemetry GenAI semantic conventions.
 	 */
 	INPUT_MESSAGES("gen_ai.input.messages"),
-	/**
-	 * The output messages received from the model, serialized as a JSON array following
-	 * the OpenTelemetry GenAI semantic conventions.
-	 */
-	OUTPUT_MESSAGES("gen_ai.output.messages"),
 
 	// GenAI Request
 
--- a/spring-ai-commons/src/main/java/org/springframework/ai/observation/conventions/AiObservationAttributes.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/observation/conventions/AiObservationAttributes.java
@@ -40,6 +40,19 @@ public enum AiObservationAttributes {
 	 */
 	AI_PROVIDER("gen_ai.system"),
 
+	// GenAI Content
+
+	/**
+	 * The input messages sent to the model, serialized as a JSON array following the
+	 * OpenTelemetry GenAI semantic conventions.
+	 */
+	INPUT_MESSAGES("gen_ai.input.messages"),
+	/**
+	 * The output messages received from the model, serialized as a JSON array following
+	 * the OpenTelemetry GenAI semantic conventions.
+	 */
+	OUTPUT_MESSAGES("gen_ai.output.messages"),
+
 	// GenAI Request
 
 	/**
--- a/.gitignore
+++ b/.gitignore
@@ -55,4 +55,3 @@ tmp
 
 
 plans
-E:\\IDEA\\Maven\\apache-maven-3.8.9\\mvn_repo/
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,4 @@ tmp
 
 
 plans
+E:\\IDEA\\Maven\\apache-maven-3.8.9\\mvn_repo/
--- a/auto-configurations/models/chat/observation/spring-ai-autoconfigure-model-chat-observation/src/main/java/org/springframework/ai/model/chat/observation/autoconfigure/ChatObservationAutoConfiguration.java
+++ b/auto-configurations/models/chat/observation/spring-ai-autoconfigure-model-chat-observation/src/main/java/org/springframework/ai/model/chat/observation/autoconfigure/ChatObservationAutoConfiguration.java
@@ -30,6 +30,7 @@ import org.springframework.ai.chat.observation.ChatModelCompletionObservationHan
 import org.springframework.ai.chat.observation.ChatModelMeterObservationHandler;
 import org.springframework.ai.chat.observation.ChatModelObservationContext;
 import org.springframework.ai.chat.observation.ChatModelPromptContentObservationHandler;
+import org.springframework.ai.chat.observation.ChatModelPromptSpanContentObservationHandler;
 import org.springframework.ai.embedding.observation.EmbeddingModelObservationContext;
 import org.springframework.ai.image.observation.ImageModelObservationContext;
 import org.springframework.ai.model.observation.ErrorLoggingObservationHandler;
@@ -104,6 +105,15 @@ public class ChatObservationAutoConfiguration {
 				Tracer tracer) {
 			logCompletionWarning();
 			return new TracingAwareLoggingObservationHandler<>(new ChatModelCompletionObservationHandler(), tracer);
+		}
+
+		@Bean
+		@ConditionalOnMissingBean(value = ChatModelPromptSpanContentObservationHandler.class,
+				name = "chatModelPromptSpanContentObservationHandler")
+		@ConditionalOnProperty(prefix = ChatObservationProperties.CONFIG_PREFIX, name = "log-prompt",
+				havingValue = "true")
+		ChatModelPromptSpanContentObservationHandler chatModelPromptSpanContentObservationHandler() {
+			return new ChatModelPromptSpanContentObservationHandler();
 		}
 
 		@Bean
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/observation/ChatModelPromptContentObservationHandler.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/observation/ChatModelPromptContentObservationHandler.java
@@ -17,18 +17,22 @@
 package org.springframework.ai.chat.observation;
 
 import java.util.List;
+import java.util.StringJoiner;
 
 import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationHandler;
+import io.micrometer.tracing.Span;
+import io.micrometer.tracing.handler.TracingObservationHandler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.springframework.ai.chat.messages.Message;
 import org.springframework.ai.content.Content;
 import org.springframework.ai.observation.ObservabilityHelper;
 import org.springframework.util.CollectionUtils;
 
 /**
- * Handler for emitting the chat prompt content to logs.
+ * Handler for emitting the chat prompt content to logs and as span attributes.
  *
  * @author Thomas Vitale
  * @author Jonatan Ivanov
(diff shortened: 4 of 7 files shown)


jewoodev:
@EvanYao826 Thanks for the PR. Looks like we may have crossed paths — I posted an analysis comment on the issue itself around the same time (my earlier issue comment: https://github.com/spring-projects/spring-ai/issues/6051#issuecomment-4468955457), so it seems we were working in parallel without seeing each other's work. Looking at that analysis side by side with this PR's diff, I'd like to share four things — three are concerns I think should be addressed before merge, and the last one is a scope question for the maintainers.

## 1. JSON escaping defect

In `serializeMessages()`, message text is concatenated directly into the JSON string without escaping:

```java
joiner.add("{\"role\":\"" + role + "\",\"content\":\"" + content + "\"}");
```

When user input contains `"`, `\`, newlines, or control characters, an invalid JSON string is emitted as the span attribute. For example, if a user types `Say "hi"`, the result is:

```
{"role":"user","content":"Say "hi""}
```

OTel-native backends could fail to parse this attribute as valid JSON. Since user input is essentially arbitrary text, this isn't a rare edge case but a predictable input pattern.

`spring-ai-model` already declares `tools.jackson.core:jackson-databind` (Jackson 3) as a compile dependency, so using Jackson's `JsonMapper`/`ObjectMapper` or an existing shared serializer in Spring AI would avoid the hand-rolled escaping here.

## 2. Optional `micrometer-tracing` classpath compatibility

`spring-ai-model/pom.xml` declares `micrometer-tracing` with `<optional>true</optional>` (L61-65), so it does not transitively reach consumers of `spring-ai-model`.

This PR adds direct imports of `io.micrometer.tracing.Span` and `io.micrometer.tracing.handler.TracingObservationHandler` to `ChatModelPromptContentObservationHandler`. However, `ChatObservationAutoConfiguration` has a fallback path — `TracerNotPresentObservationConfiguration` (L122-133) — that instantiates this handler directly even when tracing is missing from the classpath:

```java
@Configuration(proxyBeanMethods = false)
@ConditionalOnMissingClass("io.micrometer.tracing.Tracer")
static class TracerNotPresentObservationConfiguration {

    @Bean
    @ConditionalOnMissingBean
    @ConditionalOnProperty(prefix = ChatObservationProperties.CONFIG_PREFIX, name = "log-prompt",
            havingValue = "true")
    ChatModelPromptContentObservationHandler chatModelPromptContentObservationHandler() {
        logPromptContentWarning();
        return new ChatModelPromptContentObservationHandler();
    }
}
```

For consumers without `micrometer-tracing` on the classpath, loading this handler class during `ApplicationContext` refresh could throw `NoClassDefFoundError` and fail startup.

One direction consistent with the existing structure would be to move the span tagging into a separate handler (e.g., `ChatModelPromptSpanContentObservationHandler`) guarded by `@ConditionalOnClass(Tracer.class)` + `@ConditionalOnBean(Tracer.class)`, keeping the existing logging handler free of tracing imports. That mirrors the current `TracerPresentObservationConfiguration` vs `TracerNotPresentObservationConfiguration` split.

## 3. OTel spec's content-capture opt-in guidance

The OTel GenAI semantic conventions mark `gen_ai.input.messages` / `gen_ai.output.messages` as opt-in and state:

> "SHOULD NOT capture by default, but SHOULD provide an option for users to opt in."

— [gen-ai-spans](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/)

Currently, when a `TracingContext`/`Span` is present, this PR tags `gen_ai.input.messages` without a separate content-capture opt-in. Spring AI already has the `spring.ai.chat.observations.log-prompt` flag — reusing it, or introducing a dedicated flag such as `include-content`, so that the span tag is set only when the user has explicitly opted in, would be consistent with this guidance. Exposing prompt content on spans by default without an explicit content-capture opt-in is inconsistent with the spec's SHOULD-NOT guidance.

## 4. Scope question for maintainers

Two items from the issue body and my earlier analysis are not covered by this PR. I'd appreciate maintainer input on whether they belong in the same PR or in a follow-up:

- **Add `INPUT_MESSAGES`/`OUTPUT_MESSAGES` constants to `AiObservationAttributes`** — the PR currently defines these keys as package-private constants inside the handler, so other conventions/handlers cannot reuse them.
- **Counterpart for `ChatModelCompletionObservationHandler`** — the output-messages side is missing. Given that the issue covers both input and output messages, it would feel natural to pair them.

EvanYao826:
@jewoodev Thanks for the detailed analysis! You're right on all four points. Here's what I've changed:

**1. JSON escaping** — Replaced manual string concatenation with Jackson `ObjectMapper` (already a compile dependency). The serializer properly handles quotes, backslashes, and control characters.

**2. micrometer-tracing classpath** — Moved span tagging into a new `ChatModelPromptSpanContentObservationHandler` guarded by `@ConditionalOnClass(Tracer.class)`. The existing logging handler remains free of tracing imports, so `TracerNotPresentObservationConfiguration` continues to work without `micrometer-tracing` on the classpath.

**3. OTel content-capture opt-in** — The span tagging now only activates when `spring.ai.chat.observations.log-prompt=true`, consistent with the existing opt-in flag and the OTel spec's SHOULD-NOT default guidance.

**4. Constants** — Moved `INPUT_MESSAGES` and `OUTPUT_MESSAGES` to `AiObservationAttributes` so other conventions/handlers can reuse them.

Regarding scope: I agree `OUTPUT_MESSAGES` for the completion handler should be a follow-up PR. Happy to work on that next.

Pushed the updated commits. Looking forward to your feedback!

jewoodev:
@EvanYao826 Thanks for the update — the four fixes all look right from the commit message.

Heads up that the new commit isn't reflected on this PR yet. The PR's head ref `fix/chat-model-prompt-content-span-tags` is still at the original `4a6368e`, while the updated commit `8fadd0e` ("address review feedback on span attributes for input messages") landed on a separate branch `fix/span-attributes-input-messages` in your fork. Looks like a branch mix-up.

If you push (or cherry-pick) `8fadd0e` onto `fix/chat-model-prompt-content-span-tags`, the PR will pick it up automatically. Happy to re-review once it's in.

spring-ai-commons/src/main/java/org/springframework/ai/observation/conventions/AiObservationAttributes.java
```diff
@@ -40,6 +40,19 @@ public enum AiObservationAttributes {
...
+	/**
+	 * The input messages sent to the model, serialized as a JSON array following the
+	 * OpenTelemetry GenAI semantic conventions.
+	 */
+	INPUT_MESSAGES("gen_ai.input.messages"),
+	/**
+	 * The output messages received from the model, serialized as a JSON array following
+	 * the OpenTelemetry GenAI semantic conventions.
+	 */
+	OUTPUT_MESSAGES("gen_ai.output.messages"),
```
jewoodev: Adds `OUTPUT_MESSAGES` to `AiObservationAttributes` but only uses `INPUT_MESSAGES`.

jewoodev (commented):
Was the output constant added intentionally for something here? I ask because I just opened #6193, the output-side counterpart, which actually consumes `gen_ai.output.messages`. If it indeed isn't used here, would you mind dropping `OUTPUT_MESSAGES` so it lives in the PR that consumes it? That keeps each constant added by the PR that uses it and avoids both PRs adding the same enum member.

spring-ai-commons/src/main/java/org/springframework/ai/observation/conventions/AiObservationAttributes.java
EvanYao826: Good catch — I've removed the unused `OUTPUT_MESSAGES` constant. Only `INPUT_MESSAGES` is referenced in this handler. Updated the PR accordingly.

EvanYao826:
Hi @jewoodev, thanks for catching the unused `OUTPUT_MESSAGES` constant — I've removed it and pushed the fix. Could you please take another look? The latest commit should be reflected on the PR now. Let me know if there's anything else to address!

EvanYao826:
@jewoodev Thanks for the heads up! I've now pushed all the fixes to the correct PR branch `fix/chat-model-prompt-content-span-tags`. The latest commits include:

1. Address review feedback (Jackson ObjectMapper, micrometer-tracing guard, new handler class)
2. Remove unused `OUTPUT_MESSAGES` constant
3. Remove accidental .gitignore entry

Could you take another look? Thanks!

EvanYao826:
Thanks for the thorough review @jewoodev! Glad the four fixes all look correct. Your analysis was very helpful — especially the point about JSON escaping in filter output and the auto-close timing for ObservationContext.

Appreciate you catching this in parallel — great minds think alike! 😄
Let me know if there's anything else you'd like me to adjust.

jewoodev:
Thanks for the updates. From the current PR state, it looks like the review context I shared has been reflected, and the remaining issue I notice is the failing DCO check.

That said, I'm not a maintainer on this project and I won't be able to keep following this PR for further review rounds. Please rely on the Spring AI maintainers for the next review/merge decisions from here. Thanks for understanding.

Issue #5563, opened 2026-03-07
Bug: Race condition in WebClientStreamableHttpTransportErrorHandlingIT.testSessionRecoveryAfter404


galt-k:
## Problem
The `testSessionRecoveryAfter404` test has a race condition where an assertion runs before the previous async HTTP request completes, causing intermittent CI/CD failures.

## Root Cause
Missing synchronization between async operations in the test's Mono chain. The test was asserting that a session ID was cleared before the HTTP request that invalidated the session had actually completed.

## Solution
Add a 200ms delay between session loss handling and subsequent assertions to ensure proper async coordination (similar to existing pattern in `test404WithSessionId`).

## Impact
This race condition has been causing intermittent CI/CD failures in the WebFlux transport integration tests.

## Related PR
PR: feature/fix-mcp-transport-race-condition (galt-k fork)
Commit: be06d3d - "Fix race condition in MCP WebFlux transport error handling test"

## Test Results
All 6 tests in WebClientStreamableHttpTransportErrorHandlingIT now pass consistently:
- test404WithoutSessionId ✓
- test404WithSessionId ✓
- test400WithoutSessionId ✓
- test400WithSessionId ✓
- testSessionRecoveryAfter404 ✓
- testReconnectErrorHandling ✓

PR #6062, opened 2026-05-17
Fix `OllamaChatModel` dropping thinking field in multi-turn conversation history

### LINKED ISSUE #6016: `reasoning_content` / `thinking` is received from the model but silently dropped when replaying conversation history
### Summary

All current `ChatModel` implementations that support reasoning/thinking models correctly
extract `reasoning_content` (or `thinking`) from the model response and store it in
`AssistantMessage` metadata or a subclass field. However, when that same
`AssistantMessage` is serialised back into a follow-up request (i.e. replayed as
conversation history), the field is silently set to `null`. This breaks multi-turn
conversations with any model that requires `reasoning_content` to be echoed back in the
conversation history (DeepSeek-R1, Qwen3, etc.).

---

### Affected models

| ChatModel | Receives reasoning | Sends reasoning back |
|---|---|---|
| `OpenAiChatModel` (2.0.0-M4, OpenAiApi-based) | ✅ stored in `AssistantMessage` metadata key `"reasoningContent"` | ❌ hardcoded `null` in `createRequest()` |
| `OpenAiChatModel` (SNAPSHOT, openai-java SDK) | ✅ stored in `AssistantMessage` metadata key `"reasoningContent"` | ❌ `ChatCompletionAssistantMessageParam.Builder` never sets the thinking field |
| `DeepSeekChatModel` | ✅ stored in `DeepSeekAssistantMessage.reasoningContent` | ❌ hardcoded `null` as the 7th constructor argument in `createRequest()` |
| `OllamaChatModel` | ✅ stored in `AssistantMessage` metadata key `"thinking"` | ❌ `OllamaApi.Message.Builder.thinking()` is never called in `createRequest()` |

---

#### `OpenAiChatModel` (2.0.0-M4)

`createRequest()` constructs `ChatCompletionMessage` for `ASSISTANT` messages with the
`reasoningContent` constructor parameter hardcoded to `null`, regardless of what was
stored in the incoming `AssistantMessage`:

```java
// OpenAiChatModel.java (2.0.0-M4), createRequest(), ASSISTANT branch
return List.of(new ChatCompletionMessage(
    assistantMessage.getText(),
    ChatCompletionMessage.Role.ASSISTANT,
    null, null, toolCalls, null, audioOutput,
    null,
    null   // ← reasoning_content always null, metadata value ignored
));
```

The metadata is populated correctly on the receive side (line 322):
```java
"reasoningContent", choice.message().reasoningContent() != null ? choice.message().reasoningContent() : ""
```

#### `DeepSeekChatModel`

Same pattern. `DeepSeekAssistantMessage.getReasoningContent()` is never read back in
`createRequest()`. The 7th constructor argument is hardcoded `null` (line 426):

```java
return List.of(new ChatCompletionMessage(text, ChatCompletionMessage.Role.ASSISTANT,
    null, null, toolCalls, isPrefixAssistantMessage,
    null   // ← reasoning_content always null
));
```

The fix would be:
```java
String reasoningContent = (message instanceof DeepSeekAssistantMessage dsm)
    ? dsm.getReasoningContent() : null;
return List.of(new ChatCompletionMessage(text, ChatCompletionMessage.Role.ASSISTANT,
    null, null, toolCalls, isPrefixAssistantMessage, reasoningContent));
```

#### `OllamaChatModel`

`OllamaApi.Message.Builder` already has a `.thinking()` method (added for Qwen3 etc.),
but `createRequest()` never calls it (line 466):

```java
return List.of(OllamaApi.Message.builder(Role.ASSISTANT)
    .content(assistantMessage.getText())
    .toolCalls(toolCalls)
    // .thinking(...) ← never called, metadata value ignored
    .build());
```

The fix would be:
```java
String thinking = (String) assistantMessage.getMetadata().get("thinking");
return List.of(OllamaApi.Message.builder(Role.ASSISTANT)
    .content(assistantMessage.getText())
    .toolCalls(toolCalls)
    .thinking(thinking)
    .build());
```

---

### Impact

Any application using a reasoning model in a multi-turn / agentic loop will silently send
incomplete history to the model. Some providers (e.g. DeepSeek) explicitly require
`reasoning_content` to be present alongside `tool_calls` in assistant turns; omitting it
causes API errors or degraded model behaviour.

---

### Expected behaviour

When an `AssistantMessage` that was produced by a reasoning model is replayed as
conversation history, the `reasoning_content` / `thinking` field should be included in
the serialised request, exactly as it was received.

---

### Suggested fix

For each affected `ChatModel`, in the `ASSISTANT` branch of `createRequest()`:

1. Check whether the incoming `AssistantMessage` carries reasoning content
   (via subclass field or metadata key, depending on the model).
2. Pass that value to the corresponding request message field instead of `null`.

No new public API is needed; it is purely a missing read-through in the serialisation
path.

suryateja-g13:
Closes #6016 (Ollama path).

## Problem

`OllamaChatModel` stores the `thinking` field from model responses in
`ChatGenerationMetadata` (via `builder.metadata("thinking", thinking)`),
but when an `AssistantMessage` is replayed as conversation history in a
follow-up request, only `AssistantMessage.properties` is accessible —
`ChatGenerationMetadata` is not available at that point. As a result,
`thinking` was always `null` when building the outbound `OllamaApi.Message`,
silently breaking multi-turn conversations with Qwen3 and other thinking models.

## Fix

Extract `thinking` from the Ollama response at both receive sites (blocking
and streaming) and store it in `AssistantMessage.properties` under the same
`"thinking"` key, in addition to the existing `ChatGenerationMetadata` storage.
The `createRequest()` ASSISTANT branch already reads that key from
`AssistantMessage.getMetadata()` and passes it to `Message.Builder.thinking()`.

## Test plan

- [ ] Unit test: build an `AssistantMessage` with `thinking` in properties, call `ollamaChatRequest()`, verify the outbound `OllamaApi.Message` carries `thinking`
- [ ] Existing `OllamaChatModelTests` pass
- [ ] Streaming path: same round-trip test using the streaming code path

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
@@ -264,8 +264,8 @@ public class OllamaChatModel implements ChatModel {
 							.toList();
 
 				String thinking = ollamaResponse.message().thinking();
-				Map<String, Object> messageProperties = thinking != null
-						? Map.of(THINKING_METADATA_KEY, thinking) : Map.of();
+				Map<String, Object> messageProperties = thinking != null ? Map.of(THINKING_METADATA_KEY, thinking)
+						: Map.of();
 				var assistantMessage = AssistantMessage.builder()
 					.content(ollamaResponse.message().content())
 					.properties(messageProperties)
@@ -357,8 +357,8 @@ public class OllamaChatModel implements ChatModel {
 
 				boolean hasEvalCount = chunk.promptEvalCount() != null && chunk.evalCount() != null;
 				String thinking = chunk.message().thinking();
-				Map<String, Object> messageProperties = thinking != null
-						? Map.of(THINKING_METADATA_KEY, thinking) : Map.of();
+				Map<String, Object> messageProperties = thinking != null ? Map.of(THINKING_METADATA_KEY, thinking)
+						: Map.of();
 				var assistantMessage = AssistantMessage.builder()
 					.content(content)
 					.properties(messageProperties)
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
@@ -263,9 +263,12 @@ public class OllamaChatModel implements ChatModel {
 									ModelOptionsUtils.toJsonString(toolCall.function().arguments())))
 							.toList();
 
+				String thinking = ollamaResponse.message().thinking();
+				Map<String, Object> messageProperties = thinking != null
+						? Map.of(THINKING_METADATA_KEY, thinking) : Map.of();
 				var assistantMessage = AssistantMessage.builder()
 					.content(ollamaResponse.message().content())
-					.properties(Map.of())
+					.properties(messageProperties)
 					.toolCalls(toolCalls)
 					.build();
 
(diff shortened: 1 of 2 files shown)


suryateja-g13:
The failing CI check ('MCP common and annotations integration tests') is caused by a known flaky test: `WebClientStreamableHttpTransportErrorHandlingIT.testSessionRecoveryAfter404` (race condition, tracked in #5563). This failure is unrelated to the Ollama changes in this PR, which only touch `OllamaChatModel.java` and its test. The 'Build branch' check passes cleanly.

PR #6071, opened 2026-05-19
Generate optional nullable schemas for `JsonNullable<T>`

### LINKED ISSUE #6046: Support distinguishing null vs absent fields in MCP tool parameters for partial update operations
## Problem

  When building MCP tools that perform partial updates, there is currently no way to
  distinguish between a field that was explicitly set to `null` (meaning "clear this value")
  and a field that was simply omitted from the call (meaning "leave this value unchanged").

  This is a common requirement for update operations. For example, a tool that updates
  an entity with optional `description` and `parent_id` fields:

  ```java
  @Tool
  public void updateItem(String id, String description, Long parentId) {
      // Was parentId omitted (skip) or explicitly null (clear)?
      // Impossible to tell after deserialization.
  }
```

  When the LLM calls this tool with `{"id": "123", "description": "new value"}`,
  both the omitted parentId and an explicitly passed null arrive as null in Java —
  the intent is lost.

  Expected Behavior
  
  It should be possible to detect whether a parameter was explicitly set to null or
  was absent from the tool call payload.

  One well-established solution is jackson-databind-nullable (https://github.com/OpenAPITools/jackson-databind-nullable),
  which provides a `JsonNullable<T>` wrapper type for exactly this distinction:

```java
  @Tool
  public void updateItem(String id, JsonNullable<String> description, JsonNullable<Long> parentId) {
      if (JsonNullable.undefined().equals(parentId)) {
          // field was omitted — skip
      } else if (parentId.get() == null) {
          // field was explicitly null — clear it
      } else {
          // field has a value — update it
      }
  }
```

  ## Workaround (partial)

  It is possible to fix the deserialization side today by manually patching Spring AI's (v.1.1.4)
  internal ObjectMapper at application startup:

```java
  @PostConstruct
  public void registerJsonNullableModule() {
      JsonParser.getObjectMapper().registerModule(new JsonNullableModule());
  }

  @Bean(name = "mcpServerObjectMapper")
  public ObjectMapper mcpServerObjectMapper() {
      return JsonMapper.builder()
              .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
              .enable(DeserializationFeature.ACCEPT_EMPTY_STRING_AS_NULL_OBJECT)
              .disable(SerializationFeature.FAIL_ON_EMPTY_BEANS)
              .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
              .addModules(JacksonUtils.instantiateAvailableModules())
              .addModule(new JsonNullableModule()) // module registration
              .build();
  }
```

  However, this only partially solves the problem:
  - ✅ JsonNullable<T> fields are correctly deserialized from tool call payloads
  - ❌ JsonSchemaGenerator does not support JsonNullableModule, so the generated
  MCP input schema for JsonNullable<T> patameters is incorrect — the LLM receives
  a wrong or misleading schema for those fields

jewoodev:
This fixes the schema-generation gap from #6046: MCP tools that use `JsonNullable<T>` for partial updates need input schemas that let the model either omit a value or send an explicit `null`.

When `jackson-databind-nullable` is on the classpath, `JsonSchemaGenerator` now registers a `JsonNullable` schema module. It unwraps `JsonNullable<T>` to the wrapped type, adds `null` to the allowed schema type, and keeps unannotated wrappers out of `required` while explicit `@ToolParam`, `@JsonProperty`, and `@Schema` required settings still take precedence.

Tests cover top-level tool parameters and record fields for scalar, list, and object wrappers, including required overrides.

Closes #6046

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -283,6 +283,7 @@
 		<pdfbox.version>3.0.7</pdfbox.version>
 
 		<!-- Utilities -->
+		<jackson-databind-nullable.version>0.2.6</jackson-databind-nullable.version>
 		<jsonschema.version>5.0.0</jsonschema.version>
 		<jtokkit.version>1.1.0</jtokkit.version>
 		<lucene.version>9.12.3</lucene.version>
--- a/spring-ai-model/pom.xml
+++ b/spring-ai-model/pom.xml
@@ -84,6 +84,13 @@
 		</dependency>
 
 		<dependency>
+			<groupId>org.openapitools</groupId>
+			<artifactId>jackson-databind-nullable</artifactId>
+			<version>${jackson-databind-nullable.version}</version>
+			<optional>true</optional>
+		</dependency>
+
+		<dependency>
 			<groupId>org.jetbrains.kotlin</groupId>
 			<artifactId>kotlin-stdlib</artifactId>
 			<optional>true</optional>
@@ -120,7 +127,7 @@
 			<version>${mockk-jvm.version}</version>
 			<scope>test</scope>
 		</dependency>
-		
+
 	</dependencies>
 
 	<build>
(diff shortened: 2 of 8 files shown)


jewoodev:
@sdeleuze, could you take a look at this PR when you have a chance?

PR #6078, opened 2026-05-19
Fix module dependencies to harmonize with spring boot practices

ericbottard:
First PR to refactor dependencies between our modules.

The first commit of 3 has the full rationale.
This PR takes care of vector-stores only, more to come.


sdeleuze:
I propose to move it to RC1 in order to provide those changes for all modules.

ericbottard:
Closing for a slightly different approach were the dependencies towards core functionality will come from starter rather than autoconfigure module

PR #6081, opened 2026-05-20
GH-5930: Restore HTTP-layer observability for Anthropic chat model

### LINKED ISSUE #5930: Customize OkHttp client for Micrometer-based observability
We should explore on a branch how involved would be the customization for bringing back Micrometer-based observability following the following guidelines:
https://github.com/openai/openai-java#customized-okhttpclient
https://platform.claude.com/docs/en/api/sdks/java#custom-http-client

We may have to convert some Kotlin code to Java in the process.

The goal is to evaluate if we would be ok to maintain this amount of code, and if we are able to add [Micrometer-based observablity ](https://github.com/micrometer-metrics/micrometer/tree/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3).

sobychacko:
Fixes: https://github.com/spring-projects/spring-ai/issues/5930

- Swap anthropic-java (umbrella) for anthropic-java-core; supply our own OkHttp-backed HttpClient via the SDK-documented Custom-HTTP-Client path
- New SpringAiAnthropicHttpClient: Java port of the SDK's OkHttpClient.kt, plus Micrometer OkHttpObservationInterceptor and OkHttpConnectionPoolMetrics
- AnthropicChatModel.Builder now accepts MeterRegistry; auto-config wires it from the Spring Boot context
- Sync calls: okhttp.requests span correctly nested under gen_ai.client.operation; timer, traceparent, and connection-pool gauges all flow as expected
- Streaming HTTP spans fire and timer/pool gauges record, but the span is not parented under gen_ai.client.operation — SDK's thenComposeAsync hops onto ForkJoinPool.commonPool() before our client runs, dropping context. Documented with span-name-filtering workaround.

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-anthropic/src/main/java/org/springframework/ai/model/anthropic/autoconfigure/AnthropicChatAutoConfiguration.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-anthropic/src/main/java/org/springframework/ai/model/anthropic/autoconfigure/AnthropicChatAutoConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.ai.model.anthropic.autoconfigure;
 
 import com.anthropic.client.AnthropicClient;
+import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.observation.ObservationRegistry;
 
 import org.springframework.ai.anthropic.AnthropicChatModel;
@@ -53,7 +54,7 @@ public class AnthropicChatAutoConfiguration {
 	@ConditionalOnMissingBean
 	public AnthropicChatModel anthropicChatModel(AnthropicConnectionProperties connectionProperties,
 			AnthropicChatProperties chatProperties, ToolCallingManager toolCallingManager,
-			ObjectProvider<ObservationRegistry> observationRegistry,
+			ObjectProvider<ObservationRegistry> observationRegistry, ObjectProvider<MeterRegistry> meterRegistry,
 			ObjectProvider<ChatModelObservationConvention> observationConvention,
 			ObjectProvider<ToolExecutionEligibilityPredicate> anthropicToolExecutionEligibilityPredicate) {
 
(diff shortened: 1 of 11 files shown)


sobychacko:
@ThomasVitale Could you take a look at this PR and provide any feedback? Thanks in advance!

ThomasVitale:
@sobychacko overall, it looks good to me. I have a couple of suggestions.

* I would perhaps consider introducing a configuration property to control whether the additional connection pool metrics should be enabled (similar to other scenarios in spring like for MongoDB connection pool), with default value being `false`, as they're secondary telemetry compared to the ones generated via the observation for the HTTP interaction.
* the `defaultDispatcherExecutor()` currently uses platform threads. And since there's no parameter for it passed via `AnthropicChatModel`, there is no way to define a different executor unless one re-implements the whole class. It would be great if there was a way to use an executor based on virtual threads when `@ConditionalOnThreading(Threading.VIRTUAL)` (if we can capture that somehow) or when the app runs Java >= 21 or some other way. Or maybe we can re-use one of the builders auto-configured by Spring Boot in [TaskExecutorConfigurations.java](https://github.com/spring-projects/spring-boot/blob/main/core/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutorConfigurations.java). This might be a future enhancement.

sobychacko:
Thanks @ThomasVitale for the review. I will try to address both of the feedback soon. What do you think about the streaming limitation in terms of the parent-child lineage broken? https://github.com/spring-projects/spring-ai/pull/6081/changes#diff-f505c45a30e5af63c059b6bd2b5a80a7ee289571aa842e6592c03d65ef267f3fR1634

sobychacko:
@ThomasVitale introduced a property to control connection pool metrics. Also, enabled user-supplied `ExecutorService` via a builder method.

ThomasVitale:
@sobychacko looks good, thanks! I like the new builder method in AnthropicChatModel for passing an ExecutorService, opening up for future autoconfig possibilities.

ThomasVitale (approved):

sdeleuze (commented):
From a high level perspective (I am not an observability expert), LGTM and the level of code we would have to maintain looks reasonable to me, especially from the angle that it comes from Anthropic impl.

It gives us also a capability to customize even more the http client layer if needed, so may serve other purposes than just observability.

If we decide to move forward, worth evaluating if OpenAI and Anthropic OkHttp infra should share some common bits. Not sure yet we can or want as maybe that's safer to just stick to the Kotlin -> Java translation of the various bits.

I see some limitation for streaming, but I suppose those will not be a concern anymore with the future VT unified infrastructure we plan for 3.0, so just documenting those looks good enough for me.

PR #6095, opened 2026-05-21
Restore ChatClient dependency in model starters

### LINKED ISSUE #6094: `ChatClient` is missing from all starter dependencies
All model starters depend on `spring-ai-autoconfigure-model-chat-client`, which used to bring `spring-ai-client-chat` as a transitive dependency.

After the changes in https://github.com/spring-projects/spring-ai/commit/99b6bcecaf4bdc165520871fd52c01a05dfd4c4e, the transitive dependency has been removed, resulting in all the model starters to now be published without ChatClient support.

Solution: include explicit dependency on `spring-ai-client-chat` in all model starters.

ThomasVitale:
Include explicit dependency on spring-ai-client-chat in all model starters after it was removed as a transitive dependendency on spring-ai-autoconfigure-model-chat-client.

Fixes gh-6094

### DIFF
--- a/starters/spring-ai-starter-model-anthropic/pom.xml
+++ b/starters/spring-ai-starter-model-anthropic/pom.xml
@@ -56,6 +56,12 @@
 
 		<dependency>
 			<groupId>org.springframework.ai</groupId>
+			<artifactId>spring-ai-client-chat</artifactId>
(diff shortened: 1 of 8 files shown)


tzolov:
This is a regression indeed. Thanks @ThomasVitale

PR #6096, opened 2026-05-21
Fix ToolCallAdvisor ignoring auto-configured ToolCallingManager

ThomasVitale:
The `ChatClientAutoConfiguration` is now applied after `ToolCallingAutoConfiguration`, ensuring the auto-configured `ToolCallingManager` is used correctly when building a `ToolCallAdvisor` instance.

I added an integration test to cover this dependency scenario.

The issue was visible by inspecting the telemetry.

Before my fix, no tool execution span was created:

After my fix, the tool execution span is created correctly because the auto-configured ToolCallingManager is now used correctly (which is the one that includes the observability convention configuration).

Screenshot 2026-05-21 at 08 04 14


tzolov:
Nice catch. Thanks @ThomasVitale

Issue #5120, opened 2025-12-18
OpenAiChatModel streaming drops chunks due to internal switchMap when using buffered concatMap


chengong825:
**Bug description**
In our scenario, we need to perform a security check for every 2 streaming tokens using buffer(2) followed by concatMap to call an external validation API. 
When using OpenAiChatModel with its internal switchMap, some tokens or chunks are dropped unexpectedly.

The issue appears to be caused by switchMap canceling previous inner streams prematurely when new tokens arrive, which conflicts with our buffered, sequential processing logic. 
I believe the semantics of switchMap may not be ideal here; using concatMap or flatMapSequential might be more appropriate.

This issue might be related to [#876](https://github.com/spring-projects/spring-ai/issues/876)

**Environment**
Spring AI version:1.1.2
Java version: OpenJDK 21
Operating System:  Windows11

**Steps to reproduce**
1. Run the provided unit tests below.
2. Although we use the internal LLM endpoint, the same issue is expected to occur with other model endpoints, as it is caused by the internal switchMap behavior.

**Expected behavior**
All tokens should be emitted and processed in order, allowing buffered calls to the security check API without losing any chunks.

**Minimal Complete Reproducible example**

```
    @Test
    void testSwitchMapMissingPartialResult() throws InterruptedException {
        // Initialize OpenAI API client
        OpenAiApi openAiApi = OpenAiApi.builder()
                .apiKey("YOUR_API_KEY")
                .restClientBuilder(RestClient.builder())
                .webClientBuilder(WebClient.builder())
                .build();

        // Build OpenAiChatModel with default options
        OpenAiChatModel chatModel = OpenAiChatModel.builder()
                .openAiApi(openAiApi)
                .defaultOptions(OpenAiChatOptions.builder().build())
                .build();

        // Prepare a user message requesting integers from 1 to 300
        UserMessage userMessage = new UserMessage(
                "List all integers from 1 to 300, separated by spaces, with no other text.");
        Prompt prompt = new Prompt(List.of(userMessage));

        // Start streaming chat response
        Flux<ChatResponse> chatResponseFlux = chatModel.stream(prompt);

        StringBuilder answer = new StringBuilder();
        CountDownLatch latch = new CountDownLatch(1);

        chatResponseFlux
                // Append each received token to answer
                .doOnNext(chatResponse -> {
                    String responseContent = chatResponse.getResult().getOutput().getText();
                    answer.append(responseContent);
                })
                // Buffer every 2 tokens
[...]

PR #6106, opened 2026-05-21
fix: Avoid switchMap in OpenAiChatModel

chemicL:
Reactive switchMap operator eagerly subscribes to the source requesting
unlimited data from the upstream. When the downstream is busy processing items
on another thread, the next incoming item will cause a previous one to be
discarded. That can force the streaming variant of ChatModel to silently drop
parts of the streamed response and is highly undesired.

Resolves https://github.com/spring-projects/spring-ai/issues/5120

### DIFF
--- a/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
+++ b/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
@@ -834,7 +834,7 @@ public class BedrockProxyChatModel implements ChatModel {
 					converseStreamRequest, accumulatedUsage)
 				.stream();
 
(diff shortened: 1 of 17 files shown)


sdeleuze:
Would that make sense to backport related test to `main` even if not impacted by the bug to prevent potential regressions?

tzolov:
Yes back porting the tests makes sense.  
Also we have to fix this for all affected models. For example [Mistral](https://github.com/spring-projects/spring-ai/blob/97107e07f65762ba1f0b3bf4f0bf42510b80d3e7/models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatModel.java#L293), Bedrock, MinMax and DeepSeek

chemicL:
The PR is now ready for review. I applied similar changes to the other models to avoid switchMap altogether.

tzolov:
Rebased, code formatted, squashed and merged at 42b9cc72e1d665a417403e403828228bcff254d4

PR #6110, opened 2026-05-21
PgVectorStore : support halfvec, bit and sparsevec column type

R3gardless:
Related to #2903 

Extends `PgVectorStore` to support pgvector's three non-default column types - `halfvec`, `bit` and `sparsevec` in addition to the existing `vector`, Also adds the bit-only `HAMMING_DISTANCE` / `JACCARD_DISTANCE` metrics.
Existing users are unaffected: the default remains `vector` + `cosine` + `HNSW`.

- **pgvector 0.7+ required** for `halfvec`, `bit`, and `sparsevec`. Currently not validated at init, so older version fails with the error `halfvec does not exist`. 
- **bit quantization** the Spring AI internally sign-bit quantizes to bind PGvector `PGbit`. 
 
Reference 
- https://github.com/pgvector/pgvector-java
- https://github.com/pgvector/pgvector


R3gardless:
@ilayaperumalg Gentle ping for review. Thanks!

starJammer:
I support this going in and it's following the existing pattern the PgVector has. 
I have no use for it myself personally, but I have my own PgVector contributions I'd like to make but the lag on reviews here is not inspiring me to even open the PR since I feel like it would go unnoticed for months before anyone looks at it.

PR #6111, opened 2026-05-21
Enforce single ToolAdvisor invariant in DefaultChatClient

tzolov:
- Add validateSingleToolAdvisor() to throw IllegalStateException when more than one ToolAdvisor is present in the advisor chain after auto-registration.
- Delegate legacy toolCallbacks/toolContext methods to tools(ToolSpec) to consolidate validation.
- Remove the streaming boolean from buildAdvisorChain since streamToolCallResponses is now pre-configured on the builder.


sobychacko:
Merged upstream via https://github.com/spring-projects/spring-ai/commit/2f48e9ca04d847009b3cd26491550871834fb917.

Issue #5038, opened 2025-12-05
Missing `reasoning_content` field in the assistant message


sinaweibo:
model: DeepSeek-V3.1-Reasoner
action: call mcp tool and send result 
error message: Missing `reasoning_content` field in the assistant message

<img width="1652" height="245" alt="Image" src="https://github.com/user-attachments/assets/9848a6aa-7bcb-41bb-a8c9-bd4cfc2467ce" />

class org.springframework.ai.chat.messages.AssistantMessage does not have a field named reasoningContent

spring ai version : 1.1.1

abssyyy:
Hi, I saw your description about reasoning_content missing in the assistant message.
However, in my case, I can get reasoning_content from DeepSeek-V3.1-Reasoner, even with Spring AI 1.1.1.

Even a very simple call returns it correctly.

<img width="1199" height="464" alt="Image" src="https://github.com/user-attachments/assets/539d8417-5d2a-44ce-8e46-2a77cbdc4b15" />

Here is my code:
```java
@ApiResource(
    name = "AI multimodal chat",
    path = "/chat/tools/multimodal-chat",
    method = RequestMethod.POST,
    produces = MediaType.TEXT_EVENT_STREAM_VALUE
)
public Flux<ChatResponse> multimodalChat(@Validated @RequestBody MultimodalChatRequest request) {

    final ChatResponse chatResponse = mcpChatService
        .getMultimodalKnowledge(request)
        .call()
        .chatResponse();

    return mcpChatService
        .getMultimodalKnowledge(request)
        .stream()
        .chatResponse();
}

private ChatClient.ChatClientRequestSpec buildChatRequest(
        MultimodalChatRequest request,
        String userId,
        Pair<String, List<Citation>> ctxAndCitations,
        String systemMessage) {

    return deepseekV3ClientWithJdbc
        .prompt()
        .system(systemMessage)
        .user(request.getText())
        .advisors(config -> config
            .param("conversation_id", request.getConversationId())
            .param("user_id", userId)
            .param("external_system", request.getExternalSystem() ? 1 : 0)
            .param("citations", ctxAndCitations.getRight())
        );
}
```
I also have screenshots showing that reasoning_content is included in the raw response.

So I'm trying to understand —
how are you calling the API in your code?

Can you share:

your calling snippet,

or the part where Spring AI attempts to parse the response,

or how your MCP/tool-calling pipeline is built?

Maybe the issue only appears in a specific usage path or message mapping process.

[...]

PR #6113, opened 2026-05-22
Fix DeepSeek dropping assistant reasoning_content on multi-turn requests

GitZhangChi:
`DeepSeekChatModel#buildGeneration` already captures `reasoning_content` from the response into `DeepSeekAssistantMessage`, but `createRequest` then drops it when serializing the next outbound request — it hardcodes `reasoning_content = null` even when the prompt's assistant message is a `DeepSeekAssistantMessage` carrying a real reasoning trace. DeepSeek's API rejects subsequent turns that drop the field with `HTTP 400 — The reasoning_content in the thinking mode must be passed back to the API.`, which is most visible during a tool-calling round (user → assistant(tool_calls, reasoning_content) → tool result → second model call without reasoning_content → 400).

This PR forwards `reasoningContent` (and `prefix`, already special-cased) from a `DeepSeekAssistantMessage` back into the outbound `ChatCompletionMessage` record. Plain `AssistantMessage` instances continue to serialize with `reasoning_content = null` — no behavior change for non-thinking flows.

Tests:
- `createRequestPreservesReasoningContentFromDeepSeekAssistantMessage` pins the outbound roundtrip.
- `createRequestKeepsReasoningContentNullForPlainAssistantMessage` guards plain `AssistantMessage` behavior.

Related: gh-5038

### DIFF
--- a/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekChatModel.java
+++ b/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/DeepSeekChatModel.java
@@ -430,14 +430,24 @@ public class DeepSeekChatModel implements ChatModel {
 					}).toList();
 				}
 				Boolean isPrefixAssistantMessage = null;
-				if (message instanceof DeepSeekAssistantMessage
-						&& Boolean.TRUE.equals(((DeepSeekAssistantMessage) message).getPrefix())) {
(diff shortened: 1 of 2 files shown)


sunmh207:
I've also encountered this problem, and I'd like to know when a new version will be released.

sdeleuze:
Supposed to be fixed in Spring AI 2.x, please create an issue with a reproducer if that's not the case.

PR #6126, opened 2026-05-22
GH-6119: Use class literal for ChatClient autoconf ordering

### LINKED ISSUE #6119: Use .class reference for autoconf ordering
Following [this fix](https://github.com/spring-projects/spring-ai/commit/04fc1e3d8e3c232e66616b20cf6b8ca6c8727c9c), it would be good to introduce a pom dependency from chat-model to tool-calling autoconf module and use `org.springframework.ai.model.tool.autoconfigure.ToolCallingAutoConfiguration.class` literal to avoid regressions in case of renaming

Introducing such a dependency is especially ok now that everything is marked optional in autoconfs, so it doesn't drag anything in if not explicitly requested by the user

zxuhan:
## Summary

`ChatClientAutoConfiguration` references `ToolCallingAutoConfiguration` for ordering via `afterName` with a hardcoded fully qualified string. A rename or package move of `ToolCallingAutoConfiguration` would silently break the ordering without any compile-time signal. Per #6119, switching to a `.class` literal makes the reference type-safe.

This change promotes the dependency on `spring-ai-autoconfigure-model-tool` from test to optional main scope so the class is visible to the chat-client autoconf module. The dependency stays `optional` and `spring-ai-autoconfigure-model-tool` itself marks its transitive deps optional, so consumers that do not opt in are unaffected.

## Changes

- `spring-ai-autoconfigure-model-chat-client/pom.xml`: move the `spring-ai-autoconfigure-model-tool` dependency from `test` scope to optional main scope.
- `ChatClientAutoConfiguration`: replace `afterName = "org.springframework.ai.model.tool.autoconfigure.ToolCallingAutoConfiguration"` with `after = ToolCallingAutoConfiguration.class` and add the corresponding import.

The existing `toolCallAdvisorBuilderUsesAutoConfiguredToolCallingManager` test continues to verify the ordering is honored.

Fixes #6119

### DIFF
--- a/auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client/pom.xml
+++ b/auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client/pom.xml
@@ -32,6 +32,13 @@
 		</dependency>
 
 		<dependency>
+			<groupId>org.springframework.ai</groupId>
+			<artifactId>spring-ai-autoconfigure-model-tool</artifactId>
+			<version>${project.parent.version}</version>
+			<optional>true</optional>
+		</dependency>
+
+		<dependency>
 			<groupId>io.micrometer</groupId>
 			<artifactId>micrometer-tracing</artifactId>
 			<optional>true</optional>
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
@zxuhan Thanks for the PR! The changes look good. Rebased and merged via 45b0c0bddbcd8cc51c8199f3b0f0acbcfff4480c

PR #6134, opened 2026-05-23
Use class reference for tool auto-configuration order

### LINKED ISSUE #6119: Use .class reference for autoconf ordering

juanjuandog:
Closes #6119.

This updates `ChatClientAutoConfiguration` to order after `ToolCallingAutoConfiguration.class` directly instead of using a string-based `afterName` reference.

It also moves the tool auto-configuration module dependency from test scope to an optional main dependency so the class literal is available at compile time without adding a required transitive dependency for users.

Tests:
- `./mvnw -q -pl auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client -am -DskipITs -DskipDocs test`

### DIFF
--- a/auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client/pom.xml
+++ b/auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client/pom.xml
@@ -32,6 +32,13 @@
 		</dependency>
 
 		<dependency>
+			<groupId>org.springframework.ai</groupId>
+			<artifactId>spring-ai-autoconfigure-model-tool</artifactId>
+			<version>${project.parent.version}</version>
+			<optional>true</optional>
+		</dependency>
+
+		<dependency>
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
@juanjuandog Thanks for the PR! The same issue was addressed via https://github.com/spring-projects/spring-ai/pull/6126. I have cherry-picked the tests changes from your PR and applied via 7969b4d973aa4193eac5a22887215c9cdf66b0dc

PR #6139, opened 2026-05-24
GH-5987: Fix OpenAiChatModel.stream() buffering all chunks before emission

### LINKED ISSUE #5987: `OpenAiChatModel#stream` buffers the entire response before emitting chunks

gmjneko:
Closes GH-5987

This PR fixes `OpenAiChatModel.stream()` so regular content chunks are emitted as they arrive instead of being buffered until the OpenAI stream completes.

## Background

`OpenAiChatModel.stream()` previously collected the entire stream before continuing with response aggregation and tool-call handling. This caused a real streaming request to behave like a buffered response: clients would not receive partial content chunks incrementally, and all chunks were emitted only after the upstream stream completed.

The terminal aggregation flow made sense for reconstructing tool calls, because OpenAI tool-call deltas can arrive across multiple chunks and need to be merged before execution or final emission. However, the same full-stream buffering was also applied to regular content chunks. As a result, even content chunks that did not require aggregation were blocked behind stream completion.

This was especially visible for normal text streaming scenarios, where the expected behavior is that each content delta is propagated immediately.

## Solution

This change separates the streaming path into two responsibilities:

1. Emit regular content chunks immediately.
2. Keep end-of-stream aggregation only for cases that actually need it, such as tool-call reconstruction, final usage metadata, and internal tool execution.

The implementation now creates the streaming `Flux<ChatResponse>` from the async OpenAI streaming client and records received responses in a synchronized list for final aggregation. While the stream is active, responses that do not contain tool calls are passed downstream immediately.

At stream completion, the collected responses are aggregated once to preserve behavior that depends on full-stream context, including metadata, usage, tool-call reconstruction, and internal tool execution.

This change also:

- Removes the `collectList()`-based streaming path that delayed downstream emissions.
- Emits non-tool-call content chunks immediately as they are received.
- Keeps aggregation at stream completion for metadata, usage, and tool-call reconstruction.
- Preserves tool-call behavior by emitting the merged tool call only after all tool-call delta chunks have arrived.
- Extracts streaming aggregation and tool execution handling into smaller helper methods.

For tool-call chunks, the implementation intentionally does not emit partial tool-call responses immediately. Tool-call arguments can be split across multiple chunks, so emitting them early would expose incomplete tool-call data. Instead, tool calls are merged and emitted after the stream completes, matching the previous logical behavior while avoiding unnecessary buffering for normal content streaming.

## Tests

- All tests passed.

- run `./mvnw -pl models/spring-ai-openai -Dtest=OpenAiChatModelTests test`

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -18,6 +18,7 @@ package org.springframework.ai.openai;
 
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -25,6 +26,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 import com.openai.client.OpenAIClient;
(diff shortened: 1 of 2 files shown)


gmjneko:
@sdeleuze Hi, could you review this PR when you have a chance? This issue has a significant impact on users, so it would be great if we could get a fix in soon. Thanks!

posuhov:
@tzolov can you help with review for this PR? Is there any chance to include it to 2.0.0-M8?

ericbottard:
Thanks for your contribution @Akika404 

This issue is being addressed via a more thorough rewrite of the stream() method and is coming to 2.0.0.RC1

I will thus close this PR

PR #6141, opened 2026-05-24
Fix ToolCallback overload selection in ChatClient

### LINKED ISSUE #2495: The overload `tools` method in DefaultChatClient causes FunctionToolCallback not taking effect.
I have the following `FunctionToolCallback` instance created and registered to ChatClient Builder as below:

```java
FunctionToolCallback toolCallback1 = ...
FunctionToolCallback toolCallback2 = ...
ChatClient.create(chatModel)
    .prompt("What's the weather like in Copenhagen?")
    .tools(toolCallback1, toolCallback2)
    .call()
    .content();
```

The above will take effect because it actually calls the `Builder defaultTools(Object... toolObjects);` overload method instead of the `Builder defaultTools(List<ToolCallback> toolCallbacks);`.

`Builder defaultTools(Object... toolObjects);` method only works for Method tool, because inside `ToolCallbacks.from(toolObjects)` it only accepts Method tools:

```java
@Override
public ChatClientRequestSpec tools(Object... toolObjects) {
	Assert.notNull(toolObjects, "toolObjects cannot be null");
	Assert.noNullElements(toolObjects, "toolObjects cannot contain null elements");
	this.functionCallbacks.addAll(Arrays.asList(ToolCallbacks.from(toolObjects))); // here only accepts Method tools
	return this;
}
```

Only the following form is correct:

```java
ToolCallback toolCallback1 = ... // must be ToolCallback type but not FunctionToolCallback type.
ToolCallback toolCallback2 = ... // must be ToolCallback type but not FunctionToolCallback type.
ChatClient.create(chatModel)
    .prompt("What's the weather like in Copenhagen?")
    .tools(toolCallback1, toolCallback2)
    .call()
    .content();
```

It's very easy to make the above mistake. There's no compilation error because of the overload methods and it's not easy to debug and find the problem.

juanjuandog:
A direct ToolCallback argument to tools(...) binds to Object... today. That sends FunctionToolCallback through the @Tool method path and ends with "No @Tool annotated methods found".

This adds the specific ToolCallback and ToolCallbackProvider overloads for tools(...) and defaultTools(...), routed through the existing ToolSpec callback path.

Closes #2495

Tests:
- ./mvnw -Dmaven.build.cache.enabled=false -pl spring-ai-client-chat -Dtest=DefaultChatClientTests#whenToolCallbackPassedDirectlyToToolsThenReturn,DefaultChatClientBuilderTests#whenToolCallbackPassedDirectlyToDefaultToolsThenReturn test
- ./mvnw -Dmaven.build.cache.enabled=false -pl spring-ai-client-chat test

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
@@ -223,6 +223,14 @@ public interface ChatClient {
 
 		ChatClientRequestSpec tools(Consumer<ToolSpec> consumer);
 
+		default ChatClientRequestSpec tools(ToolCallback... toolCallbacks) {
+			return tools(t -> t.callbacks(toolCallbacks));
+		}
+
+		default ChatClientRequestSpec tools(ToolCallbackProvider... toolCallbackProviders) {
+			return tools(t -> t.callbacks(toolCallbackProviders));
+		}
+
 		ChatClientRequestSpec tools(Object... toolObjects);
 
 		/**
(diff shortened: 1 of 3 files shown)


juanjuandog:
This still looks clean from my side. The branch is up to date and the checks are green, so I think this is ready for review when you have a chance.

PR #6151, opened 2026-05-25
Optimize Typesense vector query construction

chabinhwang:
## Summary

- Build the Typesense vector query directly from the embedding `float[]` instead of creating a boxed `Stream<Float>` and intermediate `List` before `String.join`.
- Add focused tests to lock the generated vector query format, including the empty-embedding edge case.

## Why

This preserves the existing Typesense vector query string format while reducing temporary allocations during similarity search request construction.

## Validation

- `./mvnw -pl vector-stores/spring-ai-typesense-store -Dtest=TypesenseVectorStoreTests test`
- `./mvnw -pl vector-stores/spring-ai-typesense-store package`
- `git diff --check`

## Notes

No associated issue; this is a small internal allocation optimization with behavior-preserving tests.


chabinhwang:
Superseded by #6707, which is rebased on current `main`, corrects the buffer capacity estimate so it actually avoids the resize, and adds format tests for the sign and scientific-notation cases. Closing this one to avoid a duplicate.

PR #6156, opened 2026-05-26
GH-6130: Skip Google GenAI embedding auto-config when no credentials are configured

### LINKED ISSUE #6130: When using only Google Gen AI chat, configuring the embeddings shouldn't be required
I want to use just the Googe GenAI chat, not the embeddings. Though when including `org.springframework.ai:spring-ai-starter-model-google-genai` I am force to configure the embeddings (at least the project ID). There is no property to disable the embeddings support.

So I need to resort to
```java
@SpringBootApplication(exclude = {
        GoogleGenAiEmbeddingConnectionAutoConfiguration.class,
        GoogleGenAiTextEmbeddingAutoConfiguration.class
})
```

Please remove the need to configure the embeddings or add an option to disable them.

suryateja-g13:
## Summary

Fixes #6130.

`GoogleGenAiEmbeddingConnectionAutoConfiguration` was creating the connection-details bean unconditionally and throwing `IllegalArgumentException` when neither `api-key` nor `project-id` was set. Users who included `spring-ai-starter-model-google-genai` but only wanted the chat model were forced to supply embedding credentials or exclude the two embedding auto-configuration classes manually.

**Changes:**

- Introduces `OnGoogleGenAiEmbeddingCredentials` — an `AnyNestedCondition` that evaluates to `true` when `spring.ai.google.genai.embedding.api-key` **or** `spring.ai.google.genai.embedding.project-id` is present.
- Annotates `GoogleGenAiEmbeddingConnectionAutoConfiguration#googleGenAiEmbeddingConnectionDetails` with `@Conditional(OnGoogleGenAiEmbeddingCredentials.class)` so the bean is skipped when no credentials are configured.
- Adds `@ConditionalOnBean(GoogleGenAiEmbeddingConnectionDetails.class)` to `GoogleGenAiTextEmbeddingAutoConfiguration#googleGenAiTextEmbedding` so the model bean is silently skipped when the connection bean is absent.

**Behaviour after the fix:**

| Scenario | Before | After |
|---|---|---|
| No embedding credentials, only chat | `IllegalArgumentException` at startup | Embedding beans absent, chat works normally |
| `api-key` set | Works | Works (unchanged) |
| `project-id` + `location` set | Works | Works (unchanged) |
| User-provided `GoogleGenAiEmbeddingConnectionDetails` bean | Works | Works (unchanged, `@ConditionalOnMissingBean` still respected) |

## Test plan

- [ ] `GoogleGenAiEmbeddingConnectionAutoConfigurationTests#noEmbeddingBeansCreatedWhenNoCredentialsConfigured` — context starts cleanly with no embedding beans
- [ ] `#embeddingConnectionCreatedWhenApiKeyPresent` — connection bean created when api-key property is set
- [ ] `#embeddingModelNotCreatedWhenConnectionDetailsAbsent` — embedding model bean absent when connection details are absent

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-google-genai/src/main/java/org/springframework/ai/model/google/genai/autoconfigure/embedding/GoogleGenAiEmbeddingConnectionAutoConfiguration.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-google-genai/src/main/java/org/springframework/ai/model/google/genai/autoconfigure/embedding/GoogleGenAiEmbeddingConnectionAutoConfiguration.java
@@ -27,6 +27,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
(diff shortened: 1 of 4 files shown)


sdeleuze:
#6171 will probably fix the original issue.

suryateja-g13:
Makes sense , Thanks !!

PR #6157, opened 2026-05-26
GH-6120: Add REQUIRED constant to ToolChoiceBuilder in DeepSeek and MiniMax APIs

### LINKED ISSUE #6120: ToolChoiceBuilder: REQUIRED type
In `ToolChoiceBuilder`,we need to ask llm to call tools which are required

```java
		/**
		 * Helper factory that creates a tool_choice of type 'none', 'auto' or selected function by name.
		 */
		public static class ToolChoiceBuilder {
			/**
			 * Model can pick between generating a message or calling a function.
			 */
			public static final String AUTO = "auto";
			/**
			 * Model will not call a function and instead generates a message
			 */
			public static final String NONE = "none";

			/**
			 * Specifying a particular function forces the model to call that function.
			 */
			public static Object function(String functionName) {
				return Map.of("type", "function", "function", Map.of("name", functionName));
			}
		}
```

suryateja-g13:
## Summary

- Adds `REQUIRED` constant to `ToolChoiceBuilder` in `DeepSeekApi` and `MiniMaxApi`
- `REQUIRED` forces the model to call at least one tool, but allows it to choose which one
- Aligns with the OpenAI spec where `tool_choice: "required"` is a valid value

Fixes: #6120

## Test plan

- [ ] Build passes: `./mvnw -pl models/spring-ai-deepseek,models/spring-ai-minimax install -DskipTests`
- [ ] New constant is accessible as `DeepSeekApi.ChatCompletionRequest.ToolChoiceBuilder.REQUIRED`
- [ ] New constant is accessible as `MiniMaxApi.ChatCompletionRequest.ToolChoiceBuilder.REQUIRED`

### DIFF
--- a/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/api/DeepSeekApi.java
+++ b/models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/api/DeepSeekApi.java
@@ -547,6 +547,10 @@ public class DeepSeekApi {
 			 * Model will not call a function and instead generates a message
 			 */
 			public static final String NONE = "none";
+			/**
+			 * Model must call at least one tool, but it can choose which one.
+			 */
+			public static final String REQUIRED = "required";
 
 			/**
 			 * Specifying a particular function forces the model to call that function.
--- a/models/spring-ai-minimax/src/main/java/org/springframework/ai/minimax/api/MiniMaxApi.java
+++ b/models/spring-ai-minimax/src/main/java/org/springframework/ai/minimax/api/MiniMaxApi.java
@@ -625,6 +625,10 @@ public class MiniMaxApi {
 			 * Model will not call a function and instead generates a message
 			 */
 			public static final String NONE = "none";
+			/**
+			 * Model must call at least one tool, but it can choose which one.
+			 */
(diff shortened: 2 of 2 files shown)


lplusc:
What about other providers? Why Only DeepSeek and MiniMax APIs?

suryateja-g13:
Hi @lplusc — only **DeepSeek** and **MiniMax** define a `ToolChoiceBuilder` helper class (in `DeepSeekApi` and `MiniMaxApi` respectively). That class already has `AUTO` and `NONE` string constants, but `REQUIRED` was missing from both.

The other providers use different patterns:
- **OpenAI**: `ChatCompletionToolChoiceOption` (a sealed type from the `openai-java` SDK) — already handles `"required"` via `ChatCompletionToolChoiceOption.ofRequired(...)` 
- **Anthropic**: uses `tool_choice` as a typed object, not a string constant
- **Mistral AI**: uses `ToolChoice` enum (`AUTO`, `NONE`, `ANY`) — no `REQUIRED` concept applicable
- **Google GenAI** / **Bedrock**: no comparable string-based tool-choice API

So the scope of this PR is intentionally limited to the two providers that share the `ToolChoiceBuilder` pattern and were both missing the `REQUIRED` constant.

PR #6160, opened 2026-05-26
Consolidate JSON Schema generation documentation between Tool Calling and Structured Output

### LINKED ISSUE #5915: Consolidate JSON Schema generation documentation between Tool Calling and Structured Output Conversion
Following the consolidation in #5897, `BeanOutputConverter` now delegates to `JsonSchemaGenerator` for schema generation, aligning the behavior between tool calling and structured output conversion.

The reference documentation for these two features should be reviewed and consolidated, since they now share the same schema generation logic — in particular, the supported annotations (`@ToolParam`, `@JsonProperty`, `@Schema`, `@Nullable`) and the default required-property behavior.

Discussed in #5897.

ChoMinGi:
Fixes https://github.com/spring-projects/spring-ai/issues/5915

Following the consolidation in #5897, `BeanOutputConverter` now delegates to `JsonSchemaGenerator` for schema generation, aligning the behavior between tool calling and structured output conversion.

However, the documentation for these two features remained separate — the supported annotations and default required-property behavior were only documented in the Tool Calling page.
- Add a "JSON Schema Customization" section to `structured-output-converter.adoc` documenting the supported annotations for description (`@JsonClassDescription`, `@JsonPropertyDescription`, `@Schema`) and required/optional status (`@JsonProperty`, `@Schema`, `@Nullable`)
- Add a cross-reference from `tools.adoc` to `structured-output-converter.adoc`, noting that `BeanOutputConverter` shares the same `JsonSchemaGenerator`

### DIFF
--- a/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/structured-output/converters.adoc
+++ b/spring-ai-docs/src/main/antora/modules/ROOT/pages/api/structured-output/converters.adoc
@@ -164,8 +164,8 @@ ActorsFilms actorsFilms = this.beanOutputConverter.convert(this.generation.getOu
 
 You can provide a description for the properties of the target type using one of the following annotations:
 
-* `@JsonClassDescription(description = "...")` — Jackson
-* `@JsonPropertyDescription(description = "...")` — Jackson
+* `@JsonClassDescription("...")` — Jackson
+* `@JsonPropertyDescription("...")` — Jackson
 * `@Schema(description = "...")` — Swagger
 
 This also works recursively for nested types.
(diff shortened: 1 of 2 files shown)


spring-ai-docs/src/main/antora/modules/ROOT/pages/api/structured-output/converters.adoc
```diff
@@ -155,6 +155,51 @@ Generation generation = chatModel.call(
...
+}
+----
+
+**Required vs optional**
+
+By default, each property is considered required. You can make a property optional using one of the following annotations, in this order of precedence:
+
+* `@JsonProperty(required = false)` — Jackson
+* `@Schema(required = false)` — Swagger
+* `@Nullable` — Spring Framework / JSpecify
```
nicolaskrier: Spring Framework's one is deprecated, so I would recommend not highlighting it. Please note that Jakarta Annotations' one is a viable alternative for this use case, it may be interesting to mention it as well.

ChoMinGi: Good catch, thanks! Fixed — the bullet now reads `@Nullable` — JSpecify / Jakarta Annotations. Both are honored via `org.springframework.core.Nullness`, and I verified the Jakarta one with a runtime test. I applied the same fix to the original list in tools.adoc that this section was consolidated from, so the two documents stay consistent.

While verifying, I found and fixed a few more inaccuracies in the same lists: `@Schema(required = false)` never actually makes a property optional (`RequiredMode.AUTO` is treated as required), so both lists now use `@Schema(requiredMode = Schema.RequiredMode.NOT_REQUIRED)` — matching the tests and your #6541 examples; the Jackson description bullets used a non-existent `description` attribute; and the precedence claim is now limited to method parameters, since victools ORs required checks across modules for bean properties.

PR #6162, opened 2026-05-26
Extract AbstractJsonSchemaGenerator to remove duplication between JsonSchemaGenerator and McpJsonSchemaGenerator

tzolov:
Extract shared parameter-inspection logic into a new AbstractJsonSchemaGenerator base class. 

The JsonSchemaGenerator and McpJsonSchemaGenerator generators now implement two abstract hook methods (`getToolAnnotationRequired` / `getToolAnnotationDescription`) for their respective tool annotations (@ToolParam vs @McpToolParam), while the common required-check, description-resolution, forbidAdditionalProperties, and convertTypeValuesToUpperCase logic lives in the base class. 
Also fixes pre-existing Jackson 3 deprecations (isTextual → isString, asText → stringValue).

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/utils/McpJsonSchemaGenerator.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/utils/McpJsonSchemaGenerator.java
@@ -24,8 +24,6 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.annotation.JsonPropertyDescription;
 import com.github.victools.jsonschema.generator.Module;
 import com.github.victools.jsonschema.generator.Option;
 import com.github.victools.jsonschema.generator.OptionPreset;
@@ -41,8 +39,6 @@ import io.modelcontextprotocol.server.McpAsyncServerExchange;
 import io.modelcontextprotocol.server.McpSyncServerExchange;
 import io.modelcontextprotocol.spec.McpSchema.CallToolRequest;
 import io.modelcontextprotocol.util.Assert;
-import io.modelcontextprotocol.util.Utils;
-import io.swagger.v3.oas.annotations.media.Schema;
 import org.jspecify.annotations.Nullable;
 import tools.jackson.databind.node.ObjectNode;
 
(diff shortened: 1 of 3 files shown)


mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/utils/McpJsonSchemaGenerator.java
```diff
@@ -53,16 +49,17 @@
...
-import org.springframework.core.Nullness;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.ConcurrentReferenceHashMap;
+import org.springframework.util.StringUtils;
 
-public final class McpJsonSchemaGenerator {
+public final class McpJsonSchemaGenerator extends AbstractJsonSchemaGenerator {
 
-	private static final boolean PROPERTY_REQUIRED_BY_DEFAULT = true;
+	private static final McpJsonSchemaGenerator INSTANCE = new McpJsonSchemaGenerator();
```
sdeleuze, line 62: I don't think we should use this kind of `INSTANCE` static field, looks like an anti-pattern to me and I don't understand why it is needed, could we please remove it?

spring-ai-model/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java
```diff
@@ -75,15 +70,9 @@
...
+public final class JsonSchemaGenerator extends AbstractJsonSchemaGenerator {
 
-	/**
-	 * To ensure consistency and robustness across different model providers, all
-	 * properties in the JSON Schema are considered required by default. This behavior can
-	 * be overridden by setting the {@link ToolParam#required()},
-	 * {@link JsonProperty#required()}, or {@link Schema#requiredMode()}} annotation.
-	 */
-	private static final boolean PROPERTY_REQUIRED_BY_DEFAULT = true;
+	private static final JsonSchemaGenerator INSTANCE = new JsonSchemaGenerator();
```
sdeleuze, line 75: I don't think we should use this kind of `INSTANCE` static field, looks like an anti-pattern to me and I don't understand why it is needed, could we please remove it?

sdeleuze (changes requested):

spring-ai-model/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java
tzolov: I've tried to minimize the breaking changes. Will have another pass. Moving this to RC1

tzolov:
We will give this another try after 2.0GA

PR #6166, opened 2026-05-26
Add EntityParamSpec to ChatClient entity() methods

### LINKED ISSUE #6165: Add EntityParamSpec to ChatClient entity() methods for per-call structured output configuration
Currently, enabling provider-native structured output on an entity() call requires adding an advisor parameter separately, and enabling schema validation requires manually wiring a StructuredOutputValidationAdvisor. There is no way to configure either option directly at the entity() call site.

#### Proposed Change

Introduce `EntityParamSpec`, a fluent configuration interface accepted by new overloads of `entity()` and `responseEntity()` on `CallResponseSpec`:

```java
// provider-native schema delivery
ActorFilms films = chatClient.prompt()
    .user("Generate a filmography.")
    .call()
    .entity(ActorFilms.class, e -> e.delegateToProvider());

// schema validation with automatic retry
ActorFilms films = chatClient.prompt()
    .user("Generate a filmography.")
    .call()
    .entity(ActorFilms.class, e -> e.schemaValidation());

// both combined
ActorFilms films = chatClient.prompt()
    .user("Generate a filmography.")
    .call()
    .entity(ActorFilms.class, e -> e
        .delegateToProvider()
        .schemaValidation());
```

`EntityParamSpec` options:

- delegateToProvider() — passes the JSON schema to the AI provider as an API-level constraint via StructuredOutputChatOptions, instead of appending it as prompt text. Equivalent to the existing AdvisorParams.ENABLE_NATIVE_STRUCTURED_OUTPUT advisor parameter, but scoped to a single call.
- schemaValidation() — validates the model's JSON response against the entity schema using StructuredOutputValidationAdvisor, and retries with the validation error appended to the prompt on failure (default: 3 attempts). Streaming is not supported.

tzolov:
Introduce EntityParamSpec, a fluent configuration interface for the entity() and responseEntity() overloads, enabling two opt-in behaviours per call:

- delegateToProvider(): passes the JSON schema to the AI provider as an API-level constraint instead of appending it as prompt text (equivalent to AdvisorParams.ENABLE_NATIVE_STRUCTURED_OUTPUT).
- schemaValidation(): validates the model's JSON response against the entity schema via StructuredOutputValidationAdvisor, retrying with error feedback on failure (default: 3 attempts).

StructuredOutputValidationAdvisor.Builder gains an outputJsonSchema(String) alternative to outputType(), and BaseAdvisorChain gains mutate() and a nested Builder interface to support per-call advisor chain augmentation.

Resolves #6165

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
@@ -182,14 +182,14 @@ public interface ChatClient {
 		 * {@link org.springframework.ai.chat.model.ChatModel} does not support
 		 * {@link org.springframework.ai.model.tool.StructuredOutputChatOptions}.
 		 */
-		EntityParamSpec useProviderStructuredOutput();
+		EntityParamSpec delegateToProvider();
 
 		/**
 		 * Validates the model's JSON response against the entity schema and retries with
 		 * the error feedback on failure, up to {@code maxRepeatAttempts} times (default:
 		 * 3). Streaming is not supported.
 		 */
-		EntityParamSpec withSchemaValidation();
+		EntityParamSpec schemaValidation();
 
 	}
 
@@ -200,8 +200,8 @@ public interface ChatClient {
 		 * via the {@code entityParamSpecConsumer}.
 		 * @param type the target parameterized type
 		 * @param entityParamSpecConsumer configures options such as
-		 * {@link EntityParamSpec#useProviderStructuredOutput()} and
-		 * {@link EntityParamSpec#withSchemaValidation()}
+		 * {@link EntityParamSpec#delegateToProvider()} and
+		 * {@link EntityParamSpec#schemaValidation()}
 		 * @return the deserialized entity, or {@code null} if the response is empty
 		 */
 		<T> @Nullable T entity(ParameterizedTypeReference<T> type, Consumer<EntityParamSpec> entityParamSpecConsumer);
@@ -219,8 +219,8 @@ public interface ChatClient {
 		 * @param structuredOutputConverter the converter for parsing and schema
 		 * resolution
 		 * @param entityParamSpecConsumer configures options such as
-		 * {@link EntityParamSpec#useProviderStructuredOutput()} and
-		 * {@link EntityParamSpec#withSchemaValidation()}
+		 * {@link EntityParamSpec#delegateToProvider()} and
+		 * {@link EntityParamSpec#schemaValidation()}
 		 * @return the deserialized entity, or {@code null} if the response is empty
 		 */
 		<T> @Nullable T entity(StructuredOutputConverter<T> structuredOutputConverter,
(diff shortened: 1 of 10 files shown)


spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chatclient.adoc
```diff
@@ -273,23 +273,66 @@ List<ActorFilms> actorFilms = chatClient.prompt()
...
+This requires the underlying model to support `StructuredOutputChatOptions`.
 
 [source,java]
 ----
 ActorFilms actorFilms = chatClient.prompt()
-    .advisors(AdvisorParams.ENABLE_NATIVE_STRUCTURED_OUTPUT)
     .user("Generate the filmography for a random actor.")
     .call()
-    .entity(ActorFilms.class);
+    .entity(ActorFilms.class, spec -> spec.useProviderStructuredOutput());
```
ThomasVitale, line 291: I guess this should be

```suggestion
    .entity(ActorFilms.class, spec -> spec.delegateToProvider());
```

I'm a bit unsure about the naming. If I didn't have the additional context, reading something like "entity -> delegate to provider" would probably not be very clear to me what is happening. From that perspective, the `.useProviderStructuredOutput()` would be more expressive, or `.nativeStructuredOutput()`.

But I also acknowledge that I could write it like this and would probably get that additional context to read it correctly:

```java
    .entity(ActorFilms.class, structuredOutput -> structuredOutput.delegateToProvider());
```

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
```diff
@@ -169,12 +170,86 @@ interface AdvisorSpec {
...
+		 * {@link org.springframework.ai.model.tool.StructuredOutputChatOptions}.
+		 */
+		EntityParamSpec delegateToProvider();
+
+		/**
+		 * Validates the model's JSON response against the entity schema and retries with
+		 * the error feedback on failure, up to {@code maxRepeatAttempts} times (default:
+		 * 3). Streaming is not supported.
+		 */
+		EntityParamSpec schemaValidation();
```
ThomasVitale: About naming, it could be nice to align the format of the two methods. For example, if we go with "verb-object" like `delegateToProvider`, then I would also go with `.validateSchema()` instead of `.schemaValidation()`.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
```diff
@@ -401,6 +402,37 @@ public Map<String, Object> getParams() {
 
 	}
 
+	/**
+	 * Default implementation of {@link EntityParamSpec}.
+	 */
+	public static class DefaultEntityParamSpec<T> implements EntityParamSpec {
```
ilayaperumalg: Is generic type parameter needed here for DefaultEntityParamSpec?

spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chatclient.adoc
tzolov: renamed to useProviderStructuredOutput()

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
tzolov: ok renamed to validateSchema()

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
tzolov: You're right. This as leftover and it is cleaned with the latest commit

ilayaperumalg (approved):

ilayaperumalg:
Squashed, rebased and merged 567756715

PR #6180, opened 2026-05-27
Fix Microsoft Foundry URL path mode for proxy hosts

### LINKED ISSUE #6060: Azure support on URI construction is limited if the service is provided through proxy
Spring AI 2.0 M6 introduced the following change:

"Azure OpenAI functionality is now available through the standard spring-ai-openai module, which provides full Azure OpenAI deployment support."

However, this change does not appear to support a common enterprise deployment scenario where a proxy server sits between the application and Azure OpenAI.

Currently, Azure-specific behavior seems to be inferred from the target domain name. When requests are routed through an enterprise proxy whose hostname does not contain the expected Azure domain pattern, Spring AI does not apply Azure-specific request handling (such as deployment-based paths, query parameters, or headers). As a result, requests sent through the proxy fail, typically with HTTP 404 responses.

Expected Behavior

Azure OpenAI support should not rely solely on hostname/domain detection. It should be possible to explicitly configure Azure mode even when requests are routed through a non-Azure proxy endpoint.

Example Scenario

Application → Enterprise Proxy → Azure OpenAI

Example proxy endpoint:

https://llm-proxy.company.internal

Even though the backend target is Azure OpenAI, Spring AI treats the endpoint as a standard OpenAI-compatible API because the hostname is not an Azure domain.

Impact

This affects enterprise environments where:

outbound traffic must traverse API gateways or proxy servers
Azure OpenAI access is centralized behind internal infrastructure
applications are not allowed to directly access Azure endpoints
Suggested Improvement

Consider adding an explicit configuration option to force Azure OpenAI mode independently of endpoint hostname detection.

For example:

spring.ai.openai.azure.enabled=true

or a configurable API type/provider option similar to previous Azure-specific integrations.

jewoodev:
When `spring.ai.openai.microsoft-foundry=true` is used with an enterprise proxy `base-url`, OpenAI requests can miss the Azure deployment path and `api-version` query because the OpenAI SDK leaves `AzureUrlPathMode.AUTO` to classify the proxy host as non-Azure.

This change forwards Spring AI's resolved `MICROSOFT_FOUNDRY` provider decision to the SDK by setting `UNIFIED` for `/openai/v1` base URLs and `LEGACY` otherwise, across all credential paths. Tests cover proxy hosts, unified endpoints, explicit credential injection, non-Foundry baselines, and captured request paths with MockWebServer.

Closes #6060

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/setup/OpenAiSetup.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/setup/OpenAiSetup.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import java.util.stream.Collectors;
 
 import com.openai.azure.AzureOpenAIServiceVersion;
+import com.openai.azure.AzureUrlPathMode;
 import com.openai.azure.credential.AzureApiKeyCredential;
 import com.openai.client.OpenAIClient;
 import com.openai.client.OpenAIClientAsync;
@@ -49,6 +50,7 @@ import org.springframework.ai.openai.http.okhttp.SpringAiOpenAiHttpClient;
  *
  * @author Julien Dubois
  * @author Thomas Vitale
+ * @author Jewoo Shin
  */
 public final class OpenAiSetup {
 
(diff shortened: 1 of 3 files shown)


sdeleuze:
Please add an integration test that demonstrates the issue.

jewoodev:
Thanks for the review suggestion. I rebased on latest `main` and added MockWebServer coverage for #6060: a Foundry embeddings request with a non-Azure proxy `base-url` now verifies the Azure deployment path (`/openai/deployments/{deployment}/embeddings?api-version=...`) instead of plain `/embeddings`.

The `/openai/v1` unified endpoint remains covered separately so it is not forced into the legacy deployments path. Verified locally with the targeted OpenAI setup tests and `./mvnw -pl models/spring-ai-openai package`.

ilyakastsenevich:
Dear someone please merge this request

jewoodev:
Rebased this on current `main` and force-pushed the branch.

The targeted OpenAI setup tests pass locally, and I also re-ran the MCP integration-test command from the failing GitHub job locally with Docker available; it passes as well.

@sdeleuze, does this test shape address the integration-test concern you raised, or would you prefer a different coverage style?

sdeleuze:
@jewoodev Thanks, will check and let you know.

PR #6193, opened 2026-05-28
Add OTel `gen_ai.output.messages` to chat completion spans

jewoodev:
Chat completion content is currently available only through the opt-in SLF4J logging handler. This adds the OpenTelemetry span-side equivalent by exposing textual completions as gen_ai.output.messages.

The new handler tags the active span directly and is auto-configured only when tracing is present and spring.ai.chat.observations.log-completion=true is enabled. To avoid malformed or misaligned span attributes, this emits output messages only when every generation can be represented as text with a finish_reason; non-text output such as tool calls is skipped.

I reused the existing log-completion opt-in here. If separate controls for log output and span content are preferred, I can split this into a dedicated property. Tests cover the handler and auto-configuration opt-in matrix.

Part of #6051


jewoodev:
@jonatan-ivanov, could you take a look at this PR when you have a chance?

PR #6195, opened 2026-05-28
Simplify ToolCallAdvisor stream aggregation by removing publish() multicast

### LINKED ISSUE #6017: ToolCallAdvisor >= 1.1.3 streamWithToolCallResponses Bug
When ToolCallAdvisor is configured in ChatClient, the Flux stream continues to receive a cancel signal after receiving the completion signal.

The following code in the streamWithToolCallResponses method of the ToolCallAdvisor class:

Image

publish() converts a cold stream into a hot stream. Internally, it creates a ConnectableFlux. If doOnComplete and doOnCancel are used in the Flux stream during the model invocation or in the Flux processing of other custom advisors, the program will receive a doOnCancel signal after executing doOnComplete, thereby triggering the logic inside doOnCancel.

Image

logs:

2026-05-13T10:12:49.958+08:00 ERROR 31052 --- [nio-8080-exec-1] com.og.ai.config.model.SuperChatModel    : >>> 发生订阅: 0aeae39b72214746a137da75edc7f7a0 thread: http-nio-8080-exec-1
2026-05-13T10:12:52.085+08:00 ERROR 31052 --- [ient-2-Worker-2] com.og.ai.config.model.SuperChatModel    : 请求完成: 0aeae39b72214746a137da75edc7f7a0
2026-05-13T10:12:52.086+08:00 ERROR 31052 --- [ient-2-Worker-2] com.og.ai.config.model.SuperChatModel    : 请求结束: 0aeae39b72214746a137da75edc7f7a0 onComplete
2026-05-13T10:12:52.160+08:00 ERROR 31052 --- [oundedElastic-1] com.og.ai.config.model.SuperChatModel    : 请求被取消: 0aeae39b72214746a137da75edc7f7a0

tzolov:
Replace the publish()-based dual-branch pattern with a straightforward sequential aggregation: aggregate the response first, then defer the tool-call recursion. 
Also extract ChatClientMessageAggregator as a static field to avoid repeated instantiation per stream invocation.

Closes #6017

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ToolCallAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ToolCallAdvisor.java
@@ -55,6 +55,8 @@ import org.springframework.util.Assert;
  */
 public class ToolCallAdvisor implements CallAdvisor, StreamAdvisor, ToolAdvisor {
 
+	private static final ChatClientMessageAggregator CHAT_CLIENT_MESSAGE_AGGREGATOR = new ChatClientMessageAggregator();
+
 	/**
 	 * Default advisor order. Placed early in the chain so that all downstream advisors
 	 * (e.g. {@link org.springframework.ai.chat.client.advisor.api.BaseChatMemoryAdvisor})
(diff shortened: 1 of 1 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ToolCallAdvisor.java
```diff
@@ -255,40 +257,24 @@ private Flux<ChatClientResponse> internalStream(StreamAdvisorChain streamAdvisor
 
 			final ChatClientRequest finalRequest = processedRequest;
 
-			// Get the streaming response
+			// Propagate this round's observation as the parent for inner-chain advisors
+			// (e.g. ChatModelStreamAdvisor) so their spans are correctly nested.
```
chemicL: This comment doesn't seem justified or reflecting what actually happens? I mean it might be the case but it couples the comment with whatever happens in the `chainCopy.nextStream` if that is the internal behaviour. Let's bring back the previous comment maybe?

chemicL (approved):

PR #6198, opened 2026-05-28
Update ChatClient docs and ITs

### LINKED ISSUE #5786: [docs] ChatClient creation best practices
I am reading https://docs.spring.io/spring-ai/reference/2.0/api/chatclient.html#_creating_a_chatclient and have some opinions;

## The first section "Using an autoconfigured ChatClient.Builder"
This is spot on imo 👍 

## Then off to "Multiple ChatClients with a Single Model Type"
In this case I think one would still be able to use the auto-configured `ChatClient.Builder` and just clone and modify it, e.g.;
```java
        ChatClient chatClient = chatClientBuilder.build();

        ChatClient customChatClient = chatClientBuilder.clone()
                .defaultSystem("You are a helpful assistant.")
                .build();
```

(Btw, the current docs uses the `defaultSystemPrompt(String)` method on the builder, but it doesn't exist).

## Next up the "ChatClients for Different Model Types" section
Here I agree that we really need to bypass the auto-configured `ChatClient.Builder`. However, when we do so it has a couple of side effects;
1. Observability is completely disabled for the mode
2. `ChatClientCustomizer` beans are ignored

I think it would be a good idea if the docs clearly pointed out those side effects (especially the observability thing would be a no-no in enterprise deployments imo), and maybe also show how to avoid the side effects by creating the custom/ad-hoc builder similar to how [`ChatClientAutoConfiguration`](https://github.com/spring-projects/spring-ai/blob/main/auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client/src/main/java/org/springframework/ai/model/chat/client/autoconfigure/ChatClientAutoConfiguration.java#L88-L99) does it?

My point is that, imo, just doing `ChatClient.builder(myChatModel)` should be avoided as much as possible in order to ensure we have proper observability (and customizers applied), but it is not clear from the current docs.

ilayaperumalg:
- Improve docs for ChatClient - Clarify usage of multiple ChatClients and ChatClients for multiple model types
  - Add integration tests to verify the patterns

Resolves #5786

### DIFF
--- a/auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client/src/test/java/org/springframework/ai/model/chat/client/autoconfigure/ChatClientCreationPatternsTests.java
+++ b/auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client/src/test/java/org/springframework/ai/model/chat/client/autoconfigure/ChatClientCreationPatternsTests.java
@@ -1,0 +1,288 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.model.chat.client.autoconfigure;
+
+import io.micrometer.observation.ObservationRegistry;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.ai.chat.client.ChatClient;
+import org.springframework.ai.chat.client.ChatClientCustomizer;
+import org.springframework.ai.chat.client.advisor.ToolCallAdvisor;
+import org.springframework.ai.chat.client.advisor.observation.AdvisorObservationConvention;
+import org.springframework.ai.chat.client.observation.ChatClientObservationConvention;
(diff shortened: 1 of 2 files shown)


spring-ai-docs/src/main/antora/modules/ROOT/pages/api/chatclient.adoc
```diff
@@ -66,48 +66,90 @@ By default, Spring AI autoconfigures a single `ChatClient.Builder` bean.
...
-    .defaultSystemPrompt("You are a helpful assistant.")
-    .build();
+    @Bean
+    ChatClient defaultChatClient(ChatClient.Builder builder) {
+        return builder.build();
+    }
+
+    @Bean
+    ChatClient customChatClient(ChatClient.Builder builder) {
+        return builder.clone().defaultSystem("You are a helpful assistant.").build();
```
sdeleuze: As discussed, the `clone()` is not needed as the builder is injected with the prototype scope. To be removed in tests as well.

auto-configurations/models/chat/client/spring-ai-autoconfigure-model-chat-client/src/test/java/org/springframework/ai/model/chat/client/autoconfigure/ChatClientCreationPatternsTests.java
```diff
@@ -0,0 +1,288 @@
...
+
+	/**
+	 * A {@link ChatClientCustomizer} that sets a fixed system prompt, used to verify that
+	 * customizers are applied in both ChatClient creation patterns.
+	 */
+	@Configuration(proxyBeanMethods = false)
+	static class SystemPromptCustomizerConfig {
+
+		@Bean
+		ChatClientCustomizer systemPromptCustomizer() {
```
sdeleuze, line 282: In a different commit should be renamed to `ChatClientBuilderCustomizer` and this change should be mentioned in the upgrade notes.

sdeleuze (changes requested):

PR #6199, opened 2026-05-28
Fix streaming observability for tool calling: open observation scope across reactive thread boundaries

tzolov:
Micrometer observations in the streaming tool-call path were started with .start() but their ThreadLocal scope was never opened, streaming causing tool-call spans to appear as siblings in the trace instead of being correctly nested.

- DefaultAroundAdvisorChain.nextStream(): open scope in Flux.defer so the OTel ThreadLocal is set during the synchronous subscription chain
- DefaultChatClient.doGetObservableFluxChatResponse(): same fix so the ToolCallAdvisor observation is correctly nested under the chat-client span
- ToolCallAdvisor.handleToolCallRecursion(): open scope on the boundedElastic thread before blocking tool execution

Apply the same fix to the deprecated internal tool-execution paths in Anthropic, Bedrock, DeepSeek, GoogleGenAI, MiniMax, MistralAI, Ollama, and OpenAI chat models.

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -337,6 +337,8 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 			if (chatResponse.hasFinishReasons(java.util.Set.of("tool_use"))) {
 				return Flux.deferContextual(ctx -> {
 					ToolExecutionResult toolExecutionResult;
+					Observation parentObs = ctx.getOrDefault(ObservationThreadLocalAccessor.KEY, null);
+					Observation.Scope scope = parentObs != null ? parentObs.openScope() : null;
 					try {
 						if (this.internalToolExecutionWarned.compareAndSet(false, true)) {
 							logger.warn(
@@ -347,6 +349,9 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 						toolExecutionResult = this.toolCallingManager.executeToolCalls(prompt, chatResponse);
 					}
 					finally {
+						if (scope != null) {
+							scope.close();
+						}
 						org.springframework.ai.model.tool.internal.ToolCallReactiveContextHolder.clearContext();
 					}
 					if (toolExecutionResult.returnDirect()) {
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
@@ -719,10 +719,19 @@ public class DefaultChatClient implements ChatClient {
 
 				// @formatter:off
 				// Apply the advisor chain that terminates with the ChatModelStreamAdvisor.
-				Flux<ChatClientResponse> chatClientResponse = this.advisorChain.nextStream(chatClientRequest)
-						.doOnError(observation::error)
-						.doFinally(s -> observation.stop())
-						.contextWrite(ctx -> ctx.put(ObservationThreadLocalAccessor.KEY, observation));
+				Flux<ChatClientResponse> chatClientResponse = Flux.defer(() -> {
+					// Open observation scope at subscription time, ensuring
+					// advisor observations created during the synchronous subscription chain
+					// find this observation in ThreadLocal and attach to it as parent.
+					Observation.Scope scope = observation.openScope();
+					return this.advisorChain.nextStream(chatClientRequest)
+							.doOnError(observation::error)
+							.doFinally(s -> {
+								scope.close();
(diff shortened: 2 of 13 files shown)


models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -337,6 +337,8 @@ private Flux<ChatResponse> handleStreamingToolExecution(Prompt prompt, ChatRespo
 			if (chatResponse.hasFinishReasons(java.util.Set.of("tool_use"))) {
 				return Flux.deferContextual(ctx -> {
 					ToolExecutionResult toolExecutionResult;
+					Observation parentObs = ctx.getOrDefault(ObservationThreadLocalAccessor.KEY, null);
+					Observation.Scope scope = parentObs != null ? parentObs.openScope() : null;
```
chemicL, line 341: In case of automatic context propagation it might be that the scope is already open, so perhaps it is worth checking first before opening. Otherwise, the observation might become its own parent if I there is no protection against re-opening the same scope. Worth running a test with automatic context propagation enabled.

tzolov: Thanks for the hint. I've added a check for open observation and added tests for it.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
```diff
@@ -601,10 +601,19 @@ private Flux<ChatClientResponse> doGetObservableFluxChatResponse(ChatClientReque
...
-						.contextWrite(ctx -> ctx.put(ObservationThreadLocalAccessor.KEY, observation));
+				Flux<ChatClientResponse> chatClientResponse = Flux.defer(() -> {
+					// Open observation scope at subscription time, ensuring
+					// advisor observations created during the synchronous subscription chain
+					// find this observation in ThreadLocal and attach to it as parent.
+					Observation.Scope scope = observation.openScope();
+					return this.advisorChain.nextStream(chatClientRequest)
+							.doOnError(observation::error)
+							.doFinally(s -> {
+								scope.close();
```
chemicL: The `scope.close()` most probably will happen on a different thread than the `openScope()` leading the original scope open on a different Thread.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
```diff
@@ -601,10 +601,19 @@ private Flux<ChatClientResponse> doGetObservableFluxChatResponse(ChatClientReque
...
 				// Apply the advisor chain that terminates with the ChatModelStreamAdvisor.
-				Flux<ChatClientResponse> chatClientResponse = this.advisorChain.nextStream(chatClientRequest)
-						.doOnError(observation::error)
-						.doFinally(s -> observation.stop())
-						.contextWrite(ctx -> ctx.put(ObservationThreadLocalAccessor.KEY, observation));
+				Flux<ChatClientResponse> chatClientResponse = Flux.defer(() -> {
+					// Open observation scope at subscription time, ensuring
+					// advisor observations created during the synchronous subscription chain
+					// find this observation in ThreadLocal and attach to it as parent.
+					Observation.Scope scope = observation.openScope();
```
chemicL: This code should be removed. The parent-child relationship is established above when the observation is created. The code here is not synchronous in nature and introduces a reactive dispatch which can change threads. Since the observation lands in the reactive context you should be fine to take it from the context when you call the synchronous tools.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/DefaultAroundAdvisorChain.java
```diff
@@ -141,10 +142,18 @@ public Flux<ChatClientResponse> nextStream(ChatClientRequest chatClientRequest)
...
+				return advisor.adviseStream(chatClientRequest, this)
 						.doOnError(observation::error)
-						.doFinally(s -> observation.stop())
-						.contextWrite(ctx -> ctx.put(ObservationThreadLocalAccessor.KEY, observation)));
+						.doFinally(s -> {
+							scope.close();
+							observation.stop();
+						})
+						.contextWrite(ctx -> ctx.put(ObservationThreadLocalAccessor.KEY, observation));
+			});
```
chemicL: The code can be reverted, this code block introduces a reactive chain that can hop threads between subscription and doFinally. The parent-child relationship is already established and the only thing that is missing is the ToolCallAdvisor population of thread local scope when dispatching to synchronous tools.

chemicL (changes requested):

tzolov:
As pointed out above this approach is incorrect. 
Replacing the approach with this https://github.com/spring-projects/spring-ai/pull/6244

PR #6203, opened 2026-05-29
Eliminate calls to deprecated `HttpHeaders.asMultiValueMap()`

quaff:
And refactor `HeaderUtils::collectHeaders` to use `headerSet` instead.


Kehrlann (approved):

PR #6218, opened 2026-05-30
Refactor options for strict immutability and centralized default values

sdeleuze:
### Summary
This PR refactors the various `Options` classes (such as `ChatOptions`, `EmbeddingOptions`, etc.) across the framework to enforce strict immutability, improve null-safety representation, and centralize where default values are defined. 
Previously, default values were scattered across `Model` implementations and Auto-configuration `*Properties` classes, and options collections were sometimes mutable or defaulted to empty collections (which caused issues when merging). This PR streamlines the options architecture to be safer and more predictable.

### Key Changes
* **Strict Immutability:** Enforced strict immutability for all collections within options (e.g., `toolCallbacks`, `stopSequences`, `customHeaders`) by using unmodifiable copies. 
* **Nullable Collections:** Replaced empty collections with nullable collections (`@Nullable`) to accurately represent the absence of a value, which fixes and improves the behavior of options merging.
* **Centralized Defaults:** Moved default values (like default model names, temperatures, etc.) from the model implementations and `*Properties` configuration classes directly into the options constructors.
* **Removed Redundant Properties:** Removed default configuration properties from `*Properties` classes since they are now consistently handled at the options level.
* **API Updates:** 
  * Renamed `ChatModel#getDefaultOptions()` to `ChatModel#getOptions()` (the former is now deprecated).
  * Removed `ChatOptions#copy()` since options are now strictly immutable (users should use `mutate().build()` instead).
* **Cleanup:** Improved the consistency of ternary operators in options, cleaned up `@NullMarked` annotations, and fixed integration tests to align with the new options behavior.
* **Documentation:** Updated `upgrade-notes.adoc` for `2.0.0-RC1` to detail the breaking changes and migration paths for these options updates.

### Breaking Changes & Migration
* `ChatOptions#copy()` and `*Options#fromOptions` have been removed. Use `mutate().build()` to create a modified copy of an options instance.
* Modifying collections returned by options getters will now throw an `UnsupportedOperationException`.
* `ChatModel#getDefaultOptions()` is deprecated in favor of `ChatModel#getOptions()`.
* Default values are no longer duplicated in `*Properties` classes, which may affect code that relied on reading those default properties directly.

@tzolov @ilayaperumalg I will take care of merging it, please just provide your review.
@ThomasVitale @nicolaskrier Feel free to provide your feedback as well.
I recommend reviewing commit by commit, each has a meaningful description.


models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
```diff
@@ -58,120 +56,116 @@ public class MistralAiChatOptions implements ToolCallingChatOptions, StructuredO
 	 * ID of the model to use
 	 */
 	@SuppressWarnings("NullAway.Init")
```
nicolaskrier: Is this `@SuppressWarnings` annotation still required?

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
nicolaskrier: I think this TODO could be handled now.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
```diff
@@ -58,120 +56,116 @@ public class MistralAiChatOptions implements ToolCallingChatOptions, StructuredO
...
-	private String model;
+	private final String model;
 
 	/**
 	 * What sampling temperature to use, between 0.0 and 1.0. Higher values like 0.8 will
 	 * make the output more random, while lower values like 0.2 will make it more focused
 	 * and deterministic. We generally recommend altering this or top_p but not both.
 	 */
-	private @Nullable Double temperature;
+	private final @Nullable Double temperature;
```
nicolaskrier: Why not removing `@Nullable` annotation on defaulted attributes?

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
nicolaskrier: Is this method still required?

auto-configurations/models/spring-ai-autoconfigure-model-bedrock-ai/src/main/java/org/springframework/ai/model/bedrock/cohere/autoconfigure/BedrockCohereEmbeddingAutoConfiguration.java
```diff
@@ -66,7 +67,7 @@ public CohereEmbeddingBedrockApi cohereEmbeddingApi(AwsCredentialsProvider crede
 	public BedrockCohereEmbeddingModel cohereEmbeddingModel(CohereEmbeddingBedrockApi cohereEmbeddingApi,
 			BedrockCohereEmbeddingProperties properties) {
 
-		return new BedrockCohereEmbeddingModel(cohereEmbeddingApi, properties.getOptions());
+		return new BedrockCohereEmbeddingModel(cohereEmbeddingApi, properties.getOptions().toOptions());
```
nicolaskrier: I think it is a bit disturbing calling `getOptions` and then `toOptions`.

nicolaskrier:
@sdeleuze: I like this refactor! I have added a few comments that could be applied to other AI vendors as well.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
ericbottard: Indeed with options being immutable now, there is no point obtaining a new instance from an original. (same rationale as `copy()`)

auto-configurations/models/spring-ai-autoconfigure-model-anthropic/src/main/java/org/springframework/ai/model/anthropic/autoconfigure/AnthropicCacheProperties.java
```diff
@@ -0,0 +1,87 @@
...
+		return this.multiBlockSystemCaching;
+	}
+
+	public void setMultiBlockSystemCaching(@Nullable Boolean multiBlockSystemCaching) {
+		this.multiBlockSystemCaching = multiBlockSystemCaching;
+	}
+
+	public AnthropicCacheOptions toOptions() {
+		AnthropicCacheOptions.Builder builder = AnthropicCacheOptions.builder();
+		if (this.strategy != null) {
```
ericbottard: In general, since the behavior of builders should be to do nothing when receiving a `null` parameter, all those `if` guards can be removed, making the code way more readable (using chains)

spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
```diff
@@ -4,6 +4,55 @@
...
+
+To create a modified copy of an options instance, use the `mutate()` method instead of `copy()`:
+
+[source,java]
+----
+// Before
+ChatOptions options = originalOptions.copy();
+
+// After
+ChatOptions options = originalOptions.mutate().build();
```
ericbottard: maybe this example should add some builder changing calls between mutate() and build(), otherwise it's not clear why a user would create an (new) instance of something that is immutable

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
sdeleuze: I will remove these kind of comments as at builder level it can be null.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
sdeleuze: Indeed, I will.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
sdeleuze: Good point, will remove those as well.

auto-configurations/models/spring-ai-autoconfigure-model-bedrock-ai/src/main/java/org/springframework/ai/model/bedrock/cohere/autoconfigure/BedrockCohereEmbeddingAutoConfiguration.java
sdeleuze: Indeed this class needs a refactoring to stop using nested `Options` class at configuration properties level.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
sdeleuze: No, I will remove it.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
ilayaperumalg: It looks like some of the other Options classes need similar removal.

models/spring-ai-mistral-ai/src/main/java/org/springframework/ai/mistralai/MistralAiChatOptions.java
ilayaperumalg: I assume this will be applied across the models where required.

sdeleuze:
I applied the suggested changes.

auto-configurations/models/spring-ai-autoconfigure-model-anthropic/src/main/java/org/springframework/ai/model/anthropic/autoconfigure/AnthropicCacheProperties.java
nicolaskrier: Indeed, I think this simplification could be applied to all `toOptions` methods. What do you think about it @sdeleuze?

PR #6220, opened 2026-05-30
test: implement streamWithToolUse usage aggregation test for BedrockConverse

### LINKED ISSUE #6108: Add missing test coverage for Bedrock Converse streaming tool-use usage aggregation
## Description

`BedrockConverseUsageAggregationTests` currently covers usage aggregation for regular Bedrock Converse calls and synchronous tool-use calls, but the `streamWithToolUse()` test method is left unimplemented.

This leaves the streaming tool-use path without direct test coverage for token usage aggregation across multiple Converse Stream rounds.

## Why it matters

When a streaming response triggers a tool call, the model invocation can involve multiple Bedrock Converse Stream requests:

1. The first stream returns a `tool_use` stop reason and usage metadata.
2. The tool is executed.
3. A second stream returns the final assistant response and its own usage metadata.

The final Spring AI `ChatResponse` metadata should expose aggregated usage values across both rounds, similar to the synchronous tool-use behavior.

## Suggested test coverage

Implement `BedrockConverseUsageAggregationTests#streamWithToolUse()` to verify that:

- a mocked first stream emits a tool-use event with usage metadata
- a mocked second stream emits the final assistant response with usage metadata
- the final streamed `ChatResponse` reports aggregated prompt, completion, and total token counts

suryateja-g13:
## Summary

Implements the previously empty `streamWithToolUse()` test in `BedrockConverseUsageAggregationTests` to verify that token usage (prompt, completion, total tokens) is correctly aggregated across two streaming rounds when tool calling is involved.

Closes #6108

## What changed

- Replaced the `// TODO: Implement the test` stub with a full mock-based test
- First stream round produces tool-use events (toolUseId, tool args, TOOL_USE stop reason, usage metadata)  
- The tool callback executes and triggers a second stream round
- Second stream round returns the final text response
- Asserts that the final `ChatResponse.getMetadata().getUsage()` correctly sums both rounds: `445+540=985` prompt tokens, `119+106=225` completion tokens, `564+646=1210` total tokens

## Key findings

Two non-obvious constraints were needed to make AWS SDK mock streaming work correctly in unit tests:

**1. Push-based publisher required**  
`ConverseStreamResponseHandler` uses `SequentialSubscriber` internally which does NOT proactively call `request(n)`. A pull-based `Flux.fromIterable` never emits — events must be pushed directly:

```java
handler.onEventStream(SdkPublisher.adapt((Publisher<ConverseStreamOutput>) sub -> {
    sub.onSubscribe(noopSubscription);
    for (ConverseStreamOutput event : events) sub.onNext(event);
    sub.onComplete();
}));
```

**2. Use `ConverseStreamOutput.*builder()` static factories, not `XxxEvent.builder()`**  
The public event builder POJOs (`ContentBlockStartEvent.builder()...build()`) have `accept(Visitor)` that throws `UnsupportedOperationException`. The `ConverseStreamOutput.contentBlockStartBuilder()` static factories create internal `Default*` implementations whose `accept(Visitor)` correctly dispatches to the visitor pattern.

## Test plan

- [x] `streamWithToolUse` passes (verified with `mvnw test -Dmaven.build.cache.enabled=false`)
- [x] All 58 unit tests in `spring-ai-bedrock-converse` pass

### DIFF
--- a/models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockConverseUsageAggregationTests.java
+++ b/models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockConverseUsageAggregationTests.java
@@ -16,25 +16,38 @@
 
 package org.springframework.ai.bedrock.converse;
 
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import software.amazon.awssdk.core.async.SdkPublisher;
 import software.amazon.awssdk.core.document.internal.MapDocument;
 import software.amazon.awssdk.services.bedrockruntime.BedrockRuntimeAsyncClient;
 import software.amazon.awssdk.services.bedrockruntime.BedrockRuntimeClient;
 import software.amazon.awssdk.services.bedrockruntime.model.ContentBlock;
+import software.amazon.awssdk.services.bedrockruntime.model.ContentBlockDelta;
+import software.amazon.awssdk.services.bedrockruntime.model.ContentBlockStart;
 import software.amazon.awssdk.services.bedrockruntime.model.ConversationRole;
 import software.amazon.awssdk.services.bedrockruntime.model.ConverseMetrics;
 import software.amazon.awssdk.services.bedrockruntime.model.ConverseOutput;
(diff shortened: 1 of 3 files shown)


sobychacko:
Thanks for this @suryateja-g13. This overlaps with #6109, which also closes #6108 by implementing `streamWithToolUse()` in `BedrockConverseUsageAggregationTests`. That one was opened a bit earlier and is scoped just to the test, so we are going to go with it.

One thing worth noting: this PR also carries an unrelated production change to `OpenAiChatModel.java` (the tool-response mapping that skips responses with a blank `id`) plus its OpenAI tests. That looks independent of #6108, so if it addresses a real issue it would be great as its own PR where it can be reviewed on its own merits.

Really appreciate the effort here. Please do keep the contributions coming.

PR #6221, opened 2026-05-30
fix: use list position instead of API index when looking up chunk choice in OpenAI streaming

### LINKED ISSUE #6023: IndexOutOfBoundsException in OpenAiChatModel.internalStream when choice.index() != list position

suryateja-g13:
## Summary

Fixes an `IndexOutOfBoundsException` in `OpenAiChatModel.internalStream` that occurs when a streaming chunk contains a choice whose `index` API field value does not match its position in the `choices` list.

Closes #6023

## Root cause

In the streaming path, each `Generation`'s metadata included a `chunkChoice` entry populated by:

```java
"chunkChoice", chunk.choices().get((int) choice.index())
```

`choice.index()` is the **semantic** index from the OpenAI API response (e.g., `0` or `1`), not the **positional** index within the `choices` array. When a gateway sends a single chunk choice with `index: 1`, `chunk.choices()` has length 1 (position 0 only), so `get(1)` throws `IndexOutOfBoundsException`.

This breaks with:
- OpenAI-compatible gateways that use non-zero-based choice indices
- Usage-only final chunks with fewer choices than expected
- Responses where choices arrive in non-sequential order

## Fix

Switch from `Stream.map` to `IntStream.range` so the **list position** `i` is used for the chunk choice lookup, while `choice.index()` (the API field) continues to be stored in metadata:

```java
// Before (broken):
List<Generation> generations = chatCompletion.choices().stream().map(choice -> {
    ...
    "chunkChoice", chunk.choices().get((int) choice.index())  // uses API index as list pos
    ...
}).toList();

// After (fixed):
List<ChatCompletion.Choice> choiceList = chatCompletion.choices();
List<Generation> generations = IntStream.range(0, choiceList.size()).mapToObj(i -> {
    ChatCompletion.Choice choice = choiceList.get(i);
    ...
    "chunkChoice", chunk.choices().get(i)  // uses list position
    ...
}).toList();
```

This is safe because `chunkToChatCompletion()` converts `chunk.choices()` in order, so position `i` in `chatCompletion.choices()` always corresponds to position `i` in `chunk.choices()`.

## Test

Added `streamingDoesNotThrowWhenChoiceApiIndexExceedsListSize` in `OpenAiChatModelTests` that mocks the async streaming path with a chunk containing a single choice having `index=1` (API index != list position 0) and asserts the response completes successfully.

## Test plan

- [x] New test `streamingDoesNotThrowWhenChoiceApiIndexExceedsListSize` passes
- [x] All 8 tests in `OpenAiChatModelTests` pass

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -26,6 +26,7 @@ import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import com.openai.client.OpenAIClient;
 import com.openai.client.OpenAIClientAsync;
@@ -318,7 +319,9 @@ public final class OpenAiChatModel implements ChatModel {
 					try {
 						ChatCompletion chatCompletion = chunkToChatCompletion(chunk);
 						String id = chatCompletion.id();
-						List<Generation> generations = chatCompletion.choices().stream().map(choice -> {
+						List<ChatCompletion.Choice> choiceList = chatCompletion.choices();
+						List<Generation> generations = IntStream.range(0, choiceList.size()).mapToObj(i -> {
+							ChatCompletion.Choice choice = choiceList.get(i);
 							roleMap.putIfAbsent(id, choice.message()._role().asString().isPresent()
 									? choice.message()._role().asStringOrThrow() : "");
 
@@ -327,7 +330,7 @@ public final class OpenAiChatModel implements ChatModel {
 									choice.message().refusal().isPresent() ? choice.message().refusal() : "",
 									"annotations", choice.message().annotations().isPresent()
 											? choice.message().annotations() : List.of(),
-									"chunkChoice", chunk.choices().get((int) choice.index()));
+									"chunkChoice", chunk.choices().get(i));
 
 							return buildGeneration(choice, metadata, request);
 						}).toList();
--- a/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java
+++ b/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java
@@ -17,18 +17,24 @@
 package org.springframework.ai.openai;
 
 import java.lang.reflect.Method;
+import java.time.Duration;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 
 import com.openai.client.OpenAIClient;
 import com.openai.client.OpenAIClientAsync;
 import com.openai.core.JsonValue;
+import com.openai.core.http.AsyncStreamResponse;
 import com.openai.models.chat.completions.ChatCompletion;
+import com.openai.models.chat.completions.ChatCompletionChunk;
 import com.openai.models.chat.completions.ChatCompletionCreateParams;
 import com.openai.models.chat.completions.ChatCompletionMessage;
 import com.openai.models.completions.CompletionUsage;
+import com.openai.services.async.ChatServiceAsync;
+import com.openai.services.async.chat.ChatCompletionServiceAsync;
 import com.openai.services.blocking.ChatService;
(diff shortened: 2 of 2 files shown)


ericbottard:
Hi @suryateja-g13,

the OpenAiModel streaming code has been reworked prior to the 2.0.0 release. I don't think this issue still exists. Could you please confirm that you cannot reproduce with 2.0.0? The `chunk.index()` is no longer used where it shouldn't be, unless I missed something.

ericbottard:
As an alternative, if you do find an issue still, could you provide a PR against current `main`?

FYI, I'm also looking at https://github.com/spring-projects/spring-ai/issues/6441 but I believe this is a different issue

PR #6228, opened 2026-05-31
GH-6211: Avoid duplicate chat memory in tool prompts

### LINKED ISSUE #6211: `@Tool` tool is called endlessly with chat memory enabled
https://github.com/joshlong-attic/2026-05-29-infinite-tool-call-in-m8/blob/main/README.md

that is a reproducer

there's an oddity with infinite tool calling in Spring AI M8 + memory

you need ollama to make this run.

im using some big models but you might be able to get away with less.

them visit:

http://127.0.0.1:8080/josh/ask?question=fantastic.%20when%20can%20I%20pick%20up%20Prancer%20(dog%20ID%2045)%20from%20the%20Lisbon%20Pooch%20Palace%20location%3F&continue

it calls the tool over and over. if u change the DB to use PostgreSQL or somethign youc an see that its adding many blank rows in the chat memory table, too

goutamadwant:
Fixes #6211

## Summary

- Avoid prepending chat memory when the current prompt already contains the stored memory messages.
- Keep tool-call continuation prompts from repeating prior user, assistant, and tool messages while still storing the latest tool response in memory.
- Cover both full-history and windowed-memory cases.

## Tests

- `./mvnw -pl spring-ai-client-chat -Dtest=MessageChatMemoryAdvisorTests -Dmaven.build.cache.enabled=false test`
- `./mvnw -pl spring-ai-client-chat -Dmaven.build.cache.enabled=false test`

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/MessageChatMemoryAdvisor.java
@@ -78,8 +78,12 @@ public final class MessageChatMemoryAdvisor implements BaseChatMemoryAdvisor {
 		List<Message> memoryMessages = this.chatMemory.get(conversationId);
 
 		// 2. Advise the request messages list.
-		List<Message> processedMessages = new ArrayList<>(memoryMessages);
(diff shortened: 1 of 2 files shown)


tzolov:
Thanks you @goutamadwant

PR #5876, opened 2026-04-23
GH-1608: Add observability support to Stability AI image model

Gaurav1112:
# GH-1608: Add observability support to Stability AI image model

Fixes https://github.com/spring-projects/spring-ai/issues/1608

## Problem

`StabilityAiImageModel` was the last remaining `ImageModel` in the project without observability. Calls to `imageModel.call(...)` for Stability AI produced no `gen_ai.client.operation` observation, so they were invisible in the Actuator metrics endpoint and in any attached tracer — while equivalent calls against `OpenAiImageModel` have been fully instrumented for some time.

## What this PR does

Instruments `StabilityAiImageModel` with Micrometer observations using the portable `ImageModelObservationConvention` infrastructure, mirroring `OpenAiImageModel` so the resulting tags are consistent across providers.

### Production code

- **`AiProvider`** — added `STABILITY_AI("stability_ai")` enum value (placed alphabetically, inside the existing `@formatter:off` region).
- **`StabilityAiImageModel`**:
  - New 3-arg constructor `StabilityAiImageModel(StabilityAiApi, StabilityAiImageOptions, ObservationRegistry)`. A `null` registry is defensively swapped for `ObservationRegistry.NOOP`.
  - Existing 1-arg and 2-arg constructors delegate to the new one with `ObservationRegistry.NOOP`, preserving source and binary compatibility.
  - `call(ImagePrompt)` is wrapped with `ImageModelObservationDocumentation.IMAGE_MODEL_OPERATION.observation(...).observe(...)` using the shared `DefaultImageModelObservationConvention`. The **raw** `ImagePrompt` is passed to the observation context (matching `OpenAiImageModel`) so observation tags reflect what the user actually sent rather than post-merge defaults.
  - New `setObservationConvention(ImageModelObservationConvention)` accessor; guards against `null`.
- **`StabilityAiImageAutoConfiguration`** — accepts `ObjectProvider<ObservationRegistry>` and `ObjectProvider<ImageModelObservationConvention>` and wires them into the model bean, identical to `OpenAiImageAutoConfiguration`.

### Tests

All mocked — no network access and no API keys required.

- **`StabilityAiImageModelObservationTests`** (7 cases):
  1. Happy path: verifies `gen_ai.client.operation` name, contextual name, provider = `stability_ai`, model tag, size/style high-cardinality tags.
  2. `ImagePrompt` with no options — observation still emitted with provider + operation type.
  3. Error path — when the API throws, the observation is still started and stopped (no instrumentation lost).
[...]

PR #6229, opened 2026-05-31
Add observability metrics to StabilityAiImageModel

### LINKED ISSUE #1608: Stability AI does not produce observability metrics

oniwon:
## Summary
Wires `StabilityAiImageModel` into the existing image-model observation
infrastructure so it emits `gen_ai.client.operation` metrics and traces,
consistent with `OpenAiImageModel`. Resolves #1608, where Stability AI
produced no observability.

## Changes
- Add `STABILITY_AI("stability_ai")` to `AiProvider`.
- Wrap `StabilityAiImageModel.call(...)` in an `ImageModelObservationContext`
  via `ImageModelObservationDocumentation.IMAGE_MODEL_OPERATION`. Adds an
  `ObservationRegistry`-aware constructor and `setObservationConvention(...)`;
- `StabilityAiImageAutoConfiguration` injects `ObjectProvider<ObservationRegistry>`
  and `ObjectProvider<ImageModelObservationConvention>`, mirroring
  `OpenAiImageAutoConfiguration`.

## Test Plan
- New `StabilityAiImageModelObservationTests` (unit test, no API key) mocks
  `StabilityAiApi` and asserts via `TestObservationRegistry` that the observation
  is recorded with the expected low-cardinality keys (operation `image`,
  provider `stability_ai`, request model) and image size, and is started/stopped.
- `./mvnw -pl
models/spring-ai-stability-ai,auto-configurations/models/spring-ai-autoconfigure-model-stability-ai -am
verify` passes (0 checkstyle violations).

Fixes GH-1608

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-stability-ai/src/main/java/org/springframework/ai/model/stabilityai/autoconfigure/StabilityAiImageAutoConfiguration.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-stability-ai/src/main/java/org/springframework/ai/model/stabilityai/autoconfigure/StabilityAiImageAutoConfiguration.java
@@ -16,6 +16,9 @@
 
 package org.springframework.ai.model.stabilityai.autoconfigure;
 
+import io.micrometer.observation.ObservationRegistry;
+
+import org.springframework.ai.image.observation.ImageModelObservationConvention;
 import org.springframework.ai.model.SpringAIModelProperties;
 import org.springframework.ai.model.SpringAIModels;
 import org.springframework.ai.stabilityai.StabilityAiImageModel;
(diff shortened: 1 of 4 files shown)


quaff:
It's part of #6091.

donasman:
Hi @oniwon, thanks for picking this up — I ran into the same gap and
implemented an equivalent change locally before finding your PR. A few
observations from doing so:

**1. Needs a rebase.** This targets a base where the field was named
`defaultOptions`; on current `main` it is `options`. The hunks in
`StabilityAiImageModel` won't apply cleanly as-is.

**2. Null handling in the new constructor is inconsistent.** The two
lines above use `Assert.notNull`, but the registry uses
`Objects.requireNonNullElse(...)`. Since the parameter isn't annotated
`@Nullable`, NullAway already treats null as a contract violation, so
the fallback is unreachable under the project's own rules. Either
`Assert.notNull(observationRegistry, "ObservationRegistry must not be
null")` for consistency, or mark the parameter `@Nullable` and keep the
fallback.

**3. Tags reflect runtime options only, not the merged ones.** The
context receives the raw `imagePrompt`, so when a caller relies on the
constructor defaults, `gen_ai.request.model` reports `none` and
`gen_ai.request.image.size` is omitted — even though
`requestImageOptions` is what actually goes to the API. Worth noting
that `OpenAiImageModel` and `GoogleGenAiImageModel` behave identically,
so this is pre-existing and consistent. I'd suggest keeping it as-is
here and raising a separate issue rather than widening this PR.

One optional addition: a second test asserting that the two-arg
constructor records nothing would lock in the backward-compatibility
guarantee of the `ObservationRegistry.NOOP` default.

sobychacko:
Thanks very much for this, @oniwon. This overlaps almost entirely with #5876, which targets the same issue (#1608) and applies effectively the same fix. #5876 was opened a bit earlier and also includes the reference documentation update, an autoconfiguration observability test, and broader unit coverage, plus the `micrometer-observation-test` test dependency that the new tests rely on. Given the overlap we are going to consolidate on #5876 and close this one. We really appreciate you taking the time on this, and we would be glad to see more contributions from you.

PR #6235, opened 2026-05-31
fix(bedrock): omit null temperature and topP from InferenceConfiguration

### LINKED ISSUE #6163: BedrockProxyChatModel includes temperature in inferenceConfig even when null, causing ValidationException with Claude Opus 4.7
## Bug description

When using `BedrockProxyChatModel` (`spring-ai-bedrock-converse`) with Claude Opus 4.7 (e.g. `eu.anthropic.claude-opus-4-7`), every request fails with:

```
400 ValidationException: The model returned the following errors:
`temperature` is deprecated for this model.
```

This occurs even when `temperature` is explicitly set to `null` in the `ChatOptions` passed to `Prompt`, and even when no temperature is set in `application.yml`.

Claude Opus 4.7 rejects the `temperature` field entirely — it must be **absent** from `inferenceConfig`, not merely null.

---

## Environment

- Spring AI: `1.1.7`
- Module: `spring-ai-bedrock-converse`
- Model: `eu.anthropic.claude-opus-4-7` (Bedrock Converse API)
- Java 17 / Kotlin

---

## Steps to reproduce

```kotlin
val options = BedrockChatOptions.builder()
    .model("eu.anthropic.claude-opus-4-7")
    .maxTokens(400)
    // temperature intentionally not set
    .build()

val responseFlux = chatModel.stream(
    Prompt(messages, options)
)
```

---

## Expected behavior

When `temperature` is null (not explicitly set), it should be **omitted entirely** from the `inferenceConfig` block sent to Bedrock — not serialized as a null or default value.

---

## Actual behavior

`ModelOptionsUtils.merge()` copies `temperature` from the bean's default options into the merged options object. `BedrockProxyChatModel` then passes this value (even when null) to `InferenceConfiguration.builder()`, which includes it in the Bedrock Converse request payload. Bedrock forwards this to Anthropic, which rejects it with a 400.

---

## Root cause

The problem is that `null != absent` in the serialized request. `InferenceConfiguration.builder().temperature(null)` still results in the field being present in the payload. Spring AI has no mechanism to distinguish "this field was never set" from "this field was set to null", so there is no way for a caller to force omission of `temperature` without patching the model class itself. It cannot be set to null in yaml either; as it expects a 'java.lang.Double'. 

Possibly related: #3253 (same issue class for OpenAI search models).

---

## Suggested fix

In `BedrockProxyChatModel`,  avoid setting default values when any are set. 
or a switch to not override ChatOptions, perhaps?

Baqirrizvidev:
Fixes a bug where \BedrockProxyChatModel\ included \	emperature\ and \	opP\ in the \InferenceConfiguration\ even when they were null, which caused \ValidationException\ with models like Claude Opus 4.7 that reject the \	emperature\ field entirely if it is not supported.

By only setting these properties when they are non-null, the AWS Bedrock SDK correctly omits them from the serialized payload as expected.

Resolves #6163

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-minimax/src/main/java/org/springframework/ai/model/minimax/autoconfigure/MiniMaxChatAutoConfiguration.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-minimax/src/main/java/org/springframework/ai/model/minimax/autoconfigure/MiniMaxChatAutoConfiguration.java
@@ -40,6 +40,7 @@ import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 import org.springframework.web.client.ResponseErrorHandler;
 import org.springframework.web.client.RestClient;
+import org.springframework.web.reactive.function.client.WebClient;
 
 /**
  * {@link AutoConfiguration Auto-configuration} for MiniMax Chat Model.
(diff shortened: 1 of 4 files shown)


sdeleuze:
As requested multiple times, please add [a Signed-off-by trailer](https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring#how-to-use-developer-certificate-of-origin). Notice [we don't accept contribution from AI bots](https://github.com/spring-projects/spring-ai/blob/main/CONTRIBUTING.md#code-of-conduct).

sobychacko:
Thanks for the fix, @Baqirrizvidev. This overlaps with #6188, which targets the same issue (#6163) with the same core change (only setting `temperature` and `topP` on the `InferenceConfiguration.Builder` when they are non-null so Bedrock omits them from the payload). #6188 came first and is scoped to the `spring-ai-bedrock-converse` module with two unit tests covering both the null-omitted and explicit-value cases. One thing to note here: this PR also includes unrelated `MiniMaxApi` and MiniMax autoconfiguration changes (adding `WebClient.Builder`, deprecating constructors) that are outside the scope of #6163, so they would be better as a separate PR. We are going to go with #6188 for this one, so I will close this. Feel free to open a dedicated PR for the MiniMax `WebClient` improvement if you would like to pursue it.

PR #6241, opened 2026-06-01
GH-6219: Refresh local transformer resource cache

### LINKED ISSUE #6219: TransformersEmbeddingModel serves stale ONNX model from temp cache when running from fat jar
## Bug description

I am using `TransformersEmbeddingModel` to power a RAG pipeline backed by Redis VectorStore. I tried switching from one ONNX model to another with a different embedding dimension.

I exported the model using `optimum-cli` and saved `model.onnx` and `tokenizer.json` into `src/main/resources/models/`. The bean is wired in a `@Configuration` class:

```java
@Bean
public EmbeddingModel embeddingModel() throws Exception {
  TransformersEmbeddingModel embeddingModel = new TransformersEmbeddingModel();
  embeddingModel.setTokenizerResource("classpath:/models/tokenizer.json");
  embeddingModel.setModelResource("classpath:/models/model.onnx");
  embeddingModel.afterPropertiesSet();
  return embeddingModel;
}
```

Everything works fine the first time: the VectorStore index and Spring AI both use the same model. I am using Redis.

I then switched to a different model with a lower embedding dimension (768 → 384):

1. Re-indexed Redis with vectors from the new model (384-dim).
2. Exported the new ONNX model and overwrote `src/main/resources/models/model.onnx` and `tokenizer.json`.
3. Rebuilt the fat jar.
4. Restarted the application.

After the restart every query to the RAG produces:

```
Error invoking method: searchRAG
Error parsing vector similarity query: query vector blob size (3072) does not match index's expected size (1536).
```

`3072 bytes ÷ 4 = 768 floats` — Spring AI is still querying with the old 768-dim model despite the jar containing the new 384-dim one. Redis correctly expects 384-dim (1536 bytes).

This error **only occurs when running from the fat jar** (e.g. stdio MCP mode, `java -jar`). When running from an exploded classpath (e.g. `mvn spring-boot:run` or HTTP mode via IDE), the model file is read directly from `target/classes/models/` without any caching intermediate, so the new model is picked up immediately and it works fine. The difference is fat-jar vs exploded classpath, not the transport protocol itself.

**Root cause:**
When loading a `classpath:` resource from inside a fat jar, Spring AI extracts the ONNX model to a temp directory:

```
C:\Users\<user>\AppData\Local\Temp\spring-ai-onnx-generative\<deterministic-uuid>\
```

The UUID appears to be derived from the resource path (`classpath:/models/model.onnx`), so the same path always resolves to the same temp directory across restarts. When the jar is rebuilt with a new model, the old extracted files are still present in that temp directory and are reused — the new jar contents are never re-extracted.

**Workaround:** manually delete the temp directory. On the next startup Spring AI re-extracts the model from the jar and the correct dimensions are used.

---

## Environment

- Spring AI: 2.0.0-M4, 2.0.0-M8
- Java: Amazon Corretto 21 (JVM)
- OS: Windows 11
- Vector store: Redis (RedisJSON + HNSW index)
- Transport: Spring AI MCP Server (stdio and Streamable HTTP)

---

## Steps to reproduce

1. Export a first ONNX model (e.g. `msmarco-distilbert-base-v4`, 768-dim) to `src/main/resources/models/`.
2. Create a Redis index and ingest documents embedded with this model (768-dim vectors).
3. Wire `TransformersEmbeddingModel` as shown above and build a fat jar.
4. Run `java -jar app.jar` — queries work correctly.
5. Export a second ONNX model with a different dimension (e.g. `thenlper/gte-small`, 384-dim) to the **same** `src/main/resources/models/` directory, overwriting the previous files.
6. Re-index Redis with the new model (384-dim vectors).
7. Rebuild the fat jar (`mvn clean package`) — note that `clean` only wipes `target/`, it does not touch the temp directory.
8. Run `java -jar app.jar` again **without** deleting the temp directory from step 4.
9. Issue a vector similarity query.
10. Observe the dimension mismatch error — the old 768-dim model is loaded from the temp cache instead of the new 384-dim model from the jar.

---

## Expected behavior

When the fat jar is rebuilt with a new `model.onnx`, the next startup of the application should use the model bundled in the new jar. The embedding dimension reported at query time should match the model that was packaged into the jar, not a model that was extracted from a previous build.

---

## Minimal Complete Reproducible Example

```java
package org.example.springaiissues;

import org.springframework.ai.embedding.EmbeddingModel;
import org.springframework.ai.transformers.TransformersEmbeddingModel;
import org.springframework.boot.CommandLineRunner;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;

@SpringBootApplication
public class SpringaiIssuesApplication {

    SpringApplication.run(SpringaiIssuesApplication.class, args);
  }

  @Bean
  public EmbeddingModel embeddingModel() throws Exception {
    TransformersEmbeddingModel m = new TransformersEmbeddingModel();
    m.setTokenizerResource("classpath:/models/tokenizer.json");
    m.setModelResource("classpath:/models/model.onnx");
    m.afterPropertiesSet();
    return m;
  }

  @Bean
  CommandLineRunner run(EmbeddingModel model) {
    return args -> {
      float[] embedding = model.embed("test sentence");
      System.out.println("Embedding dimension: " + embedding.length);
      // First run with 768-dim model  → prints 768
      // Second run after swapping to 384-dim model + rebuild → still prints 768
      // Expected: 384
    };
  }

}

```

Place the ONNX model and tokenizer under `src/main/resources/models/`. Run once with a 768-dim model, then swap for a 384-dim model and rebuild the jar (`mvn clean package`). `mvn clean` only wipes `target/` — it does not affect the system temp directory. Run again without manually deleting `%TEMP%\spring-ai-onnx-generative\`. The printed dimension will still be 768.

A ready-to-use reproduction repository is available at: https://github.com/olegbilovus/springai-issues

jyt6640:
## Summary

Fixes GH-6219 (https://github.com/spring-projects/spring-ai/issues/6219) by refreshing cached local Transformer resources when the source resource changes.

`ResourceCacheService` previously reused an existing cached file solely because the resolved cache path already existed. For resources extracted from an executable jar, that meant a new application version could keep loading an old `model.onnx` or `tokenizer.json` from `${java.io.tmpdir}/spring-ai-onnx-generative` when the classpath resource path stayed the same.

This change stores lightweight metadata next to cached local resources and compares the current resource URI, last-modified timestamp, and content length before reusing the cached file. If the metadata differs, the cached resource is refreshed. Existing `http` and `https` cache behavior remains unchanged to avoid introducing extra remote checks.

## Validation

```bash
./mvnw -pl models/spring-ai-transformers -DskipITs -DskipNative -DskipTests=false -Dtest=ResourceCacheServiceTests test
```

### DIFF
--- a/models/spring-ai-transformers/src/main/java/org/springframework/ai/transformers/ResourceCacheService.java
+++ b/models/spring-ai-transformers/src/main/java/org/springframework/ai/transformers/ResourceCacheService.java
@@ -19,8 +19,11 @@ package org.springframework.ai.transformers;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Properties;
 import java.util.UUID;
 
 import org.apache.commons.logging.Log;
@@ -100,16 +103,18 @@ public class ResourceCacheService {
 	 */
 	public Resource getCachedResource(Resource originalResource) {
 		try {
-			if (this.excludedUriSchemas.contains(originalResource.getURI().getScheme())) {
+			URI originalResourceUri = originalResource.getURI();
+			if (this.excludedUriSchemas.contains(originalResourceUri.getScheme())) {
 				logger.info("The " + originalResource.toString() + " resource with URI schema ["
-						+ originalResource.getURI().getScheme() + "] is excluded from caching");
+						+ originalResourceUri.getScheme() + "] is excluded from caching");
 				return originalResource;
 			}
 
 			File cachedFile = getCachedFile(originalResource);
-			if (!cachedFile.exists()) {
+			if (shouldCacheResource(originalResource, originalResourceUri, cachedFile)) {
 				FileCopyUtils.copy(StreamUtils.copyToByteArray(originalResource.getInputStream()), cachedFile);
 				logger.info("Caching the " + originalResource.toString() + " resource to: " + cachedFile);
+				writeCacheMetadata(originalResource, originalResourceUri, cachedFile);
 			}
 			return new FileUrlResource(cachedFile.getAbsolutePath());
 		}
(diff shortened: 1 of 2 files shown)


ericbottard:
The default location for the cache service will soon be changed to be a new temp directory on each JVM startup, so I will close this PR as this solves the reported issue (across restarts).

Issue #6097, opened 2026-05-21
MCP server starters SHOULD use spring.jackson properties.


fbus:
**Bug description**
When using `spring.jackson.default-property-inclusion=non_null` and spring-ai-starter-mcp-server-webmvc
tool calls are returning null values.
For example `"end_date": null,`
I want to avoid this noise in the context window.

**Expected behavior**
The starter should create a ToolCallResultConverter based on the autoconfigured jackson json mapper and use it.

**Workaround**
Of course I can use this annotation : 
```
@Tool(description = "Retrieve customer information", resultConverter = MyNonNullJacksonConverter.class)
```
But this does not fit well in the spring boot autoconfiguration philosophy. 

Another solution could be to allow injecting a ToolCallResultConverter using `MethodToolCallbackProvider.build().toolCallResultConverter( ... )`, but this method does not exists.

My feeling is that the root problem is that **DefaultToolCallResultConverter and JsonParser are not IoC friendly**. Hardcoding a static JsonParser is imho a bad practice.

quaff:
It may works by registering `McpSyncServerCustomizer` :
```java
		@Bean
		McpSyncServerCustomizer jacksonMcpSyncServerCustomizer(JsonMaper jsonMaper) {
			return serverBuilder -> serverBuilder.jsonMapper(new JacksonMcpJsonMapper(jsonMaper));
		}
```

fbus:
> It may works by registering `McpSyncServerCustomizer` :
> 
> 		@Bean
> 		McpSyncServerCustomizer jacksonMcpSyncServerCustomizer(JsonMaper jsonMaper) {
> 			return serverBuilder -> serverBuilder.jsonMapper(new JacksonMcpJsonMapper(jsonMaper));
> 		}

Nice suggestion, but didn't work. The message is serialized by DefaultToolCallResultConverter instead: 
[MethodToolCallbackProvider.java#L98](https://github.com/spring-projects/spring-ai/blob/9bceb110e6bc34798c27ad203e9e005b85a4e9fd/spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallbackProvider.java#L98)
-> https://github.com/spring-projects/spring-ai/blob/9bceb110e6bc34798c27ad203e9e005b85a4e9fd/spring-ai-model/src/main/java/org/springframework/ai/tool/support/ToolUtils.java#L92-L97

quaff:
You are right, it's handled by `JsonParser`:

[...]

PR #6242, opened 2026-06-01
Allow custom JSON serialization for `@Tool` method results

jewoodev:
Tool call result serialization could not use a customized `JsonMapper`, because `DefaultToolCallResultConverter` held a static `JsonHelper` and `MethodToolCallbackProvider` had no provider-level override.

This adds an explicit opt-in hook instead of automatically applying `spring.jackson` to all tool results: `DefaultToolCallResultConverter` now accepts a `JsonMapper`, and `MethodToolCallbackProvider.Builder` can set a default `ToolCallResultConverter` for emitted callbacks. A per-tool `@Tool` `resultConverter` other than `DefaultToolCallResultConverter` still takes precedence.

Tests cover custom `JsonMapper` serialization and provider builder precedence. The annotation-driven `@McpTool` path is a separate API path and remains out of scope.

Addresses #6097.


jewoodev:
@sdeleuze, could you take a look at this PR when you have a chance?

Issue #5971, opened 2026-05-07
Observations not working properly in streaming mode


cijaaimee:
**Bug description**
When using ChatClient's streaming mode, observations are broken in two different ways:
* Observations are stopped in reverse order
* Tool calling observations have `null` parent
**Environment**
Spring AI 2.0.0-M5.

**Steps to reproduce**
Sample code to see how it breaks:
```kotlin
  @Bean
  fun observationRegistryCustomizer() =
    ObservationRegistryCustomizer<ObservationRegistry> {
      it.observationConfig().observationHandler(LoggingObservationHandler())
    }

class LoggingObservationHandler : ObservationHandler<Observation.Context> {
  override fun supportsContext(context: Observation.Context) = true

  private val logger = LoggerFactory.getLogger(javaClass)

  override fun onStart(context: Observation.Context) {
    logger.info(
      "start ${context.name ?: context.javaClass.name} with parent ${context.parentObservation?.contextView?.name ?: context.parentObservation?.contextView?.javaClass?.name}")
    super.onStart(context)
  }

  override fun onStop(context: Observation.Context) {
    logger.info("stop ${context.name ?: context.javaClass.name}")
  }
}
```

You'll see something like this:
```
start spring.ai.chat.client with parent <something>
start spring.ai.advisor with parent spring.ai.chat.client
start gen_ai.client.operation with parent spring.ai.advisor
start spring.ai.tool with parent null
stop spring.ai.tool
start gen_ai.client.operation with parent gen_ai.client.operation
stop gen_ai.client.operation
stop gen_ai.client.operation
stop spring.ai.chat.client
stop spring.ai.advisor
```
Notice how (1) spring.ai.tool has no parent, and (2) observations are stopped in reverse order

**Expected behavior**
With the example above, we should be seeing this:
```
start spring.ai.chat.client with parent <something>
start spring.ai.advisor with parent spring.ai.chat.client
start gen_ai.client.operation with parent spring.ai.advisor
stop gen_ai.client.operation
start spring.ai.tool with parent spring.ai.advisor
stop spring.ai.tool
start gen_ai.client.operation with parent gen_ai.client.operation
stop gen_ai.client.operation
stop spring.ai.advisor
stop spring.ai.chat.client
```

sangkyoonnam:
Both symptoms are gone on main, and I think this can be closed as fixed in 2.0.0.

[...]

PR #6244, opened 2026-06-01
Resolve null parent observation for tool calls in streaming mode

tzolov:
- Document reactive streaming context propagation and observation tree
- Wire the ObservationRegistry into the default ToolCallingManager so that spring.ai.tool observations are recorded.
- Fix a parent-observation gap in DefaultToolCallingManager where tool execution on a boundedElastic thread could not find its parent span when automatic context propagation was disabled
- Clean up deprecated DefaultChatClientBuilder and DefaultChatClientRequestSpec constructors.
- Add streaming observation integration tests and document the reactive context propagation behaviour in the observability guide.

### DIFF
--- a/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/chat/client/OpenAIToolCallObservation2IT.java
+++ b/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/chat/client/OpenAIToolCallObservation2IT.java
@@ -1,0 +1,349 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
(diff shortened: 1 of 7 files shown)


cijaaimee:
Somewhat related to this issue, I'd also like to point out #5971, where observations in streaming mode are stopped in the wrong order.
This seems to be caused by the use of `doFinally()` instead of `#doOnTerminate()` and `#doOnCancel()`. An example of how this could be solved can be found in #5979.

tzolov:
replaced by #6282 , #6276, #6249

PR #6245, opened 2026-06-01
Restore HTTP-layer observability for OpenAI models

### LINKED ISSUE #5930: Customize OkHttp client for Micrometer-based observability

ilayaperumalg:
- New SpringAiOpenAiHttpClient: Java port of the SDK's OkHttpClient, plus Micrometer OkHttpObservationInterceptor and OkHttpConnectionPoolMetrics
  - Extract shared wiring (interceptor, context-wrapped dispatcher, pool metrics) into SpringAiOkHttpObservabilityUtils in spring-ai-commons; refactor SpringAiAnthropicHttpClient to delegate to the same utility
  - OpenAiChatModel.Builder accepts MeterRegistry and dispatcherExecutor; all model auto-configs wire them from the Spring Boot context
  - Add connectionPoolMetricsEnabled property to AbstractOpenAiProperties (default false) to opt in to OkHttp connection-pool gauges
  - Add OpenAiChatModelHttpObservabilityIT

Fixes #5930


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/http/okhttp/SpringAiOpenAiHttpClient.java
```diff
@@ -0,0 +1,605 @@
...
+		public OkHttpClient buildOkHttpClient() {
+			OkHttpClient.Builder okBuilder = new OkHttpClient.Builder().retryOnConnectionFailure(false)
+				.pingInterval(Duration.ofMinutes(1))
+				.connectTimeout(this.timeout.connect())
+				.readTimeout(this.timeout.read())
+				.writeTimeout(this.timeout.write())
+				.callTimeout(this.timeout.request())
+				.proxy(this.proxy);
+
+			SpringAiOkHttpObservabilityUtils.configureObservability(okBuilder, this.observationRegistry,
```
sobychacko: It looks like we are dropping the return `boolean` value here. When no user executor is supplied, `configureObservability` method creates an internal thread pool and returns `true` to signal that the framework must shut down the executor (same way we do in the `build()` method above). Because it's discarded here and `buildWithClient` hardcodes to `false`, from what I can see we never shuts the pool down. Could you confirm?

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/http/okhttp/SpringAiOpenAiHttpClient.java
```diff
@@ -0,0 +1,605 @@
...
+				okBuilder.sslSocketFactory(this.sslSocketFactory, this.trustManager);
+			}
+
+			if (this.hostnameVerifier != null) {
+				okBuilder.hostnameVerifier(this.hostnameVerifier);
+			}
+
+			OkHttpClient okClient = okBuilder.build();
+			okClient.dispatcher().setMaxRequestsPerHost(okClient.dispatcher().getMaxRequests());
+			return okClient;
```
sobychacko: Why are we not binding to the connection pool metrics machinery here? The `build()` method correctly binds.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/http/okhttp/SpringAiOpenAiHttpClient.java
```diff
@@ -0,0 +1,605 @@
...
+		}
+
+		public SpringAiOpenAiHttpClient buildWithClient(OkHttpClient okClient) {
+			return new SpringAiOpenAiHttpClient(okClient, false);
```
sobychacko: The signal to clean up the resource is hard-coded here. See my comment above. We should capture the correct dispatcher threadpool ownership.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/setup/OpenAiSetup.java
```diff
@@ -225,17 +202,28 @@ else if (modelProvider == ModelProvider.MICROSOFT_FOUNDRY) {
...
+			.timeout(timeout)
+			.proxy(proxy)
+			.dispatcherExecutorService(dispatcherExecutor);
+
+		// Build the OkHttpClient directly (no intermediate SpringAiOpenAiHttpClient
+		// wrapper) so no executor is created and discarded before we add the
+		// Authorization-stripping interceptor.
+		okhttp3.OkHttpClient baseOkClient = httpBuilder.buildOkHttpClient();
+
+		okhttp3.OkHttpClient strippedOkClient = baseOkClient.newBuilder().addInterceptor(chain -> {
```
sobychacko: @ilayaperumalg Maybe instead of this whole detour, we add a `buildWithInterceptor` method that calls the normal `build()` and just adds the interceptor at the end? That way, we keep a single clean `build()` path and avoid the leak/metrics issues. Basically, call `buildWithInterceptor()` right after the call to `dispatcherExecutorService(dispatcherExecutor)` above. What do you think?

spring-ai-commons/src/main/java/org/springframework/ai/observation/conventions/SpringAiOkHttpObservabilityUtils.java
```diff
@@ -0,0 +1,103 @@
...
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Utility class for configuring Micrometer observability on OkHttp clients.
+ *
+ * @author Soby Chacko
+ * @author Ilayaperumal Gopinathan
+ * @since 2.0.0
+ */
+public final class SpringAiOkHttpObservabilityUtils {
```
sobychacko: @ilayaperumalg I understand that we have some duplication between the modules. However, I think we might be better off keeping them in the respective modules (opeanai and anthropic) and live with the duplication. Here are the reasons for that:
1. First of all, we are now pulling okttp packages and classes into spring-ai-common, and even though its an optional dep, we now compile against it whether we need it or not. 
2. This `convention` package seems to be for micrometer observation-related conventions, and a utils class doesn't fit here 
3. This situation with okhttp is a temporary one since there is a chance that the SDIK might fix it eventually, then this whole thing becomes throw-away code. So, no reason for duplication imho.

What do you think?

sobychacko (changes requested):
@ilayaperumalg Added some comments. Pls let me know what you think? Also, could you rebase the PR against the latest main?

spring-ai-commons/src/main/java/org/springframework/ai/observation/conventions/SpringAiOkHttpObservabilityUtils.java
ilayaperumalg: Sure Soby, I agree with your suggestion. I have updated the PR accordingly.

sobychacko:
Merged upstream.

PR #6265, opened 2026-06-02
Trim whitespace from evaluation response check

rajadilipkolli:
image

if condition is failing due to whitespace in the evaluationResponse

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/RelevancyEvaluator.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/RelevancyEvaluator.java
@@ -80,7 +80,7 @@ public class RelevancyEvaluator implements Evaluator {
 
 		boolean passing = false;
 		float score = 0;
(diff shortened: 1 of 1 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/evaluation/RelevancyEvaluator.java
```diff
@@ -80,7 +80,7 @@ public EvaluationResponse evaluate(EvaluationRequest evaluationRequest) {
 
 		boolean passing = false;
 		float score = 0;
-		if ("yes".equalsIgnoreCase(evaluationResponse)) {
```
ilayaperumalg, line 83: @rajadilipkolli Thanks for the PR! We need to guard against null here. Also, the same fix needs to be applied for FactCheckingEvaluator as well. I will apply the changes along with some tests.

ilayaperumalg:
Rebased and merged via 847fc0e4481cec0d9cbdf5482675adc75de65264. Polished the changes to include the changes discussed [here](https://github.com/spring-projects/spring-ai/pull/6265#discussion_r3350700643).

PR #6268, opened 2026-06-03
GH-6261: Cache tool result messages during tool-calling rounds

### LINKED ISSUE #6261: [Anthropic] Place a `cache_control` breakpoint on the last tool result message during tool-calling rounds
Related to (and a more specific follow-up of) #4325 *(Make Anthropic prompt caching message-type aware)*, which was closed as completed. The general message-type-aware machinery landed, but `MessageType.TOOL` (tool result) messages still never receive a `cache_control` breakpoint. We verified this is unchanged on `main` (`2.0.0-SNAPSHOT`, post-`M8`) — the `MessageType.TOOL` branch in `createRequest()` is identical to `2.0.0-M6`, and none of the five `AnthropicCacheStrategy` values target tool result messages (they address tool *definitions*, system, and the last user message only).

**Expected Behavior**

When a turn triggers one or more tool calls, the conversation grows by an assistant `tool_use` message and a user
`tool_result` message on each round, and every round is sent back to Anthropic as a new request. To benefit from prompt
caching across these rounds, the request should be able to place a `cache_control` breakpoint on the **last tool result block** of the current request.

That way, on the next round Anthropic reads the previous round's tool results from cache (`cache_read_input_tokens`) instead
of re-processing them as uncached input. Whether this is done automatically by an existing/new `AnthropicCacheStrategy` or
exposed as a configurable option is up to the maintainers.

**Current Behavior**

In `AnthropicChatModel.createRequest()`, tool result messages (`MessageType.TOOL` / `ToolResponseMessage`) are converted to
content blocks **without any `cache_control`**, and no `AnthropicCacheStrategy` places a breakpoint on them.

The `CONVERSATION_HISTORY` strategy only places a breakpoint on the **last user message**. During a multi-round tool-calling
turn, the tool result messages are appended *after* that last user message, so they fall outside every breakpoint the
strategy sets. None of the strategy values (`NONE`, `TOOLS_ONLY`, `SYSTEM_ONLY`, `SYSTEM_AND_TOOLS`, `CONVERSATION_HISTORY`)
targets a tool result message.

Consequence: each tool result is billed as uncached `input_tokens` on the round that introduces it. Because there is no
explicit breakpoint at/after the tool results, the growing tool-result prefix is not reliably written to / read from the
cache across rounds.

This is not configurable with the current API: `cache_control` placement is strategy-driven, and no strategy targets tool
result messages (cf. #5494, which requested similar granularity for system messages).

**Context**

We run an assistant on Spring AI + Anthropic (Claude) where most questions trigger 1–4 tool calls per turn, some of which
return large payloads. We analyzed the raw request logs for a two-turn conversation and confirmed the stable prefix (tools +
system + prior user message) is served from cache correctly. But within the tool-calling rounds of a single turn we saw the
tool results never cached:

| Request | Round                 | `cache_read` | `cache_creation` | `input` (uncached) |
|---------|-----------------------|-------------:|-----------------:|-------------------:|
| 1       | initial user question |        18045 |               69 |                  3 |
| 2       | after 1st tool result |        18114 |                0 |          **11836** |
| 3       | after 2nd tool result |        18114 |            11987 |                672 |
| 4       | after 3rd tool result |        18114 |            11987 |                918 |

The ~11.8k-token tool result in Request 4 shows up entirely as uncached `input_tokens` rather than `cache_creation`,
precisely because no `cache_control` breakpoint is placed on (or after) the tool result. With large tool outputs and several
rounds per turn, this is a meaningful, repeated cost on exactly the content most worth caching.

We considered switching to a different `AnthropicCacheStrategy`, but none place a breakpoint on tool result messages, so
there is no way to address this without bypassing the framework. Native support for caching the last tool result of a
tool-calling round would let long tool outputs be cached across rounds without working around Spring AI.

**Environment**

- spring-ai-anthropic `2.0.0-M6` (also verified present on `main` / `2.0.0-SNAPSHOT`, post-`M8`)
- anthropic-java `2.30.0`
- `AnthropicCacheStrategy.CONVERSATION_HISTORY`, `multiBlockSystemCaching(true)`

sobychacko:
Closes #6261

https://github.com/spring-projects/spring-ai/issues/6261

Add a `cacheToolResults` flag to `AnthropicCacheOptions`. When enabled, place a `cache_control` breakpoint on the last block of the last tool result message of a request, so prior tool outputs are written to and read from the cache across tool-calling rounds instead of being billed as uncached input on every round.

Previously the `MessageType.TOOL` branch in `createRequest()` never consulted the cache resolver, and `CONVERSATION_HISTORY` only places a breakpoint on the last user message. During a multi-round tool-calling turn the tool results are appended after that user message, so they fell outside every breakpoint and large tool outputs were reprocessed as uncached input each round.

The flag is additive and takes effect with `CONVERSATION_HISTORY` (the strategy whose eligible message types already include tool results); no resolver change was needed, only the missing call site. The breakpoint moves to the latest tool result each round, consuming one of the four allowed breakpoints. TTL and minimum content length are configurable.

Adds unit tests for breakpoint placement and an integration test that verifies tool results are cached and read back across rounds.

Documentation updated accordingly.

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicCacheOptions.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicCacheOptions.java
@@ -58,11 +58,13 @@ public class AnthropicCacheOptions {
 
 	private final boolean multiBlockSystemCaching;
 
+	private final boolean cacheToolResults;
+
 	protected AnthropicCacheOptions(@Nullable AnthropicCacheStrategy strategy,
 			@Nullable Function<@Nullable String, Integer> contentLengthFunction,
 			@Nullable Map<MessageType, AnthropicCacheTtl> messageTypeTtl,
-			@Nullable Map<MessageType, Integer> messageTypeMinContentLengths,
-			@Nullable Boolean multiBlockSystemCaching) {
+			@Nullable Map<MessageType, Integer> messageTypeMinContentLengths, @Nullable Boolean multiBlockSystemCaching,
+			@Nullable Boolean cacheToolResults) {
 		this.strategy = (strategy != null ? strategy : AnthropicCacheStrategy.NONE);
 		this.contentLengthFunction = (contentLengthFunction != null ? contentLengthFunction
 				: s -> s != null ? s.length() : 0);
@@ -75,6 +77,7 @@ public class AnthropicCacheOptions {
 					mt -> (messageTypeMinContentLengths != null && messageTypeMinContentLengths.containsKey(mt))
 							? messageTypeMinContentLengths.get(mt) : DEFAULT_MIN_CONTENT_LENGTH));
 		this.multiBlockSystemCaching = (multiBlockSystemCaching != null ? multiBlockSystemCaching : false);
+		this.cacheToolResults = (cacheToolResults != null ? cacheToolResults : false);
 	}
 
 	public static AnthropicCacheOptions.Builder builder() {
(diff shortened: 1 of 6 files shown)


sobychacko:
This PR has been merged upstream.

PR #6288, opened 2026-06-04
Fix OpenAiChatModel.stream() to only buffer tool calls.

### LINKED ISSUE #5987: `OpenAiChatModel#stream` buffers the entire response before emitting chunks

ericbottard:
Previously, the new implementation of OpenAiChatModel.stream() buffered
 the entire response, which meant that the caller would not receive
 any tokens until the entire response was received. This change modifies
the implementation to only buffer tool calls, allowing the caller to
 receive tokens as they are generated.

Fix https://github.com/spring-projects/spring-ai/issues/5987

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -19,13 +19,16 @@ package org.springframework.ai.openai;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.openai.client.OpenAIClient;
 import com.openai.client.OpenAIClientAsync;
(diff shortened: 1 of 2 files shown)


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -306,193 +299,156 @@ private Flux<ChatResponse> internalStream(Prompt prompt, @Nullable ChatResponse
...
+					.map(OpenAiChatModel::mergeChunks)
+					.map(this::chunkToChatCompletion);
+
+			Flux<ChatResponse> chatResponses = aggregatedChatCompletions.map(chatCompletion -> {
+				String id = chatCompletion.id();
+				List<Generation> generations = chatCompletion.choices().stream().map(choice -> {
+					roleMap.putIfAbsent(id, choice.message()._role().asString().isPresent()
+							? choice.message()._role().asStringOrThrow() : "");
+
+					Map<String, Object> metadata = Map.of("id", id, //
```
guanxuc, line 336: Hi, @ericbottard , seems like the `chunkChoice` and `reasoningContent` properties are missed in metadata. 
The current main branch code logic is
```
Map<String, Object> metadata = Map.of("id", id, "role", roleMap.getOrDefault(id, ""),
		"index", choice.index(), "finishReason", choice.finishReason().value(), "refusal",
		choice.message().refusal().orElse(""), "annotations",
		choice.message().annotations().orElse(List.of()), "chunkChoice",
		chunk.choices().get((int) choice.index()), "reasoningContent",
		getReasoningContent(choice));
```

ericbottard: Hi @guanxuc,

the whole choice object had no place in the metadata in the first place, so I removed it.
The reasoningContent on the other hand was a mistake on my side, good catch!

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -306,193 +298,156 @@ private Flux<ChatResponse> internalStream(Prompt prompt, @Nullable ChatResponse
...
+				CompletionUsage usageVal = usage.orElse(null);
+				Usage currentUsage = usageVal != null ? getDefaultUsage(usageVal) : new EmptyUsage();
+				Usage accumulated = UsageCalculator.getCumulativeUsage(currentUsage, null);
+				return new ChatResponse(generations, from(chatCompletion, accumulated));
+
 			});
 
-			Flux<ChatResponse> flux = chatResponses
+			Flux<ChatResponse> observedResponses = chatResponses.doOnError(observation::error)
+				.doFinally(s -> observation.stop())
```
ericbottard, line 356: @tzolov pay attention to [this comment](https://github.com/spring-projects/spring-ai/pull/6244#issuecomment-4613406700) as well, maybe this needs changing.

tzolov:
Rebased, fix tests, squashed and merged at 12f5cbe618fdadb284e10ea1f0c2fc792f0ab5e4
Thanks @ericbottard !

PR #6263, opened 2026-06-02
fix(core): allow tool callbacks and providers directly in ChatClient tools()

Baqirrizvidev:
Resolves spring-projects/spring-ai#6206. Permitting ChatClient.prompt().tools(...) to accept ToolCallback, ToolCallbackProvider, collections, and arrays of them directly, matching user expectations after deprecation of toolCallbacks(...).

Issue #6206, opened 2026-05-29
Deprecated method toolCallbacks cannot be replaced with method tools when using mcp client


hansdesmet:
My code
``` java
package be.vdab.mcpclientprobleem;

import org.springframework.ai.chat.client.ChatClient;
import org.springframework.ai.tool.ToolCallbackProvider;
import org.springframework.boot.CommandLineRunner;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;

@SpringBootApplication
public class McpclientprobleemApplication implements CommandLineRunner {
    private final ChatClient chatClient;
    private final ToolCallbackProvider toolCallbackProvider;

    public McpclientprobleemApplication(ChatClient.Builder builder, ToolCallbackProvider toolCallbackProvider) {
        this.chatClient = builder.build();
        this.toolCallbackProvider = toolCallbackProvider;
    }

    public static void main(String[] args) {
        SpringApplication.run(McpclientprobleemApplication.class, args);
    }

    @Override
    public void run(String... args) throws Exception {
        System.out.println(
                chatClient.prompt("What is the capital of France ?")
                        .toolCallbacks(toolCallbackProvider)
                        .call().content());
    }
}
```
works with Spring AI 2.0.0-M8 and follwing reference to MCP server in application.properties:
spring.ai.mcp.client.streamable-http.connections.googelmaps.url=https://mapscodeassist.googleapis.com/mcp

However, the method **toolCallbacks** is deprecated and it is advised to replace it with the method **tools**.
However, when I use the method tools (line 25 of the code becomes .tools(toolCallbackProvider)),
I get this exception:

java.lang.IllegalStateException: No @Tool annotated methods found in org.springframework.ai.mcp.SyncMcpToolCallbackProvider@2f0ed952.Did you mean to pass a ToolCallback or ToolCallbackProvider? If so, you have to use .toolCallbacks() instead of .tool()
	at org.springframework.ai.tool.method.MethodToolCallbackProvider.assertToolAnnotatedMethodsPresent(MethodToolCallbackProvider.java:78) ~[spring-ai-model-2.0.0-M8.jar:2.0.0-M8]
	at org.springframework.ai.tool.method.MethodToolCallbackProvider.<init>(MethodToolCallbackProvider.java:62) ~[spring-ai-model-2.0.0-M8.jar:2.0.0-M8]
[...]

PR #6292, opened 2026-06-04
Remove `ToolSpec` consumer API from `ChatClient`

tzolov:
Follow-up to #6263: now that tools(Object...) accepts ToolCallback and ToolCallbackProvider directly, the ToolSpec consumer overloads are no longer needed. 
Remove tools(Consumer<ToolSpec>), defaultTools(Consumer<ToolSpec>), and the ChatClient.ToolSpec interface entirely.

The toolContext() and defaultToolContext() are promoted to first-class non-deprecated methods with full key/value validation. mutate() is fixed to propagate toolCallbackProviders and toolContext to the cloned builder.

Update docs and tests accordingly.

See #6206

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
@@ -29,7 +29,6 @@ import reactor.core.publisher.Flux;
 
 import org.springframework.ai.chat.client.advisor.ToolCallAdvisor;
 import org.springframework.ai.chat.client.advisor.api.Advisor;
-import org.springframework.ai.chat.client.advisor.api.ToolAdvisor;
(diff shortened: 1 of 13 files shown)


spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/ChatClient.java
```diff
@@ -389,43 +388,35 @@ interface ChatClientRequestSpec {
 
 		<B extends ChatOptions.Builder<?>> ChatClientRequestSpec options(B customizer);
 
-		ChatClientRequestSpec tools(Consumer<ToolSpec> consumer);
-
 		ChatClientRequestSpec tools(Object... toolObjects);
```
sdeleuze: The parameter type does not give any hint on what to use here, so please provide some some information on what types or annotations are supported.

PR #6294, opened 2026-06-04
Make OpenAI HTTP Client configurable

ThomasVitale:
The official OpenAI Java SDK doesn't support any customization of the underlying HTTP Client, which makes it unsuitable for production scenarios.

In a previous PR, a custom `SpringAiOpenAiHttpClient` has been introduced to add support for observability instrumentation and context propagation. This PR builds on top of that and introduces a `OpenAiHttpClientBuilderCustomizer` interface that can be used to customize the `SpringAiOpenAiHttpClient.Builder`.

Examples of real-world customizations:

* add an interceptor to integrate Spring Security and configure OAuth2 Client Credentials authentication instead of the weaker API Key approach
* configure a custom certificate for TLS communication based on a private PKI
* use a virtual thread-based executor instead of platform threads.

Any `OpenAiHttpClientBuilderCustomizer` bean is used in the auto-configuration and passed down to each `OpenAi*Model` class for uniform customization support.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiImageModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiImageModel.java
@@ -16,6 +16,7 @@
 
 package org.springframework.ai.openai;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
@@ -37,6 +38,7 @@ import org.springframework.ai.image.observation.ImageModelObservationContext;
 import org.springframework.ai.image.observation.ImageModelObservationConvention;
 import org.springframework.ai.image.observation.ImageModelObservationDocumentation;
 import org.springframework.ai.observation.conventions.AiProvider;
+import org.springframework.ai.openai.http.okhttp.OpenAiHttpClientBuilderCustomizer;
 import org.springframework.ai.openai.metadata.OpenAiImageGenerationMetadata;
 import org.springframework.ai.openai.metadata.OpenAiImageResponseMetadata;
 import org.springframework.ai.openai.setup.OpenAiSetup;
@@ -132,22 +134,24 @@ public class OpenAiImageModel implements ImageModel {
 	 */
 	public OpenAiImageModel(@Nullable OpenAIClient openAiClient, @Nullable OpenAiImageOptions options,
 			@Nullable ObservationRegistry observationRegistry) {
+		this(builder().openAiClient(openAiClient).options(options).observationRegistry(observationRegistry));
+	}
 
-		if (options == null) {
-			this.options = OpenAiImageOptions.builder().build();
-		}
-		else {
-			this.options = options;
-		}
-		this.openAiClient = Objects.requireNonNullElseGet(openAiClient,
+	public static Builder builder() {
+		return new Builder();
+	}
+
+	private OpenAiImageModel(Builder builder) {
+		this.options = builder.options != null ? builder.options : OpenAiImageOptions.builder().build();
+		this.observationRegistry = Objects.requireNonNullElse(builder.observationRegistry, ObservationRegistry.NOOP);
+		this.openAiClient = Objects.requireNonNullElseGet(builder.openAiClient,
 				() -> OpenAiSetup.setupSyncClient(this.options.getBaseUrl(), this.options.getApiKey(),
 						this.options.getCredential(), this.options.getMicrosoftDeploymentName(),
 						this.options.getMicrosoftFoundryServiceVersion(), this.options.getOrganizationId(),
 						this.options.isMicrosoftFoundry(), this.options.isGitHubModels(), this.options.getModel(),
 						this.options.getTimeout(), this.options.getMaxRetries(), this.options.getProxy(),
-						this.options.getCustomHeaders(),
-						observationRegistry != null ? observationRegistry : ObservationRegistry.NOOP, null, null));
-		this.observationRegistry = Objects.requireNonNullElse(observationRegistry, ObservationRegistry.NOOP);
+						this.options.getCustomHeaders(), this.observationRegistry, null,
+						builder.httpClientCustomizers));
 	}
 
 	/**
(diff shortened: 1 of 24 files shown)


ThomasVitale:
@tzolov this is a proposal to address the problems I shared yesterday about not being able to customize the HTTP Client used by the OpenAI module after the switch to the official SDK. I hope it can make it to 2.0.0.

@sobychacko @ilayaperumalg since you worked on the custom client module for Anthropic and OpenAI to add observability, it'd be great to get your feedback on this proposal

ThomasVitale:
The test failing in `OpenAiChatModelTests.preserveRateLimitAndPromptMetadataInAggregation:215` is caused by this recent change just pushed to the main branch: https://github.com/spring-projects/spring-ai/commit/12f5cbe618fdadb284e10ea1f0c2fc792f0ab5e4#diff-e5cb77ca54425cc7c57b08f1d70da4118eaf2c0377d381b1d184ace96d6f6e89L556

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/http/okhttp/SpringAiOpenAiHttpClient.java
```diff
@@ -527,6 +561,15 @@ public SpringAiOpenAiHttpClient build() {
 				.build();
 			okBuilder.addInterceptor(observationInterceptor);
 
+			if (this.noApiKey) {
+				okBuilder.addInterceptor(
+						chain -> chain.proceed(chain.request().newBuilder().removeHeader("Authorization").build()));
+			}
+
+			for (Interceptor userInterceptor : this.userInterceptors) {
```
sobychacko: If a customizer calls `.noApiKey(false)`, then it could potentially override if it was set to `true` before and any associated interceptors above. no? For instance, the builder method `noApiKey(boolean)` is publicly settable via customization, which means that an end user can control it while we intend that to be set only via `OpenAiSetup`. Maybe we should remove that new builder method (`noApiKey`)  and let only the `OpenAiSetup` can flip it. Then we can remove that `if (this.noApiKey)` conditional above altogether along with the builder method. We should then modify `OpenaiSetup#buildNoAuthClientOptions` by directly calling the `.interceptor(...)` method. What do you think?

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/http/okhttp/SpringAiOpenAiHttpClient.java
```diff
@@ -511,6 +519,32 @@ public Builder meterTags(Iterable<Tag> meterTags) {
...
+
+		/**
+		 * Enables no-API key mode. When set, an OkHttp interceptor is inserted that
+		 * strips the {@code Authorization} header from every outbound request before any
+		 * {@link #interceptor(Interceptor) user interceptor} runs. Used by Spring AI's
+		 * {@code OpenAiSetup} when an empty API key is supplied to bypass the OpenAI
+		 * SDK's mandatory-credential check while letting a user interceptor (e.g. OAuth2)
+		 * attach the real Authorization header afterwards.
+		 */
+		public Builder noApiKey(boolean noApiKey) {
```
sobychacko: What about removing this method and let `OpenAiSetup#buildNoAuthClientOptions` directly handle this so that the user will not be inadvertently flip this setting? See my other related comment.

sobychacko (changes requested):

sobychacko (changes requested):
@ThomasVitale The PR looks great! I have one question/comment that I added as in-line on the PR. Please take a look and see what you think. Maybe its a non-issue, but i will let you decide. Other than that, we definitely need to add docs for this as this a user-facing feature. But, overall, I think this is a great enhancement. Thank you!

sdeleuze:
Moved to 2.0.0 GA in order to have more time to review and add a similar Anthropic extension point.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/http/okhttp/SpringAiOpenAiHttpClient.java
ThomasVitale: I think it's a great idea to remove the `.noApiKey(boolean)` builder method and handle the no-api-key interceptor in `OpenAiSetup`.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/http/okhttp/SpringAiOpenAiHttpClient.java
ThomasVitale: I really like that idea, working on it.

ThomasVitale:
I have made the changes suggested by @sobychacko and added documentation to the OpenAI pages for each model type. I used a "partial" to avoid duplicating the content everywhere.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/http/okhttp/SpringAiOpenAiHttpClient.java
ThomasVitale: Change applied. Now `OpenAiSetup#buildNoAuthClientOptions` handles directly the case when no API key is defined, and adds an interceptor.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/http/okhttp/SpringAiOpenAiHttpClient.java
ThomasVitale: Change applied. Now `OpenAiSetup#buildNoAuthClientOptions` handles directly the case when no API key is defined, and adds an interceptor. The `.noApiKey(boolean)` method has been removed.

ThomasVitale:
The failing test `ToolCallAdvisorSpanHierarchyTests.modelCallSpansNestUnderToolCallAdvisorEvenWhenAnOuterScopeIsOpen` seems to be flaky and it's not related to the changes in this PR.

sobychacko (approved):

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiImageModel.java
```diff
@@ -221,4 +225,63 @@ public void setObservationConvention(ImageModelObservationConvention observation
...
+		 * Registers an {@link OpenAiHttpClientBuilderCustomizer} that mutates the
+		 * underlying OkHttp client builder before the OpenAI clients are constructed. Use
+		 * this to attach OkHttp interceptors (e.g. OAuth2 bearer-token injection), swap
+		 * the dispatcher executor, or tweak any other OkHttp setting. Customizers are
+		 * applied in the order they are registered, after Spring AI's own defaults, so
+		 * user code wins.
+		 */
+		public Builder httpClientBuilderCustomizer(OpenAiHttpClientBuilderCustomizer customizer) {
+			Assert.notNull(customizer, "customizer cannot be null");
+			this.httpClientCustomizers.add(Objects.requireNonNull(customizer, "customizer"));
```
ilayaperumalg: redundant null check here.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiModerationModel.java
```diff
@@ -197,6 +202,31 @@ public Builder options(OpenAiModerationOptions options) {
...
+		 * Registers an {@link OpenAiHttpClientBuilderCustomizer} that mutates the
+		 * underlying OkHttp client builder before the OpenAI clients are constructed. Use
+		 * this to attach OkHttp interceptors (e.g. OAuth2 bearer-token injection), swap
+		 * the dispatcher executor, or tweak any other OkHttp setting. Customizers are
+		 * applied in the order they are registered, after Spring AI's own defaults, so
+		 * user code wins.
+		 */
+		public Builder httpClientBuilderCustomizer(OpenAiHttpClientBuilderCustomizer customizer) {
+			Assert.notNull(customizer, "customizer cannot be null");
+			this.httpClientCustomizers.add(Objects.requireNonNull(customizer, "customizer"));
```
ilayaperumalg: redundant null check here as well.

ilayaperumalg (approved):

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiModerationModel.java
ThomasVitale: Fixed

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiImageModel.java
ThomasVitale: Fixed

ThomasVitale:
@tzolov I addressed all the comments. If there's nothing more, could this be included in the RC2 release today? That would give us the possibility for doing some extra testing and share feedback.

tzolov (approved):
LGTM, thanks @ThomasVitale

tzolov:
rebased, added additional tests, squashed and merged at b677c7fca2a8e462a3f4dca4afe5c6f2f77c7597

ilayaperumalg:
@ThomasVitale Reflecting your changes on OpenAI, I have submitted a PR for Anthropic https://github.com/spring-projects/spring-ai/pull/6354 Could you review? Thank you

PR #6303, opened 2026-06-05
Rename ToolCallAdvisor to ToolCallingAdvisor with backward-compat shims and bug fixes

tzolov:
Retain ToolCallAdvisor as a @Deprecated subclass with builder() factory, Builder<T> inner class, and AdvisorParams.toolCallAdvisorAutoRegister() alias so existing callers get a deprecation warning rather than a compile error.

Also fix bugs in ToolCallingAdvisor:
- Builder.copy() now uses a newCopy() template method so subclass builders preserve their type through auto-registration
- doGetNextInstructions* guards against empty conversationHistory
- The streaming filter delegates to toolExecutionEligibilityChecker instead of hardcodin hasToolCalls()


ilayaperumalg (approved):

ilayaperumalg:
Rebased and merged as 3c527d69716714da28bab28d4216f15514bf7518 with some polishing via dfb5076037df6ee7e3c280d2020c714ebe161ce7

PR #6309, opened 2026-06-05
Move advisors to their proper modules

sdeleuze:
- Rename `spring-ai-advisors-vector-store` to `spring-ai-vector-store-advisor`
- Move `ToolSearchToolCallingAdvisor` to a `spring-ai-tool-search-advisor` module


ilayaperumalg (approved):

spring-ai-rag/src/main/java/org/springframework/ai/rag/preretrieval/query/expansion/MultiQueryExpander.java
```diff
@@ -73,19 +74,19 @@ public final class MultiQueryExpander implements QueryExpander {
 
 	private static final Integer DEFAULT_NUMBER_OF_QUERIES = 3;
 
-	private final ChatClient chatClient;
+	private final ChatModel chatModel;
```
ThomasVitale: @sdeleuze Replacing `ChatClient` with `ChatModel` has a series of consequences.

* developers have to autowire both a `ChatClient.Builder` and a `ChatModel` instance whenever they use the RAG advisor, which is not ideal. See for example https://github.com/ThomasVitale/llm-apps-java-spring-ai/blob/main/rag/rag-sequential/rag-advanced/src/main/java/com/thomasvitale/ai/spring/RagControllerQueryTranslation.java#L28
* it's not possible to configure the `ChatOptions`, which is an essential step for ensuring the right performance of these components. For example, the rewrite and translation components require a different temperature value for optimal results. If I can only pass a `ChatModel`, there's no way for me to pass custom options (unlike with `ChatClient.Builder`), unless I build a `ChatModel` from scratch (which is really not a good developer experience, since I wouldn't even benefit from auto-configuration).
* it won't be possible to enhance the LLM interactions performed by the RAG components. Using a `ChatClient.Builder` made it possible to define guardrails, evaluation, or other compliance/resilience checks via Advisors (e.g. for audit logging of all LLM interactions), while also allowing the component implementation to add additional feature-specific advisors. Now, that won't be possible anymore.
* in some advanced RAG scenarios, tool calling capabilities is needed. After removing the `ChatClient` dependency, no tool calling will be possible.

In the first implementation I made of the Modular RAG components, I was actually using `ChatModel`. But after running a few concrete scenarios, I realised that I would have had to re-implement lots of things that `ChatClient` offers, so I switched to using `ChatClient.Builder` in the final version I contributed.

My recommendation would be to keep using `ChatClient.Builder`.

ilayaperumalg: @ThomasVitale Thanks for catching this!

sdeleuze: Yeah thanks, this is unfortunate but required with current design, so I will remove the second commit and keep `spring-ai-rag-advisor` unchanged.

ThomasVitale: Thanks, Sébastien!

tzolov (approved):
LGTM! Thanks @sdeleuze

PR #6319, opened 2026-06-06
Update OllamaModel with Mistral AI's recent models

nicolaskrier:
## Description

- Remove Mistral deprecated model
- Add Ministral 3 model family
- Polish unit tests

## Explanations

Recently integration tests have failed for Ollama model due to model `mistral` not found:

```shell
Error:  Errors: 
Error:  org.springframework.ai.ollama.OllamaChatModelIT.jsonStructuredOutputWithFormatOption
Error:    Run 1: OllamaChatModelIT.jsonStructuredOutputWithFormatOption:284 » NonTransientAi 404 - {"error":"model 'mistral' not found"}
Error:    Run 2: OllamaChatModelIT.jsonStructuredOutputWithFormatOption:284 » NonTransientAi 404 - {"error":"model 'mistral' not found"}
Error:    Run 3: OllamaChatModelIT.jsonStructuredOutputWithFormatOption:284 » NonTransientAi 404 - {"error":"model 'mistral' not found"}
[INFO] 
Error:  org.springframework.ai.ollama.OllamaChatModelIT.jsonStructuredOutputWithOutputSchemaOption
Error:    Run 1: OllamaChatModelIT.jsonStructuredOutputWithOutputSchemaOption:300 » NonTransientAi 404 - {"error":"model 'mistral' not found"}
Error:    Run 2: OllamaChatModelIT.jsonStructuredOutputWithOutputSchemaOption:300 » NonTransientAi 404 - {"error":"model 'mistral' not found"}
Error:    Run 3: OllamaChatModelIT.jsonStructuredOutputWithOutputSchemaOption:300 » NonTransientAi 404 - {"error":"model 'mistral' not found"}
[INFO] 
Error:  org.springframework.ai.ollama.OllamaChatModelIT.roleTest
Error:    Run 1: OllamaChatModelIT.roleTest:130 » NonTransientAi 404 - {"error":"model 'mistral' not found"}
Error:    Run 2: OllamaChatModelIT.roleTest:130 » NonTransientAi 404 - {"error":"model 'mistral' not found"}
Error:    Run 3: OllamaChatModelIT.roleTest:130 » NonTransientAi 404 - {"error":"model 'mistral' not found"}
```

## Proposed milestone

**2.0.0**

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaChatOptions.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaChatOptions.java
@@ -93,7 +93,7 @@ public class OllamaChatOptions implements ToolCallingChatOptions, StructuredOutp
 		this.mirostatEta = mirostatEta;
 		this.penalizeNewline = penalizeNewline;
 		this.stop = stop != null ? List.copyOf(stop) : null;
-		this.model = model != null ? model : OllamaModel.MISTRAL.id();
+		this.model = model != null ? model : OllamaModel.MINISTRAL_3_8B.id();
 		this.format = format;
 		this.keepAlive = keepAlive;
 		this.truncate = truncate;
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
@@ -115,15 +115,25 @@ public enum OllamaModel implements ChatModelDescription {
 	LLAMA3_2_3B("llama3.2:3b"),
 
 	/**
-	 * The 7B parameters model
-	 */
-	MISTRAL("mistral"),
-
-	/**
 	 * A 12B model with 128k context length, built by Mistral AI in collaboration with
 	 * NVIDIA.
 	 */
 	MISTRAL_NEMO("mistral-nemo"),
+
+	/**
+	 * The Ministral 3 3B language model from Mistral AI.
+	 */
+	MINISTRAL_3_3B("ministral-3:3b"),
+
+	/**
+	 * The Ministral 3 8B language model from Mistral AI.
+	 */
+	MINISTRAL_3_8B("ministral-3:8b"),
+
+	/**
+	 * The Ministral 3 14B language model from Mistral AI.
+	 */
+	MINISTRAL_3_14B("ministral-3:14b"),
 
 	/**
 	 * A small vision language model designed to run efficiently on edge devices.
--- a/models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatModelTests.java
+++ b/models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatModelTests.java
@@ -24,7 +24,7 @@ import io.micrometer.observation.ObservationRegistry;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.ValueSource;
+import org.junit.jupiter.params.provider.EnumSource;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
@@ -57,19 +57,21 @@ import static org.mockito.Mockito.when;
  * @author Alexandros Pappas
  * @author Thomas Vitale
  * @author Sebastien Deleuze
+ * @author Nicolas Krier
  * @since 1.0.0
  */
 @ExtendWith(MockitoExtension.class)
 class OllamaChatModelTests {
 
 	@Mock
-	OllamaApi ollamaApi;
+	private OllamaApi ollamaApi;
 
 	@Test
 	void buildOllamaChatModelWithConstructor() {
 		ChatModel chatModel = new OllamaChatModel(this.ollamaApi,
-				OllamaChatOptions.builder().model(OllamaModel.MISTRAL).build(), ToolCallingManager.builder().build(),
-				ObservationRegistry.NOOP, ModelManagementOptions.builder().build());
+				OllamaChatOptions.builder().model(OllamaModel.MINISTRAL_3_8B).build(),
+				ToolCallingManager.builder().build(), ObservationRegistry.NOOP,
+				ModelManagementOptions.builder().build());
 		assertThat(chatModel).isNotNull();
 	}
 
@@ -242,9 +244,8 @@ class OllamaChatModelTests {
 	}
 
 	@ParameterizedTest
-	@ValueSource(strings = { "LLAMA2", "MISTRAL", "CODELLAMA", "LLAMA3", "GEMMA" })
-	void buildOllamaChatModelWithDifferentModels(String modelName) {
-		OllamaModel model = OllamaModel.valueOf(modelName);
+	@EnumSource(value = OllamaModel.class, names = { "LLAMA2", "MINISTRAL_3_8B", "CODELLAMA", "LLAMA3", "GEMMA" })
+	void buildOllamaChatModelWithDifferentModels(OllamaModel model) {
 		OllamaChatOptions options = OllamaChatOptions.builder().model(model).build();
 
 		ChatModel chatModel = OllamaChatModel.builder().ollamaApi(this.ollamaApi).options(options).build();
@@ -314,7 +315,10 @@ class OllamaChatModelTests {
 	@Test
 	void buildOllamaChatModelImmutability() {
 		// Test that the builder creates immutable instances
-		OllamaChatOptions options = OllamaChatOptions.builder().model(OllamaModel.MISTRAL).temperature(0.5).build();
+		OllamaChatOptions options = OllamaChatOptions.builder()
+			.model(OllamaModel.MINISTRAL_3_14B)
+			.temperature(0.5)
+			.build();
 
 		ChatModel chatModel1 = OllamaChatModel.builder().ollamaApi(this.ollamaApi).options(options).build();
 
--- a/models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaChatOptionsTests.java
+++ b/models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaChatOptionsTests.java
@@ -49,9 +49,8 @@ class OllamaChatOptionsTests extends AbstractChatOptionsTests<OllamaChatOptions,
 
 	@Test
 	void testBasicOptions() {
-		var b1 = OllamaChatOptions.builder().model("model").mainGPU(12);
-
-		var b = OllamaChatOptions.builder().mainGPU(12).model("model");
+		OllamaChatOptions.builder().model("model").mainGPU(12);
+		OllamaChatOptions.builder().mainGPU(12).model("model");
 
 		var options = OllamaChatOptions.builder().temperature(3.14).topK(30).stop(List.of("a", "b", "c")).build();
 
@@ -236,9 +235,9 @@ class OllamaChatOptionsTests extends AbstractChatOptionsTests<OllamaChatOptions,
 
 		var optionsMap = options.toMap();
 		assertThat(optionsMap).containsOnlyKeys("model");
-		assertThat(optionsMap.get("model")).isEqualTo(OllamaModel.MISTRAL.id());
+		assertThat(optionsMap.get("model")).isEqualTo(OllamaModel.MINISTRAL_3_8B.id());
 
-		assertThat(options.getModel()).isEqualTo(OllamaModel.MISTRAL.id());
+		assertThat(options.getModel()).isEqualTo(OllamaModel.MINISTRAL_3_8B.id());
 
 		// Verify all getters return null
 		assertThat(options.getTemperature()).isNull();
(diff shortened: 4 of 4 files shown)


kuntal1461:
## Code Review

The motivation here is correct — the `mistral` model being unavailable on Ollama's registry was breaking integration tests and this needs to be fixed. The new Ministral 3 additions are the right direction. However, there are two critical issues that need to be addressed before this can merge.

---

### 🔴 Critical: Hard removal of public enum constants without deprecation

`MISTRAL` and `MISTRAL_NEMO` are removed outright, which breaks compilation for any downstream code referencing `OllamaModel.MISTRAL` or `OllamaModel.MISTRAL_NEMO`.

The project's own precedent in `MistralAiApi.ChatModel` is clear:

```java
@Deprecated(forRemoval = true) // Retirement planned the 31st of July 2026
MAGISTRAL_MEDIUM("magistral-medium-latest"),
```

The new constants should be **added alongside** the old ones (marked `@Deprecated(forRemoval = true)`) in this PR, and the hard removal should happen in a subsequent release.

---

### 🔴 Critical: Silent behavioral change to the default model

`OllamaChatOptions.java` line 96:

```java
// Before
this.model = model != null ? model : OllamaModel.MISTRAL.id();

// After
this.model = model != null ? model : OllamaModel.MINISTRAL_3_8B.id();
```

This changes the fallback model for every application that builds `OllamaChatOptions` without setting a model explicitly. Users who relied on the default will silently receive responses from a different model with no warning and no migration path.

The root cause of the failing tests is in the integration test setup, not in the production default. The correct fix is to update `OllamaChatModelIT` to set an explicit model, leaving the production fallback unchanged.

---

### 🟡 Major: Naming convention inconsistency

The existing Qwen3 constants in the same file use `QWEN3_7B` (no underscore between family name and generation number). The new constants use `MINISTRAL_3_8B`, which is inconsistent. Consistent naming would be `MINISTRAL3_3B`, `MINISTRAL3_8B`, `MINISTRAL3_14B` to match the established pattern.

---

### 🟡 Minor: Removed test coverage without replacement

In `OllamaChatOptionsTests.testBasicOptions()`, two builder lines were removed:

```java
// Removed
var b1 = OllamaChatOptions.builder().model("model").mainGPU(12);
var b  = OllamaChatOptions.builder().mainGPU(12).model("model");
```

These were the only lines exercising `mainGPU()` and verifying that the fluent builder accepts method calls in any order. If they were considered dead code, they should be replaced with explicit assertions rather than silently deleted:

```java
assertThat(OllamaChatOptions.builder().model("model").mainGPU(12).build()).isNotNull();
assertThat(OllamaChatOptions.builder().mainGPU(12).model("model").build()).isNotNull();
```

---

### 🟢 Good changes

- `@EnumSource` over `@ValueSource(strings = {...})` in the parameterized test is a genuine improvement — it catches invalid enum names at compile time rather than at runtime via `OllamaModel.valueOf()`.
- Adding `private` to the `@Mock` field is correct Java style.
- The `<p>` tag fix in javadoc is correct HTML formatting.
- Author attribution (`@author Nicolas Krier`) follows project contribution guidelines.

---

### Suggested path forward

1. Keep `MISTRAL` and `MISTRAL_NEMO`, annotate with `@Deprecated(forRemoval = true)`, add the new constants alongside them.
2. Revert the default model change in `OllamaChatOptions`; update `OllamaChatModelIT` directly to use a non-deprecated model.
3. Rename `MINISTRAL_3_8B` → `MINISTRAL3_8B` etc. to match `QWEN3_7B`.
4. Restore or assert on the removed builder lines.

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
```diff
@@ -115,15 +115,19 @@ public enum OllamaModel implements ChatModelDescription {
 	LLAMA3_2_3B("llama3.2:3b"),
 
 	/**
-	 * The 7B parameters model
+	 * The Ministral 3 3B language model from Mistral AI.
 	 */
-	MISTRAL("mistral"),
+	MINISTRAL_3_3B("ministral-3:3b"),
```
kuntal1461, line 127: 🔴 **Critical — hard removal without deprecation cycle**

`MISTRAL` is deleted outright. Any downstream code referencing `OllamaModel.MISTRAL` will **fail to compile**.

The project's own precedent in `MistralAiApi.ChatModel` uses `@Deprecated(forRemoval = true)` before removal:

```java
@Deprecated(forRemoval = true) // Retirement planned the 31st of July 2026
MAGISTRAL_MEDIUM("magistral-medium-latest"),
```

Please keep `MISTRAL("mistral")` in this PR, mark it `@Deprecated(forRemoval = true)`, and add `MINISTRAL_3_3B` alongside it. Hard removal can follow in a later release.

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
```diff
@@ -115,15 +115,19 @@ public enum OllamaModel implements ChatModelDescription {
...
 
 	/**
-	 * A 12B model with 128k context length, built by Mistral AI in collaboration with
-	 * NVIDIA.
+	 * The Ministral 3 8B language model from Mistral AI.
 	 */
-	MISTRAL_NEMO("mistral-nemo"),
+	MINISTRAL_3_8B("ministral-3:8b"),
+
```
kuntal1461, line 132: 🔴 **Critical — same issue for `MISTRAL_NEMO`**

Same as above: `MISTRAL_NEMO` should be kept and annotated `@Deprecated(forRemoval = true)` rather than removed outright.

---

🟡 **Major — naming convention inconsistency**

The existing Qwen3 constants use `QWEN3_7B` (no underscore between family name and generation number). This constant introduces `MINISTRAL_3_8B` which breaks that pattern:

| Existing | New (this PR) | Consistent would be |
|----------|--------------|---------------------|
| `QWEN3_7B` | `MINISTRAL_3_8B` | `MINISTRAL3_8B` |
| `QWEN3_4B` | `MINISTRAL_3_3B` | `MINISTRAL3_3B` |

Please rename to `MINISTRAL3_3B`, `MINISTRAL3_8B`, `MINISTRAL3_14B`.

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
```diff
@@ -115,15 +115,19 @@ public enum OllamaModel implements ChatModelDescription {
...
+	/**
+	 * The Ministral 3 14B language model from Mistral AI.
+	 */
```
kuntal1461, line 135: 🟡 **Minor — sparse Javadoc**

The old `MISTRAL_NEMO` entry documented meaningful facts: *"A 12B model with 128k context length, built by Mistral AI in collaboration with NVIDIA."*

The new entries only say *"The Ministral 3 14B language model from Mistral AI."* — missing: context window size, key capabilities, and an Ollama registry link. Suggestion:

```java
/**
 * The Ministral 3 14B language model from Mistral AI.
 * Supports a 128k context window.
 * @see <a href="https://ollama.com/library/ministral-3">Ollama: ministral-3</a>
 */
```

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaChatOptions.java
```diff
@@ -93,7 +93,7 @@ protected OllamaChatOptions(@Nullable Boolean useNUMA, @Nullable Integer numCtx,
 		this.mirostatEta = mirostatEta;
 		this.penalizeNewline = penalizeNewline;
 		this.stop = stop != null ? List.copyOf(stop) : null;
-		this.model = model != null ? model : OllamaModel.MISTRAL.id();
+		this.model = model != null ? model : OllamaModel.MINISTRAL_3_8B.id();
 		this.format = format;
```
kuntal1461, line 97: 🔴 **Critical — silent behavioral change to the production default**

This changes the fallback model for every application that builds `OllamaChatOptions` without an explicit model. Users relying on the Mistral 7B default will silently receive responses from a completely different model — with no warning and no migration path.

The root cause (failing integration tests) is in the **test setup**, not the production default. The correct fix is to update `OllamaChatModelIT` to specify the new model explicitly, leaving this fallback unchanged.

If a default *must* change here, it also needs a migration note in the release notes and ideally a `@since` comment.

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatModelTests.java
```diff
@@ -51,19 +51,21 @@
...
  * @author Sebastien Deleuze
+ * @author Nicolas Krier
  * @since 1.0.0
  */
 @ExtendWith(MockitoExtension.class)
 class OllamaChatModelTests {
 
 	@Mock
-	OllamaApi ollamaApi;
+	private OllamaApi ollamaApi;
```
kuntal1461, line 68: 🟢 **Good** — adding `private` to the `@Mock` field is correct Java style for test classes.

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatModelTests.java
```diff
@@ -236,9 +238,8 @@ void buildChatResponseMetadataAggregationWithNullPrevious() {
 	}
 
 	@ParameterizedTest
-	@ValueSource(strings = { "LLAMA2", "MISTRAL", "CODELLAMA", "LLAMA3", "GEMMA" })
-	void buildOllamaChatModelWithDifferentModels(String modelName) {
-		OllamaModel model = OllamaModel.valueOf(modelName);
+	@EnumSource(value = OllamaModel.class, names = { "LLAMA2", "MINISTRAL_3_8B", "CODELLAMA", "LLAMA3", "GEMMA" })
```
kuntal1461, line 247: 🟢 **Good improvement** — replacing `@ValueSource(strings = {...})` + `OllamaModel.valueOf(modelName)` with `@EnumSource` gives compile-time safety. Typos in enum names now fail at compile time instead of at runtime.

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/OllamaChatModelTests.java
```diff
@@ -308,7 +309,10 @@ void buildChatResponseMetadataAggregationOverflowHandling() {
 	@Test
 	void buildOllamaChatModelImmutability() {
 		// Test that the builder creates immutable instances
-		OllamaChatOptions options = OllamaChatOptions.builder().model(OllamaModel.MISTRAL).temperature(0.5).build();
+		OllamaChatOptions options = OllamaChatOptions.builder()
+			.model(OllamaModel.MINISTRAL_3_14B)
```
kuntal1461, line 319: 🟡 **Minor — inconsistent model size between tests**

This test uses `MINISTRAL_3_14B` while `buildOllamaChatModelWithConstructor` uses `MINISTRAL_3_8B`. Since both tests only verify construction (not model behavior), there is no clear rationale for using two different sizes. Consider standardizing to one constant, or add a comment explaining the intent.

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaChatOptionsTests.java
```diff
@@ -49,10 +49,6 @@ protected Builder readyToBuildBuilder() {
 
 	@Test
 	void testBasicOptions() {
-		var b1 = OllamaChatOptions.builder().model("model").mainGPU(12);
-
```
kuntal1461, line 53: 🟡 **Minor — removed test coverage without replacement**

These two lines were the only ones exercising `mainGPU()` and verifying that the fluent builder accepts method calls in any order without throwing. Even as passive smoke-tests, removing them silently reduces coverage of the builder API.

Suggestion — convert to explicit assertions instead of deleting:
```java
assertThat(OllamaChatOptions.builder().model("model").mainGPU(12).build()).isNotNull();
assertThat(OllamaChatOptions.builder().mainGPU(12).model("model").build()).isNotNull();
```

models/spring-ai-ollama/src/test/java/org/springframework/ai/ollama/api/OllamaChatOptionsTests.java
```diff
@@ -272,9 +268,10 @@ void testZeroValuesIncludedInMap() {
```
kuntal1461, line 271: 🟢 **Good** — correct HTML Javadoc paragraph formatting. Nice catch.

kuntal1461 (changes requested):
The motivation is correct — the `mistral` model being unavailable on Ollama's registry breaks integration tests and the new Ministral 3 family is the right replacement. However, there are **two critical API-breaking issues** that must be addressed before merge, plus a naming convention mismatch. Inline comments below point to the exact lines.

nicolaskrier:
Hi @ilayaperumalg, could you have a look at this PR? I have tried to fix Ollama's integration tests failures by using Mistral AI's recent models.

https://github.com/spring-projects/spring-ai-integration-tests/actions/runs/27118143948/job/80029169371

models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/api/OllamaModel.java
nicolaskrier: Maybe we should keep Mistral Nemo for now and only deprecate it.
There are Qwen models following the same naming convention than the one I used for Ministral models.

kuntal1461: Thanks for checking in on the naming! You're right — `QWEN_2_5_3B` and `QWEN_3_1_7_B` do follow the same underscore-separated pattern as `MINISTRAL_3_8B`, so I withdraw that part of the comment. The existing naming is already mixed in the enum (`QWEN3_7B` vs `QWEN_2_5_3B`), so I'll defer to the maintainers on whichever they prefer .

On deprecation: glad we agree on keeping `MISTRAL_NEMO`. The same applies to `MISTRAL` as well — the PR diff hard-removes both. Please keep `MISTRAL("mistral")` too and annotate it with `@Deprecated(forRemoval = true)`, not just `MISTRAL_NEMO`.

nicolaskrier:
Hi @sdeleuze, could you have a look at this PR? It is an attempt to fix failing integration tests for Ollama. I have removed only the model that isn't available anymore.

PR #6326, opened 2026-06-07
GH-2588 Add name attribite to @ToolParam

### LINKED ISSUE #2588: @ToolParam  need a name Attribute !!!
Why doesn't the @ToolParam have a name attribute? If the actual names of method parameters are lost after compilation, it will result in MCP being unable to correctly assign values to the parameters during invocation.

Grandys:
Fixes #2588
Adds explicit parameter name support for `@ToolParam` and `@McpToolParam`, allowing stable tool schema property names and argument binding even when Java parameter metadata is unavailable.

Also adds duplicate effective-name validation for tool method parameters, keeps schema generation and invocation lookup aligned, updates documentation, and covers core and MCP paths with tests.

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/McpToolParam.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/McpToolParam.java
@@ -31,6 +31,11 @@ import java.lang.annotation.Target;
 public @interface McpToolParam {
 
 	/**
+	 * The name of the tool argument.
+	 */
+	String name() default "";
+
+	/**
 	 * Whether the tool argument is required.
 	 */
 	boolean required() default true;
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/McpToolParam.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/McpToolParam.java
@@ -31,7 +31,9 @@ import java.lang.annotation.Target;
 public @interface McpToolParam {
 
 	/**
-	 * The name of the tool argument.
+	 * The name of the tool argument. Only effective on method parameters. For fields or
+	 * record components, use {@code @JsonProperty} or {@code @Schema} to customize schema
+	 * property names.
 	 */
 	String name() default "";
 
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
@@ -62,6 +62,8 @@ public abstract class AbstractMcpToolMethodCallback<T, RC extends McpRequestCont
 		this.toolMethod = toolMethod;
 		this.toolObject = toolObject;
 		this.returnMode = returnMode;
+		// MCP callbacks can be built independently from schema generation, so runtime
+		// argument binding must validate duplicate effective names independently.
 		McpToolMethodParameterUtils.validateUniqueParameterNames(toolMethod, this::isInfrastructureParameter);
 	}
 
@@ -155,7 +157,7 @@ public abstract class AbstractMcpToolMethodCallback<T, RC extends McpRequestCont
 	 * @param type The target type
 	 * @return The typed argument
 	 */
-	protected Object buildTypedArgument(@Nullable Object value, Type type) {
+	protected @Nullable Object buildTypedArgument(@Nullable Object value, Type type) {
 		if (value == null) {
 			return null;
 		}
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
@@ -18,6 +18,7 @@ package org.springframework.ai.mcp.annotation.method.tool;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
 import java.util.Map;
 import java.util.Objects;
@@ -61,6 +62,7 @@ public abstract class AbstractMcpToolMethodCallback<T, RC extends McpRequestCont
 		this.toolMethod = toolMethod;
 		this.toolObject = toolObject;
 		this.returnMode = returnMode;
+		McpToolMethodParameterUtils.validateUniqueParameterNames(toolMethod, this::isInfrastructureParameter);
 	}
 
 	/**
@@ -131,9 +133,20 @@ public abstract class AbstractMcpToolMethodCallback<T, RC extends McpRequestCont
 				return exchangeOrContext;
 			}
 
-			Object rawArgument = toolInputArguments.get(parameter.getName());
+			Object rawArgument = toolInputArguments.get(McpToolMethodParameterUtils.getParameterName(parameter));
 			return buildTypedArgument(rawArgument, parameter.getParameterizedType());
 		}).toArray();
+	}
+
+	private boolean isInfrastructureParameter(Parameter parameter) {
+		// Check if parameter is a common MCP infrastructure parameter
+		if (McpToolMethodParameterUtils.isInfrastructureParameter(parameter)) {
+			return true;
+		}
+
+		// Check if parameter is an injectable exchange or context type for this
+		// callback
+		return isExchangeOrContextType(parameter.getType());
 	}
 
 	/**
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/McpToolMethodParameterUtils.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/McpToolMethodParameterUtils.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.mcp.annotation.method.tool;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.function.Predicate;
+
+import io.modelcontextprotocol.common.McpTransportContext;
+import io.modelcontextprotocol.spec.McpSchema.CallToolRequest;
+
+import org.springframework.ai.mcp.annotation.McpMeta;
+import org.springframework.ai.mcp.annotation.McpProgressToken;
+import org.springframework.ai.mcp.annotation.McpToolParam;
+import org.springframework.ai.mcp.annotation.context.McpAsyncRequestContext;
+import org.springframework.ai.mcp.annotation.context.McpSyncRequestContext;
+import org.springframework.util.StringUtils;
+
+/**
+ * Utility methods for handling MCP tool method parameters.
+ * <p>
+ * This class contains only the common parameter handling shared by MCP tool schema
+ * generation and method invocation. Caller-specific exchange/context parameter handling
+ * remains in the caller.
+ *
+ * @author Michał Grandys
+ */
+public final class McpToolMethodParameterUtils {
+
+	private McpToolMethodParameterUtils() {
+	}
+
+	public static boolean isInfrastructureParameter(Parameter parameter) {
+		// Check if parameter is a request context type
+		if (McpSyncRequestContext.class.isAssignableFrom(parameter.getType())
+				|| McpAsyncRequestContext.class.isAssignableFrom(parameter.getType())) {
+			return true;
+		}
+
+		// Check if parameter is annotated with @McpProgressToken
+		if (parameter.isAnnotationPresent(McpProgressToken.class)) {
+			return true;
+		}
+
+		// Check if parameter is McpMeta type
+		if (McpMeta.class.isAssignableFrom(parameter.getType())) {
+			return true;
+		}
+
+		// Check if parameter is CallToolRequest type
+		if (CallToolRequest.class.isAssignableFrom(parameter.getType())) {
+			return true;
+		}
+
+		// Check if parameter is McpTransportContext type
+		return McpTransportContext.class.isAssignableFrom(parameter.getType());
+	}
+
+	public static String getParameterName(Parameter parameter) {
+		McpToolParam toolParamAnnotation = parameter.getAnnotation(McpToolParam.class);
+		if (toolParamAnnotation != null && StringUtils.hasText(toolParamAnnotation.name())) {
+			return toolParamAnnotation.name();
+		}
+		return parameter.getName();
+	}
+
+	public static void validateUniqueParameterNames(Method method,
+			Predicate<Parameter> infrastructureParameterPredicate) {
+		Set<String> parameterNames = new HashSet<>();
+		for (Parameter parameter : method.getParameters()) {
+			if (infrastructureParameterPredicate.test(parameter)) {
+				continue;
+			}
+			String parameterName = getParameterName(parameter);
+			if (!parameterNames.add(parameterName)) {
+				throw new IllegalArgumentException(
+						"Duplicate tool parameter name '" + parameterName + "' in method " + method);
+			}
+		}
+	}
+
+}
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/McpToolMethodParameterUtils.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/McpToolMethodParameterUtils.java
@@ -18,8 +18,6 @@ package org.springframework.ai.mcp.annotation.method.tool;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
-import java.util.HashSet;
-import java.util.Set;
 import java.util.function.Predicate;
 
 import io.modelcontextprotocol.common.McpTransportContext;
@@ -30,6 +28,7 @@ import org.springframework.ai.mcp.annotation.McpProgressToken;
 import org.springframework.ai.mcp.annotation.McpToolParam;
 import org.springframework.ai.mcp.annotation.context.McpAsyncRequestContext;
 import org.springframework.ai.mcp.annotation.context.McpSyncRequestContext;
+import org.springframework.ai.tool.support.ToolMethodParameterUtils;
 import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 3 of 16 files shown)


kuntal1461 (commented):
## Code Review — GH-2588 Add `name` attribute to `@ToolParam`

The core feature is implemented correctly and the approach is sound. Schema generation and invocation both use the same `getParameterName()` call so the LLM-visible name and the binding key stay aligned. Backward compatibility is preserved via the `""` default. A few issues worth addressing before merge:

---

### Critical

**1. `name()` is silently ignored when `@ToolParam` is used on a field or annotation type**

`@ToolParam` targets `PARAMETER`, `FIELD`, and `ANNOTATION_TYPE`, but `ToolMethodParameterUtils.getParameterName()` only reads `name()` for method parameters. A developer who writes `@ToolParam(name = "foo")` on a record field gets no rename and no error — silent misbehaviour.

The `name()` Javadoc should be explicit:

```java
/**
 * The name of the tool argument. Only effective on method parameters.
 * For fields or record components, use {@link com.fasterxml.jackson.annotation.JsonProperty}
 * or {@code @Schema} to customize schema property names.
 */
String name() default "";
```

Same change needed in `McpToolParam`.

---

**2. Duplicate validation fires twice in the normal tool-registration flow**

`MethodToolCallbackProvider.getToolCallbacks()` calls `ToolDefinitions.from(method)` → `JsonSchemaGenerator.generateForMethodInput()` which runs `validateUniqueParameterNames()` (call #1). Then `MethodToolCallback` is built and its constructor runs `validateUniqueParameterNames()` again on the same method (call #2). The same duplication exists in the MCP path (`AbstractMcpToolMethodCallback` constructor + `McpJsonSchemaGenerator`).

Not a correctness bug but a design smell — same check twice with no benefit. Please pick one location: either the constructor (fail-fast at object construction) or schema gen (covers standalone usage too), not both.

---

### Moderate

**3. `validateUniqueParameterNames` body is copy-pasted between the two utility classes**

`ToolMethodParameterUtils` and `McpToolMethodParameterUtils` have byte-for-byte identical loop implementations. Since `mcp-annotations` already depends on `spring-ai-model`, `McpToolMethodParameterUtils` should delegate:

```java
public static void validateUniqueParameterNames(Method method,
        Predicate<Parameter> infrastructureParameterPredicate) {
    ToolMethodParameterUtils.validateUniqueParameterNames(method, infrastructureParameterPredicate);
}
```

---

**4. `McpToolMethodParameterUtils` is not covered by `@NullMarked`**

The `tool/support` package has `@NullMarked` in its `package-info.java` so `ToolMethodParameterUtils` is protected. The `mcp/annotation/method/tool` package-info does not have `@NullMarked`, leaving `McpToolMethodParameterUtils` unguarded under JSpecify — inconsistent with the rest of the project.

---

**5. Documentation examples use `name` redundantly where Java names already match**

In both MCP adoc files the examples now show:

```java
@McpToolParam(name = "a", description = "First number", required = true) int a,
@McpToolParam(name = "b", description = "Second number", required = true) int b
```

The parameter is already named `a`/`b` in Java — `name = "a"` adds nothing and implies `name` is always required. These examples should either drop the `name` attribute (it is not needed here) or use a meaningful override (e.g., `name = "first_number"` on a parameter called `a`) to demonstrate when the attribute is actually useful.

---

**6. New public utility classes are missing `@since` tags**

`ToolMethodParameterUtils` and `McpToolMethodParameterUtils` have `@author` but no `@since 1.0.0`, unlike other public classes in the project.

---

### Minor

**7. `McpToolMethodParameterUtils.isInfrastructureParameter()` silently excludes server exchange types**

The method name implies completeness but it deliberately does not cover `McpSyncServerExchange`/`McpAsyncServerExchange` (callers handle those). The class Javadoc explains this at the class level but not on the method itself. A one-line Javadoc note on the method would prevent a future maintainer from trusting it as a complete check and silently introducing a bug.

---

**8. No invocation test for a simple named `String` parameter**

The only invocation test that exercises `@ToolParam(name)` binding uses `Map<String, Integer>` (`named_map`). A simple `@ToolParam(name = "customer_id") String customerId` invocation test directly matches the issue reporter's scenario and would round out coverage.

---

### What is done well

- Blank-name fallback via `StringUtils.hasText()` is correct — `""` and `" "` both fall back to the reflective name
- Fail-fast duplicate detection at construction time is the right place to catch programmer errors
- `name` placed first in the annotation — consistent with `@Tool` and `@McpTool` attribute ordering
- Both `spring-ai-model` and `mcp-annotations` paths updated symmetrically
- All new tests use AssertJ — matches project standard
- Fully backward compatible

kuntal1461 (commented):
Inline file-by-file review comments for GH-2588 / PR #6326. Each finding is pinned to the relevant line.

mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/McpToolParam.java
```diff
@@ -30,6 +30,11 @@
 @Documented
 public @interface McpToolParam {
 
+	/**
+	 * The name of the tool argument.
+	 */
+	String name() default "";
```
kuntal1461, line 38: **[Critical #1] `name()` is silently ignored on fields and annotation-type uses**

`@McpToolParam` targets `PARAMETER`, `FIELD`, and `ANNOTATION_TYPE`. But `McpToolMethodParameterUtils.getParameterName()` only reads `name()` for method parameters. A developer who writes `@McpToolParam(name = "foo")` on a field gets no rename and no error — silent misbehaviour.

Please update the Javadoc:
```java
/**
 * The name of the tool argument. Only effective on method parameters.
 * For fields or record components, use {@literal @}JsonProperty or {@literal @}Schema instead.
 */
String name() default "";
```

spring-ai-model/src/main/java/org/springframework/ai/tool/annotation/ToolParam.java
```diff
@@ -33,6 +33,11 @@
 @Documented
 public @interface ToolParam {
 
+	/**
+	 * The name of the tool argument.
+	 */
+	String name() default "";
```
kuntal1461, line 41: **[Critical #1] Same issue as in `McpToolParam` — `name()` is silently ignored on fields**

`@ToolParam` also targets `FIELD` and `ANNOTATION_TYPE`, but `ToolMethodParameterUtils.getParameterName()` only reads `name()` for method parameters. Using `@ToolParam(name = "foo")` on a record field does nothing — no rename, no warning.

Please update the Javadoc to say it is only effective on method parameters, and direct field/record users to `@JsonProperty` or `@Schema`.

spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallback.java
```diff
@@ -79,6 +81,7 @@ public MethodToolCallback(ToolDefinition toolDefinition, @Nullable ToolMetadata
 		this.toolObject = toolObject;
 		this.toolCallResultConverter = toolCallResultConverter != null ? toolCallResultConverter
 				: DEFAULT_RESULT_CONVERTER;
+		validateUniqueParameterNames(toolMethod);
```
kuntal1461, line 86: **[Critical #2] Duplicate validation — same check fires twice in normal tool-registration flow**

In the normal path `MethodToolCallbackProvider.getToolCallbacks()` → `ToolDefinitions.from(method)` → `JsonSchemaGenerator.generateForMethodInput()` already calls `validateUniqueParameterNames()`. Then this constructor fires it again on the same method.

Please pick one location — either the constructor (fail-fast at object construction) or `generateForMethodInput()` (covers standalone schema gen too) — not both. Having both with no rationale is a maintenance smell.

spring-ai-model/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java
```diff
@@ -140,18 +141,19 @@ public static String generateForMethodInput(Method method, SchemaOption... schem
 
 		ObjectNode properties = schema.putObject("properties");
 		List<String> required = new ArrayList<>();
+		ToolMethodParameterUtils.validateUniqueParameterNames(method, JsonSchemaGenerator::isInfrastructureParameter);
```
kuntal1461, line 144: **[Critical #2 continued] Second call-site for the duplicated validation**

This is the other half of the double-validation issue flagged at `MethodToolCallback` line 84. When the full tool-registration path runs, this call and the constructor call both execute on the same method. Choose one authoritative location.

mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/McpToolMethodParameterUtils.java
```diff
@@ -0,0 +1,98 @@
...
+
+	public static String getParameterName(Parameter parameter) {
+		McpToolParam toolParamAnnotation = parameter.getAnnotation(McpToolParam.class);
+		if (toolParamAnnotation != null && StringUtils.hasText(toolParamAnnotation.name())) {
+			return toolParamAnnotation.name();
+		}
+		return parameter.getName();
+	}
+
+	public static void validateUniqueParameterNames(Method method,
```
kuntal1461: **[Moderate #3] `validateUniqueParameterNames` is copy-pasted from `ToolMethodParameterUtils`**

The loop body here is byte-for-byte identical to `ToolMethodParameterUtils.validateUniqueParameterNames()`. Since `mcp-annotations` already depends on `spring-ai-model`, this method should simply delegate instead of reimplementing:

```java
public static void validateUniqueParameterNames(Method method,
        Predicate<Parameter> infrastructureParameterPredicate) {
    ToolMethodParameterUtils.validateUniqueParameterNames(method, infrastructureParameterPredicate);
}
```

mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/McpToolMethodParameterUtils.java
```diff
@@ -0,0 +1,98 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * Utility methods for handling MCP tool method parameters.
+ * <p>
+ * This class contains only the common parameter handling shared by MCP tool schema
+ * generation and method invocation. Caller-specific exchange/context parameter handling
+ * remains in the caller.
+ *
+ * @author Michał Grandys
```
kuntal1461, line 41: **[Moderate #4 & #6] Missing `@NullMarked` coverage and missing `@since` tag**

Two issues here:
1. The `tool/support` package has `@NullMarked` in its `package-info.java`, so `ToolMethodParameterUtils` is protected. The `mcp/annotation/method/tool` `package-info.java` does **not** have `@NullMarked`, leaving this class unguarded under JSpecify. Please add `@NullMarked` to the package-info or to this class directly.
2. Other public classes in the project carry `@since 1.0.0`. Please add it here.

mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/McpToolMethodParameterUtils.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+public final class McpToolMethodParameterUtils {
+
+	private McpToolMethodParameterUtils() {
+	}
+
+	public static boolean isInfrastructureParameter(Parameter parameter) {
```
kuntal1461, line 53: **[Minor #7] Method name `isInfrastructureParameter` overpromises — silently excludes server exchange types**

This method deliberately does **not** cover `McpSyncServerExchange` or `McpAsyncServerExchange` (those are handled by callers). The class-level Javadoc explains the design, but this method has no documentation. A future maintainer will assume the method is complete and may introduce a subtle bug by relying on it exclusively.

Please add a brief note:
```java
/**
 * Returns true for common MCP infrastructure parameters (request contexts, progress tokens,
 * McpMeta, CallToolRequest, McpTransportContext).
 * Does NOT include server exchange types (McpSyncServerExchange, McpAsyncServerExchange);
 * callers must handle those separately.
 */
```

mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
```diff
@@ -61,6 +62,7 @@ protected AbstractMcpToolMethodCallback(ReturnMode returnMode, Method toolMethod
 		this.toolMethod = toolMethod;
 		this.toolObject = toolObject;
 		this.returnMode = returnMode;
+		McpToolMethodParameterUtils.validateUniqueParameterNames(toolMethod, this::isInfrastructureParameter);
```
kuntal1461, line 67: **[Critical #2 — MCP path] Same double-validation issue as in `MethodToolCallback`**

`McpJsonSchemaGenerator.internalGenerateFromMethodArguments()` already calls `validateUniqueParameterNames()` during schema generation. This constructor call duplicates that check on the same method. Pick one authoritative location.

kuntal1461 (commented):
Moderate and minor issues — continued from the critical issues review above.

spring-ai-model/src/main/java/org/springframework/ai/tool/support/ToolMethodParameterUtils.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Set;
+import java.util.function.Predicate;
+
+import org.springframework.ai.tool.annotation.ToolParam;
+import org.springframework.util.StringUtils;
+
+/**
+ * Utility methods for handling tool method parameters.
+ *
+ * @author Michał Grandys
```
kuntal1461, line 32: **[Moderate #6] Missing `@since` tag**

Other public classes in the project carry `@since 1.0.0`. Please add it to match project convention:
```java
 * @author Michał Grandys
 * @since 1.0.0
```

spring-ai-docs/src/main/antora/modules/ROOT/pages/api/mcp/mcp-annotations-overview.adoc
```diff
@@ -104,20 +104,22 @@ public class CalculatorTools {
 
     @McpTool(name = "add", description = "Add two numbers together")
     public int add(
-            @McpToolParam(description = "First number", required = true) int a,
-            @McpToolParam(description = "Second number", required = true) int b) {
+            @McpToolParam(name = "a", description = "First number", required = true) int a,
```
kuntal1461: **[Moderate #5] Redundant `name` attribute in documentation examples**

The Java parameters are already named `a`, `b`, `x`, `y` — adding `name = "a"` when the name already matches adds nothing and misleads readers into thinking `name` is always required.

The whole point of this feature is to override names when they **differ** (e.g., a parameter named `a` in Java that should appear as `first_number` in the schema). Please either:
- Remove `name` from these examples (it is not needed here), or
- Use a realistic override — e.g., rename `int a` → `@McpToolParam(name = "first_number") int a` — to demonstrate the actual value of the attribute.

spring-ai-docs/src/main/antora/modules/ROOT/pages/api/mcp/mcp-annotations-server.adoc
```diff
@@ -18,13 +18,16 @@ public class CalculatorTools {
 
     @McpTool(name = "add", description = "Add two numbers together")
     public int add(
-            @McpToolParam(description = "First number", required = true) int a,
-            @McpToolParam(description = "Second number", required = true) int b) {
+            @McpToolParam(name = "a", description = "First number", required = true) int a,
```
kuntal1461: **[Moderate #5 continued] Same redundant `name` issue in server adoc**

Same feedback as `mcp-annotations-overview.adoc`: `name = "a"` and `name = "b"` are redundant because the Java parameter names already match. This example teaches the wrong lesson. Show a case where the override is actually needed.

kuntal1461 (commented):
Remaining issues: missing @since, redundant name in docs, and missing invocation test.

spring-ai-model/src/test/java/org/springframework/ai/tool/method/MethodToolCallbackGenericTypesTest.java
```diff
@@ -90,7 +92,7 @@ void testGenericMapType() throws Exception {
 		// Create a JSON input with a map of string to integer
```
kuntal1461, line 92: **[Minor #8] No invocation test for a simple named `String` parameter**

This is the only invocation test that exercises `@ToolParam(name)` argument binding, and it uses a `Map<String, Integer>` (a generic type). While this works, the canonical use case from the issue is a simple type like `String`. A companion test with `@ToolParam(name = "customer_id") String customerId` called via `{"customer_id": "abc"}` would directly cover the reporter's scenario and round out the test suite.

spring-ai-model/src/main/java/org/springframework/ai/tool/annotation/ToolParam.java
Grandys: I added comment with an explanation, that parameter is effective only on method parameters.

spring-ai-model/src/main/java/org/springframework/ai/tool/method/MethodToolCallback.java
Grandys: This was intentional, both validation paths are separate entry points, depending if schema generation was standalone or caller provided schema (https://github.com/spring-projects/spring-ai/blob/95d99e666398d5ffff137fbbaab43e04787e8eda/spring-ai-model/src/main/java/org/springframework/ai/tool/definition/DefaultToolDefinition.java#L64). I added comments to clarify why the constructor validation is duplicated.

spring-ai-model/src/main/java/org/springframework/ai/tool/support/ToolMethodParameterUtils.java
Grandys: Fixed, added `@since 2.0.0`

spring-ai-model/src/main/java/org/springframework/ai/util/json/schema/JsonSchemaGenerator.java
Grandys: As I explained, it's intentional. Removing either check would leave one of those paths able to accept
  an invalid method

spring-ai-model/src/test/java/org/springframework/ai/tool/method/MethodToolCallbackGenericTypesTest.java
Grandys: I added new test `testExplicitToolParamNameForSimpleString`

Grandys:
@kuntal1461, thank you for your review! I addressed all your comments.

mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/McpToolParam.java
Grandys: I added comment with an explanation, that parameter is effective only on method parameters.

mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/McpToolMethodParameterUtils.java
Grandys: I added `@NullMarked` to the package info, and I added missing `@since` in javadoc

mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/McpToolMethodParameterUtils.java
Grandys: Missing coverage is now documented

mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/McpToolMethodParameterUtils.java
Grandys: Addressed, now validation is delegated to the `ToolMethodParameterUtils`

spring-ai-docs/src/main/antora/modules/ROOT/pages/api/mcp/mcp-annotations-overview.adoc
Grandys: Documentation was updated

kuntal1461 (approved):
All 8 issues from the previous review have been addressed.

On the duplicate `validateUniqueParameterNames` call (Critical #2): after reviewing the code, the explanation is correct. `MethodToolCallback` and `AbstractMcpToolMethodCallback` can be constructed with a caller-provided schema that bypasses `JsonSchemaGenerator`, so the constructor validation is genuinely needed as an independent check. The added comments make this clear.

The `@since 2.0.0` tag is also correct — the project is on `2.0.0-SNAPSHOT`.

Good work addressing everything thoroughly.

@tzolov @ericbottard @sdeleuze — ready for final review and merge.

PR #6349, opened 2026-06-09
Always auto-register `ToolCallingAdvisor` to support runtime-injected tools

### LINKED ISSUE #6325: DefaultChatClient#autoRegisterToolCallingAdvisor fails to register the ToolCallingAdvisor for dynamically injected tools
Currently  the`DefaultChatClient#autoRegisterToolCallingAdvisor()` will register the `ToolCallingAdvisor` only if there are existing ToolCallbacks defined by the time the `.call()`/`.stream()` is invoked.  

There are use cases where tool can be defined dynamically at run-time by some advisor.  If there are no other explicitly defined tools then the `ToolCallingAdvisor` won't be provisioned and there will be no support for the dynamically registered tools.

To prevent this we should consider always auto-registering the `ToolCallingAdvisor` unless the `spring.ai.chat.client.tool-calling.enabled=false` or (`.advisors(AdvisorParams.toolCallingAdvisorAutoRegister(false))`) are set.

WORKAROUND: At the moment, the workaround is to register the  `ToolCallingAdvisor` explicitly or ensure that there is at least one (`dummy`) tool that will trigger the auto-registration.

tzolov:
- ToolCallingAdvisor is now always auto-registered (unless explicitly disabled via AdvisorParams.toolCallingAdvisorAutoRegister(false) or the spring.ai.chat.client.tool-calling.enabled=false property).
- When the prompt options do not implement ToolCallingChatOptions the advisor passes through gracefully instead of throwing IllegalArgumentException.
- Update related tests and docs
- Add auto-registration ITs for google and anthropic models

Closes #6325

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/AdvisorParams.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/AdvisorParams.java
@@ -38,10 +38,11 @@ public final class AdvisorParams {
 	/**
 	 * Controls whether a
 	 * {@link org.springframework.ai.chat.client.advisor.ToolCallingAdvisor} is
-	 * automatically added to the chain when tools are configured on the
-	 * {@code ChatClient} and no explicit
-	 * {@link org.springframework.ai.chat.client.advisor.api.ToolAdvisor} is already
-	 * present. Auto-registration is enabled by default; pass {@code false} to opt out:
+	 * automatically added to the chain. Auto-registration is enabled by default so that
+	 * tools injected at runtime by another advisor are handled correctly even when no
+	 * static tools are configured. No explicit
+	 * {@link org.springframework.ai.chat.client.advisor.api.ToolAdvisor} must be present.
+	 * Pass {@code false} to opt out:
 	 *
 	 * <pre>{@code
 	 * client.prompt()
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ToolCallingAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ToolCallingAdvisor.java
@@ -51,6 +51,11 @@ import org.springframework.util.Assert;
  * <p>
  * This enables intercepting the tool calling loop by the rest of the advisors next in the
  * chain.
+ * <p>
+ * When the prompt options do not implement
+ * {@link org.springframework.ai.model.tool.ToolCallingChatOptions} (e.g. for models that
+ * don't support tool calling), this advisor passes the request through to the next
+ * advisor without modification.
  *
  * @author Christian Tzolov
  * @since 2.0.0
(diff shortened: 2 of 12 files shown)


spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc
```diff
@@ -62,7 +62,7 @@ Code that sets `.internalToolExecutionEnabled(true)` (explicitly enabling the in
 
 Remove all calls to `.internalToolExecutionEnabled(...)`. The recommended approaches are:
 
-* **`ToolCallingAdvisor` via `ChatClient`** (recommended) — auto-registered when tools are present; no flag needed.
```
sdeleuze, line 65: Those `upgrade-notes.adoc` changes should likely be reverted and moved to a proper "Upgrading to 2.0.0-RC2" section.

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/AdvisorParams.java
```diff
@@ -38,10 +38,11 @@ private AdvisorParams() {
 	/**
 	 * Controls whether a
 	 * {@link org.springframework.ai.chat.client.advisor.ToolCallingAdvisor} is
-	 * automatically added to the chain when tools are configured on the
-	 * {@code ChatClient} and no explicit
-	 * {@link org.springframework.ai.chat.client.advisor.api.ToolAdvisor} is already
-	 * present. Auto-registration is enabled by default; pass {@code false} to opt out:
+	 * automatically added to the chain. Auto-registration is always enabled by default so
```
sdeleuze: "Auto-registration is always enabled by default" should be replaced by either "Auto-registration is enabled by default" or "Auto-registration is always enabled".

spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ToolCallingAdvisor.java
```diff
@@ -228,16 +233,15 @@ public Flux<ChatClientResponse> adviseStream(ChatClientRequest chatClientRequest
...
+		if (!(options instanceof ToolCallingChatOptions toolCallingChatOptions)) {
+			// No tool calling options - skip the tool calling advisor
+			return streamAdvisorChain.nextStream(chatClientRequest);
 		}
 
 		ChatClientRequest initializedRequest = this.doInitializeLoopStream(chatClientRequest, streamAdvisorChain);
 
-		var optionsCopy = ((ToolCallingChatOptions.Builder<?>) chatClientRequest.prompt().getOptions().mutate())
-			.build();
+		var optionsCopy = toolCallingChatOptions.mutate().build();
```
sdeleuze: options are now immutable so creating a copy with `toolCallingChatOptions.mutate().build()` is likely not needed.

sdeleuze (changes requested):

PR #6365, opened 2026-06-09
Preserve OpenAI tool call additional properties

### LINKED ISSUE #6364: `OpenAiChatModel` drops tool-call additional properties, causing Gemini `thought_signature` 400
## Bug description

When Gemini is used through its OpenAI-compatible Chat Completions endpoint, tool-calling fails on the follow-up request.

With Gemini 3.x and newer models, the OpenAI-compatible endpoint returns the required thought signature on the assistant tool call:

```text
tool_calls[].extra_content.google.thought_signature
```

Spring AI maps the response into `AssistantMessage.ToolCall`, but the root tool-call additional properties are not preserved. When the tool result is sent back, the assistant tool call is replayed without the thought signature and Gemini rejects the request.

This affects enterprise applications that use `OpenAiChatModel` with OpenAI-compatible endpoints instead of binding directly to each provider-native integration, so providers or models can be switched at runtime.

## Environment

- This happens on both Spring AI `1.x` and `2.x`.
- Provider endpoint: Gemini OpenAI-compatible Chat Completions endpoint
- Base URL: `https://generativelanguage.googleapis.com/v1beta/openai/`
- Model used in the manual test: `gemini-3.5-flash`

## Steps to reproduce

1. Configure `OpenAiChatModel` with the Gemini OpenAI-compatible endpoint.
2. Register a function tool.
3. Send a prompt that triggers the tool.
4. Run the Spring AI tool-calling flow.

The first model call returns a tool call, but the follow-up call after tool execution fails:

```text
com.openai.errors.BadRequestException: 400: [{"error":{"code":400,"message":"Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:get_current_weather` , position 2. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.","status":"INVALID_ARGUMENT"}}]
```

## Expected behavior

Spring AI should preserve root additional properties attached to assistant tool calls and replay them when converting `AssistantMessage.ToolCall` back into OpenAI SDK request objects.

ultramancode:
Fixes #6364 

`OpenAiChatModel` currently drops additional properties attached to assistant tool calls while converting OpenAI SDK responses into Spring AI `AssistantMessage.ToolCall`s.

That breaks Gemini OpenAI-compatible tool-calling flows for models that require provider metadata such as:

```text
tool_calls[].extra_content.google.thought_signature
```

to be sent back on the follow-up model call after tool execution. Without it, Gemini 3.x models can fail with:

```text
Function call is missing a thought_signature in functionCall parts.
```

The OpenAI Java SDK already keeps these provider fields in `_additionalProperties()`. This PR carries the root tool-call additional properties through `AssistantMessage` metadata and restores them on the OpenAI SDK request object when conversation history is replayed.

I also checked several Gemini OpenAI-compatible response shapes, including `gemini-2.5-flash-lite` and `gemini-3.5-flash`, covering responses with and without `thought_signature`. When no additional properties are present, nothing extra is replayed.

There is no Gemini-specific handling here.
Spring AI only replays additional fields that were already present in the provider response.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -117,12 +117,15 @@ import org.springframework.util.StringUtils;
  * @author Ilayaperumal Gopinathan
  * @author Thomas Vitale
  * @author Eric Bottard
+ * @author Taewoong Kim
  */
 public final class OpenAiChatModel implements ChatModel {
 
 	private static final ChatModelObservationConvention DEFAULT_OBSERVATION_CONVENTION = new DefaultChatModelObservationConvention();
 
 	private static final String REASONING_CONTENT = "reasoningContent";
+
+	private static final String TOOL_CALL_ADDITIONAL_PROPERTIES_METADATA_KEY = "openai.tool_calls.additional_properties";
 
 	private final Log logger = LogFactory.getLog(OpenAiChatModel.class);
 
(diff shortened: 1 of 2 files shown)


sdeleuze:
@ultramancode Please add a related integration test that fails when the fix is not present.

tzolov:
@ultramancode , can you point us to where the OpenAI API spec defines additionalProperties: https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create ?
If it is not part of the OpenAI Spec then it is an issue with the Google OpenAI support i guess not with Spring AI

ultramancode:
I replied with more context in #6364.

If this still seems like the right direction, I can add the requested integration test.

tzolov:
Thanks @ultramancode , I've made some optimizations and merged it

tzolov:
rebased, optimized, squashed and merged at 798d4332fab638e288d8b8585801df639190d6e5

ultramancode:
Thanks for taking care of this!

PR #6366, opened 2026-06-09
Make buildRequestPrompt as default method in ChatModel

guanxuc:


ericbottard:
Hi @guanxuc,

we just merged a change that goes exactly in the opposite direction for the following reason: even though there is "opportunity" for factorisation, `buildRequestPrompt()` (which started as a complicated method when merging was handled at the model level) is now a small implementation detail (which could even be inlined if it wasn't shared by `call()` and `stream()`).
As such, it should NOT be pulled up in the super interface, as that makes it part of the public API, which we don't want.

As a consequence, I am going to close this PR without merging. I hope the explanations are clear.

guanxuc:
Hi @ericbottard, Thanks for your explanations!

PR #6370, opened 2026-06-09
Fix BedrockProxyChatModel model option handling

ilayaperumalg:
- Chat model to fall back to default model when not set in prompt options


models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
```diff
@@ -446,8 +446,9 @@ else if (message.getMessageType() == MessageType.TOOL) {
 		Map<String, String> requestMetadata = ConverseApiUtils
 			.getRequestMetadata(prompt.getUserMessage().getMetadata());
 
+		String modelId = options.getModel() != null ? options.getModel() : this.options.getModel();
```
ilayaperumalg, line 449: After discussion with @tzolov, we agreed to handle this with merging of options at the tests. Automatic model options merging should only happen at the Chat client level. Otherwise, the merge needs to be handled manually when using the chat model alone which is what happening at this case.

ilayaperumalg:
Closing this PR to address this separately.

Issue #5898, opened 2026-04-28
Reasoning Content Missing Error in Stream API Calls with Tool Usage


ningmd:
When calling the model via streaming API with both deep thinking enabled and tool calling functionality, the API returns an error indicating that reasoning content is missing in assistant tool call messages.

{"error":{"message":"thinking is enabled but reasoning_content is missing in assistant tool call message at index 4","type":"invalid_request_error"}}

i-zoufeng:
i can fix this

jewoodev:
Hi @ningmd, I traced the 400 to where a streaming turn loses `reasoning_content` — the streaming sibling of the non-streaming reasoning round-trip in #6026.

In the OpenAI-compatible path (`OpenAiChatModel`), the round-trip breaks during chunk merge, before aggregation:

- `ChunkMerger.chunkToChatCompletion` rebuilds the message from `content`/`refusal`/`toolCalls` only, and `mergeDeltas` merges tool calls only, so the `reasoning_content` carried on each `delta` is never copied onto the merged message — the chunk-level `putAllAdditionalProperties` lands on the `ChatCompletion`, not the message that `getReasoningContent` reads.
- `getReasoningContent` then finds no `reasoning_content` in `message._additionalProperties()` and returns empty, so the assistant message's `reasoningContent` metadata stays empty and the replay in `createRequest` has nothing to re-attach on the next turn — hence the 400.

This is independent of tool calling; the tool case is just where DeepSeek requires `reasoning_content` to be present. On current `main` the non-streaming (`call`) path does capture it (via `getReasoningContent`) and replay it in `createRequest` (wired in `b1df49934`) — only streaming drops it before capture.

I saw @i-zoufeng offered to fix this — happy to support that, or to take a focused PR carrying `reasoning_content` through to the aggregated `AssistantMessage` so the existing replay can re-attach it, with a test asserting it survives aggregation. One thing that would pinpoint the fix: are you on the `spring-ai-starter-model-openai` starter or the native `-deepseek` one? They take different streaming routes — `spring-ai-starter-model-openai` hits the `ChunkMerger` gap above, while the native `-deepseek` path does capture `reasoning_content` on `DeepSeekAssistantMessage`, but the shared `MessageAggregator` rebuilds the message into a plain `AssistantMessage` and drops that typed field, so the next-turn replay has nothing to re-attach. That tells us where the fix belongs. Happy to go a different direction if you'd prefer.

jewoodev:
[...]

PR #6373, opened 2026-06-10
Align OpenAI-compatible streaming `reasoning_content` with non-streaming responses

jewoodev:
The OpenAI-compatible streaming path drops provider reasoning metadata that `call()` already preserves. `ChunkMerger.chunkToChatCompletion` rebuilds streamed messages from `content`/`refusal`/`toolCalls` only, so fields such as DeepSeek's `reasoning_content` and OpenRouter's `reasoning` never reach `getReasoningContent`. On next-turn replay, `createRequest` then has nothing to re-attach, so providers that require reasoning on assistant tool-call messages reject the request as missing `reasoning_content`.

This keeps the fix inside `OpenAiChatModel` and reuses the existing `reasoning_content` / `reasoning` mapping already used by `getReasoningContent` and assistant-message replay:

- Streamed chunks carry delta additional properties into rebuilt messages.
- Merged tool-call chunks concatenate reasoning fragments.
- The streaming pipeline accumulates reasoning per choice so the final response survives `MessageAggregator`-based advisor replay.
- Intermediate streamed responses now expose the reasoning accumulated so far.

Related scope: #6402 was closed as superseded by this PR. #6484 reports the same `ChunkMerger.chunkToChatCompletion(...)` stream-path boundary, and #6483 covers the first delta-to-message additional-properties copy from that report. This PR also covers reasoning fragments merged across tool-call chunks and the final aggregated assistant message used for advisor/chat-memory replay. The confirmed RC2 `stream()` symptom in #6375 maps to the same gap.

Tests cover DeepSeek/OpenRouter reasoning fields with split tool calls, reasoning without tools, and `createRequest` replay. The native DeepSeek starter loses reasoning through a separate `MessageAggregator` flattening path and remains out of scope.

See #5898, #6375, and #6484

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -119,6 +119,7 @@ import org.springframework.util.StringUtils;
  * @author Thomas Vitale
  * @author Eric Bottard
  * @author Taewoong Kim
+ * @author Jewoo Shin
  */
 public final class OpenAiChatModel implements ChatModel {
 
@@ -264,6 +265,7 @@ public final class OpenAiChatModel implements ChatModel {
 		return Flux.deferContextual(contextView -> {
 			ChatCompletionCreateParams request = createRequest(prompt, true);
 			ConcurrentHashMap<String, String> roleMap = new ConcurrentHashMap<>();
+			ConcurrentHashMap<String, String> reasoningMap = new ConcurrentHashMap<>();
 			final ChatModelObservationContext observationContext = ChatModelObservationContext.builder()
 				.prompt(prompt)
 				.provider(AiProvider.OPENAI.value())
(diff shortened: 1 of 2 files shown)


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -1074,7 +1082,23 @@ private static Delta mergeDeltas(Delta left, Delta right) {
...
+					deltaBuilder.putAdditionalProperty(reasoningKey,
+							JsonValue.from(stringProperty(left, reasoningKey) + rightFragment));
+				}
+			}
+			return deltaBuilder.build();
+		}
+
+		private static String stringProperty(Delta delta, String key) {
+			JsonValue value = delta._additionalProperties().get(key);
+			return value == null ? "" : (String) value.asString().orElse("");
```
ericbottard: I'm not very keen on transforming a `null` into `""`. If a value isn't there, it isn't there. Or does the downstream code rely on a value being present in all cases?

jewoodev: Good point. The downstream replay path only needs the value when actual reasoning content is present, so missing provider fields do not need to be modeled as empty strings here. I updated the merge logic to leave missing fields absent and only concatenate/propagate reasoning fields when the incoming chunk carries a non-empty fragment.

ericbottard:
Thanks @jewoodev for a very extensive analysis and fix.

You can see I pushed fda9ed4ce containing also integration tests against DeepSeek using OpenAiChatModel

PR #6381, opened 2026-06-10
Merge streaming tool call deltas by index in `OpenAiChatModel`

### LINKED ISSUE #6374: Using the DeepSeek model that is called by OpenAI, it was discovered that the results returned by the streaming calling tool were not aggregated, resulting in the parameters of the calling tool always being empty.
**Bug description**
Using the DeepSeek model that is called by OpenAI, it was discovered that the results returned by the streaming calling tool were not aggregated, resulting in the parameters of the calling tool always being empty.

**Environment**
Spring Boot 4.5
Spring AI 2.0.0-RC2

**Steps to reproduce**

1. Configure OpenAiChatModel with the Gemini deepseek endpoint.
2. Register a tool with the parameter being List<String>
3. Send a prompt that triggers the tool.
4. Run the Spring AI tool-calling flow.

The model invocation tool returned arguments are null or empty for tool: show_options. Using empty JSON object as default

**Expected behavior**
The parameters of the Spring AI streaming invocation tool should be of type List<String> instead of null.

Debugging revealed that the streaming invocation tool's results were not aggregated together, resulting in the parameters of the invocation tool remaining always empty.

`
[ToolCall[id=call_1d70a7ee36be43c286e4ddc7, type=function, name=show_options, arguments=], ToolCall[id=, type=function, name=, arguments={], ToolCall[id=, type=function, name=, arguments="options": ["美人], ToolCall[id=, type=function, name=, arguments=鱼在], ToolCall[id=, type=function, name=, arguments=海底唱歌"], ToolCall[id=, type=function, name=, arguments=, "], ToolCall[id=, type=function, name=, arguments=美人鱼], ToolCall[id=, type=function, name=, arguments=和小], ToolCall[id=, type=function, name=, arguments=鱼一起], ToolCall[id=, type=function, name=, arguments=游泳"], ToolCall[id=, type=function, name=, arguments=, "], ToolCall[id=, type=function, name=, arguments=美人鱼], ToolCall[id=, type=function, name=, arguments=坐在海], ToolCall[id=, type=function, name=, arguments=星], ToolCall[id=, type=function, name=, arguments=上晒太阳"]}]]
`
`org.springframework.ai.tool.execution.ToolExecutionException: Cannot invoke "java.util.List.iterator()" because "options" is null`

```java

import org.springframework.ai.tool.annotation.ToolParam;
import org.springframework.stereotype.Component;

import java.util.List;

@Component
public class ShowOptionsTool {
    @Tool(name = "show_options",
            description = "Present 3 clickable choices/extensions to the user. Used to provide creative ideas, guide imagination, or offer safe alternatives.",
            returnDirect = false)
    public String showOptions(@ToolParam(description = "An array of exactly 3 short creative options in the user's language. " +
                                          "CRITICAL LIMITATION: Each option MUST be a complete phrase that includes the subject. " +
                                          "Never return isolated verbs. Example: If the subject is 'cat', " +
                                          "return ['Sleeping cat', 'Cat playing with yarn', 'Cat eating fish'], NOT ['Sleeping', 'Playing', 'Eating'].")
                              List<String> options) {
         return  "success";
    }

        var clientOptions = ClientOptions.builder()
                .apiKey(config.getApiKey())
                .baseUrl(config.getApiUrl())
                .httpClient(SpringAiOpenAiHttpClient.builder()
                        .build())
                .build();
        var openAIClient = new OpenAIClientImpl(clientOptions);

        var options = OpenAiChatOptions.builder()
                .apiKey(config.getApiKey())
                .baseUrl(config.getApiUrl())
                .model(config.getConfigName())
                .temperature(config.getTemperature())
                .topP(config.getTopP())
                // 开启流式返回 token 的消耗数量
                .streamUsage(true)
                .build();

        OpenAiChatModel chatModel = OpenAiChatModel.builder()
                .openAiClient(openAIClient)
                .options(options)
                .build();

        var chatClient = ChatClient.builder(chatModel).defaultAdvisors(ToolCallingAdvisor
                .builder()
                .toolCallingManager(customToolCallingManager)
                .streamToolCallResponses(false)
                .build()).build();

        Flux<ChatResponse> flux = chatClient.prompt().system("xxx")
                .user("Mermaid")
                .tools(new ShowOptionsTool())
                .stream()
                .chatResponse();
    }
}
```

jewoodev:
OpenAI-compatible streaming endpoints can send continuation tool-call deltas with `id: ""` instead of omitting `id`. `OpenAiChatModel.ChunkMerger` treats any present id as a new tool-call boundary, so argument fragments become separate `ToolCall`s with empty id/name and tool execution receives empty or null arguments.

This PR merges tool-call deltas by the required `index` field instead. It fixes the empty-id case from gh-6374, keeps OpenAI-style omitted ids working, and removes the one-tool-call-per-chunk assumption so parallel or multi-delta chunks merge correctly. #6369 covers the empty-id continuation case; this PR uses the index-keyed path so those related cases stay covered together.

Tests cover omitted ids, empty ids, parallel tool calls, and multiple deltas per chunk. Verified with `./mvnw -pl models/spring-ai-openai package`.

Fixes gh-6374

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -1039,7 +1039,7 @@ public final class OpenAiChatModel implements ChatModel {
 		}
 	}
 
-	private static final class ChunkMerger {
+	static final class ChunkMerger {
 
 		static boolean hasToolCall(ChatCompletionChunk chunk) {
 			return !chunk.choices().isEmpty()
@@ -1077,35 +1077,18 @@ public final class OpenAiChatModel implements ChatModel {
 		}
 
 		private static Delta mergeDeltas(Delta left, Delta right) {
+			// Deltas of the same logical tool call share the required 'index' field.
+			// Some OpenAI-compatible providers (e.g. DeepSeek) send an empty-string id
+			// on continuation deltas instead of omitting it, so id presence cannot be
+			// used to detect the start of a new tool call.
 			var tcs = Stream.of(left.toolCalls(), right.toolCalls()).flatMap(Optional::stream).reduce((tcs1, tcs2) -> {
 				if (tcs2.isEmpty()) {
 					return tcs1;
 				}
-				Assert.isTrue(tcs2.size() == 1, "no more than one tool call per message currently supported");
-				ToolCall toolCall = tcs2.get(0);
-				if (toolCall.id().isPresent()) {
-					List<ToolCall> result = new ArrayList<>(tcs1);
-					result.add(toolCall);
-					return result;
-				}
-				else {
-					ToolCall lastFromTc1 = tcs1.get(tcs1.size() - 1);
-					Function lastFromTc1F = lastFromTc1.function().get();
-
-					var concatenatedArgs = Stream
-						.of(lastFromTc1F.arguments(), toolCall.function().flatMap(Function::arguments))
-						.flatMap(Optional::stream)
-						.reduce((args1, args2) -> args1 + args2)
-						.orElse("");
-
-					List<ToolCall> result = new ArrayList<>(tcs1);
-					result.set(tcs1.size() - 1,
-							lastFromTc1.toBuilder()
-								.putAllAdditionalProperties(toolCall._additionalProperties())
-								.function(lastFromTc1F.toBuilder().arguments(concatenatedArgs).build())
-								.build());
-					return result;
-				}
+				Map<Long, ToolCall> mergedByIndex = new LinkedHashMap<>();
+				tcs1.forEach(tc -> mergedByIndex.merge(tc.index(), tc, ChunkMerger::mergeToolCalls));
+				tcs2.forEach(tc -> mergedByIndex.merge(tc.index(), tc, ChunkMerger::mergeToolCalls));
+				return List.copyOf(mergedByIndex.values());
 			}).orElse(List.of());
 
 			Delta.Builder deltaBuilder = left.toBuilder().toolCalls(tcs);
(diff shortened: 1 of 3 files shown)


jewoodev:
Rebased onto main to resolve a conflict with #6365.

#6365 changed `mergeDeltas` to preserve tool-call additional properties. This PR replaces that merge logic with by-index merging, so I carried the preservation over: `ChunkMerger.mergeToolCalls` now copies each delta's additional properties via `putAllAdditionalProperties`.

The streaming test added in #6365 (`preserveUnmappedToolCallAdditionalProperties`) verifies this. It fails without that call and passes with it, along with the rest of the chat model tests.

jewoodev:
Thanks for the feedback.

Added a small follow-up commit that guards the remaining `chunkToChatCompletion` `Optional.get()` calls and covers both arguments-only tool call deltas and deltas without a `function` object.

The `mergeDeltas` case is already covered by the by-index merge in this branch, since `mergeToolCalls` now builds a function when the previous delta does not carry one.

I'll keep follow-up changes on this branch on my side to keep the PR history straightforward. Feel free to leave any additional feedback in this PR.

freshman-wang:
Hi  @jewoodev ,

I noticed this PR fixes tool call delta merging in the model implementation layer (OpenAiChatModel). 

However, according to Spring AI's design philosophy (see https://docs.spring.io/spring-ai/reference/api/advisors-recursive.html), the tool calling loop and related logic should be implemented in the advisor layer (e.g., ToolCallingAdvisor) rather than in the model implementation.

This approach:
1. Keeps model implementations focused on API response parsing
2. Makes tool calling logic reusable across different model implementations
3. Allows other advisors to observe/intercepts the tool calling process

Would you consider moving this logic to the advisor layer, or opening a discussion with the maintainers about the best approach?

Thanks!

jewoodev:
Thanks for raising this. I agree that the tool-calling loop belongs in the advisor layer.

My intent here is narrower: this PR does not move tool execution or the tool-calling loop into `OpenAiChatModel`. It fixes the OpenAI streaming response parsing boundary so split `tool_calls` deltas are reconstructed into a valid assistant tool call before `ToolCallingAdvisor` or a user-controlled loop sees the response.

Without that normalization, the advisor layer receives fragmented or invalid tool calls and cannot reliably run the loop. So I think this merge should stay in the OpenAI model implementation, while the loop and tool execution remain in `ToolCallingAdvisor`.

If the maintainers prefer a different boundary here, I’m happy to adjust.

myifeng:
https://github.com/spring-projects/spring-ai/issues/6762

Issue #6375, opened 2026-06-10
`reasoningContent` in metadata is always null when calling DeepSeek model via OpenAI API


lyn-my:
I am calling the DeepSeek model through the OpenAI API. When trying to retrieve the reasoning/reasoning content from the response metadata, it always returns null.
﻿
**Code snippet:**
﻿
```java
// Get the assistant message
AssistantMessage message = response.getResult().getOutput();
﻿
// Access the reasoning content from metadata
Object reasoning = message.getMetadata().get("reasoningContent");
```
﻿
**Expected behavior:**
The `reasoningContent` should contain the reasoning text from the DeepSeek model's response.
﻿
**Actual behavior:**
The `reasoning` object is always `null`.
﻿
**Additional context:**
- API endpoint: OpenAI compatible API calling DeepSeek model
- The model I am using supports reasoning output (e.g., deepseek-reasoner)
- The response from the API does contain reasoning content when checking the raw JSON response, but it is not being populated into the metadata map
﻿
**Question:**
Is `reasoningContent` the correct metadata key? Or is there another way to access the reasoning content when using the OpenAI API format with DeepSeek models?
﻿
---

demo-zexuan:
Could you please provide the framework version, logs and screenshots?

demo-zexuan:
Ref：https://github.com/spring-projects/spring-ai/issues/6374

wolf8334:
using the dedicated DeepSeek model instead: https://docs.spring.io/spring-ai/reference/2.0/api/chat/deepseek-chat.html

sdeleuze:
Indeed please use the dedicated DeepSeek support instead.

sdeleuze:
Reopening this one to re-evaluate it in connection to #6364.

jewoodev:
Hi, I looked into this. The main split is `null` vs `""`.

A literal `null` means the metadata key is absent, which points to older milestone builds or the 1.1.x `call()` path; #6382 covers the 1.1.x variant. On 2.0.0-RC1+ the key normally exists, and the known streaming gap is that provider reasoning metadata can be dropped while streamed chunks are rebuilt, so the value becomes `""`; #6373 targets that path.

@lyn-my could you confirm the Spring AI version and whether this uses `call()` or `stream()`?

lyn-my:
Sorry, my description was inaccurate. I am using version **2.0.0-R2**.

[...]

PR #5711, opened 2026-03-30
Fix: Add missing reasoningContent to metadata in internalCall

YashRL:
Fixes #5693

Problem

When using the non-streaming API (call()), the reasoningContent field was not included in the AssistantMessage metadata. However, the streaming API (stream()) correctly included it. This inconsistency affected models like DeepSeek-R1 and Qwen3.5-plus that return reasoning content in their API responses.

Root Cause

In OpenAiChatModel.java, the internalCall method built the metadata map without the reasoningContent field, while internalStream included it.

Solution

Added reasoningContent to the metadata map in the internalCall method to ensure consistency with the streaming API.

Changes

- Added reasoningContent to metadata in internalCall
- Added unit tests to verify behavior

Testing

- Verified reasoningContent is included in non-streaming responses
- All tests pass

PR #6382, opened 2026-06-10
Fix `reasoningContent` always `null` on `OpenAiChatModel` call path

jewoodev:
On the 1.1.x line, `OpenAiChatModel` already exposes `reasoningContent` metadata for streaming responses, but the blocking `call()` path omits the same entry from its generation metadata map. This leaves OpenAI-compatible reasoning responses with `message.getMetadata().get("reasoningContent") == null` even when the wire payload contains `reasoning_content`.

This PR mirrors the streaming metadata entry on the `call()` path and adds a `MockRestServiceServer` regression test for a `reasoning_content` response.

Verified with `./mvnw package -pl models/spring-ai-openai`.

See #6375

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -105,6 +105,7 @@ import org.springframework.util.StringUtils;
  * @author Alexandros Pappas
  * @author Soby Chacko
  * @author Jonghoon Park
+ * @author Jewoo Shin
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks for this fix, @jewoodev. This overlaps with #5711, which was opened earlier and makes the identical change to `internalCall` in `OpenAiChatModel.java`, adding the `reasoningContent` entry to the generation metadata. That PR also carries a test for the empty-reasoning fallback in addition to the populated case. Since the two address the same issue with the same production change, I am going to consolidate on #5711 and close this one. Thank you for the contribution, and I hope you will send more our way.

PR #6391, opened 2026-06-11
Remove `streamToolCallResponses` from advisor builders

tzolov:
When streamToolCallResponses=true, tool-call request chunks were streamed downstream without the paired ToolResponseMessage, corrupting any downstream memory advisor's conversation history. The flaw cannot be fixed cleanly, so the option is removed from ToolCallingAdvisor, ToolCallAdvisor, and ToolSearchToolCallingAdvisor builders along with the corresponding auto-configuration properties.
Two replacement patterns are documented:
- Placing a custom advisor inside the tool-call loop (order > `ToolCallingAdvisor.DEFAULT_ORDER) to observe every iteration via a side channel
- Disabling the auto-registered advisor with AdvisorParams.toolCallingAdvisorAutoRegister(false) and driving the loop manually.
Integration tests for the manual pattern are added . Upgrade notes for 2.0.0 GA cover impact and migration steps.


ilayaperumalg:
The module `spring-ai-tool-search-advisor` needs `lucene-core` as a test dependency. Also, the test `AbstractToolSearchToolCallingAdvisorIT` requires changes to enable the Chat memory inside the tool call advisor loop. Will fix them when merging.

ilayaperumalg:
Rebased, merged along with the test fixes via a753fde6c23cf1912104b9d833467b624e9c50d2

PR #6395, opened 2026-06-11
feat(tool): allow returnDirect tool calls to bypass serialization

### LINKED ISSUE #3191: Allow Tool Calls that use returnDirect to return the result object without serialisation
Currently, when doing tool calls, the response is a converted to JSON so it can be sent back to the model to process the result. This works for the normal tool calling flow, but is overly restrictive when using the return-direct functionality.

In some of my tool calls, I use return direct and I have to create complex objects that are not easily serialisable. It would be great if we could return the actual object when returning the result of a tool call direct.

**Expected Behavior**

When performing a tool call with return direct, the result object is stored within the `ToolResponse` item contained within `ToolResponseMessage`.

e.g. 

```java
public record ToolResponse(
  String id, String name, String responseData, @Nullable Object responseObj
) {}
```

Would be happy to just use an object and let the user cast as needed, or some generic way to implement this.

**Current Behavior**

Tool calls that return direct must return an object that can be serialised to JSON, even if not practical.

**Context**

I've hit this within a chatbot i'm building that needs to integrate with a third-party. It's possible to create intermediate objects to delay creating the unserialisable objects until I have control of the code-flow, but it's a lot of ceremony. 

I've taken to using `AtomicRefs` within `ToolContext` in the meantime, but obviously it's not ideal!

Baqirrizvidev:
Resolves #3191. This PR introduces support for passing raw objects directly from returnDirect tool calls to DefaultChatClient output converters and entity return methods, bypassing the intermediate JSON serialization step.

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
@@ -517,6 +517,16 @@ public class DefaultChatClient implements ChatClient {
 			}
 
 			var chatResponse = doGetObservableChatClientResponse(this.request, advisorChain).chatResponse();
+			if (chatResponse != null && chatResponse.getResult() != null
+					&& chatResponse.getResult().getOutput() != null) {
+				var assistantMessage = chatResponse.getResult().getOutput();
+				if (assistantMessage.getMetadata().containsKey("tool-output")) {
+					Object toolOutputObj = assistantMessage.getMetadata().get("tool-output");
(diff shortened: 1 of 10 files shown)


sdeleuze:
As requested multiple times, please add [a Signed-off-by trailer](https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring#how-to-use-developer-certificate-of-origin). Notice [we don't accept contribution from AI bots](https://github.com/spring-projects/spring-ai/blob/main/CONTRIBUTING.md#code-of-conduct).

PR #6398, opened 2026-06-11
Filter unsupported tool messages in `JdbcChatMemoryRepository`

tzolov:
Previously, ToolResponseMessage and AssistantMessage instances with tool calls were silently saved with corrupted data and read back as empty stubs, polluting the LLM conversation context.

This commit explicitly filters out those message types on save to prevent corrupted messages from being injected into the LLM context. Legacy TOOL rows already in the database are skipped on read.


memory-repositories/spring-ai-model-chat-memory-repository-jdbc/src/main/java/org/springframework/ai/chat/memory/repository/jdbc/JdbcChatMemoryRepository.java
```diff
@@ -151,6 +160,7 @@ public int getBatchSize() {
 	private static class MessageRowMapper implements RowMapper<Message> {
 
 		@Override
+		@SuppressWarnings("NullAway") // TOOL rows return null; caller filters them out
```
ericbottard, line 163: should change return type to `@Nullable Message` instead to represent reality, or did I miss some exceptional situation that legitimizes this?

PR #6404, opened 2026-06-11
GH-6403: Preserve single-line code block content in MarkdownCodeBlockCleaner

### LINKED ISSUE #6403: MarkdownCodeBlockCleaner discards content of single-line fenced code blocks
**Bug description**

`MarkdownCodeBlockCleaner` is part of the default cleaning pipeline used by `BeanOutputConverter` to strip markdown code fences from LLM responses before parsing (`WhitespaceCleaner` → `ThinkingTagCleaner` → `MarkdownCodeBlockCleaner` → `WhitespaceCleaner`).

When a fenced code block fits on a **single line** and has **no language identifier**, the cleaner returns an empty string and silently discards the payload. The now-empty text then fails structured-output parsing.

**Affected code**

`spring-ai-model/src/main/java/org/springframework/ai/converter/MarkdownCodeBlockCleaner.java`

```java
if (firstLine.length() > 3) {
    // Has language identifier like ```json
    text = lines.length > 1 ? lines[1] : "";
}
```

`text.split("\n", 2)` yields a single element for a single-line block, so `lines.length > 1` is `false` and the content is replaced with `""`.

**Minimal reproducible example**

```java
String result = new MarkdownCodeBlockCleaner().clean("```{\"key\": \"value\"}```");
```

**Expected:** `{"key": "value"}`
**Actual:** `""` (content discarded)

Multi-line blocks (```` ```json\n{...}\n``` ```` and ```` ```\n{...}\n``` ````) are handled correctly; only the single-line, no-language case is affected.

**Root cause**

The `firstLine.length() > 3` check assumes that any first line longer than ```` ``` ```` is a ```` ```lang ```` identifier whose body lives on the following lines. For a single-line block there is no separate identifier line, so only the leading fence should be stripped instead of returning `""`.

**Environment**

* Spring AI version: `2.0.0-SNAPSHOT` (`main`); class is `@since 1.1.0`
* Java version: 17+

I'd be happy to submit a PR (with regression tests) for this.

gaoxiaolei-s59:
This fixes `MarkdownCodeBlockCleaner` silently discarding the content of a
single-line fenced code block that has no language identifier (for example
`` ```{"key":"value"}``` ``). The cleaner returned an empty string, so the
payload was lost and structured-output parsing through `BeanOutputConverter`'s
default cleaning pipeline failed.

`text.split("\n", 2)` returns a single element for a single-line block, so the
`lines.length > 1 ? lines[1] : ""` branch replaced the content with `""`. A
single-line block has no separate language-identifier line, so only the leading
fence is stripped now. Multi-line blocks are unaffected.

Regression tests are added covering single-line, multi-line, language-tagged,
plain-text, `null` and empty inputs.

Closes gh-6403

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/converter/MarkdownCodeBlockCleaner.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/converter/MarkdownCodeBlockCleaner.java
@@ -48,8 +48,9 @@ public class MarkdownCodeBlockCleaner implements ResponseTextCleaner {
 				// Extract language identifier if present
 				String firstLine = lines[0].trim();
 				if (firstLine.length() > 3) {
-					// Has language identifier like ```json
-					text = lines.length > 1 ? lines[1] : "";
+					// A single-line block has no language line:
+					// strip the leading fence, keeping the content.
+					text = lines.length > 1 ? lines[1] : text.substring(3);
 				}
 				else {
 					// Just ``` without language
--- a/spring-ai-model/src/test/java/org/springframework/ai/converter/MarkdownCodeBlockCleanerTest.java
+++ b/spring-ai-model/src/test/java/org/springframework/ai/converter/MarkdownCodeBlockCleanerTest.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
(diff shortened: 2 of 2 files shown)


ericbottard:
This has been fixed in 6fcace9a2d4d017de5b9c33a91fd3652caceb7ae already

Issue #3388, opened 2025-05-30
AssistantMessage supports reasoningContent


leooolee:
**Expected Behavior**
Currently, many model providers offer reason models, and users choose whether to use them depending on the specific use case. Spring AI needs to improve its support for reason  models. add reasoningContent to AssistantMessage
```java
public class AssistantMessage extends AbstractMessage implements MediaContent {

	private final List<ToolCall> toolCalls;

	protected final List<Media> media;
        private String reasoningContent 
        // other codes        
       ...
```

**Current Behavior**
At present, supporting reason models requires adapting to different model vendors. This makes it difficult to standardize and unify reasoning messages, and also results in relatively high adaptation costs.

simsnow:
+1
The openai chatModel support toolCall and media，but not support reasoningContent，
The deepseek chatModel support reasoningContent，but not support toolCall and media

sha2kyou:
> +1 The openai chatModel support toolCall and media，but not support reasoningContent， The deepseek chatModel support reasoningContent，but not support toolCall and media

Same with Bedrock converse models—it doesn’t support reasoning either, but support tool calling.

AllyBully:
+1
Now I'm recording the reasoning content into the database, and I still need to rewrite a lot of content to adapt to the reasoningContent of different large models.

mistreckless:
+1,
I have to disable reasoning_parser in vllm inference, and parse reasoning content in a ChatResponse text output. 
Also, lI'm confused about the library roadmap. It has a lot of unnecessary staff (like Framework-Controlled Tool Execution, with limitations that make it useless), and doesn't simply have full support for API fields

PR #6414, opened 2026-06-13
Preserve Bedrock reasoning content on tool replay

### LINKED ISSUE #6413: Bedrock Converse drops reasoningContent, breaking extended thinking with tool use
**Bug description**
  
  The Bedrock Converse chat model (`BedrockProxyChatModel`) does not handle Anthropic
  `reasoningContent` (extended thinking) blocks. They are silently dropped when parsing
  the response and are never sent back when rebuilding the assistant turn. As a result:

  1. With thinking enabled and **no** tools, the reasoning text is never surfaced
     (reasoning tokens are billed but the content is discarded).
  2. With thinking enabled **and** tools, every tool-use turn fails with a Bedrock
     `ValidationException`, because Bedrock requires the assistant message containing a
     `toolUse` block to also carry the preceding signed `reasoningContent` block when the
     follow-up `toolResult` is sent. Since Spring AI's internal tool loop (`ToolCallAdvisor`
     re-invokes the model with the tool result) rebuilds that assistant turn without the
     reasoning block, the next request is rejected.
     
  Two code paths are involved (line numbers from 2.0.0-M8,
  `models/spring-ai-bedrock-converse/.../BedrockProxyChatModel.java`):

  - **Response parsing** (`toChatResponse`, ~L653): content is filtered to
    `text` + `toolUse` only — `ContentBlock.Type.REASONING_CONTENT` is dropped:
    ```java
    message.content().stream()
        .filter(content -> content.type() != ContentBlock.Type.TOOL_USE)
        .filter(content -> content.text() != null)   // reasoningContent has null text → dropped
  - Request building (assistant branch of createRequest, ~L366): the assistant turn is
  reconstructed from message.getText() + assistantMessage.getToolCalls() only — there is
  no path that re-emits a reasoningContent block (with its signature).

  This appears related to #3388 (AssistantMessage has no field to carry reasoning), but
  that issue is generic; this is specifically the Bedrock Converse drop + tool-replay failure.
  The same gap is present on main.

  Environment

  - Spring AI: 2.0.0-M8 (also reproduces on main)
  - Model: eu.anthropic.claude-sonnet-4-6 (any thinking-capable Claude model on Bedrock)
  - Java: 25
  - AWS SDK bedrockruntime via spring-ai-starter-model-bedrock-converse
  - No vector store

  Steps to reproduce

  1. Build a BedrockProxyChatModel for a thinking-capable Claude model.
  2. Enable extended thinking by passing the Anthropic thinking object through
  BedrockChatOptions.requestParameters (→ Converse additionalModelRequestFields),
  e.g. {"thinking": {"type": "enabled", "budget_tokens": 2048}}, with temperature 1.0.
  3. Register at least one tool and send a prompt that triggers a tool call.
  4. The first model response returns a reasoningContent + toolUse; Spring AI executes the
  tool and re-invokes the model with the toolResult. That second request throws a Bedrock
  ValidationException (the assistant turn is missing the required reasoningContent block).

  Expected behavior

  - reasoningContent blocks parsed from the Converse response are preserved on the resulting
  AssistantMessage (text + signature), and
  - re-emitted as a reasoningContent ContentBlock (ahead of the toolUse block, with the
  original signature) when that assistant turn is rebuilt for the next request,

  so that extended thinking works on the Converse API both standalone and with tool use, matching
  the first-party Anthropic provider's thinking support.

  Minimal Complete Reproducible example

  Drop-in `@Test` for `models/spring-ai-bedrock-converse/.../BedrockProxyChatModelIT.java`
  (reuses the existing `MockWeatherService` and `@RequiresAwsCredentials` setup in that class).
  It mirrors the existing `functionCallTest`, only adding extended thinking via
  `requestParameters`. It passes today with thinking disabled and fails with thinking enabled:
  the tool-result turn throws `software.amazon.awssdk.services.bedrockruntime.model.ValidationException`
  because the rebuilt assistant turn is missing the required `reasoningContent` block.

  ```java
  @Test
  void functionCallWithThinkingEnabled() {

      UserMessage userMessage = new UserMessage(
              "What's the weather like in San Francisco, Tokyo and Paris? Return the result in Celsius.");

      List<Message> messages = new ArrayList<>(List.of(userMessage));

      var promptOptions = BedrockChatOptions.builder()
          // Anthropic constraint: temperature must be 1.0 when thinking is enabled
          .temperature(1.0)
          // Inject the Anthropic `thinking` object via Converse additionalModelRequestFields.
          // requestParameters is typed Map<String,String> but the converter accepts nested maps.
          .requestParameters(Map.of("thinking",
                  Map.of("type", "enabled", "budget_tokens", 2048)))
          .toolCallbacks(List.of(FunctionToolCallback.builder("getCurrentWeather", new MockWeatherService())
              .description("Get the weather in location. Return in 36°C format")
              .inputType(MockWeatherService.Request.class)
              .build()))
          .build();

      // Fails: ValidationException on the second (tool-result) turn, because the assistant
      // message replayed by the internal tool loop omits the signed reasoningContent block
      // that Bedrock requires to precede the toolUse block.
      ChatResponse response = this.chatModel.call(new Prompt(messages, promptOptions));

      assertThat(response.getResult().getOutput().getText()).contains("30", "10", "15");
  }
  ```

  (`requestParameters(Map<String,String>)` is the public M8 way to reach `additionalModelRequestFields`;
  the same failure occurs through any path that sets the `thinking` field. Note the
  `Map.of(...)` argument relies on the converter handling nested maps, which it does.)

jewoodev:
Bedrock Converse responses can include signed `reasoningContent` before `toolUse`. The current adapter drops those blocks, so the next request sends the `toolResult` without replaying the signed reasoning block that Bedrock validates for that assistant turn.

This PR keeps the preserved reasoning on a Bedrock-specific assistant message and replays it, unchanged and before `toolUse`, when the tool-calling loop builds the next request. The scope stays Bedrock-specific; it does not add a generic `AssistantMessage` reasoning API.

AWS-free mapping tests cover signed reasoning replay, redacted reasoning replay, reasoning-only responses, and the existing non-reasoning `text` + `toolUse` path.

Fixes #6413

### DIFF
--- a/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockAssistantMessage.java
+++ b/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockAssistantMessage.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.bedrock.converse;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.ai.chat.messages.AssistantMessage;
+import org.springframework.ai.content.Media;
+import org.springframework.util.CollectionUtils;
+
+/**
+ * Bedrock-specific {@link AssistantMessage} that carries the signed
+ * {@code reasoningContent} blocks returned alongside an assistant turn so the
+ * tool-calling loop can replay them, unmodified, on the next Bedrock request.
+ * <p>
+ * Amazon Bedrock requires that the assistant turn containing a {@code toolUse} block also
+ * replays the preceding signed {@code reasoningContent} block when the matching
+ * {@code toolResult} is sent. Keeping the reasoning state on the same assistant message
+ * as the tool calls makes that replay invariant explicit: the reasoning is associated
+ * with the same turn that is selected and re-emitted during tool calling.
+ *
+ * @author Jewoo Shin
+ * @since 2.0.1
+ * @see BedrockReasoningContent
+ */
+public class BedrockAssistantMessage extends AssistantMessage {
+
+	private final List<BedrockReasoningContent> reasoningContents;
+
+	protected BedrockAssistantMessage(@Nullable String content, List<BedrockReasoningContent> reasoningContents,
+			Map<String, Object> properties, List<ToolCall> toolCalls, List<Media> media) {
+		super(content, properties, toolCalls, media);
+		this.reasoningContents = reasoningContents;
+	}
+
+	public List<BedrockReasoningContent> getReasoningContents() {
+		return this.reasoningContents;
+	}
+
+	public boolean hasReasoningContents() {
+		return !CollectionUtils.isEmpty(this.reasoningContents);
+	}
+
+	@Override
+	public boolean equals(@Nullable Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (!(o instanceof BedrockAssistantMessage that)) {
+			return false;
+		}
+		if (!super.equals(o)) {
+			return false;
+		}
+		return Objects.equals(this.reasoningContents, that.reasoningContents);
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(super.hashCode(), this.reasoningContents);
+	}
+
+	@Override
+	public String toString() {
+		return "BedrockAssistantMessage [messageType=" + this.messageType + ", toolCalls=" + super.getToolCalls()
+				+ ", textContent=" + this.textContent + ", reasoningContents=" + this.reasoningContents + ", metadata="
+				+ this.metadata + "]";
+	}
+
+	public static Builder builder() {
+		return new Builder();
+	}
+
+	// Public Builder class exposed to users. Avoids having to deal with noisy generic
+	// parameters.
+	public static class Builder extends AbstractBuilder<Builder> {
+
+	}
+
+	public static class AbstractBuilder<B extends AbstractBuilder<B>> extends AssistantMessage.Builder<B> {
+
+		protected List<BedrockReasoningContent> reasoningContents = List.of();
+
+		public B reasoningContents(List<BedrockReasoningContent> reasoningContents) {
+			this.reasoningContents = reasoningContents;
+			return self();
+		}
+
+		@Override
+		public BedrockAssistantMessage build() {
+			return new BedrockAssistantMessage(this.content, this.reasoningContents, this.properties, this.toolCalls,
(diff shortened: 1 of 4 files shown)


jewoodev:
A note on test scope: I intentionally kept the coverage AWS-free instead of adding a live Bedrock integration test for the extended-thinking + tool-use scenario.

The regression is in Spring AI's request/response mapping and replay path: Bedrock returns a `reasoningContent` block next to `toolUse`, and `DefaultToolCallingManager` replays the selected assistant message into the next request. The added tests exercise that path by mapping the Bedrock response into an assistant message and then converting it back to Bedrock `ContentBlock`s, including signed and redacted reasoning content.

A live integration test would require AWS credentials, model access in a supported region, extended thinking support, quota/cost, and could be unavailable or flaky for contributors. If maintainers would prefer live coverage as well, I can add a narrow env-gated IT, but I kept this PR focused on deterministic replay coverage first.

sobychacko:
@jewoodev Thanks for the PR. I am reviewing it now. I have a general question before we proceed too deep into the PR. Both this PR and the Anthropic counterpart (#6551) introduce a provider-specific assistant message subtype to carry reasoning/thinking content for tool replay. Here BedrockAssistantMessage and BedrockReasoningContent are public, but in #6551 AnthropicAssistantMessage and AnthropicThinkingContent are package-private. The use case is identical on both sides - the integration test in the issue (#6413) confirms users want to inspect whether reasoning was preserved on a tool-call turn. Was the difference intentional? If not, we should align them before 2.0.1 ships - either both public or both internal. What do you think?

jewoodev:
Thanks for raising this. I agree that the Bedrock and Anthropic replay carrier visibility should not diverge before 2.0.1.

The difference was contextual rather than intended as a long-term provider-specific API policy. In #6414, Bedrock did not have an existing reasoning/thinking response surface, so I used a small Bedrock-specific message/value object to keep the replay state attached to the tool-call turn and inspectable without introducing a generic `AssistantMessage` reasoning API.

For the Anthropic counterpart, the context was different because Anthropic already exposed `thinking` / `redacted_thinking` blocks as separate `Generation`s in 2.0.x. I kept the additional replay carrier internal there to limit the PR scope and avoid changing the existing observable Anthropic response surface while fixing the replay path.

Given the inspection use case, I think the better alignment is public provider-specific carriers for both Bedrock and Anthropic, while keeping this out of the generic `AssistantMessage` contract and avoiding raw SDK objects in the public surface. I opened #6582 to show that direction concretely. If you prefer minimizing new public API for 2.0.1, I can also make the Bedrock carrier internal, but my preference is the public provider-specific alignment.

sobychacko:
Thanks for the explanation, that clarifies things. 

After thinking it through a bit more though, I'd prefer to keep both `BedrockAssistantMessage` and  `BedrockReasoningContent` and the Anthropic counterparts internal for `2.0.1`. The bug fix is the priority here, and the replay works identically regardless of visibility. The inspection use case is real but not urgent for a patch release.

The bigger concern is #3388. We don't know yet what the right generic reasoning API on `AssistantMessage` looks like, and locking in public provider-specific carrier types now could constrain that design in `2.1.0`. If the generic API ends up conflicting with what we shipped in `2.0.1`, we'd be stuck maintaining both or breaking one of them. Keeping things internal preserves full design freedom when we tackle that properly.

 So for this PR, could you make `BedrockAssistantMessage` and `BedrockReasoningContent` package-private before we merge? How about closing #6582 also? when we tackle #3388 we can expose clean, consistent public types across both providers at the right level rather than doing it piecemeal in a patch release. What do you think?

jewoodev:
Thanks for the thoughtful explanation. I enjoyed working through this API boundary with you, and I agree.

I’ll keep the Bedrock replay carriers internal for `2.0.1` and close #6582 so the public reasoning API can be handled through #3388.

sobychacko:
@jewoodev Thanks for the PR. It has been merged upstream. Great contribution! I added some polish commits on top of yours. Please take a look. Thanks again!

PR #6423, opened 2026-06-15
Fix MarkdownCodeBlockCleaner discarding content of single-line fenced code blocks

### LINKED ISSUE #6403: MarkdownCodeBlockCleaner discards content of single-line fenced code blocks

MKP999:
## Problem

When a fenced code block fits on a **single line** without a language identifier, `MarkdownCodeBlockCleaner` returns an empty string and silently discards the payload.

**Affected code:** `spring-ai-model/src/main/java/org/springframework/ai/converter/MarkdownCodeBlockCleaner.java`

**Minimal reproducible example:**
```java
String result = new MarkdownCodeBlockCleaner().clean("```{\"key\": \"value\"}```");
// Expected: {"key": "value"}
// Actual: "" (content discarded)
```

## Root Cause

`text.split("\n", 2)` yields a single-element array for a single-line block. When `firstLine.length() > 3` (because the content is included after the opening fence), the code enters the "has language identifier" branch and evaluates `lines.length > 1 ? lines[1] : ""`. Since `lines.length == 1`, it returns `""`.

## Fix

Add an explicit check for `lines.length == 1` in the single-line case. When the block has no line break, strip only the opening fence from `firstLine` instead of returning an empty string. The trailing fence is already removed by the existing code below.

## Test Coverage

Added `MarkdownCodeBlockCleanerTest` with 7 test cases:
- Single-line code block without language identifier (the bug case)
- Single-line code block with language identifier
- Multi-line code block with language identifier
- Multi-line code block without language identifier
- Null input
- Empty input
- Plain text without code blocks

Closes #6403

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/converter/MarkdownCodeBlockCleaner.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/converter/MarkdownCodeBlockCleaner.java
@@ -42,14 +42,19 @@ public class MarkdownCodeBlockCleaner implements ResponseTextCleaner {
 
 		// Check for and remove triple backticks
 		if (text.startsWith("```") && text.endsWith("```")) {
-			// Remove the first line if it contains "```json" or similar
 			String[] lines = text.split("\n", 2);
 			if (lines[0].trim().toLowerCase().startsWith("```")) {
 				// Extract language identifier if present
 				String firstLine = lines[0].trim();
-				if (firstLine.length() > 3) {
-					// Has language identifier like ```json
-					text = lines.length > 1 ? lines[1] : "";
+				if (firstLine.length() > 3 && lines.length > 1) {
+					// Has language identifier like ```json and content on following lines
+					text = lines[1];
+				}
+				else if (firstLine.length() > 3 && lines.length == 1) {
+					// Single-line fenced block without line break, e.g.
+					// ```{"key": "value"}```
+					// Strip the opening fence only; the trailing fence is removed below
+					text = firstLine.substring(3);
 				}
 				else {
 					// Just ``` without language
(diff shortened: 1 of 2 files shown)


ericbottard:
Fixed via 6fcace9a2

ericbottard:
Thanks for your contribution @MKP999,

I ended up simplifying the logic quite a bit while re-using (most) of the tests you provided.

That being said, when you contribute a PR to the project, please make sure that all your tests pass (all did not) and that the code style and formatting used in the project is respected.

This is as simple as running a final
```
./mvnw clean package
```

PR #6424, opened 2026-06-15
Accumulate tool calling advisor usage

### LINKED ISSUE #6411: ToolCallingAdvisor no longer accumulates Usage across internal tool-call round trips
# `ToolCallingAdvisor` no longer accumulates `Usage` across internal tool-call round trips — what's the supported way to track total token usage?

## Spring AI version
2.0.0 (GA). Also reproduced on 2.0.0-M7/M8/RC1.

## Background

Prior to the M7 tool-calling rework, `OpenAiChatModel.internalCall()` (and the
equivalent in other providers) handled the tool-execution loop internally and
accumulated `Usage` across every internal LLM round trip via
`UsageCalculator.getCumulativeUsage(...)`. The `ChatResponse` returned to the
caller therefore had a `getMetadata().getUsage()` that already represented the
**total** prompt/completion tokens for the whole tool-calling exchange.

Since the tool-calling loop moved out of the model into
`org.springframework.ai.chat.client.advisor.ToolCallingAdvisor` (a
`do { ... } while (...)` loop in `adviseCall`/`adviseStream` over
`callAdvisorChain.nextCall(...)`), each iteration of the loop produces its own
`ChatResponse` with its own (per-round-trip) `Usage`. None of
`doInitializeLoop` / `doAfterCall` / `doFinalizeLoop` merge these together —
`doFinalizeLoop` simply returns the `ChatClientResponse` from the **last**
iteration as-is.

The net effect: for any conversation that triggers ≥1 tool call,
`chatClient.prompt(...).call().chatResponse().getMetadata().getUsage()` now
only reflects the **final** round trip, silently dropping the prompt/completion
tokens consumed by every earlier round trip (including the tokens spent on the
tool-call request itself). For a typical "one tool call then answer" exchange
this roughly halves the reported usage compared to pre-M7 behavior.

## Repro sketch

```java
ChatClient client = ChatClient.builder(chatModel)
        .defaultTools(someToolWithOneCallback)
        .build();

ChatResponse response = client.prompt()
        .user("question that triggers exactly one tool call")
        .call()
        .chatResponse();

Usage usage = response.getMetadata().getUsage();
// usage.getPromptTokens() / getCompletionTokens() only reflect the SECOND
// (post-tool-result) round trip; the first round trip's tokens (prompt for
// the user question + completion for the tool-call request) are lost.
```

`org.springframework.ai.support.UsageCalculator.getCumulativeUsage(Usage, ChatResponse)`
still exists and does exactly the summation that's needed, but nothing in
`ToolCallingAdvisor` calls it across loop iterations anymore.

## Question

Is there a supported/intended pattern for getting the **cumulative** usage for
the whole tool-calling exchange from `ChatClient`?  Specifically:

- Is `UsageCalculator.getCumulativeUsage` meant to be wired in by application
  code (e.g. via a custom `ToolCallingAdvisor` subclass that accumulates usage
  across `doAfterCall` and applies the total in `doFinalizeLoop`), or is this
  considered an internal utility that the framework itself should be using?
- If the former, is overriding `ToolCallingAdvisor` the recommended extension
  point, or is there a less invasive hook (observation/metrics listener, etc.)
  intended for this?
- Is this considered a regression vs. pre-M7 behavior, or an intentional
  change where callers are now expected to sum `Usage` themselves per round
  trip (e.g. via `TokenUsageAwareAdvisor`/observation hooks during the loop)?

This feels like a lot of internal-API knowledge to require for what seems like
a fairly common need (accurate per-request token accounting when tools are
involved). Happy to share the full implementation if useful as a starting
point for a built-in solution.

jewoodev:
This update makes `ToolCallingAdvisor` carry cumulative usage across framework-controlled tool-calling loop iterations.

Previously, each internal model call returned its own `Usage`, so the final non-streaming `ChatClientResponse` could expose only the usage from the last model call. The advisor now composes usage with the existing `UsageCalculator` semantics before returning the final call response, including `returnDirect` responses where the generation is replaced with the tool result.

For streaming responses, the accumulator is created per subscription and is updated from each streamed iteration's aggregate response. When a later iteration emits chunks, only chunks that already carry usage metadata are rewritten to include the prior loop accumulator; chunks without usage metadata are left unchanged. No state is kept on the advisor instance.

Validation:

```
./mvnw -pl spring-ai-client-chat -am test -Dtest=ToolCallingAdvisorTests,ToolCallingAdvisorSpanHierarchyTests -Dsurefire.failIfNoSpecifiedTests=false -Dmaven.build.cache.enabled=false
./mvnw -pl spring-ai-client-chat -am package -Dmaven.build.cache.enabled=false
```

Closes #6411

### DIFF
--- a/models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockConverseUsageAggregationTests.java
+++ b/models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockConverseUsageAggregationTests.java
@@ -288,19 +288,15 @@ public class BedrockConverseUsageAggregationTests {
 
 		assertThat(result).isNotNull();
 
-		// With external tool calling each chatModel.call() is independent, so only the
-		// last round's usage is reported (no cross-round accumulation).
-		assertThat(result.getMetadata().getUsage().getPromptTokens()).isEqualTo(300);
-		assertThat(result.getMetadata().getUsage().getCompletionTokens()).isEqualTo(30);
-		assertThat(result.getMetadata().getUsage().getTotalTokens()).isEqualTo(330);
+		// ToolCallingAdvisor accumulates usage across framework-controlled tool
+		// calling rounds.
+		assertThat(result.getMetadata().getUsage().getPromptTokens()).isEqualTo(500);
+		assertThat(result.getMetadata().getUsage().getCompletionTokens()).isEqualTo(80);
+		assertThat(result.getMetadata().getUsage().getTotalTokens()).isEqualTo(580);
 
-		// Verify cache metrics in native usage object (last round only)
-		Object nativeUsage = result.getMetadata().getUsage().getNativeUsage();
-		assertThat(nativeUsage).isInstanceOf(TokenUsage.class);
-
-		TokenUsage tokenUsage = (TokenUsage) nativeUsage;
-		assertThat(tokenUsage.cacheReadInputTokens()).isEqualTo(150);
-		assertThat(tokenUsage.cacheWriteInputTokens()).isEqualTo(0);
+		// Verify cache metrics in the accumulated Usage abstraction.
+		assertThat(result.getMetadata().getUsage().getCacheReadInputTokens()).isEqualTo(300L);
+		assertThat(result.getMetadata().getUsage().getCacheWriteInputTokens()).isEqualTo(0L);
 	}
 
 	public record Request(String location, String unit) {
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/StructuredOutputValidationAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/StructuredOutputValidationAdvisor.java
@@ -139,7 +139,7 @@ public final class StructuredOutputValidationAdvisor implements CallAdvisor, Str
 
 			// After Call
 			ChatResponse chatResponse = chatClientResponse.chatResponse();
-			usageAccumulator.add(chatResponse);
+			usageAccumulator.addRoundResponse(chatResponse);
 
 			// We should not validate tool call requests, only the content of the final
 			// response.
@@ -175,7 +175,7 @@ public final class StructuredOutputValidationAdvisor implements CallAdvisor, Str
 		}
 		while (!isValidationSuccess && repeatCounter <= this.maxRepeatAttempts);
 
-		return usageAccumulator.applyTotalUsage(chatClientResponse);
+		return usageAccumulator.applyAccumulatedUsage(chatClientResponse);
 	}
 
 	@SuppressWarnings("null")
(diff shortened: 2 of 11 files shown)


ilayaperumalg:
Hi @jewoodev, Please refer to the contributor guidelines as mentioned [here](https://github.com/spring-projects/spring-ai/blob/main/CONTRIBUTING.md#submit-a-pull-request). When the issue is assigned, Could you please mention at the issue if you are working on it? This will avoid us to avoid duplicate work. I had assigned this issue to myself and started working on it. Now, I see your PR :-) I will go ahead and review your PR. Next time, please check the issue first, and follow the guideline. Thanks!

jewoodev:
Thanks for pointing this out, Ilayaperumal.

Sorry for the duplicate work. I had already been working on a fix locally, but I should have checked the issue state again and left a note on the issue before opening the PR.

I'll follow that process next time. Thanks for still taking the time to review this.

jewoodev:
One more note: I found an additional streaming usage edge case while reviewing the final diff, so I'm going to update the PR before review.

Could you please hold off on reviewing this for a bit? I'll push a follow-up update shortly. Thanks!

jewoodev:
I pushed an update and revised the PR description to match the implementation more precisely.

The previous PR body described the streaming path too broadly as applying cumulative usage to final response metadata. More precisely, the loop accumulator is updated from each streamed iteration's aggregate response. When a later iteration emits chunks, only chunks that already carry usage metadata are rewritten to include the prior loop accumulator; chunks without usage metadata are left unchanged.

This update also makes the stream accumulator subscription-local and adds coverage for that case.

Validation:

```
./mvnw -pl spring-ai-client-chat -am test -Dtest=ToolCallingAdvisorTests,ToolCallingAdvisorSpanHierarchyTests -Dsurefire.failIfNoSpecifiedTests=false -Dmaven.build.cache.enabled=false
./mvnw -pl spring-ai-client-chat -am package -Dmaven.build.cache.enabled=false
```

tzolov:
@jewoodev @ilayaperumalg I've pushed a 2c206b424db92a8eb19540230aa3636177d8c5c0 that heavily refactors this PR. 
The point is that this issues is common for any recursive-advisor impl. So I've extracted the common logic into utils and applied to the ToolCallingAdvisor and the StructuredOutputValidatorAdvisor. 
Also fixed some edge cases in the streaming use-case to match the old behavior and updated the reference docs. 

Please review my changes.

jewoodev:
Thanks for extracting `UsageAccumulator`; that refactoring makes the follow-up changes easier to apply.

While reviewing the recursive-advisor coverage again, I noticed that `StructuredOutputValidationAdvisor` was still not using the shared accumulation path, so I added that missing coverage.

I also clarified the helper API names and docs so the responsibility split is easier to follow: `UsageCalculator` handles stateless usage arithmetic and metadata copying, while `UsageAccumulator` owns recursive-advisor loop state and final response stamping.

Please review these changes when you get a chance.

tzolov:
Thanks for the update @jewoodev,
Yesterday, I missed to push the commit with the StructuredOutputValidationAdvisor change, but you have filled the gap. Other naming improvements make sense to.

tzolov (approved):
Thanks @jewoodev , it LGTM

PR #6434, opened 2026-06-16
Align `Document` Jackson deserialization with serialized properties

### LINKED ISSUE #6433: `@JsonProperty` name incorrect on Document constructor
**Bug description**
The `Document` class' constructor has a `@JsonProperty` with a name "content", but it should be "text" to align with the actual property's name. Because of this, when a `Document` is serialized to JSON, the JSON has a "text" field, but can't be deserialized back into the "content" field.

Affects Spring AI 2.0.0 (and several versions back of 1.1.x).

To reproduce, create a `Document` object, serialize it to JSON, then try to deserialize it back into a `Document`. It will fail because of the content/text mismatch.

qixiangme:
## Summary

Align `Document` Jackson deserialization with the properties emitted
during serialization.

Fixes #6433

## Problem

`Document` serializes text content as `text`, but deserialization used
a Jackson creator that expected `content`. That breaks round-trip
serialization/deserialization and does not align deserialization with
serialized properties such as `id`, `metadata`, and `score`.

## Changes

- replace the Jackson creator with a property-based creator for `id`,
  `text`, `media`, `metadata`, and `score`
- keep Jackson-only defaults for missing `id` and `metadata`
- preserve the existing validation behavior of the public constructors
- add a regression test covering `Document` round-trip serialization
  and deserialization

## Testing

- `./mvnw -pl spring-ai-commons -Dmaven.build.cache.enabled=false test`
- `./mvnw -ntp -B -U -Dmaven.build.cache.enabled=false package`

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/document/Document.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/document/Document.java
@@ -126,39 +126,50 @@ public class Document {
 	@JsonIgnore
 	private ContentFormatter contentFormatter = DEFAULT_CONTENT_FORMATTER;
 
-	@JsonCreator(mode = JsonCreator.Mode.PROPERTIES)
-	public Document(@JsonProperty("content") @Nullable String content) {
-		this(content, new HashMap<>());
+	public Document(@Nullable String text) {
+		this(text, new HashMap<>());
 	}
 
 	public Document(@Nullable String text, Map<String, Object> metadata) {
-		this(new RandomIdGenerator().generateId(), text, null, metadata, null);
+		this(new RandomIdGenerator().generateId(), text, null, metadata, null, false);
 	}
 
 	public Document(String id, @Nullable String text, Map<String, Object> metadata) {
-		this(id, text, null, metadata, null);
+		this(id, text, null, metadata, null, false);
 	}
 
 	public Document(@Nullable Media media, Map<String, Object> metadata) {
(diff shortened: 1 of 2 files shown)


spring-ai-commons/src/main/java/org/springframework/ai/document/Document.java
```diff
@@ -126,39 +126,50 @@ public class Document {
...
+	@JsonCreator(mode = JsonCreator.Mode.PROPERTIES)
+	private Document(@JsonProperty("id") @Nullable String id, @JsonProperty("text") @Nullable String text,
+			@JsonProperty("media") @Nullable Media media,
+			@JsonProperty("metadata") @Nullable Map<String, Object> metadata,
+			@JsonProperty("score") @Nullable Double score) {
+		this(id, text, media, metadata, score, true);
+	}
+
+	private Document(@Nullable String id, @Nullable String text, @Nullable Media media,
+			@Nullable Map<String, Object> metadata, @Nullable Double score, boolean applyJacksonDefaults) {
```
ilayaperumalg, line 158: Not sure if we want to introduce this flag. We are currently discussing a better design to handle this. I will keep you posted.

ilayaperumalg:
@qixiangme After discussing with the team, we have addressed this issue via https://github.com/spring-projects/spring-ai/commit/fe4e4069256198fca49ee0ff37dd7ca366859b74. Hope that makes sense. Please share your feedback. Thank you!

PR #6445, opened 2026-06-17
Support page ranges in `PagePdfDocumentReader`

### LINKED ISSUE #6353: Support specific page ranges in `PagePdfDocumentReader`
**Expected Behavior**
It would be highly beneficial if `PagePdfDocumentReader` supported extracting text from specific page ranges rather than processing the entire PDF document. This feature would allow users to easily define which pages to parse, effectively skipping useless pages such as covers, tables of contents, copyright pages, blank pages, and appendices.

Ideally, this could be configured via `PdfDocumentReaderConfig`. The page range data type could be an array of integer pairs (e.g., `int[][]` like `{{3, 50}, {55, 100}}`) or a string expression.

**Code Example:**
```java
PdfDocumentReaderConfig config = PdfDocumentReaderConfig.builder()
    // Only process pages 3-50 and 55-100, skipping the cover, TOC, and blank pages
    .withPageRanges(new int[][]{{3, 50}, {55, 100}}) 
    .withPagesPerDocument(1)
    .build();

PagePdfDocumentReader reader = new PagePdfDocumentReader(new FileSystemResource("document.pdf"), config);
List<Document> documents = reader.get(); 
```

**Current Behavior**
Currently, `PagePdfDocumentReader` iterates through the entire `PDPageTree` from the first page to the last page. The `PdfDocumentReaderConfig` only provides `pagesPerDocument` to group parsed pages, but there is no way to skip or filter out unwanted pages during the parsing phase.

Furthermore, attempting to extend `PagePdfDocumentReader` to implement this feature manually is difficult because critical internal methods like `handleSinglePage` and the `config` field are marked as `private`, forcing developers to copy-paste and rewrite the core extraction logic.

**Context**
- **Why do you need this feature?** In enterprise RAG (Retrieval-Augmented Generation) pipelines, PDFs often contain irrelevant information such as covers, tables of contents, prefaces, blank pages, or appendices. Parsing these pages introduces significant noise into the vector database, degrades retrieval accuracy, and wastes Embedding API costs and LLM context windows. We need a native way to define page ranges to exclude these sections and improve data quality.
- **Alternatives considered:** I considered subclassing `PagePdfDocumentReader` and overriding the `get()` method. However, due to the `private` access modifiers on `handleSinglePage` and `config`, the subclass ends up duplicating a lot of boilerplate code, which is not maintainable.
- **Workarounds:** The only current workaround is to use external tools (like `qpdf` or `pdftk`) to physically split the PDF or remove specific pages before feeding them to Spring AI, which adds unnecessary I/O overhead and complexity to the ETL pipeline.

---

**Expected Behavior (期望行为)**
如果 `PagePdfDocumentReader` 能够支持仅提取特定页码范围的文本，而不是强制解析整个 PDF 文档，那将非常实用。此功能将允许用户自定义需要解析的页面，从而轻松跳过无用的页面，如封面、目录、版权页、空白页和附录。

理想情况下，这可以通过 `PdfDocumentReaderConfig` 进行配置。页码范围的数据类型可以是整数对数组（例如 `int[][]`，如 `{{3, 50}, {55, 100}}`）或字符串表达式。

**代码示例：**
```java
PdfDocumentReaderConfig config = PdfDocumentReaderConfig.builder()
    // 仅处理 3-50 页和 55-100 页，跳过封面、目录和空白页
    .withPageRanges(new int[][]{{3, 50}, {55, 100}}) 
    .withPagesPerDocument(1)
    .build();

PagePdfDocumentReader reader = new PagePdfDocumentReader(new FileSystemResource("document.pdf"), config);
List<Document> documents = reader.get(); 
```

**Current Behavior (当前行为)**
目前，`PagePdfDocumentReader` 会遍历整个 `PDPageTree`，从第一页一直处理到最后一页。`PdfDocumentReaderConfig` 仅提供了 `pagesPerDocument` 用于将解析后的页面分组，但无法在解析阶段跳过或过滤掉不需要的页面。

此外，尝试通过继承 `PagePdfDocumentReader` 来手动实现此功能非常困难，因为关键的内部方法（如 `handleSinglePage`）和 `config` 属性都被标记为 `private`，这迫使开发者必须复制并重写核心的文本提取逻辑。

**Context (背景/上下文)**
- **为什么需要这个功能？** 在企业级 RAG（检索增强生成）数据管道中，PDF 文件通常包含大量无关信息，如封面、目录、前言、空白页或附录。解析这些页面会给向量数据库引入严重的噪音，降低检索准确率，并白白浪费 Embedding API 的成本和 LLM 的上下文窗口。我们需要一种原生的方式来定义页码范围，以剔除这些无关部分，从而提升数据质量。
- **考虑过的替代方案：** 我曾考虑继承 `PagePdfDocumentReader` 并重写 `get()` 方法。但是，由于 `handleSinglePage` 和 `config` 的 `private` 访问修饰符，子类最终会重复大量样板代码，非常不利于维护。
- **目前的变通方法（Workarounds）：** 目前唯一的变通方法是在将 PDF 喂给 Spring AI 之前，使用外部工具（如 `qpdf` 或 `pdftk`）物理切割 PDF 或删除特定页面。这给 ETL 管道增加了不必要的 I/O 开销和复杂性。

subhashpolisetti:
Adds an optional page-range filter so you can read only certain pages of a PDF (e.g. skip the cover, TOC, or appendix) instead of always parsing the whole document.

### Why

`PagePdfDocumentReader` always reads every page. `pagesPerDocument` only controls how parsed pages are grouped, not which ones get read, so there's no clean way to skip pages. Subclassing to do it is awkward because the page loop is internal.

This is a coarse, manual filter on physical page numbers, for when you already know which pages you want. It's not meant to replace content-based page classification (per the discussion on the issue).

### What changed

- New `addPageRange(int startPage, int endPage)` on `PdfDocumentReaderConfig.Builder`. Inclusive, 1-based; call it multiple times to read multiple, possibly disjoint, ranges (their union is read).
- Ranges past the end of the document are clamped to the last page; overlapping ranges are deduped so each page is read once.
- Builder validates each range (`startPage >= 1`, `endPage >= startPage`).
- `PagePdfDocumentReader` keeps its single sequential pass over the page tree. Out-of-range pages are skipped, so there's no perf hit and no change in output when no range is set.
- With disjoint ranges, a Document never spans an excluded gap: each contiguous run of selected pages becomes its own Document, so the `page_number`/`end_page_number` metadata stays accurate (`pagesPerDocument` acts as a maximum across gaps).
- The setting lives on the shared `PdfDocumentReaderConfig` but is honored only by `PagePdfDocumentReader`. `ParagraphPdfDocumentReader` does not support page ranges and now fails fast if a config sets them.

### Example

```java
PdfDocumentReaderConfig config = PdfDocumentReaderConfig.builder()
    // read only pages 3-50 and 55-100
    .addPageRange(3, 50)
    .addPageRange(55, 100)
    .withPagesPerDocument(1)
    .build();

List<Document> documents =
    new PagePdfDocumentReader(new FileSystemResource("document.pdf"), config).get();
```

### Tests

Added cases for a single range, disjoint ranges, all-pages grouping with a range, upper-bound clamping, overlap dedupe, disjoint ranges splitting at the gap under all-pages-per-document, a content-level check that filtered output matches the corresponding unfiltered pages, invalid input, and rejection by `ParagraphPdfDocumentReader`. Existing tests are unchanged.

```
./mvnw -pl document-readers/spring-ai-pdf-document-reader clean test \
  io.spring.javaformat:spring-javaformat-maven-plugin:validate
[INFO] You have 0 Checkstyle violations.
[INFO] Tests run: 41, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Closes #6353

### DIFF
--- a/document-readers/spring-ai-pdf-document-reader/src/main/java/org/springframework/ai/reader/pdf/config/PdfDocumentReaderConfig.java
+++ b/document-readers/spring-ai-pdf-document-reader/src/main/java/org/springframework/ai/reader/pdf/config/PdfDocumentReaderConfig.java
@@ -16,6 +16,9 @@
 
 package org.springframework.ai.reader.pdf.config;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import org.springframework.ai.reader.ExtractedTextFormatter;
 import org.springframework.ai.reader.pdf.PagePdfDocumentReader;
 import org.springframework.ai.reader.pdf.ParagraphPdfDocumentReader;
@@ -41,12 +44,21 @@ public final class PdfDocumentReaderConfig {
 
 	public final ExtractedTextFormatter pageExtractedTextFormatter;
 
+	/**
+	 * Inclusive, one-based page ranges to read, each stored as a {@code {start, end}}
+	 * pair. An empty array means the whole document is read. Currently honored only by
+	 * {@link PagePdfDocumentReader}; readers that do not support page ranges reject a
+	 * configuration that sets them.
+	 */
+	private final int[][] pageRanges;
+
 	private PdfDocumentReaderConfig(PdfDocumentReaderConfig.Builder builder) {
 		this.pagesPerDocument = builder.pagesPerDocument;
 		this.pageBottomMargin = builder.pageBottomMargin;
 		this.pageTopMargin = builder.pageTopMargin;
 		this.pageExtractedTextFormatter = builder.pageExtractedTextFormatter;
 		this.reversedParagraphPosition = builder.reversedParagraphPosition;
+		this.pageRanges = builder.pageRanges.toArray(new int[0][]);
 	}
 
 	/**
@@ -65,6 +77,34 @@ public final class PdfDocumentReaderConfig {
 		return builder().build();
 	}
 
+	/**
+	 * Whether the given one-based page number should be read. When no page ranges are
+	 * configured every page is read; otherwise a page is read when it falls within any of
+	 * the configured inclusive ranges.
+	 * @param pageNumber the one-based page number
+	 * @return {@code true} if the page should be read
+	 */
+	public boolean isPageSelected(int pageNumber) {
+		if (this.pageRanges.length == 0) {
+			return true;
+		}
+		for (int[] range : this.pageRanges) {
+			if (pageNumber >= range[0] && pageNumber <= range[1]) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	/**
+	 * Whether any page ranges are configured. Readers that do not support page ranges can
+	 * use this to reject such a configuration instead of silently ignoring it.
+	 * @return {@code true} if at least one page range has been configured
+	 */
+	public boolean hasPageRanges() {
+		return this.pageRanges.length > 0;
+	}
+
 	public static final class Builder {
 
 		private int pagesPerDocument = 1;
@@ -76,6 +116,8 @@ public final class PdfDocumentReaderConfig {
 		private ExtractedTextFormatter pageExtractedTextFormatter = ExtractedTextFormatter.defaults();
 
 		private boolean reversedParagraphPosition = false;
+
+		private final List<int[]> pageRanges = new ArrayList<>();
 
 		private Builder() {
 		}
@@ -134,6 +176,29 @@ public final class PdfDocumentReaderConfig {
 		 */
 		public Builder withReversedParagraphPosition(boolean reversedParagraphPosition) {
 			this.reversedParagraphPosition = reversedParagraphPosition;
+			return this;
+		}
+
+		/**
+		 * Restricts reading to only the given inclusive, one-based page range, allowing
+		 * callers to skip pages such as covers, tables of contents or appendices. May be
+		 * called multiple times to read multiple, possibly disjoint, ranges; their union
+		 * is read. A range whose {@code endPage} exceeds the document length is clamped
+		 * to the last page. When no range is added (the default), the whole document is
+		 * read.
+		 * <p>
+		 * Note: page ranges are a coarse, manual filter based on physical page numbers
+		 * and are intended for cases where the caller already knows which pages to
+		 * include. They are currently honored only by {@link PagePdfDocumentReader};
+		 * readers that do not support page ranges reject a configuration that sets them.
+		 * @param startPage the inclusive, one-based first page to read
+		 * @param endPage the inclusive, one-based last page to read
+		 * @return this builder
+		 */
+		public PdfDocumentReaderConfig.Builder addPageRange(int startPage, int endPage) {
+			Assert.isTrue(startPage >= 1, "startPage must be a positive (one-based) value.");
+			Assert.isTrue(endPage >= startPage, "endPage must be greater than or equal to startPage.");
+			this.pageRanges.add(new int[] { startPage, endPage });
 			return this;
 		}
 
--- a/document-readers/spring-ai-pdf-document-reader/src/main/java/org/springframework/ai/reader/pdf/PagePdfDocumentReader.java
+++ b/document-readers/spring-ai-pdf-document-reader/src/main/java/org/springframework/ai/reader/pdf/PagePdfDocumentReader.java
@@ -20,7 +20,6 @@ import java.awt.Rectangle;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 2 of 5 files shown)


document-readers/spring-ai-pdf-document-reader/src/main/java/org/springframework/ai/reader/pdf/config/PdfDocumentReaderConfig.java
```diff
@@ -137,6 +166,36 @@ public Builder withReversedParagraphPosition(boolean reversedParagraphPosition)
...
+		 * no ranges are supplied (the default), the whole document is read.
+		 * <p>
+		 * Note: page ranges are a coarse, manual filter based on physical page numbers
+		 * and are intended for cases where the caller already knows which pages to
+		 * include. They are currently honored by {@link PagePdfDocumentReader}.
+		 * @param pageRanges the inclusive, one-based {@code {startPage, endPage}} ranges
+		 * to read
+		 * @return this builder
+		 */
+		public PdfDocumentReaderConfig.Builder withPageRanges(int[]... pageRanges) {
```
ericbottard: Maybe `addPageRange(start, end)` (callable multiple times but with the drawback of not being able to reset) would be better? Otherwise, indeed introducing a PageRange record would be better, as this array of arrays is a bit unwieldy (from the POV of users at least, internal representation is ok)

document-readers/spring-ai-pdf-document-reader/src/main/java/org/springframework/ai/reader/pdf/config/PdfDocumentReaderConfig.java
```diff
@@ -137,6 +166,36 @@ public Builder withReversedParagraphPosition(boolean reversedParagraphPosition)
...
+		 * Restricts reading to the given inclusive, one-based page ranges, allowing
+		 * callers to skip pages such as covers, tables of contents or appendices. Each
+		 * range is a two-element {@code {startPage, endPage}} array; multiple, possibly
+		 * disjoint, ranges may be supplied and their union is read. A range whose
+		 * {@code endPage} exceeds the document length is clamped to the last page. When
+		 * no ranges are supplied (the default), the whole document is read.
+		 * <p>
+		 * Note: page ranges are a coarse, manual filter based on physical page numbers
+		 * and are intended for cases where the caller already knows which pages to
+		 * include. They are currently honored by {@link PagePdfDocumentReader}.
```
ericbottard: Maybe add the word "only" in this description.

The readers that don't support this feature should fail early if the value is set, too.

ericbottard (changes requested):

ericbottard:
Hi @subhashpolisetti,

thanks for your contribution. Looks good overall, I left a couple of comments, should be easy to address

subhashpolisetti:
Thanks for the review, Eric  all good suggestions. Updated the PR:

API: replaced withPageRanges(int[]...) with addPageRange(int startPage, int endPage) (accumulating, callable multiple times). Cleaner from the caller's side; the internal representation stays int[][].

Javadoc: reworded to "Restricts reading to only the given … range".

Fail-fast: ParagraphPdfDocumentReader now rejects a config that sets page ranges (it's the only reader that doesn't support them), with a message pointing to PagePdfDocumentReader. Added a test.

Default behavior is unchanged when no range is set. Happy to switch to a PageRange record instead of the accumulating method if you'd prefer.

document-readers/spring-ai-pdf-document-reader/src/main/java/org/springframework/ai/reader/pdf/config/PdfDocumentReaderConfig.java
subhashpolisetti: Done -- switched to addPageRange(int startPage, int endPage). Happy to use a PageRange record instead if you prefer.

document-readers/spring-ai-pdf-document-reader/src/main/java/org/springframework/ai/reader/pdf/config/PdfDocumentReaderConfig.java
subhashpolisetti: Done -- added "only" to the wording, and ParagraphPdfDocumentReader now fails fast if page ranges are set (with a test).

fw-coder-tl (commented):
Thanks for putting this together. I checked out commit `64f53667c` and ran:

```shell
JAVA_HOME=$(/usr/libexec/java_home -v 21) ./mvnw -am -pl document-readers/spring-ai-pdf-document-reader test
```

All 39 tests passed locally.

One behavior seems worth clarifying before I propose a patch: when disjoint ranges are combined with `pagesPerDocument > 1` (especially `ALL_PAGES`), selected pages on opposite sides of a gap are grouped into the same `Document`. For example, selecting only pages 1 and 64 with `ALL_PAGES` produces metadata `page_number=1` and `end_page_number=64`, even though pages 2-63 were excluded. Downstream consumers cannot tell from the existing metadata whether that span is contiguous.

Would it be preferable to flush the current group when the next selected page is non-contiguous? That keeps the existing start/end metadata accurate, at the cost of treating `pagesPerDocument` as a maximum for disjoint ranges. An alternative would be new metadata describing the exact selected ranges, but that feels broader than this PR.

Two smaller suggestions:

- Track the selected-page count and log something like `Processed 3 of 64 pages`; currently range-filtered runs still log `Processed total 64 pages`.
- Add a content-level assertion comparing filtered page text with the corresponding unfiltered pages. The current range tests mainly verify document counts and metadata, so they do not directly guard against excluded page text leaking into output.

I am happy to implement the agreed behavior and tests once we settle the grouping semantics.

subhashpolisetti:
Thanks for the thorough review and for running it locally. Agreed on the grouping — merging non-contiguous pages into one Document made the start/end metadata misleading. Pushed:

Flush the current group when the next selected page isn't contiguous, so a Document never spans a gap and its start/end page numbers stay accurate. As you noted, pagesPerDocument now acts as a maximum across gaps.

Log the selected-page count instead of the total on range-filtered runs.
Added two tests: disjoint ranges + all-pages-per-document now split at the gap (with correct metadata), and a content-level check that a range-filtered read matches the same pages from a full read.

I kept the "start/end stays accurate" behavior rather than adding new range metadata, to keep this focused -- happy to go the metadata route instead if the maintainers prefer. Thanks again.

fw-coder-tl (approved):
Thanks for addressing these points. I pulled commit `33bf4b6c3` and reran the PDF reader module tests on JDK 21 with the Maven build cache disabled.

All 41 tests passed, and Checkstyle reported 0 violations. I also reviewed the gap-flush behavior, selected-page logging, and the new content-level coverage; they all align with the discussion and keep the existing metadata semantics clear.

No further issues from my side.

subhashpolisetti:
Thanks for the careful review and for re-running everything — appreciate it.

ericbottard (approved):

ericbottard:
Thanks for this great collaboration @subhashpolisetti @fw-coder-tl

subhashpolisetti:
Thanks @ericbottard and @fw-coder-tl — appreciated the thorough review and the collaboration. Glad to contribute.

PR #6448, opened 2026-06-17
Add ContextAwareToolCallback and ToolContextHolder

zz-zhi54:
## Summary

Add `ContextAwareToolCallback` and `ToolContextHolder` so that `@Tool` methods whose signatures do **not** declare a `ToolContext` parameter can still access the `ToolContext` (e.g. user id, tenant id, trace id).

- `ToolContextHolder` — an `InheritableThreadLocal` holder for the `ToolContext` flowing through a tool call invocation.
- `ContextAwareToolCallback` — a `ToolCallback` decorator that publishes the `ToolContext` received in `call(String, ToolContext)` to the holder for the duration of the delegate invocation. The holder is cleared automatically in a `finally` block.

## Motivation

Today, if a `@Tool` method does not declare a `ToolContext` parameter, the framework has no way to surface the `ToolContext` to it. This is especially limiting for tools defined in third-party libraries whose method signatures cannot be modified — for example tools provided via MCP or by community projects like [spring-ai-agent-utils](https://github.com/spring-ai-community/spring-ai-agent-utils).

This change introduces a thread-local bridge: the decorator binds the `ToolContext` for the duration of the delegate call, and the tool method can read it via `ToolContextHolder.get()`. No changes to `MethodToolCallback` or the `ToolCallback` interface are required.

## Usage

```java
ToolCallback wrapped = new ContextAwareToolCallback(originalCallback);
```

```java
@Tool
String myTool(String input) {   // no ToolContext parameter needed
    ToolContext ctx = ToolContextHolder.get();
    String userId = ctx != null ? (String) ctx.getContext().get("userId") : null;
    return "hello " + userId;
}
```

## Testing

- 13 new unit tests covering: holder behaviour, constructor validation, single-arg and two-arg `call`, holder clearing on exception, and an integration scenario.
- `./mvnw -pl spring-ai-model process-sources` — 0 checkstyle violations.
- `./mvnw -pl spring-ai-model -Dtest=ContextAwareToolCallbackTest test` — all tests pass.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/tool/context/ContextAwareToolCallback.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/tool/context/ContextAwareToolCallback.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.tool.context;
+
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.ai.chat.model.ToolContext;
+import org.springframework.ai.tool.ToolCallback;
+import org.springframework.ai.tool.definition.ToolDefinition;
+import org.springframework.ai.tool.metadata.ToolMetadata;
+import org.springframework.util.Assert;
+
+/**
+ * A {@link ToolCallback} decorator that publishes the {@link ToolContext} received in
+ * {@link #call(String, ToolContext)} to the current thread's {@link ToolContextHolder}
+ * for the duration of the delegate invocation.
+ *
+ * <p>
+ * This allows {@code @Tool} methods whose signatures do <strong>not</strong> declare a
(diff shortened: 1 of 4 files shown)


ericbottard:
I'll let other members of the team chime in, but I don't understand the motivation for this change.

If the argument is that tools are in third party code that can't be changed, then how can they make use of `ToolContext ctx = ToolContextHolder.get();`, or even use the `@Tool` annotation ?

Moreover, I'm very wary of code that relies on ThreadLocals. More often than not it brings limitations down the road (assuming everything runs on the same thread throughout, etc).

So in conclusion, given that this particular behavior can be implemented on a per-need _ad hoc_ basis, I will close this pull request.

zz-zhi54:
Thanks for the review @ericbottard. I realize my PR description was misleading — let me clarify the actual motivation.                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                               
  This isn't meant for third-party @Tool methods to call ToolContextHolder.get(). The real scenario is from spring-ai-community/spring-ai-agent-utils#57: framework-level tools like AskUserQuestionTool and TodoWriteTool need access to the ToolContext during a tool call. These are framework callbacks, not third-party @Tool methods, and currently      
  there's no clean way for them to obtain the context they need.                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                               
  I agree ThreadLocal has drawbacks with virtual threads and structured concurrency. Happy to explore alternatives like ScopedValue if you think this direction is worth pursuing.

zz-zhi54:
@ericbottard  I really like this project. If you believe this doesn't belong in Spring AI but rather in `spring-ai-community`, I will close this issue—but I am indeed encountering this problem right now.

zz-zhi54:
I agree completely with what you said. While the lifecycle of this `ThreadLocal` is fully controllable, it would indeed make future extensibility very poor; however, if you believe this PR is valid, I will give further thought to alternatives to `ThreadLocal`.

ericbottard:
I would suggest that we consider adding the context to `AskUserQuestionTool` and propagate from there then.
cc @tzolov

PR #6452, opened 2026-06-17
Validate `TokenTextSplitter` builder arguments

### LINKED ISSUE #6447: `TokenTextSplitter` enters an infinite loop when `chunkSize` is 0
**Bug description**

`TokenTextSplitter.Builder` doesn't validate `chunkSize`. If it's set to `0`, `doSplit(...)` never terminates — it hangs in an infinite loop. A negative value fails with a confusing internal message instead of a clear error.

**Affected code**

`spring-ai-commons/src/main/java/org/springframework/ai/transformer/splitter/TokenTextSplitter.java`, in `doSplit`:

```java
while (!tokens.isEmpty() && num_chunks < this.maxNumChunks) {
    List<Integer> chunk = tokens.subList(0, Math.min(chunkSize, tokens.size())); // min(0, N) = 0 -> empty list
    String chunkText = decodeTokens(chunk);                                       // decode([]) = ""
    if (chunkText.trim().isEmpty()) {
        tokens = tokens.subList(chunk.size(), tokens.size());                     // chunk.size() == 0 -> tokens unchanged
        continue;                                                                 // num_chunks NOT incremented
    }
    ...
    num_chunks++;
}
```

With `chunkSize == 0`, `chunk` is always empty, `tokens` never shrinks, and `num_chunks` is never incremented, so the `while` loop can never exit.

**Steps to reproduce / Minimal example**

```java
var splitter = TokenTextSplitter.builder().withChunkSize(0).build();
splitter.apply(List.of(new Document("any non-empty text"))); // hangs forever
```

Negative size:

```java
TokenTextSplitter.builder().withChunkSize(-1).build()
    .apply(List.of(new Document("any non-empty text")));
// throws java.lang.IllegalArgumentException: fromIndex(0) > toIndex(-1)
```

**Expected behavior**

A non-positive `chunkSize` should fail fast with a clear `IllegalArgumentException` at build time — consistent with the existing `encodingType must not be null` check in the same constructor — instead of hanging or throwing a confusing internal error.

**Environment**

- Spring AI version: `2.0.0-SNAPSHOT` (`main`)
- Java version: 21

I'm happy to submit a PR adding the validation (with tests) if this looks reasonable.

subhashpolisetti:
Fixes an infinite loop in `TokenTextSplitter` and adds fail-fast validation for the builder's numeric arguments.

### Why

`TokenTextSplitter.Builder` doesn't validate `chunkSize`. With `chunkSize == 0`, `doSplit(...)` loops forever: the chunk sublist is empty, so the remaining tokens never shrink and `num_chunks` is never incremented. A negative `chunkSize` fails later with a confusing `IllegalArgumentException: fromIndex(0) > toIndex(-1)` from `List.subList`.

```java
// hangs forever
TokenTextSplitter.builder().withChunkSize(0).build()
    .apply(List.of(new Document("any non-empty text")));
```

### What changed

- Validate the numeric builder arguments in the constructor, next to the existing `encodingType` check:
  - `chunkSize` must be greater than zero
  - `maxNumChunks` must be greater than zero
  - `minChunkSizeChars` must not be negative
  - `minChunkLengthToEmbed` must not be negative
- Misconfiguration now fails fast at build time with a clear message instead of hanging or throwing an internal error.
- The checks run once at construction, so the hot splitting path (`doSplit`) is unchanged — no runtime/throughput impact.

### Tests

Added regression tests for each rejected argument (matching the existing `testTokenTextSplitterWithNullEncodingTypeThrows` style) plus a `chunkSize == 1` boundary test proving the smallest valid value terminates and produces chunks. Existing tests are unchanged.

```
./mvnw -pl spring-ai-commons clean test -Dtest=TokenTextSplitterTest \
  io.spring.javaformat:spring-javaformat-maven-plugin:validate
[INFO] You have 0 Checkstyle violations.
[INFO] Tests run: 13, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Closes #6447

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/transformer/splitter/TokenTextSplitter.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/transformer/splitter/TokenTextSplitter.java
@@ -121,6 +121,10 @@ public class TokenTextSplitter extends TextSplitter {
 	private TokenTextSplitter(EncodingType encodingType, int chunkSize, int minChunkSizeChars,
 			int minChunkLengthToEmbed, int maxNumChunks, boolean keepSeparator, List<Character> punctuationMarks) {
 		Assert.notNull(encodingType, "encodingType must not be null");
+		Assert.isTrue(chunkSize > 0, "chunkSize must be greater than zero");
+		Assert.isTrue(maxNumChunks > 0, "maxNumChunks must be greater than zero");
+		Assert.isTrue(minChunkSizeChars >= 0, "minChunkSizeChars must not be negative");
+		Assert.isTrue(minChunkLengthToEmbed >= 0, "minChunkLengthToEmbed must not be negative");
 		this.encoding = this.registry.getEncoding(encodingType);
 		this.chunkSize = chunkSize;
 		this.minChunkSizeChars = minChunkSizeChars;
--- a/spring-ai-commons/src/test/java/org/springframework/ai/transformer/splitter/TokenTextSplitterTest.java
+++ b/spring-ai-commons/src/test/java/org/springframework/ai/transformer/splitter/TokenTextSplitterTest.java
@@ -216,6 +216,51 @@ public class TokenTextSplitterTest {
 	}
 
 	@Test
+	public void testTokenTextSplitterWithZeroChunkSizeThrows() {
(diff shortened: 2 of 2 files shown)


ericbottard (approved):
Thanks for this useful contribution @subhashpolisetti

PR #6453, opened 2026-06-18
Fix double-encoding of String tool results in MCP content[].text

### LINKED ISSUE #6446: MCP tool String responses are double-encoded in content[].text
**Bug description**  
When using Spring AI MCP tools, a tool method returning a plain `String` is serialized as JSON before being written into MCP `content[].text`. This causes the text payload to contain extra quotes/escaping.

Example tool:

```java
@Tool(description = "Returns markdown")
public String jwtDecode(String token) {
    return "# Header";
}
```

Actual MCP response content:

```json
{
  "type": "text",
  "text": "\"# Header\""
}
```

Expected MCP response content:

```json
{
  "type": "text",
  "text": "# Header"
}
```

This looks like the tool return value is being passed through Jackson `writeValueAsString(...)` and then stored in the MCP `text` field, which results in double encoding for plain text responses.

I did not find an existing issue for this exact MCP tool `String` response behavior, but it may already be known under a broader MCP serialization issue.

---

**Environment**  
- Spring AI version: `2.0`
- Spring Boot version: `4.1`
- Java version: `25`
- OS: `WSL`

---

**Steps to reproduce**  
1. Create a Spring AI MCP tool that returns a plain `String`.
2. Example:

   ```java
   @Tool(description = "Returns markdown")
   public String jwtDecode(String token) {
       return "# Header";
   }
   ```

3. Invoke the tool through MCP.
4. Inspect the JSON-RPC response.
5. Observe that the returned content is:

   ```json
   "text": "\"# Header\""
   ```

   instead of:

   ```json
   "text": "# Header"
   ```

---

**Expected behavior**  
If a tool returns a Java `String`, the MCP response should place that exact value into `content[].text`.

Expected:

```json
{
  "result": {
    "content": [
      {
        "type": "text",
        "text": "# Header"
      }
    ],
    "isError": false
  }
}
```

Actual behavior appears to JSON-encode the string first, producing an extra quoting layer.

---

```java
import org.springframework.stereotype.Component;

@Component
public class MyTools {

    @Tool(description = "Returns markdown")
    public String jwtDecode(String token) {
        return "# Header";
    }
}
```

Observed response:

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "\"# Header\""
      }
    ],
    "isError": false
  }
}
```

Relevant implementation appears to be related to unconditional JSON serialization of non-forwarded strings:

```java
public String toJson(@Nullable Object object, boolean forwardIfValidJson) {
    if (forwardIfValidJson && object instanceof String str) {
        if (this.isValidJson(str)) {
            return str;
        }
    }

    return this.jsonMapper.writeValueAsString(object);
}
```

For plain text strings, this produces a JSON string literal, which seems incorrect when the target MCP field is already a text field.

---

i-zoufeng:
A tool returning a plain String is JSON-encoded by the default ToolCallResultConverter, so the MCP server placed the quoted JSON literal (e.g. the two-character header markup wrapped in quotes) into content[].text instead of the raw text. Decode JSON string scalars back to raw text when building TextContent; JSON objects/arrays/numbers and non-JSON text are left unchanged. Chat tool-calling behavior is untouched.

Fixes gh-6446

Thank you upfront!

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/util/JsonHelper.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/util/JsonHelper.java
@@ -25,6 +25,7 @@ import org.jspecify.annotations.Nullable;
 import tools.jackson.core.JacksonException;
 import tools.jackson.core.type.TypeReference;
 import tools.jackson.databind.DatabindException;
+import tools.jackson.databind.JsonNode;
 import tools.jackson.databind.json.JsonMapper;
 
 import org.springframework.core.ParameterizedTypeReference;
@@ -153,6 +154,32 @@ public class JsonHelper {
 		catch (JacksonException e) {
 			return false;
 		}
+	}
+
+	/**
+	 * If the given value is a JSON string scalar (e.g. {@code "\"text\""}), returns the
+	 * decoded raw text ({@code text}); otherwise returns the value unchanged.
+	 * <p>
+	 * This is the inverse of the String encoding performed by
+	 * {@link #toJson(Object, boolean)} and is used to avoid double-encoding plain-text
+	 * tool results placed into MCP {@code content[].text}. JSON objects, arrays, numbers,
+	 * booleans and non-JSON text are all returned unchanged.
+	 * @param value the (possibly JSON-encoded) value to decode
+	 * @return the raw text if {@code value} is a JSON string scalar, otherwise
+	 * {@code value} unchanged
+	 */
+	public String decodeJsonStringScalar(String value) {
+		Assert.notNull(value, "value cannot be null");
+		try {
+			JsonNode node = this.jsonMapper.readTree(value);
+			if (node != null && node.isString()) {
+				return node.asString();
+			}
+		}
+		catch (JacksonException ex) {
+			// Not JSON: return the original value unchanged.
+		}
+		return value;
 	}
 
 	/**
--- a/spring-ai-commons/src/main/java/org/springframework/ai/util/JsonHelper.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/util/JsonHelper.java
@@ -25,7 +25,6 @@ import org.jspecify.annotations.Nullable;
 import tools.jackson.core.JacksonException;
 import tools.jackson.core.type.TypeReference;
 import tools.jackson.databind.DatabindException;
-import tools.jackson.databind.JsonNode;
 import tools.jackson.databind.json.JsonMapper;
 
 import org.springframework.core.ParameterizedTypeReference;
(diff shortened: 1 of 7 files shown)


ericbottard:
Hi @i-zoufeng,

I have yet to dig into this more deeply, but looking at this from the outside, wouldn't it be better to NOT encode Strings in the first place. From what I understand from your comment (again, I haven't read the code thoroughly yet), Strings are handled differently from other types (arrays, numbers, etc). IMO we should be consistent (if possible)

spring-ai-commons/src/main/java/org/springframework/ai/util/JsonHelper.java
```diff
@@ -155,6 +156,32 @@ private boolean isValidJson(String input) {
...
+	 * <p>
+	 * This is the inverse of the String encoding performed by
+	 * {@link #toJson(Object, boolean)} and is used to avoid double-encoding plain-text
+	 * tool results placed into MCP {@code content[].text}. JSON objects, arrays, numbers,
+	 * booleans and non-JSON text are all returned unchanged.
+	 * @param value the (possibly JSON-encoded) value to decode
+	 * @return the raw text if {@code value} is a JSON string scalar, otherwise
+	 * {@code value} unchanged
+	 */
+	public String decodeJsonStringScalar(String value) {
```
ericbottard: I am very wary of these kind of utility methods. IMHO, calling code should know what it is dealing with.
I'm afraid the more we introduce these kind of helpers, the more the codebase will use them "because we never know" and the harder it is to reason about what type of results we're dealing with.

I would rather try to understand the issue from the beginning, seeing where we wrongfully assume something is JSON where it is not (or vice versa) and be consistent throughout (for Strings AND any other type)

i-zoufeng:
Thanks @ericbottard, that makes sense — you're right that decoding after the fact is the wrong layer. I've reworked the PR to fix it at the source.

Instead of decoding in the MCP path, DefaultToolCallResultConverter now returns String results directly without JSON-encoding them, so nothing downstream needs to un-quote anything. The JsonHelper.decodeJsonStringScalar helper and the MCP-side decode call are removed.

This also mirrors what the annotation-based path already does in AbstractMcpToolMethodCallback#convertValueToCallToolResult, which special-cases String and returns it verbatim — so the two tool-result paths are now consistent.

Worth noting the previous behavior was already inconsistent: toJson(result, true) only forwarded a String unchanged when it parsed as valid JSON, and since the mapper ignores trailing tokens, a result like "3 strings processed: [...]" (starting with 3) passed through raw while "# Header" got quoted. Returning all String results raw removes that inconsistency.

This does change the chat tool-calling path too (a String tool result is no longer JSON-quoted in the tool message). I updated FunctionToolCallbackTest, which was the only test asserting the old quoted behavior; the rest of the tool-calling tests already expected the raw value. All affected module tests pass locally.

spring-ai-model/src/main/java/org/springframework/ai/tool/execution/DefaultToolCallResultConverter.java
```diff
@@ -61,6 +61,13 @@ public String convert(@Nullable Object result, @Nullable Type returnType) {
 			return jsonHelper.toJson(Map.of("mimeType", "image/png", "data", imgB64));
 		}
 		else {
+			// Mirror AbstractMcpToolMethodCallback: return String results directly,
+			// without JSON serialization, to avoid double-encoding plain text (e.g. a
+			// tool returning "# Header" must not become the quoted literal "\"#
+			// Header\"").
+			if (result instanceof String string) {
```
ericbottard, line 68: I actually disagree with this.
All other codepaths of this method clearly return JSON, so this should return JSON too. If Java had type aliases, then the `String` return value from this method would be changed to `JSONAsAString`. Let's follow that type along and see where the inconsistency is introduced

ericbottard: To clarify my chain of thought: I think we should challenge WHY the `ToolCallResultConverter` is called in the first place. Maybe putting the `!instanceof String` check in the caller

tzolov:
Here are my thoughts: https://github.com/spring-projects/spring-ai/issues/6446#issuecomment-4778182981
IMO we  should not change the default impl.

ericbottard:
Hello @i-zoufeng,

after further discussion with the team, it appears the current behavior is intended and the OP has a simple workaround available.
I will this close this PR without merging. Thanks nevertheless for your interest in the matter.

PR #6458, opened 2026-06-18
Fix tool-search advisor auto-config ordering

tzolov:
Order ToolSearchAdvisorAutoConfiguration after ToolCallingAutoConfiguration so ToolCallingManager exists when its @ConditionalOnBean is evaluated. 
Without it, the tool-search builder backs off and ChatClientAutoConfiguration contributes the default ToolCallingAdvisor.Builder instead.

Add regression tests loading all three auto-configurations together.

### DIFF
--- a/auto-configurations/advisors/spring-ai-autoconfigure-tool-search-advisor/pom.xml
+++ b/auto-configurations/advisors/spring-ai-autoconfigure-tool-search-advisor/pom.xml
@@ -46,6 +46,16 @@
 			<optional>true</optional>
 		</dependency>
 
+		<!-- Optional compile-time dependency so @AutoConfiguration can reference
+		     ToolCallingAutoConfiguration.class for ordering (afterName equivalent).
+		     Marked optional to avoid leaking onto consumers' classpaths. -->
+		<dependency>
+			<groupId>org.springframework.ai</groupId>
+			<artifactId>spring-ai-autoconfigure-model-tool</artifactId>
+			<version>${project.parent.version}</version>
+			<optional>true</optional>
+		</dependency>
+
 		<!-- Optional: VectorStore support for VectorToolIndex auto-configuration -->
 		<dependency>
 			<groupId>org.springframework.ai</groupId>
@@ -75,14 +85,6 @@
 		<dependency>
 			<groupId>org.springframework.ai</groupId>
 			<artifactId>spring-ai-test</artifactId>
-			<version>${project.parent.version}</version>
-			<scope>test</scope>
-		</dependency>
-
-		<!-- Provides ToolCallingAutoConfiguration to verify auto-configuration ordering -->
-		<dependency>
-			<groupId>org.springframework.ai</groupId>
-			<artifactId>spring-ai-autoconfigure-model-tool</artifactId>
 			<version>${project.parent.version}</version>
 			<scope>test</scope>
 		</dependency>
--- a/auto-configurations/advisors/spring-ai-autoconfigure-tool-search-advisor/pom.xml
+++ b/auto-configurations/advisors/spring-ai-autoconfigure-tool-search-advisor/pom.xml
@@ -79,6 +79,14 @@
 			<scope>test</scope>
 		</dependency>
 
+		<!-- Provides ToolCallingAutoConfiguration to verify auto-configuration ordering -->
+		<dependency>
+			<groupId>org.springframework.ai</groupId>
+			<artifactId>spring-ai-autoconfigure-model-tool</artifactId>
+			<version>${project.parent.version}</version>
+			<scope>test</scope>
+		</dependency>
+
 		<dependency>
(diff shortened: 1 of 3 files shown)


auto-configurations/advisors/spring-ai-autoconfigure-tool-search-advisor/src/main/java/org/springframework/ai/chat/client/advisor/toolsearch/autoconfigure/ToolSearchAdvisorAutoConfiguration.java
```diff
@@ -74,7 +74,8 @@
  * @author Christian Tzolov
  * @since 2.0.0
  */
-@AutoConfiguration(beforeName = "org.springframework.ai.model.chat.client.autoconfigure.ChatClientAutoConfiguration")
+@AutoConfiguration(afterName = "org.springframework.ai.model.tool.autoconfigure.ToolCallingAutoConfiguration",
```
ericbottard: Please add comments here that justify each attribute

ericbottard:
> Order ToolSearchAdvisorAutoConfiguration after ToolCallingAutoConfiguration so ToolCallingManager exists when its @ConditionalOnBean is evaluated.

Your justification talks about ONE reason to order, yet the change adds TWO attributes. Could you please clarify?

auto-configurations/advisors/spring-ai-autoconfigure-tool-search-advisor/pom.xml
```diff
@@ -79,6 +79,14 @@
 			<scope>test</scope>
 		</dependency>
 
+		<!-- Provides ToolCallingAutoConfiguration to verify auto-configuration ordering -->
+		<dependency>
+			<groupId>org.springframework.ai</groupId>
+			<artifactId>spring-ai-autoconfigure-model-tool</artifactId>
+			<version>${project.parent.version}</version>
+			<scope>test</scope>
```
ericbottard: Would be even better IMO to add this as compile-time, optional dependency so that we can refer to the `.class` literal. With a comment explanation in the pom here.

auto-configurations/advisors/spring-ai-autoconfigure-tool-search-advisor/src/main/java/org/springframework/ai/chat/client/advisor/toolsearch/autoconfigure/ToolSearchAdvisorAutoConfiguration.java
```diff
@@ -74,7 +74,8 @@
...
+		beforeName = "org.springframework.ai.model.chat.client.autoconfigure.ChatClientAutoConfiguration")
```
ericbottard: After more thorough review, this would be better handled as a `.class` `after` in `ChatClientAutoConfiguration` (it already has `ToolCallingAutoConfiguration`)

ericbottard (changes requested):

auto-configurations/advisors/spring-ai-autoconfigure-tool-search-advisor/src/main/java/org/springframework/ai/chat/client/advisor/toolsearch/autoconfigure/ToolSearchAdvisorAutoConfiguration.java
tzolov: The ChatClientAutoConfiguration has already the `@AutoConfiguration(after = ToolCallingAutoConfiguration.class)`, but apparently this still leaves room for ambiguity. I guess this is the reason we missed this

tzolov:
> > Order ToolSearchAdvisorAutoConfiguration after ToolCallingAutoConfiguration so ToolCallingManager exists when its @ConditionalOnBean is evaluated.
> 
> Your justification talks about ONE reason to order, yet the change adds TWO attributes. Could you please clarify?

I've added javadoc. This PR adds only one extra attribute, the before one already existed. 
The last 2 tests in ToolSearchAdvisorAutoConfigurationTests.java‎ illustrate the issue. Remove the after and run the tests

auto-configurations/advisors/spring-ai-autoconfigure-tool-search-advisor/src/main/java/org/springframework/ai/chat/client/advisor/toolsearch/autoconfigure/ToolSearchAdvisorAutoConfiguration.java
ericbottard: Careful, I said it has `ToolCallingAutoConfiguration`, not `ToolSearchAdvisorAutoConfiguration`, so that was completely normal. My comment still stands.

Issue #6457, opened 2026-06-18
Tool calling observations don't have a parent in non-streaming mode


cijaaimee:
**Bug description**
When using the blocking ChatClient API, the tool calling observations generated when executing tools always have their parent set to `null`.

This seems to be because the `DefaultToolCallingManager` implementation ALWAYS sets the parent observation to whatever is in `ToolCallReactiveContextHolder.getContext()`:
```java
Observation parent = ToolCallReactiveContextHolder.getContext()
				.getOrDefault(ObservationThreadLocalAccessor.KEY, null);
// ... some more code ...

String toolCallResult = ToolCallingObservationDocumentation.TOOL_CALL
    .observation(this.observationConvention, DEFAULT_OBSERVATION_CONVENTION, () -> observationContext, this.observationRegistry)
    .parentObservation(parent)
```
In the blocking API, `ToolCallReactiveContextHolder` is never updated, so it's always `Context.empty()`.

**Environment**
This bug was introduced in Spring AI 2.0.0.

**Steps to reproduce**
1. Create a ChatClient with at least one tool
2. Call the ChatClient using the blocking API
3. Log the observations generated by this call

You'll see observations with context of type `ToolCallingObservationContext` with a null parent.

**Expected behavior**
I expect the tool call context to have the parent context properly set.

**Bandaid fix**
As a temporary patch, I'm using my own `ToolAdvisor`:
```kotlin
class FixObservationToolCallingAdvisor(toolCallingManager: ToolCallingManager, toolExecutionEligibilityChecker: ToolExecutionEligibilityChecker, advisorOrder: Int, conversationHistoryEnabled: Boolean, private val observationRegistry: ObservationRegistry) :
  ToolCallingAdvisor(toolCallingManager, toolExecutionEligibilityChecker, advisorOrder, conversationHistoryEnabled) {
  override fun adviseCall(
    chatClientRequest: ChatClientRequest,
    callAdvisorChain: CallAdvisorChain
  ) = try {
    val context = observationRegistry.currentObservation?.let { Context.of(ObservationThreadLocalAccessor.KEY, it) } ?: Context.empty()
    ToolCallReactiveContextHolder.setContext(context)

    super.adviseCall(chatClientRequest, callAdvisorChain)
  } finally {
    ToolCallReactiveContextHolder.clearContext()
  }
```
**This fix working proves the hypothesis behind the root cause of this issue.**

bbungjin:
[...]

PR #6459, opened 2026-06-18
Set the correct parent observation in tool call observations when using the blocking ChatClient API

cijaaimee:
This is a potential fix for issue #6457.


cijaaimee:
Superseded by #6475.

PR #6460, opened 2026-06-18
Use builder pattern for `OpenAiEmbeddingModel` and `OpenAiImageModel`

guanxuc:


sdeleuze:
The is no codebase wise consistency on this, which should probably happen in 2.1 or 3.0, but having consistency on OpenAI side makes sense to me.

PR #6463, opened 2026-06-19
Add unit tests for MutableResponseMetadata

kamalcis:
**What**
Add comprehensive unit tests for MutableResponseMetadata in the spring-ai-model module.

**Why**
MutableResponseMetadata is a core mutable key-value container used for response metadata, but it previously had no unit test coverage. This left all public API behavior unverified, especially around mutation, retrieval, and edge cases.

**How**
Added MutableResponseMetadataTests covering all public methods and behaviors.

**All operations**
put() stores values correctly and supports chaining
put() overwrites existing values
get() returns stored values with type safety
get() returns null for missing keys
getRequired() returns value when present
getRequired() throws IllegalArgumentException when key is missing
getOrDefault() returns stored value or default when absent
containsKey() correctly reflects key presence
remove() deletes entry and returns previous value
remove() returns null for missing keys
clear() removes all entries
isEmpty() correctly reflects state
keySet() returns all keys and is unmodifiable
entrySet() returns all entries and is unmodifiable
computeIfAbsent() inserts value when missing
computeIfAbsent() does not overwrite existing values
getRawMap() reflects current internal state

**Checklist**
 Unit tests added
 All tests pass (/mvnw clean test -pl spring-ai-model -Dtest=MutableResponseMetadataTests -Dmaven.build.cache.enabled=
false)
 Signed-off-by included
 Covers edge cases and mutation behavior

 @sdeleuze requesting your review on this change.

### DIFF
--- a/spring-ai-model/src/test/java/org/springframework/ai/model/MutableResponseMetadataTests.java
+++ b/spring-ai-model/src/test/java/org/springframework/ai/model/MutableResponseMetadataTests.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


kamalcis:
MutualResponseMetadata

sdeleuze:
@kamalcis I will merge this one as it does not hurt, but the added value is pretty low, so I will not accept future PRs like this. If you want to contribute, we prefer contributions on fixing open issues (including related test), reviewing existing PR, etc.

PR #6464, opened 2026-06-19
Add tool-search advisor modules to BOM

tzolov:
Add managed dependencies for the tool-search-advisor autoconfiguration and Spring Boot starter


sdeleuze (changes requested):
Could you please revert the formatting changes that are unrelated?

PR #6468, opened 2026-06-19
Propagate MCP errors from annotated tools

### LINKED ISSUE #6456: MCP tool method callbacks remap MCP errors into tool call responses
**Bug description**

MCP tool method callback classes from the mcp-annotations package remap exceptions thrown from methods that implement tool calls. Example of one such case with sync stateless callback, similar happens in other cases: [SyncStatelessMcpToolMethodCallback#apply method](https://github.com/spring-projects/spring-ai/blob/v2.0.0/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/SyncStatelessMcpToolMethodCallback.java#L83-L88).

This is OK for most exceptions, but should not be done for instances of `io.modelcontextprotocol.spec.McpError`. These exceptions contain the `jsonRpcError` property which should be returned directly to the MCP client, instead of getting remapped into a `CallToolResult`.

**Environment**

Tested with Spring AI version 2.0.0, with streamable protocol and sync type.
Looking at the code, the same happens in all (streamable / stateless x sync / async) combinations.

**Steps to reproduce**

The exception thrown by this implementation will be remapped into a tool call response, and will not trigger the expected [error flow of the URL mode elicitation](https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation#url-mode-with-elicitation-required-error-flow):

```java
package com.example;

import io.modelcontextprotocol.spec.McpError;
import io.modelcontextprotocol.spec.McpSchema;
import org.springframework.ai.mcp.annotation.McpTool;
import org.springframework.stereotype.Service;

import java.util.List;
import java.util.UUID;

@Service
public class ToolService {
    @McpTool
    public String notOk() {
        var urlElicitation = McpSchema.ElicitUrlRequest.builder(
                "Visit the docs first!",
                "https://docs.spring.io/spring-ai/reference/index.html",
                UUID.randomUUID().toString()
        ).build();
        throw McpError.URL_ELICITATION_REQUIRED.apply(List.of(urlElicitation));
    }
}
```

**Expected behavior**

This specific exception should not be converted into tool call response and the `jsonRpcError` should be returned to the client.

[Tools registered using the low-level API](https://docs.spring.io/spring-ai/reference/api/mcp/mcp-streamable-http-server-boot-starter-docs.html#_tools) work correctly.

For example, this will work as expected:

```java
package com.example;

import io.modelcontextprotocol.server.McpServerFeatures;
import io.modelcontextprotocol.spec.McpError;
import io.modelcontextprotocol.spec.McpSchema;
import org.springframework.context.annotation.Configuration;

import java.util.List;
import java.util.Map;
import java.util.UUID;

@Configuration
class Config {
    @Bean
    public List<McpServerFeatures.SyncToolSpecification> ok() {
        return List.of(McpServerFeatures.SyncToolSpecification
                .builder()
                .tool(McpSchema.Tool.builder(
                                "ok",
                                Map.of(
                                        "type", "object",
                                        "properties", Map.of()
                                )
                        ).build()
                ).callHandler((exchange, req) -> {
                    var urlElicitation = McpSchema.ElicitUrlRequest.builder(
                            "Visit the docs first!",
                            "https://docs.spring.io/spring-ai/reference/index.html",
                            UUID.randomUUID().toString()
                    ).build();
                    throw McpError.URL_ELICITATION_REQUIRED.apply(List.of(urlElicitation));
                }).build());
    }
}
```

See the steps to reproduce and expected behavior sections for code examples. The method annotation example should produce the same result the using the low-level APIs.

DragonFSKY:
## Problem
`@McpTool` methods that throw `McpError` can have their JSON-RPC error
converted to a `CallToolResult`, which breaks MCP protocol error flows such
as URL elicitation. The low-level MCP API already propagates these errors
correctly; the annotation-based callback path should do the same.

## Fix
Detect `McpError` in the exception cause chain and propagate it instead of
mapping it to an error `CallToolResult`.

This covers:
- sync and async annotated tool callbacks
- stateful and stateless callback variants
- wrapped `McpError` causes
- reactive async failures emitted by `Mono`, `Flux`, and generic `Publisher`

Ordinary exceptions still become error `CallToolResult` values as before.

## Tests
Added MCP error propagation and ordinary-exception regression coverage for:
- sync/async stateful/stateless callbacks
- wrapped `McpError` causes
- reactive `Mono`, `Flux`, and `Publisher` error paths
- reactive `CallToolResult` and `Void` return paths

Local verification:

```shell
./mvnw -pl mcp/mcp-annotations -Dmaven.build.cache.enabled=false test
```

Result: 1378 tests run, 0 failures, 0 errors, 2 skipped.

Fixes #6456

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractAsyncMcpToolMethodCallback.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractAsyncMcpToolMethodCallback.java
@@ -18,6 +18,7 @@ package org.springframework.ai.mcp.annotation.method.tool;
 
 import java.lang.reflect.Method;
 
+import io.modelcontextprotocol.spec.McpError;
 import io.modelcontextprotocol.spec.McpSchema.CallToolRequest;
 import io.modelcontextprotocol.spec.McpSchema.CallToolResult;
 import org.reactivestreams.Publisher;
@@ -66,21 +67,18 @@ public abstract class AbstractAsyncMcpToolMethodCallback<T, RC extends McpReques
 
 			// Check if the Mono contains CallToolResult
 			if (ReactiveUtils.isReactiveReturnTypeOfCallToolResult(this.toolMethod)) {
-				return (Mono<CallToolResult>) monoResult;
+				return ((Mono<CallToolResult>) monoResult).onErrorResume(this::propagateMcpError);
 			}
 
 			// Handle Mono<Void> for VOID return type
 			if (ReactiveUtils.isReactiveReturnTypeOfVoid(this.toolMethod)) {
 				return monoResult
-					.then(Mono.just(CallToolResult.builder().addTextContent(jsonHelper.toJson("Done")).build()));
+					.then(Mono.just(CallToolResult.builder().addTextContent(jsonHelper.toJson("Done")).build()))
+					.onErrorResume(this::propagateMcpError);
 			}
 
 			// Handle other Mono types - map the emitted value to CallToolResult
-			return monoResult.map(this::mapValueToCallToolResult)
-				.onErrorResume(e -> Mono.just(CallToolResult.builder()
-					.isError(true)
-					.addTextContent("Error invoking method: %s".formatted(e.getMessage()))
-					.build()));
+			return monoResult.map(this::mapValueToCallToolResult).onErrorResume(this::mapReactiveError);
 		}
 
 		// Handle Flux by taking the first element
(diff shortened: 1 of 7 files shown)


tzolov (approved):
LGTM, thanks @DragonFSKY

tzolov:
Hi @DragonFSKY, thank you for taking this on, the cause-chain scanning is a solid fix for the McpError propagation case.

We are opening a replacement PR: #6534 that addresses the full scope of issue #6456 rather than just McpError. 
The core problem is that callMethod() wraps every InvocationTargetException cause — including declared checked exceptions and Error subtypes — in a plain RuntimeException, destroying type information before it ever reaches the dispatch logic. The fix restores type-fidelity at the source: RuntimeException is re-thrown as-is, Error is re-thrown, and checked exceptions are wrapped in UndeclaredThrowableException (a RuntimeException that stays distinguishable). The apply() catch blocks then do a three-way dispatch on type: UndeclaredThrowableException → hard failure, McpError → hard failure, plain RuntimeException → error CallToolResult.

The result is the same McpError propagation you implemented, plus correct handling of checked exceptions and Errors, and it avoids cause-chain scanning (which silently swallows McpErrors wrapped by an intervening RuntimeException). We also aligned the client side (SyncMcpToolCallback/AsyncMcpToolCallback) to re-throw McpError as a hard failure.

I'm reference your work and you as a co-author in the replacement PR: #6534 . Feel free to review it.

PR #6475, opened 2026-06-20
Fix null parent for tool call observations in blocking mode

### LINKED ISSUE #6457: Tool calling observations don't have a parent in non-streaming mode

tzolov:
DefaultToolCallingManager read the tool call observation parent only from ToolCallReactiveContextHolder, which is populated solely in the streaming/reactive path. In blocking mode the holder is empty, so the explicit parentObservation(null) overwrote the parent that Micrometer had already auto-resolved from the current observation at creation time, leaving tool call observations without a parent and breaking the observability hierarchy.

Fall back to the observation currently in scope on the calling thread when the reactive context holder has no parent. This restores the pre-2.0.0 behavior for the blocking path while preserving the reactor-context parent required for the streaming hierarchy.

Fixes #6457

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
@@ -213,8 +213,12 @@ public final class DefaultToolCallingManager implements ToolCallingManager {
 				returnDirect = returnDirect && toolCallback.getToolMetadata().returnDirect();
 			}
 
+			// In streaming/reactive mode the parent observation is propagated through the
+			// Reactor context captured in ToolCallReactiveContextHolder. In blocking mode
+			// that holder is never populated, so fall back to the observation currently
+			// in scope on the calling thread to keep the observation hierarchy intact.
 			Observation parent = ToolCallReactiveContextHolder.getContext()
-				.getOrDefault(ObservationThreadLocalAccessor.KEY, null);
+				.getOrDefault(ObservationThreadLocalAccessor.KEY, this.observationRegistry.getCurrentObservation());
 
 			ToolCallingObservationContext observationContext = ToolCallingObservationContext.builder()
 				.toolDefinition(toolCallback.getToolDefinition())
--- a/spring-ai-model/src/test/java/org/springframework/ai/model/tool/DefaultToolCallingManagerTests.java
+++ b/spring-ai-model/src/test/java/org/springframework/ai/model/tool/DefaultToolCallingManagerTests.java
@@ -17,10 +17,14 @@
 package org.springframework.ai.model.tool;
 
 import java.lang.reflect.Method;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationHandler;
 import io.micrometer.observation.ObservationRegistry;
+import io.micrometer.observation.ObservationView;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.ai.chat.messages.AssistantMessage;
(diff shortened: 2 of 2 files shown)


spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
```diff
@@ -213,8 +213,12 @@ private InternalToolExecutionResult executeToolCall(Prompt prompt, AssistantMess
 				returnDirect = returnDirect && toolCallback.getToolMetadata().returnDirect();
 			}
 
+			// In streaming/reactive mode the parent observation is propagated through the
+			// Reactor context captured in ToolCallReactiveContextHolder. In blocking mode
+			// that holder is never populated, so fall back to the observation currently
+			// in scope on the calling thread to keep the observation hierarchy intact.
 			Observation parent = ToolCallReactiveContextHolder.getContext()
-				.getOrDefault(ObservationThreadLocalAccessor.KEY, null);
+				.getOrDefault(ObservationThreadLocalAccessor.KEY, this.observationRegistry.getCurrentObservation());
```
chemicL, line 221: AFAIU this operation is unnecessary. This leads to reading the ThreadLocal `Observation` twice. Instead, do not call `parentObservation(parent)` below (line 234) since it will lead to the same outcome of setting the parent to the current one but saving one TL read.

chemicL (changes requested):

spring-ai-model/src/main/java/org/springframework/ai/model/tool/DefaultToolCallingManager.java
tzolov: @chemicL, the DefaultToolCallingManger is used by both the reactive and non-reactive paths. 
The reactive path necessities the  

```java
Observation parent = ToolCallReactiveContextHolder.....
String toolCallResult = ToolCallingObservationDocumentation.TOOL_CALL
	.observation(this.observationConvention, DEFAULT_OBSERVATION_CONVENTION, () -> 
             observationContext, this.observationRegistry)
	.parentObservation(parent)
     .....
```
So we can remove the `	.parentObservation(parent)`. 

For the non-reactive path this leads to null parent and we need to handle it?

chemicL: Oh, my bad. Somehow I confused this with other places with `call()` vs `stream()` variants and thought this is only used in the non-reactive case like typically `call()` is. In this case, I don't see a good way to avoid the ThreadLocal read unless reactive tool execution is implemented separately. Approving the PR, sorry for the confusion.

chemicL (approved):

PR #6483, opened 2026-06-22
Preserve delta additional properties during streaming chunk conversion

Dhongli:
Fixes:

During streaming aggregation, delta additionalProperties were not copied into ChatCompletionMessage.

This causes reasoning_content from OpenAI-compatible providers (such as DashScope Qwen) to be lost before getReasoningContent(...) is executed.

This change preserves delta additional properties when converting ChatCompletionChunk into ChatCompletion.


jewoodev:
For review context, this is covered by #6373, which includes the same delta-to-message additional-properties copy and extends it to merged tool-call chunks and final advisor/chat-memory aggregation.

PR #6490, opened 2026-06-23
Fix incorrect assertions in `CouchbaseSearchVectorStore.Builder`

chaechaepower:
`CouchbaseSearchVectorStore.Builder` contains several validation inconsistencies compared to other vector stores.

* String fields (`collectionName`, `scopeName`, `bucketName`, `vectorIndexName`) used `Assert.notNull` twice with different messages. Since `Assert.notNull` does not reject empty or blank strings, replaced the duplicate checks with a single `Assert.hasText`, matching the validation used in `GemFireVectorStore` and `PineconeVectorStore`.

* `dimensions` used two `Assert.notNull` checks, including one with the message `"must not be empty"`, which is not meaningful for an integer and still allows non-positive values. Replaced these with `Assert.isTrue(dimensions > 0, ...)` and changed the type from `Integer` to primitive `int`, matching `OpenSearchVectorStore` and `OracleVectorStore`.

* Enum fields (`similarityFunction`, `indexOptimization`) also contained duplicate `Assert.notNull` checks with the misleading message `"must not be empty"`. Removed the redundant assertions.

### DIFF
--- a/vector-stores/spring-ai-couchbase-store/src/main/java/org/springframework/ai/vectorstore/couchbase/CouchbaseSearchVectorStore.java
+++ b/vector-stores/spring-ai-couchbase-store/src/main/java/org/springframework/ai/vectorstore/couchbase/CouchbaseSearchVectorStore.java
@@ -79,7 +79,7 @@ public class CouchbaseSearchVectorStore extends AbstractObservationVectorStore
 
 	private final String vectorIndexName;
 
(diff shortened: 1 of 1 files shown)


sdeleuze:
Merged without the `Integer` to `int` change that lead to some unwanted side effects at auto-configuration side and this is likely to agressive for post 2.0.0.

PR #6498, opened 2026-06-23
Fix shared mutable state in chat options builder `clone()`

subhashpolisetti:
The builder `clone()` methods in several `ChatOptions` implementations copied mutable collection fields by reference instead of creating independent copies. A builder obtained via `clone()` therefore shared those collections with its source, so mutating one affected the other.

This mirrors the fix applied to `AnthropicChatOptions` in #6487 and aligns each `clone()` with the defensive copying these builders already perform in `combineWith()`, and that the base `ChatOptions` builder (`stopSequences`) and the tool-calling builder (`toolCallbacks`, `toolContext`) already perform in their own `clone()`.

Affected builders:

- `BedrockChatOptions` — `requestParameters`
- `DeepSeekChatOptions` — `tools`
- `GoogleGenAiChatOptions` — `safetySettings`, `labels`
- `MistralAiChatOptions` — `tools`
- `OpenAiChatOptions` — `customHeaders`, `logitBias`, `outputModalities`, `metadata`, and the previously uncopied `extraBody`

Testing:

Added regression tests for each builder that explicitly fail on the previous aliasing behavior.

Added null-safety tests for each modified clone() method.

Note: I am happy to split this into per-module PRs if you would prefer separate reviews.

See #6487

### DIFF
--- a/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockChatOptions.java
+++ b/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockChatOptions.java
@@ -212,7 +212,7 @@ public class BedrockChatOptions implements ToolCallingChatOptions, StructuredOut
 		@Override
 		public B clone() {
 			B copy = super.clone();
-			copy.requestParameters = this.requestParameters;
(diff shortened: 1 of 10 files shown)


ericbottard:
LGTM

Issue #6441, opened 2026-06-17
There is a bug in the openai streaming mode


1233567890:
[demo.zip](https://github.com/user-attachments/files/29027782/demo.zip)
Look at my test class.
The result after running is
```
Hello! How can I assist you today? 😊
ChatResponse [metadata={ id: chatcmpl-daaa81a3731546c49cb4d610165a0452, usage: DefaultUsage{promptTokens=0, completionTokens=0, totalTokens=0}, rateLimit: org.springframework.ai.chat.metadata.EmptyRateLimit@5484c2c }, generations=[Generation[assistantMessage=AssistantMessage [messageType=ASSISTANT, toolCalls=[], textContent=, metadata={role=assistant, messageType=ASSISTANT, finishReason=_UNKNOWN, refusal=, annotations=[], index=0, id=chatcmpl-daaa81a3731546c49cb4d610165a0452, reasoningContent=}], chatGenerationMetadata=DefaultChatGenerationMetadata[finishReason='_UNKNOWN', filters=0, metadata=0]]]]
2026-06-17T10:08:24.999+08:00 ERROR 31608 --- [demo] [andler-thread-0] o.s.ai.chat.model.MessageAggregator      : Aggregation Error

java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100) ~[na:na]
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106) ~[na:na]
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302) ~[na:na]
	at java.base/java.util.Objects.checkIndex(Objects.java:385) ~[na:na]
	at java.base/java.util.ArrayList.get(ArrayList.java:427) ~[na:na]
	at org.springframework.ai.openai.OpenAiChatModel$ChunkMerger.lambda$mergeDeltas$4(OpenAiChatModel.java:1065) ~[spring-ai-openai-2.0.0.jar:2.0.0]
	at java.base/java.util.stream.ReduceOps$2ReducingSink.accept(ReduceOps.java:123) ~[na:na]
	at java.base/java.util.stream.Streams$StreamBuilderImpl.forEachRemaining(Streams.java:411) ~[na:na]
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762) ~[na:na]
[...]

PR #6504, opened 2026-06-24
Fix OpenAI stream handling for empty tool_calls

liushenling2001:
This fixes streaming for OpenAI-compatible endpoints that emit an empty `tool_calls` array on normal text delta chunks.

Previously, `ChunkMerger.hasToolCall` treated the presence of the optional `tool_calls` field as an active tool call, even when the list was empty. Some vLLM/OpenAI-compatible endpoints include `tool_calls: []` while streaming regular text. That caused Spring AI to enter the tool-call aggregation path and buffer normal text chunks, which could collapse the streamed response to only the first text fragment.

The fix only treats a chunk as containing tool calls when the `tool_calls` list is present and non-empty.

Validation:
- `./mvnw.cmd -pl models/spring-ai-openai -am -DskipTests -DskipITs -Dspring-javaformat.skip=true -Dcheckstyle.skip=true install`

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -1034,7 +1034,8 @@ public final class OpenAiChatModel implements ChatModel {
 	private static final class ChunkMerger {
 
 		static boolean hasToolCall(ChatCompletionChunk chunk) {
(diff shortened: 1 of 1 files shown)


ericbottard:
Thanks for this PR @liushenling2001,

it was indeed a miss in my fix for https://github.com/spring-projects/spring-ai/issues/6441#issuecomment-4787776328. Shall I understand that this fixes the remaining issue you mentioned in that comment?

PR #6508, opened 2026-06-24
Remove fastjson2 dependency from `spring-ai-azure-store`

ilayaperumalg:
- Replace the `fastjson2` dependency in `AzureVectorStore` with the Jackson 3.x `JsonMapper` to make it consistent across the usages.
  - Add tests

### DIFF
--- a/vector-stores/spring-ai-azure-store/src/main/java/org/springframework/ai/vectorstore/azure/AzureVectorStore.java
+++ b/vector-stores/spring-ai-azure-store/src/main/java/org/springframework/ai/vectorstore/azure/AzureVectorStore.java
@@ -24,8 +24,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.stream.IntStream;
 
-import com.alibaba.fastjson2.JSONObject;
-import com.alibaba.fastjson2.TypeReference;
 import com.azure.core.util.Context;
 import com.azure.search.documents.SearchClient;
 import com.azure.search.documents.SearchDocument;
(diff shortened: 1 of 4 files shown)


vector-stores/spring-ai-azure-store/src/main/java/org/springframework/ai/vectorstore/azure/AzureVectorStore.java
```diff
@@ -180,7 +180,13 @@ public void doAdd(List<Document> documents) {
 			searchDocument.put(ID_FIELD_NAME, document.getId());
 			searchDocument.put(this.embeddingFieldName, embeddings.get(i));
 			searchDocument.put(this.contentFieldName, document.getText());
-			searchDocument.put(this.metadataFieldName, new JSONObject(document.getMetadata()).toJSONString());
+			try {
+				searchDocument.put(this.metadataFieldName,
+						JacksonUtils.getDefaultJsonMapper().writeValueAsString(document.getMetadata()));
```
sdeleuze: Maybe we can use `JsonHelper#toJson(java.lang.Object)` here instead?

With Jackson 3, we don't have to catch exception for regular use case, and the old code does not, so I am not sure we need to catch exception here. If you think we should (due to frequent error where we want to provide a better error message), probably better to use a more specific exception.

vector-stores/spring-ai-azure-store/src/main/java/org/springframework/ai/vectorstore/azure/AzureVectorStore.java
```diff
@@ -344,8 +350,9 @@ static Map<String, Object> parseMetadataToMutable(@Nullable String metadataJson)
 			return new HashMap<>();
 		}
 		try {
```
sdeleuze: Maybe we can use `JsonHelper#fromJsonToMap(String)` here?

sdeleuze (changes requested):

PR #6514, opened 2026-06-25
Add streaming support for OpenAI audio transcription model

guanxuc:
# Summary
Introduce `StreamingTranscriptionModel` interface and implement streaming audio transcription in `OpenAiAudioTranscriptionModel` using the async OpenAI client

# Changes
 - Add `StreamingTranscriptionModel` interface with stream methods
 - Implement stream method in `OpenAiAudioTranscriptionModel` via `OpenAIClientAsync`
 - Register `OpenAIClientAsync` in `OpenAiAudioTranscriptionAutoConfiguration`
 - Add unit and integration tests for streaming transcription

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
@@ -151,23 +151,21 @@ public final class OpenAiAudioTranscriptionModel implements TranscriptionModel {
 			logger.trace("OpenAiAudioTranscriptionModel stream with model: " + mergedOptions.getModel());
 		}
 
-		Flux<TranscriptionStreamEvent> chunk = Flux.create(sink -> {
-			this.openAiClientAsync.audio()
-				.transcriptions()
-				.createStreaming(params)
-				.subscribe(sink::next)
-				.onCompleteFuture()
-				.whenComplete((unused, throwable) -> {
-					if (throwable != null) {
-						sink.error(throwable);
-					}
-					else {
-						sink.complete();
-					}
-				});
-		});
+		Flux<TranscriptionStreamEvent> chunks = Flux.create(sink -> this.openAiClientAsync.audio()
+			.transcriptions()
+			.createStreaming(params)
+			.subscribe(sink::next)
+			.onCompleteFuture()
+			.whenComplete((unused, throwable) -> {
+				if (throwable != null) {
+					sink.error(throwable);
+				}
+				else {
+					sink.complete();
+				}
+			}));
 
-		return chunk.map(event -> {
+		return chunks.map(event -> {
 			String text = extractStreamEventText(event);
 			AudioTranscription transcript = new AudioTranscription(text);
 			return new AudioTranscriptionResponse(transcript, new AudioTranscriptionResponseMetadata());
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
@@ -24,13 +24,17 @@ import java.util.List;
 import java.util.Objects;
 
 import com.openai.client.OpenAIClient;
+import com.openai.client.OpenAIClientAsync;
 import com.openai.core.MultipartField;
+import com.openai.core.http.AsyncStreamResponse;
 import com.openai.models.audio.transcriptions.TranscriptionCreateParams;
 import com.openai.models.audio.transcriptions.TranscriptionCreateResponse;
+import com.openai.models.audio.transcriptions.TranscriptionStreamEvent;
 import io.micrometer.observation.ObservationRegistry;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.jspecify.annotations.Nullable;
+import reactor.core.publisher.Flux;
 
 import org.springframework.ai.audio.transcription.AudioTranscription;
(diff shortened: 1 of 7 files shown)


guanxuc:
Hi @ericbottard @ilayaperumalg ,

Could you please take a look, thanks!

spring-ai-model/src/main/java/org/springframework/ai/audio/transcription/TranscriptionModel.java
```diff
@@ -26,9 +26,11 @@
  * known as Speech-to-Text.
  *
  * @author Mudabir Hussain
+ * @author guan xu
  * @since 1.0.0
  */
-public interface TranscriptionModel extends Model<AudioTranscriptionPrompt, AudioTranscriptionResponse> {
+public interface TranscriptionModel
```
ilayaperumalg, line 32: This would break the existing TranscriptionModel implementations (if there are custom ones). Hence, I suggest to keep TranscriptionModel independent by having the OpenAiAudioTranscriptionModel implementing TranscriptionModel along with StreamingTranscriptionModel. What do you think?

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
```diff
@@ -128,6 +135,46 @@ public AudioTranscriptionResponse call(AudioTranscriptionPrompt transcriptionPro
...
+		TranscriptionCreateParams params = buildParams(mergedOptions, audioBytes, filename);
+		if (logger.isTraceEnabled()) {
+			logger.trace("OpenAiAudioTranscriptionModel stream with model: " + mergedOptions.getModel());
+		}
+
+		Flux<TranscriptionStreamEvent> chunk = Flux.create(sink -> {
+			AsyncStreamResponse<TranscriptionStreamEvent> response = this.openAiClientAsync.audio()
+				.transcriptions()
+				.createStreaming(params);
+			response.subscribe(sink::next);
```
ilayaperumalg: Can we follow the pattern used in OpenAIChatModel streaming implementation for the chaining? Something like: 
```
Flux.create(sink -> this.openAiClientAsync.audio()
      .transcriptions()
      .createStreaming(params)
      .subscribe(sink::next)
      .onCompleteFuture()
      .whenComplete((unused, throwable) -> { ... }));
```

spring-ai-model/src/main/java/org/springframework/ai/audio/transcription/TranscriptionModel.java
guanxuc: Good point! However, this would lead to inconsistencies with the existing implementations of `ChatModel` and `TextToSpeechModel`, which could also cause confusion for users. Could we maintain consistency across the codebase and document this as a breaking change in `upgrade-notes.adoc`?

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
guanxuc: Thanks for the suggestion! I did look into chaining the calls like OpenAiChatModel does. However, the current implementation intentionally uses a local response variable to register `sink.onDispose(response::close)`. This ensures the underlying HTTP stream is properly closed if the subscriber cancels early. This cleanup step is missing in OpenAiChatModel's streaming and would be lost if we adopted the fluent chain.

I'm happy to switch to the chained style if we decide the onDispose cleanup is unnecessary here and subscribe() is confirmed to be fluent. But for safety, I'd prefer to keep the explicit resource management.

guanxuc:
Hi Ilayaperumal, 

Thanks for your review and the great suggestions! I've addressed your comments. Could you please take another look? Feel free to reach out if you have any further questions.

spring-ai-model/src/main/java/org/springframework/ai/audio/transcription/TranscriptionModel.java
ilayaperumalg: @guanxuc I understand your perspective to bring in consistency. But, I am sorry that we can't bring breaking changes to 2.0.1. Thanks for the understanding.

guanxuc: Hi Ilayaperumal,

  Thanks for the feedback! Could we add a default `stream` method to maintain consistency and avoid bringing breaking changes as shown below? Please let me know if this approach is acceptable. If not, I'd be more than happy to implement your previous excellent suggestions instead.

```java
public interface StreamingTranscriptionModel
		extends StreamingModel<AudioTranscriptionPrompt, AudioTranscriptionResponse> {

	default Flux<AudioTranscriptionResponse> stream(AudioTranscriptionPrompt transcriptionPrompt) {
		return Flux.error(new UnsupportedOperationException("Streaming transcription is not supported"));
	}

}
```

ilayaperumalg:
> Thanks for your review and the great suggestions! I've addressed your comments. Could you please take another look? Feel free to reach out if you have any further questions.

Hi @guanxuc Sorry about the delay in getting to back to this. Your approach to add a default `stream` method indeed is a better idea. Thanks for the follow up! 

The changes look good but I see the following tests fail:
```
streamTest()
streamTranscribeWithResourceAndOptionsTest()
streamTranscribeWithResourceTest()
```
Could you check? Thanks!

ilayaperumalg:
@guanxuc After the changing the IT to use GPT_4O_MINI_TRANSCRIBE, all the stream tests passed. It appeared the previous model didn't support streaming. Squashed all the commits and merged. Thanks once again for adding this feature!

guanxuc:
@ilayaperumalg Thanks for merging!

PR #6525, opened 2026-06-27
Fix OpenAI streaming tool call chunk merging

liushenling2001:
## Summary

Fix OpenAI streaming tool call chunk merging for OpenAI-compatible providers.

Some OpenAI-compatible servers, observed with vLLM-backed endpoints, stream one logical tool call across multiple chunks. The tool call id, function name, and function arguments can arrive in separate chunks, and some providers repeat the same tool call id on later chunks.

The previous `ChunkMerger` appended every chunk that contained a tool call id as a new tool call. That could split one logical tool call into multiple incomplete tool calls. In practice this caused stream aggregation or downstream tool execution failures, including errors like:

- `NoSuchElementException` / aggregation errors while converting merged chunks
- `toolName cannot be null or empty` when an incomplete merged tool call reached tool execution
- `No ToolCallback found` when the logical function metadata was split incorrectly across chunks

## Changes

- Merge streaming tool call chunks by tool call `index` or repeated `id` instead of appending every id-bearing chunk.
- Preserve the first available function name.
- Concatenate streamed function argument fragments.
- Keep conversion strict so a final merged tool call must still contain id/function/name.
- Add regression tests for:
  - `id`, `function.name`, and `function.arguments` arriving in separate chunks
  - providers repeating the same tool call id across chunks

## Verification

Ran:

```bash
mvn -pl models/spring-ai-openai -Dtest=OpenAiChatModelTests test
```

Result: `Tests run: 24, Failures: 0, Errors: 0, Skipped: 0`.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -1034,7 +1034,8 @@ public final class OpenAiChatModel implements ChatModel {
 	private static final class ChunkMerger {
 
 		static boolean hasToolCall(ChatCompletionChunk chunk) {
-			return !chunk.choices().isEmpty() && chunk.choices().get(0).delta().toolCalls().isPresent();
+			return !chunk.choices().isEmpty()
+					&& chunk.choices().get(0).delta().toolCalls().filter(toolCalls -> !toolCalls.isEmpty()).isPresent();
 		}
 
 		static boolean toolCallsDone(ChatCompletionChunk chunk) {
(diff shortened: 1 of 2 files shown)


ericbottard:
Hi @liushenling2001,

is this a follow up to https://github.com/spring-projects/spring-ai/issues/6441#issuecomment-4789676670?

jewoodev:
For review context, this overlaps with #6381 in the `OpenAiChatModel.ChunkMerger` streaming tool-call merge path.

#6504 already covered the empty `tool_calls: []` handling, while this PR and #6381 both change `mergeDeltas(...)` / `chunkToChatCompletion(...)` for split streaming tool-call deltas. It may be useful to review them together and choose one merge strategy.

ericbottard:
Hi @liushenling,

I just merged #6381 which fixes the same issue with a slightly more readable codeflow IMO. 
Thanks for your analysis and PR proposal, as always

PR #6534, opened 2026-07-01
Align @McpTool exception handling with @Tool contract

### LINKED ISSUE #6456: MCP tool method callbacks remap MCP errors into tool call responses

tzolov:
Mirror the @Tool exception-handling contract in all @McpTool server callbacks: RuntimeExceptions are conveyed to the model as error CallToolResults; declared checked exceptions (wrapped in UndeclaredThrowableException), Error subtypes, and McpError are propagated as hard failures and never reach the model.

On the client side, McpError thrown by a remote MCP server is now re-thrown from SyncMcpToolCallback / AsyncMcpToolCallback instead of being wrapped in a ToolExecutionException, so it bypasses DefaultToolExecutionExceptionProcessor and fails the model interaction.

The toolCallExceptionClass constructor parameter and doGetToolCallException() on AbstractMcpToolProvider are deprecated (scheduled for removal in 2.1.0) and retained for binary/source compatibility; the value is ignored at runtime.

Closes #6456

### DIFF
--- a/mcp/common/src/main/java/org/springframework/ai/mcp/AsyncMcpToolCallback.java
+++ b/mcp/common/src/main/java/org/springframework/ai/mcp/AsyncMcpToolCallback.java
@@ -19,6 +19,7 @@ package org.springframework.ai.mcp;
 import java.util.Map;
 
 import io.modelcontextprotocol.client.McpAsyncClient;
+import io.modelcontextprotocol.spec.McpError;
 import io.modelcontextprotocol.spec.McpSchema.CallToolRequest;
 import io.modelcontextprotocol.spec.McpSchema.CallToolResult;
 import io.modelcontextprotocol.spec.McpSchema.Tool;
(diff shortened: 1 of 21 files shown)


DragonFSKY:
Thanks Christian, and thanks for the explanation and attribution. I think #6534 fixes the core `McpError` propagation issue.

One small edge case I noticed: in `AbstractAsyncMcpToolMethodCallback.convertToCallToolResult`, the `Mono`/`Flux`/`Publisher` branches for `CallToolResult` and `Void` seem to bypass `toErrorResultOrPropagate`. My understanding is that this creates an inconsistency for plain `RuntimeException` handling across reactive return types: `Mono<String>` would be converted to an error `CallToolResult`, while `Mono<Void>` or `Mono<CallToolResult>` would hard-fail.

I verified this locally with small `Mono<Void>` and `Mono<CallToolResult>` tests. Would it make sense to apply the same `onErrorResume(this::toErrorResultOrPropagate)` to those branches too?

Small docs nit: the upgrade note seems to imply checked exceptions were previously hard failures under the default `Exception.class` behavior, but my understanding is that they were previously converted to error `CallToolResult`s.

tzolov:
Thanks for the feedback @DragonFSKY , I've applied it in the last commit

PR #6538, opened 2026-07-01
Support customizing MCP server `serverInfo` via an overridable `Implementation` bean

herder:
The MCP server autoconfigurations build `serverInfo` inline as `new Implementation(name, version)` in all four server beans (`mcpSyncServer`, `mcpAsyncServer`, `mcpStatelessSyncServer`, `mcpStatelessAsyncServer`), so a display title or icons can't be advertised in the `initialize` handshake and there is no seam to customize it — the server customizers were removed in the autoconfig modularization.

This adds an overridable `@ConditionalOnMissingBean McpSchema.Implementation serverInfo` bean to both `McpServerAutoConfiguration` and `McpServerStatelessAutoConfiguration`; all four server beans inject it. Default behavior is unchanged (`Implementation(name, version)`); a user can now supply their own `Implementation` bean (e.g. with `.title(...)` / `.icons(...)`) to customize what the server advertises — consistently across stateful/stateless and sync/async.

Includes a test in `McpStatelessServerAutoConfigurationIT` asserting a user `Implementation` bean overrides the default; a matching test for the stateful autoconfig can follow if preferred.

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-server-common/src/main/java/org/springframework/ai/mcp/server/common/autoconfigure/McpServerAutoConfiguration.java
@@ -99,20 +99,23 @@ public class McpServerAutoConfiguration {
 	}
 
 	@Bean
+	@ConditionalOnMissingBean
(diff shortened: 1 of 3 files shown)


herder:
Superseded by #6539. That PR takes a narrower, convention-aligned approach: dedicated `McpStatelessSyncServerCustomizer` / `McpStatelessAsyncServerCustomizer` hooks (mirroring the existing stateful customizers) instead of overriding `serverInfo` across all transports via a shared bean.

PR #6099, opened 2026-05-21
Allow to define multiple McpSyncServerCustomizer and McpAsyncServerCustomizer beans

quaff:

PR #6539, opened 2026-07-01
Add customizers for stateless MCP servers

herder:
## What

Stateful MCP servers can be tuned with `McpSyncServerCustomizer` / `McpAsyncServerCustomizer` beans, which `McpServerAutoConfiguration` applies to the server builder before `build()`. The stateless (stateless/streamable transport) servers had no equivalent hook.

This adds, symmetric with the stateful side:

- `McpStatelessSyncServerCustomizer` and `McpStatelessAsyncServerCustomizer` in `org.springframework.ai.mcp.customizer`.
- Application of these beans in `McpServerStatelessAutoConfiguration` (`mcpStatelessSyncServer` / `mcpStatelessAsyncServer`) via `ObjectProvider` + `orderedStream()`, just before `build()` — so any number of customizers are applied, honouring `@Order`, consistent with the other `ObjectProvider` collaborators already injected into those methods.

## Why

Without a customizer, the only way to change anything on the stateless server builder — e.g. `serverInfo` (server name/version/title/icons for branding), capabilities, or request timeout — is to `spring.autoconfigure.exclude` `McpServerStatelessAutoConfiguration` and re-declare the entire server bean, duplicating all of the autoconfiguration's wiring. The stateful autoconfig already avoids this via customizer beans; this brings the stateless autoconfig in line.

`ObjectProvider` is used rather than `Optional`: `Optional<T>` supports only zero-or-one bean (a second customizer fails startup with `NoUniqueBeanDefinitionException`, with no ordering), whereas `ObjectProvider<T>.orderedStream()` applies all customizers honouring `@Order` — the idiomatic Boot customizer convention. See the [comment below](https://github.com/spring-projects/spring-ai/pull/6539#issuecomment-4866736114) for a note on the stateful side, which still uses `Optional`.

## Notes

- The servlet `immediateExecution(true)` default in the stateless sync path is left inline and runs before the customizers, so a user customizer can still override it.
- Tests added in `McpStatelessServerAutoConfigurationIT`: a user-provided customizer of each type is invoked during server construction, and multiple `@Order`ed customizers are all invoked in order. Full `spring-ai-autoconfigure-mcp-server-common` module build (incl. checkstyle + spring-javaformat) is green; 26/26 tests pass.

Replaces #6538, which changed all transports via a shared `serverInfo` bean; this is the narrower, convention-aligned approach.

### DIFF
--- a/mcp/common/src/main/java/org/springframework/ai/mcp/customizer/McpStatelessAsyncServerCustomizer.java
+++ b/mcp/common/src/main/java/org/springframework/ai/mcp/customizer/McpStatelessAsyncServerCustomizer.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.mcp.customizer;
+
+import io.modelcontextprotocol.server.McpServer;
+
+/**
+ * Interface for customizing stateless asynchronous MCP server configurations.
+ *
+ * @author Niklas Herder
+ * @since 2.0.1
+ * @see io.modelcontextprotocol.server.McpServer.StatelessAsyncSpecification
+ */
+public interface McpStatelessAsyncServerCustomizer {
+
+	void customize(McpServer.StatelessAsyncSpecification serverBuilder);
+
+}
--- a/mcp/common/src/main/java/org/springframework/ai/mcp/customizer/McpStatelessAsyncServerCustomizer.java
+++ b/mcp/common/src/main/java/org/springframework/ai/mcp/customizer/McpStatelessAsyncServerCustomizer.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


quaff:
GH-6099 is related.

herder:
While aligning the stateless customizers with the stateful convention, I switched them from `Optional<Customizer>` to `ObjectProvider<Customizer>` + `orderedStream()` rather than copying the stateful side verbatim.

Rationale: `Optional<T>` as an injection point supports only zero-or-one bean — a second customizer of the same type fails context startup with `NoUniqueBeanDefinitionException`, and there's no way to order them. `ObjectProvider<T>.orderedStream()` applies all customizers honouring `@Order`, which is the idiomatic Boot customizer convention and also matches the `ObjectProvider<List<...>>` collaborators already injected into these very same factory methods. The added test registers two `@Order`ed customizers and asserts both run in order — which isn't expressible under `Optional`.

Worth flagging for a possible follow-up: the **stateful** `McpServerAutoConfiguration` still injects `Optional<McpSyncServerCustomizer>` / `Optional<McpAsyncServerCustomizer>`, and it *itself* contributes an internal `servletMcpSyncServerCustomizer` bean on the servlet + `SYNC` path. So a user who declares their own `McpSyncServerCustomizer` there currently hits `NoUniqueBeanDefinitionException` — two beans, one `Optional` injection point. Happy to bring the stateful side to the same `ObjectProvider` pattern in this PR if you'd prefer them fixed together, or open a separate issue — whichever you'd rather review.

herder:
Hi @sdeleuze - do i need to do anything more to get this moving? No rush - just want to follow the correct workflow

herder:
Would you have time to take a look at this, @dimitarproynov ?

mcp/common/src/main/java/org/springframework/ai/mcp/customizer/McpStatelessSyncServerCustomizer.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.springframework.ai.mcp.customizer;
+
+import io.modelcontextprotocol.server.McpServer;
+
+/**
+ * Interface for customizing stateless synchronous MCP server configurations.
+ *
+ * @author Niklas Herder
+ * @since 2.0.1
```
dimitarproynov: Can you, please, have it as 2.0.2

mcp/common/src/main/java/org/springframework/ai/mcp/customizer/McpStatelessAsyncServerCustomizer.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.springframework.ai.mcp.customizer;
+
+import io.modelcontextprotocol.server.McpServer;
+
+/**
+ * Interface for customizing stateless asynchronous MCP server configurations.
+ *
+ * @author Niklas Herder
+ * @since 2.0.1
```
dimitarproynov: Can you, please, have it as 2.0.2

dimitarproynov (changes requested):
LGTM! When the versions are updated I'll merge it

herder:
@dimitarproynov Thanks for the review! Rebased and set the version to 2.0.2 now.

PR #6540, opened 2026-07-01
Fix strict mode payload placement in `OpenAiChatModel`

### LINKED ISSUE #6536: OpenAI/Azure tool calling: strict is emitted inside parameters instead of at the function level, so strict mode is a silent no-op for tools
> Searched existing issues/PRs first. The `strict` items I found all concern the **`response_format` / structured-output** path (#3928, #4045, PR #3931, open #4422 "Generated JSON schema fails strict mode checks of Azure", commit `a1b7d5e`). I could not find any issue/PR about **tool/function** `strict`. This report is specifically about the tool-calling path.

**Bug description**

When building OpenAI tool/function definitions, `strict` is written **inside the `parameters` JSON-Schema object** (via `FunctionParameters.Builder.putAdditionalProperty("strict", true)`) and the **function-level** `FunctionDefinition.Builder.strict(...)` is **never called**.

Per the OpenAI/Azure API, `strict` is a field of the **function object** — a sibling of `parameters`, alongside `name`/`description` — not a JSON-Schema keyword. Placed inside `parameters`, `strict` is an unrecognized schema keyword and is **ignored** by the provider. The net effect is that **strict mode is a silent no-op for tool calling**: the framework hardcodes `strict = true`, but the model runs best-effort tool-argument generation with no schema enforcement. No error is raised, so the misbehavior is invisible.

This affects both `spring-ai-openai` (`OpenAiChatModel`) and `spring-ai-openai-sdk` (`OpenAiSdkChatModel`).

Relevant code (`OpenAiChatModel#getChatCompletionTools`, `main`):

```java
parametersBuilder.putAdditionalProperty("strict", JsonValue.from(true)); // TODO allow non-strict
...
FunctionDefinition functionDefinition = FunctionDefinition.builder()
    .name(toolDefinition.name())
    .description(toolDefinition.description())
    .parameters(parametersBuilder.build())   // strict ends up INSIDE parameters
    .build();                                 // FunctionDefinition.Builder.strict(...) never called
```

The OpenAI Java SDK already exposes `FunctionDefinition.Builder.strict(Boolean)` — it is simply not used here.

**Environment**

- Spring AI: 2.0.0 — and the defect is still present on `main`.
- Affected module(s): `spring-ai-openai` (`OpenAiChatModel#getChatCompletionTools`); the same defect is in `spring-ai-openai-sdk` (`OpenAiSdkChatModel`).
- Provider: reproducible on both OpenAI and Azure OpenAI — the misplacement is provider-agnostic.

(Java/Spring Boot versions and vector store are not relevant here: this is a deterministic tool-schema serialization defect, independent of runtime or retrieval configuration.)

**Steps to reproduce**

1. Register any `@Tool`/`ToolCallback` and call `OpenAiChatModel` (or `OpenAiSdkChatModel`) with a prompt that triggers the tool.
2. Capture the outgoing `/chat/completions` request body.
3. Inspect `tools[0].function`.

Observed request (abbreviated):

```json
{
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_weather",
        "description": "Get weather",
        "parameters": {
          "type": "object",
          "properties": { "location": { "type": "string" } },
          "strict": true            // ← placed here; ignored by the API
        }
      }
    }
  ]
}
```

There is no `tools[0].function.strict`. Because `strict` sits inside the JSON schema, the provider ignores it and the request is treated as non-strict. (This also explains why non-strict-compliant tool schemas — e.g. objects without `additionalProperties:false`/`required` — are accepted rather than rejected the way the `response_format` path rejects them in #4422.)

**Expected behavior**

`strict` should be set at the **function level** and omitted from the JSON schema:

```json
{
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_weather",
        "description": "Get weather",
        "strict": true,             // ← function-level, honored by the API
        "parameters": {
          "type": "object",
          "properties": { "location": { "type": "string" } },
          "required": ["location"]
        }
      }
    }
  ]
}
```

i.e. call `FunctionDefinition.builder()....parameters(...).strict(true).build()` instead of injecting `"strict"` into the parameters object. Ideally the value is configurable (per the existing `// TODO allow non-strict`) via `OpenAiChatOptions`, since strict imposes schema requirements (`additionalProperties:false` and all fields `required` at every level) that not every tool schema satisfies.

A test that captures the request body and asserts `strict` placement (fails on current behavior):

```java
@Test
void toolStrictIsEmittedAtFunctionLevel_notInsideParameters() throws Exception {
  try (var server = new MockWebServer()) {
    server.enqueue(new MockResponse()
        // minimal valid assistant response, no tool call needed to inspect the request
        .setBody("""
            {"id":"x","object":"chat.completion","created":0,"model":"gpt-4.1",
             "choices":[{"index":0,"finish_reason":"stop",
                         "message":{"role":"assistant","content":"ok"}}]}"""));
    server.start();

    var api = OpenAiApi.builder()
        .apiKey(new SimpleApiKey("test"))
        .build();
    var chatModel = OpenAiChatModel.builder()
        .openAiApi(api)
        .defaultOptions(OpenAiChatOptions.builder().model("gpt-4.1").build())
        .build();

    ToolCallback tool = FunctionToolCallback
        .builder("get_weather", (WeatherRequest req) -> "sunny")
        .description("Get weather")
        .inputType(WeatherRequest.class)
        .build();

    chatModel.call(new Prompt("What is the weather?",
        ToolCallingChatOptions.builder().toolCallbacks(tool).build()));

    var recorded = server.takeRequest();
    var body = new ObjectMapper().readTree(recorded.getBody().readUtf8());
    var function = body.at("/tools/0/function");

    // Expected: strict at function level
    assertThat(function.path("strict").asBoolean(false)).isTrue();          // FAILS today
    // And NOT inside the parameters schema
    assertThat(function.path("parameters").has("strict")).isFalse();        // FAILS today
  }
}

record WeatherRequest(String location) {}
```

(Builder API names may need minor adjustment to the exact release; the assertion on `tools[0].function.strict` vs `tools[0].function.parameters.strict` is the point.)

**Suggested fix**

In `OpenAiChatModel#getChatCompletionTools` (and the `OpenAiSdkChatModel` equivalent): drop `parametersBuilder.putAdditionalProperty("strict", ...)` and instead call `FunctionDefinition.Builder.strict(<configurable>)`. Expose the flag on `OpenAiChatOptions` (default configurable) to resolve the existing `// TODO allow non-strict`.

prachipanditrao:
This PR fixes the tool calling defect where the `strict` configuration property was incorrectly serialized inside the JSON-Schema parameters object rather than at the function-level wrapper, causing it to act as a silent no-op.

It moves the property mapping to the `FunctionDefinition` builder block within `OpenAiChatModel` to align with the official OpenAI/Azure API specifications and exposes a strict configuration flag in `OpenAiChatOptions`.

Fixes #6536


models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java
```diff
@@ -1118,5 +1118,43 @@ void streamingFinishReasonSurvivesRoundTripUnderTurkishLocale() {
...
+
+		var tools = request.tools().orElseThrow();
+		assertThat(tools).hasSize(1);
+
+		com.openai.models.chat.completions.ChatCompletionFunctionTool functionTool = tools.get(0)
+			.function()
+			.orElseThrow();
+		com.openai.models.FunctionDefinition functionDef = functionTool.function();
+
+		assertThat(functionDef.strict()).contains(true);
```
dimitarproynov, line 1241: Please, add a test case for when strict is explicitly set to false

models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java
```diff
@@ -1118,5 +1118,43 @@ void streamingFinishReasonSurvivesRoundTripUnderTurkishLocale() {
 			Locale.setDefault(original);
 		}
 	}
+	
+	@Test
+	void toolStrictIsEmittedAtFunctionLevel() {
+		org.springframework.ai.model.tool.ToolCallingManager mockToolCallingManager = mock(
```
dimitarproynov: Please, use import statements, rather than fully-qualified class names to be consistent with the rest of the test.

models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java
```diff
@@ -1118,5 +1118,43 @@ void streamingFinishReasonSurvivesRoundTripUnderTurkishLocale() {
 			Locale.setDefault(original);
 		}
 	}
+
```
dimitarproynov: Please, remove the tab character as per CONTRIBUTING.md: "eliminate all trailing whitespace"

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatOptions.java
```diff
@@ -608,7 +621,8 @@ public boolean equals(@Nullable Object o) {
 				&& Objects.equals(this.stop, options.stop) && Objects.equals(this.topP, options.topP)
 				&& Objects.equals(this.toolChoice, options.toolChoice) && Objects.equals(this.user, options.user)
 				&& Objects.equals(this.parallelToolCalls, options.parallelToolCalls)
-				&& Objects.equals(this.store, options.store) && Objects.equals(this.metadata, options.metadata)
+				&& Objects.equals(this.store, options.store) && Objects.equals(this.strict, options.strict)
```
dimitarproynov, line 624: Please, place the Objects.equals(this.strict, options.strict) on a new line for readability

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -978,30 +978,27 @@ private List<ChatCompletionTool> getChatCompletionTools(List<ToolDefinition> too
...
-																								// allow
-																								// non-strict
-																								// mode
 				}
 				catch (Exception e) {
 					logger.error("Failed to parse tool schema", e);
 				}
 			}
 
+			Boolean strictMode = (this.options != null && this.options.getStrict() != null) ? this.options.getStrict()
```
dimitarproynov: This new code reads this.options, which is the model-level default set once at construction time, ignoring the merger per-call requestOptions.

A caller who does chatClient.prompt().options(OpenAiChatOptions.builder().strict(false).build()) will have the value of strict silently overridden.

Please, pass OpenAiChatOptions requestOptions as a parameter of getChatCompletionTools and read from there.

Can you also, please, add a test case for that.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -978,30 +978,27 @@ private List<ChatCompletionTool> getChatCompletionTools(List<ToolDefinition> too
 			FunctionParameters.Builder parametersBuilder = FunctionParameters.builder();
 
 			if (!toolDefinition.inputSchema().isEmpty()) {
-				// Parse the schema and add its properties directly
+
 				try {
 					@SuppressWarnings("unchecked")
 					Map<String, Object> schemaMap = objectMapper.readValue(toolDefinition.inputSchema(), Map.class);
 
-					// Add each property from the schema to the parameters
```
dimitarproynov, line 986: Please, do not remove comments unrelated to the change.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -978,30 +978,27 @@ private List<ChatCompletionTool> getChatCompletionTools(List<ToolDefinition> too
 			FunctionParameters.Builder parametersBuilder = FunctionParameters.builder();
 
 			if (!toolDefinition.inputSchema().isEmpty()) {
-				// Parse the schema and add its properties directly
```
dimitarproynov, line 981: Please, do not remove comments unrelated to the change.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatOptions.java
prachipanditrao: Refactoring for readability is done

models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java
prachipanditrao: Requested test cases are now added

models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java
prachipanditrao: I've cleaned those up and moved them to proper import statements to keep things consistent.

models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java
prachipanditrao: sorry, added this accidentally. It is now resolved

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
prachipanditrao: Fixed in the latest commit. The method now properly reads from requestOptions to honor per-call overrides, and comprehensive test cases have been added to secure this behavior.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
prachipanditrao: Done. Reverted the accidental comment removals so everything unrelated remains untouched. Thanks for catching that.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -973,10 +973,15 @@ else if (mediaContentData instanceof String text) {
 		}
 	}
 
-	private List<ChatCompletionTool> getChatCompletionTools(List<ToolDefinition> toolDefinitions) {
+	private List<ChatCompletionTool> getChatCompletionTools(List<ToolDefinition> toolDefinitions, @Nullable OpenAiChatOptions requestOptions) {
```
dimitarproynov: From build:

Error:  Failed to execute goal io.spring.javaformat:spring-javaformat-maven-plugin:0.0.47:validate (format-check) on project spring-ai-openai: Formatting violations found in the following files:
Error:   * /home/runner/work/spring-ai/spring-ai/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
Error:   * /home/runner/work/spring-ai/spring-ai/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatOptions.java
Error:   * /home/runner/work/spring-ai/spring-ai/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelTests.java



Run `spring-javaformat:apply` to fix.

dimitarproynov (changes requested):

There are some formatting errors, that will fail the build, please, take care of them.

Best Regards,
Proynov

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
prachipanditrao: Java Format has fixed layout formatting rules across models and tests. It re-aligned Objects.equals(this.strict, options.strict) to the same line as store and restored the missing trailing whitespace at the end of the test class as well

PR #6543, opened 2026-07-02
Fix group precedence in `MilvusFilterExpressionConverter`

Seol-JY:
The `doGroup` override in `MilvusFilterExpressionConverter` converted a parenthesized filter group into `content && content`, duplicating the group body and dropping the parentheses that expressed the grouping.

Because Milvus binds `&&` tighter than `||`, a filter such as:

```
(year >= 2020 OR country == 'BG') AND city != 'Sofia'
```

was emitted without parentheses:

```
metadata["year"] >= 2020 || metadata["country"] == "BG" && metadata["year"] >= 2020 || metadata["country"] == "BG" && metadata["city"] != "Sofia"
```

so the trailing `AND` no longer applied to the whole group, and the query silently matched the wrong documents.

This replaces the override with the conventional `doStartGroup`/`doEndGroup` hooks that wrap the group content in parentheses, consistent with the other filter expression converters (for example `PgVectorFilterExpressionConverter`) and with the base `AbstractFilterExpressionConverter` contract. Making the grouping explicit also means the result no longer relies on Milvus's operator precedence.

The existing `testGroup` assertion codified the incorrect output and has been corrected; a `testGroupAsRightOperand` test was added to cover a group used as the right-hand operand.

### DIFF
--- a/vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/milvus/MilvusFilterExpressionConverter.java
+++ b/vector-stores/spring-ai-milvus-store/src/main/java/org/springframework/ai/vectorstore/milvus/MilvusFilterExpressionConverter.java
@@ -17,7 +17,6 @@
 package org.springframework.ai.vectorstore.milvus;
 
 import org.springframework.ai.vectorstore.filter.Filter.Expression;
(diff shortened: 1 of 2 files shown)


ericbottard:
Thanks for yet another useful contribution @Seol-JY !

PR #6551, opened 2026-07-03
Preserve Anthropic thinking blocks for tool replay

jewoodev:
This is the Anthropic direct-provider counterpart to #6414's Bedrock replay fix.

When an Anthropic response contains `thinking` or `redacted_thinking` blocks before `tool_use`, Spring AI now keeps that provider-specific replay state on the assistant message selected by the tool-calling loop. Follow-up Anthropic requests re-emit those thinking blocks before the corresponding tool-use blocks when tool results continue the same assistant turn.

The change stays scoped to the Anthropic provider and does not introduce a generic reasoning API.

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -1009,10 +1009,20 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 				ThinkingBlock thinkingBlock = block.asThinking();
 				thinkingContents
 					.add(AnthropicThinkingContent.thinking(thinkingBlock.thinking(), thinkingBlock.signature()));
+				Map<String, Object> thinkingProperties = new HashMap<>();
+				thinkingProperties.put("signature", thinkingBlock.signature());
+				generations.add(new Generation(AssistantMessage.builder()
+					.content(thinkingBlock.thinking())
+					.properties(thinkingProperties)
+					.build(), generationMetadata));
 			}
 			else if (block.isRedactedThinking()) {
 				RedactedThinkingBlock redactedBlock = block.asRedactedThinking();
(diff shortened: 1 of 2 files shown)


models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -985,23 +1006,13 @@ else if (block.isToolUse()) {
...
 			else if (block.isThinking()) {
-				// ThinkingBlock: stored as a separate Generation with the thinking
-				// text as content and signature in metadata properties.
 				ThinkingBlock thinkingBlock = block.asThinking();
-				Map<String, Object> thinkingProperties = new HashMap<>();
-				thinkingProperties.put("signature", thinkingBlock.signature());
-				generations.add(new Generation(AssistantMessage.builder()
-					.content(thinkingBlock.thinking())
-					.properties(thinkingProperties)
-					.build(), generationMetadata));
```
sobychacko, line 996: This is breaking change for `2.0.x` users: before this PR, Claude's extended thinking blocks (`thinking`, `redacted_thinking`) were each provided as a separate `Generation` in `response.getResults()`, with the thinking text and signature accessible via message metadata. These lines remove that and everything is consolidated into a single AnthropicAssistantMessage, so any user reading extended thinking content from the response will silently get nothing.

I suggest, restore the `generations.add(...)` above for those thinking blocks alongside the new replay collection. The tool-call replay only cares about the final `AssistantMessage` that has tool calls, and the extended thinking generations are independent and purely for user access. Restoring them preserves backward compatibility without touching any other part of the PR.

sobychacko (changes requested):

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
jewoodev: Thanks for catching this. I restored the `Generation` exposure as suggested.

PR #6555, opened 2026-07-04
Fix unhandled content_block_stop event in Anthropic streaming

### LINKED ISSUE #6372: Unhandled event type: CONTENT_BLOCK_STOP
- spring ai：1.1.7
- spring boot: 3.4.1
- jdk 21

There is a warning about:
```
 2026-06-09 17:07:38.001 WARN  [HttpClient-2-Worker-1] [traceId=,spanId=] [run=,ns=,sess=] o.s.ai.anthropic.api.StreamHelper - Unhandled event type: CONTENT_BLOCK_STOP
```

I looked for many issues but found no similar ones, so I can only submit this issue to find the answer. thanks!

nikita-kibitkin:
Fixes #6372

For non-tool streams the content_block_stop event passes through the tool-merge reduce untouched, and StreamHelper.eventToChatCompletionResponse has no branch for it, so every text content block logs "Unhandled event type: CONTENT_BLOCK_STOP" and emits an empty ChatResponse chunk to consumers. main/2.0 is not affected, so this targets 1.1.x only.

The event cannot simply be dropped before conversion: resetting the aggregated builder content is its load-bearing side effect, without it the message_delta chunk re-emits the last text delta. So StreamHelper now handles content_block_stop explicitly (resets content, no warning) and the converted chunk is dropped in the existing terminal filter of chatCompletionStream. Same behavior as the 2.0 module, tool-use aggregation unchanged.

Testing: new AnthropicApiChatCompletionStreamTests (MockWebServer, no live API) covers a text stream (no CONTENT_BLOCK_STOP chunk, deltas intact; verified to fail without the fix) and a tool-use stream (aggregated tool call still reaches the message_delta chunk). Module tests, spring-javaformat and checkstyle clean.

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/AnthropicApi.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/api/AnthropicApi.java
@@ -265,7 +265,11 @@ public final class AnthropicApi {
 			})
 			.flatMap(mono -> mono)
 			.map(event -> this.streamHelper.eventToChatCompletionResponse(event, chatCompletionReference))
-			.filter(chatCompletionResponse -> chatCompletionResponse.type() != null);
+			// content_block_stop carries no content and has no response chunk
+			// equivalent, but it must still reach the stream helper to reset the
+			// aggregated content before the message_delta event.
+			.filter(chatCompletionResponse -> chatCompletionResponse.type() != null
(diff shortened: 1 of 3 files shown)


sobychacko:
@nikita-kibitkin `1.1.x` is out of OSS support. `AnthropicApi` is no longer present on `main` as we moved onto base the anthropic module on the new Java SDK. If its still an issue on the `main` branch, please open another PR. Closing it now. Thanks!

PR #6557, opened 2026-07-05
Allow empty and null `ToolContext` in tool method invocation

### LINKED ISSUE #6545: Support empty ToolContext, or null values
**Current Behavior**

As of Spring AI 2.0.0, ToolContext cannot contain null values or be empty.

A null tool context map value fails with an `IllegalArgumentException` in `ChatClientRequestSpec.toolContext.toolContext(...)`.

Empty tool context fails during tool execution with an `IllegalArgumentException` in `MethodToolCallback.validateToolContextSupport(...)`.

So when I have to handle optional data (an "orderNumber" in the next example), I have to add some fake data ("bar" under the "foo" key) to prevent an exception :

```java
// Build the context passed to the tools
Map<String, Object> toolContext = new HashMap<>();
if (orderNumber != null) {
    toolContext.put("orderNumber", orderNumber);
} else {
    toolContext.put("foo", "bar"); // Workaround (tool call will fail if the tool context is empty)
}

// Execute the request
return chatClient.prompt(request).toolContext(toolContext).call().content();
```

**Suggested changes**

What are the reasons of these limitations ?

To simplify application code, could you support empty tool context (or else, as a second choice, null values) ?

kuntal1461:
## Problem

Fixes #6545

Two related but independent bugs prevent callers from expressing an optional or absent tool context:

**Bug 1 — `MethodToolCallback` (spring-ai-model)**

`validateToolContextSupport()` treated an *empty* `ToolContext` the same as a *missing* one, and threw `IllegalArgumentException` in both cases. This forced callers to inject fake placeholder entries (e.g. `Map.of("foo", "bar")`) just to satisfy the guard, even when no real context was needed:

```java
// Workaround required before this fix:
Map<String, Object> toolContext = new HashMap<>();
if (orderNumber != null) {
    toolContext.put("orderNumber", orderNumber);
} else {
    toolContext.put("foo", "bar"); // fake entry to prevent exception
}
```

**Bug 2 — `DefaultChatClient` (spring-ai-client-chat)**

`ChatClientRequestSpec.toolContext()` rejected maps that contained `null` values via `Assert.noNullElements()`, making it impossible to represent optional context entries as `null`.

### `spring-ai-model` — `MethodToolCallback`

- Remove `validateToolContextSupport()`. Empty context is valid; the tool method decides which keys it requires.
- In `buildMethodArguments()`, convert a `null` `ToolContext` to `new ToolContext(Map.of())` so tool methods that declare the parameter never receive `null`, preventing NPE without throwing.
- Remove now-unused imports `ClassUtils` and `CollectionUtils`.

### `spring-ai-client-chat` — `DefaultChatClient`

- Remove the `Assert.noNullElements()` guard on context map **values** only.
- The following guards are **preserved unchanged**:
  - `null` map reference → still rejected
  - `null` keys → still rejected
  - blank/empty string keys → still rejected

## Behaviour matrix

| Scenario | Before | After |
|---|---|---|
| Method declares `ToolContext` + `null` context | throws | tool receives `ToolContext(Map.of())` |
| Method declares `ToolContext` + empty context | throws | tool receives the empty context |
| Method declares `ToolContext` + non-empty context | works | works (unchanged) |
| Method does not declare `ToolContext` + any context | silently ignored | silently ignored (unchanged) |
| `toolContext(map)` with `null` value | throws | accepted |
| `toolContext(null)` map reference | throws | throws (unchanged) |
| `toolContext(map)` with `null` key | throws | throws (unchanged) |

## Tests

**New: `MethodToolCallbackToolContextTests`** — 13 tests covering:
- `null` context (implicit via `call(input)`) → tool receives empty `ToolContext`
- explicit `null` context → tool receives empty `ToolContext`
- empty `HashMap` and `Map.of()` are accepted
- single and multiple context entries pass through unchanged (backward compatibility)
- `null` value in context map is forwarded to the tool
- key case is preserved
- method without `ToolContext` param: `null` / empty / non-empty context are all silently ignored (original behaviour preserved)
- exception path with empty and `null` context wraps correctly in `ToolExecutionException`

**Updated: `DefaultChatClientTests`** — 2 tests renamed from `*ThenThrow` to `*ThenSucceed` to reflect the new allowed behaviour for `null` map values.

## Test results

| Module | Tests | Failures |
|---|---|---|
| `spring-ai-model` | 829 | 0 |
| `spring-ai-client-chat` | 569 | 0 |

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
@@ -1080,6 +1080,8 @@ public class DefaultChatClient implements ChatClient {
 		public ChatClientRequestSpec toolContext(Map<String, Object> toolContext) {
 			Assert.notNull(toolContext, "context cannot be null");
 			Assert.noNullElements(toolContext.keySet(), "context keys cannot contain null elements");
+			Assert.noNullElements(toolContext.values(), "context values cannot contain null elements; "
+					+ "to represent an optional entry, omit the key rather than setting it to null");
 			toolContext.keySet().forEach(key -> Assert.hasText(key, "context key cannot be null or empty"));
 			this.toolContext.putAll(toolContext);
 			return this;
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/DefaultChatClient.java
@@ -1080,7 +1080,6 @@ public class DefaultChatClient implements ChatClient {
 		public ChatClientRequestSpec toolContext(Map<String, Object> toolContext) {
 			Assert.notNull(toolContext, "context cannot be null");
 			Assert.noNullElements(toolContext.keySet(), "context keys cannot contain null elements");
-			Assert.noNullElements(toolContext.values(), "context values cannot contain null elements");
 			toolContext.keySet().forEach(key -> Assert.hasText(key, "context key cannot be null or empty"));
 			this.toolContext.putAll(toolContext);
 			return this;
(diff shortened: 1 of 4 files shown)


kuntal1461 (commented):
~~Review comments have been retracted.~~

PR #6563, opened 2026-07-06
Stop registering internal RedisClient as bean

quaff:
Align `chat-memory-repository-redis` and `vector-store-redis-semantic-cache` with `vector-store-redis`, keep `JedisClient` isolated with each other module.


sdeleuze:
@quaff Could you please share more context about this ask? Or is it only for consistency?

quaff:
> @quaff Could you please share more context about this ask? Or is it only for consistency?

Not only for consistency, `chat-memory-repository-redis` is using it's own host and port from  `spring.ai.chat.memory.repository.redis.*`, its `RedisClient` presence may backoff `vector-store-redis-semantic-cache`'s `RedisClient` which using host and port from `spring.data.redis.*`, it's unexpected behavior.

quaff:
I have a better solution: https://github.com/spring-projects/spring-ai/compare/main...quaff:spring-ai:patch-80?expand=1

Should I close this and create another one? @sdeleuze

PR #6566, opened 2026-07-06
Use Locale.ROOT for String uppercase/lowercase operations on wire/protocol values

### LINKED ISSUE #6478: PineconeFilterExpressionConverter corrupts $in/$nin operators on Turkish locale
**Bug description**

`PineconeFilterExpressionConverter` builds Pinecone metadata-filter operators by lowercasing the `Filter.ExpressionType` enum name **without specifying a `Locale`**:

```java
private String getOperationSymbol(Expression exp) {
    return "\"$" + exp.type().toString().toLowerCase() + "\": ";
}
```

`exp.type().toString()` is always the uppercase enum constant. On a JVM running the Turkish locale (`tr_TR`), `"IN".toLowerCase()` returns `"ın"` (dotless ı) and `"NIN".toLowerCase()` returns `"nın"`, so the converter emits `"$ın"` / `"$nın"` instead of `"$in"` / `"$nin"`. These are not valid Pinecone operators, so any metadata filter using `IN` / `NIN` fails on Turkish-locale JVMs.

**Environment**

- Spring AI: `main` (2.0.0-SNAPSHOT); the pattern also exists in 1.x
- JVM locale: `tr_TR`

**Steps to reproduce**

Run with `-Duser.language=tr -Duser.country=TR` and convert an `IN` filter:

```java
new PineconeFilterExpressionConverter()
    .convertExpression(new Expression(IN, new Key("genre"), new Value(List.of("comedy", "drama"))));
// expected: {"genre": {"$in": ["comedy","drama"]}}
// actual:   {"genre": {"$ın": ["comedy","drama"]}}
```

**Expected behavior**

Operator symbols are protocol tokens and must be locale-independent. The conversion should use `Locale.ROOT`.

**Related**

Same bug class as #5340 (Gemini tool type normalization) and #6476 / #6477 (OpenAI protocol values). This is a different, unreported code path in `spring-ai-vector-store`.

dimitarproynov:
On Turkish locale the operation "I".toLowerCase() results in "ı", which breaks string comparisons. Similarly, the "i".toUpperCase() results in "İ".

This change uses .toUpperCase(Locale.ROOT) and  .toLowerCase(Locale.ROOT) to address all such code occurrences where protocol/wire values are uppercased/lowercased, including test code, to achieve a stable build with "tr-TR" locale JVMs.

RegexToolIndex and RedisVectorStore use Locale.getDefault to pass a future implementation of checkstyle check that will prevent this bug in the future, as they operate on user-supplied text,

Closes https://github.com/spring-projects/spring-ai/issues/6478 , https://github.com/spring-projects/spring-ai/issues/5340

Signed-off-by: Dimitar Proynov <dimitar.proynov@broadcom.com>

### DIFF
--- a/advisors/spring-ai-vector-store-advisor/src/test/java/org/springframework/ai/chat/client/advisor/vectorstore/VectorStoreChatMemoryAdvisorTests.java
+++ b/advisors/spring-ai-vector-store-advisor/src/test/java/org/springframework/ai/chat/client/advisor/vectorstore/VectorStoreChatMemoryAdvisorTests.java
@@ -17,6 +17,7 @@
 package org.springframework.ai.chat.client.advisor.vectorstore;
 
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 import org.junit.jupiter.api.Test;
@@ -25,8 +26,11 @@ import org.mockito.Mockito;
 import reactor.core.scheduler.Scheduler;
 
 import org.springframework.ai.chat.client.ChatClient;
+import org.springframework.ai.chat.client.ChatClientRequest;
 import org.springframework.ai.chat.memory.ChatMemory;
 import org.springframework.ai.chat.messages.AssistantMessage;
+import org.springframework.ai.chat.messages.SystemMessage;
+import org.springframework.ai.chat.messages.UserMessage;
 import org.springframework.ai.chat.model.ChatModel;
 import org.springframework.ai.chat.model.ChatResponse;
 import org.springframework.ai.chat.model.Generation;
@@ -239,6 +243,33 @@ class VectorStoreChatMemoryAdvisorTests {
 			.contains("&apos;");
 		assertThat(systemText).doesNotContain("AT&T").doesNotContain("<fine>");
 		assertThat(systemText).containsOnlyOnce("<memory-entry type=\"assistant\">");
+	}
+
+	@Test
+	void memoryEntryTypeUsesRootLocaleUnderTurkishLocale() {
+		Locale defaultLocale = Locale.getDefault();
+		try {
+			// "ASSISTANT".toLowerCase() under tr-TR yields "assıstant" (dotless ı);
+			// Correct handling must produce ASCII "assistant".
+			Locale.setDefault(Locale.forLanguageTag("tr-TR"));
+
+			Document doc = Document.builder().text("hi").metadata(Map.of("messageType", "ASSISTANT")).build();
+
+			VectorStore vectorStore = Mockito.mock(VectorStore.class);
+			given(vectorStore.similaritySearch(any(SearchRequest.class))).willReturn(List.of(doc));
+
+			ChatClientRequest request = ChatClientRequest.builder()
+				.context(ChatMemory.CONVERSATION_ID, "test-session")
+				.prompt(new Prompt(List.of(new SystemMessage("instructions"), new UserMessage("query"))))
+				.build();
+
+			ChatClientRequest result = VectorStoreChatMemoryAdvisor.builder(vectorStore).build().before(request, null);
+
+			assertThat(result.prompt().getSystemMessage().getText()).contains("<memory-entry type=\"assistant\">");
+		}
+		finally {
+			Locale.setDefault(defaultLocale);
+		}
 	}
 
 }
--- a/mcp/common/src/test/java/org/springframework/ai/mcp/ToolUtilsTests.java
+++ b/mcp/common/src/test/java/org/springframework/ai/mcp/ToolUtilsTests.java
@@ -19,6 +19,7 @@ package org.springframework.ai.mcp;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 import io.modelcontextprotocol.client.McpSyncClient;
@@ -208,6 +209,21 @@ class ToolUtilsTests {
 		// Compatibility
 		String result = McpToolUtils.prefixedToolName("前缀\u3400", "缀\\u3400", "工具\u9fff名称\uf900");
 		assertThat(result).isEqualTo("前_缀u3400_工具鿿名称豈");
+	}
+
+	@Test
+	void prefixedToolNameShouldHandleTrLocale() {
+		Locale defaultLocale = Locale.getDefault();
+		try {
+			Locale.setDefault(Locale.forLanguageTag("tr-TR"));
+			// Prefix "Inventory" would be lowercased to "ı" (dotless i) in turkish locale
+			// if not properly handled.
+			String result = McpToolUtils.prefixedToolName("Inventory", "server1", "toolName");
+			assertThat(result).isEqualTo("i_server1_toolName");
+		}
+		finally {
+			Locale.setDefault(defaultLocale);
+		}
 	}
 
 	@Test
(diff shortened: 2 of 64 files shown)


advisors/spring-ai-vector-store-advisor/src/test/java/org/springframework/ai/chat/client/advisor/vectorstore/VectorStoreChatMemoryAdvisorTests.java
```diff
@@ -241,4 +242,51 @@ void whenDocumentTextContainsSpecialXmlCharactersThenAllAreEscaped() {
 		assertThat(systemText).containsOnlyOnce("<memory-entry type=\"assistant\">");
 	}
 
+	@Test
+	void whenDocumentTextContainsSpecialXmlCharactersThenAllAreEscapedTrLocale() {
+		Locale defaultLocale = Locale.getDefault();
+		try {
+			Locale.setDefault(Locale.forLanguageTag("tr-TR"));
+			String textWithSpecialChars = "AT&T said \"it's <fine>\"";
```
ericbottard: I think this test should clarify why tr-TR is interesting here, maybe via a comment.
Are all of the "i"s problematic here, etc? Also, isn't this test testing several things all at the same time (maybe some are tested elsewhere already)? Like xml chars but also "i"s that change to other chars in turkish, etc

mcp/common/src/test/java/org/springframework/ai/mcp/ToolUtilsTests.java
```diff
@@ -210,6 +211,19 @@ void prefixedToolNameShouldHandleAllHanScriptCharacters() {
 		assertThat(result).isEqualTo("前_缀u3400_工具鿿名称豈");
 	}
 
+	@Test
+	void prefixedToolNameShouldHandleShouldSupportTrLocale() {
```
ericbottard: method name is incoherent here

models/spring-ai-google-genai/src/test/java/org/springframework/ai/google/genai/CreateGeminiRequestTests.java
```diff
@@ -506,6 +508,35 @@ public void createRequestWithRuntimeThinkingLevelOverrideOnProModelThrows() {
 			.hasMessageContaining("not supported");
 	}
 
+	@Test
+	public void createRequestWithRuntimeThinkingLevelOverrideOnProModelThrowsLocaleTr() {
```
ericbottard: It's not clear to me how turkish locale affects the test here

spring-ai-model/src/test/java/org/springframework/ai/util/json/JsonSchemaGeneratorTests.java
```diff
@@ -291,6 +292,42 @@ void generateSchemaForMethodWithUpperCaseTypes() throws Exception {
 		assertThat(schema).isEqualToIgnoringWhitespace(expectedJsonSchema);
 	}
 
+	@Test
+	void generateSchemaForMethodWithUpperCaseTypesTrLocale() throws Exception {
```
ericbottard: It's not clear to me, reading the test, what could be a problem if ROOT wasn't used (and yet I'm in the mental context of reviewing the PR). Maybe add a comment?

Is is the "i" in the method name? But it doesn't appear in the generated schema. Is it the "i"s in STRING and INTEGER?

spring-ai-tool-search-tool/src/main/java/org/springframework/ai/tool/toolsearch/index/regex/RegexToolIndex.java
```diff
@@ -165,7 +166,8 @@ protected String convertQueryToRegexPattern(String query) {
 				"just", "also", "very", "too", "so");
 
 		// Tokenize: split by whitespace and common delimiters
-		String[] tokens = query.toLowerCase().split("[\\s,;:.!?()\\[\\]{}\"']+");
+		// Use toLowerCase(Locale.getDefault()) to pass checkstyle
```
ericbottard: I don't understand this comment, please clarify

ericbottard: I think I understand (from the one below). But there is no checkstyle rule added by this PR, is there? In any case, this does not warrant a comment, which IMO adds more confusion than clarification

ericbottard: Also, I'm not sure Locale.getDefault() is the correct choice here. This may be more complicated than this.
If the text is user provided, a 100% correct thing to do would be to use the locale the text is in (which is not the locale of the running JVM).

Maybe there is a way to not rely on toLowerCase() at all?

spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
```diff
@@ -128,7 +129,9 @@ public Filter.Expression parse(String textFilterExpression) {
 		Assert.hasText(textFilterExpression, "Expression should not be empty!");
 
 		// Prefix the expression with the compulsory WHERE keyword.
-		if (!textFilterExpression.toUpperCase().startsWith(WHERE_PREFIX)) {
+		// "where" has no difference when uppercased in any locale, but add locale to pass
```
ericbottard: I think Locale.ROOT would make more sense here

advisors/spring-ai-vector-store-advisor/src/test/java/org/springframework/ai/chat/client/advisor/vectorstore/VectorStoreChatMemoryAdvisorTests.java
dimitarproynov: Rewrote the test and added comments.

mcp/common/src/test/java/org/springframework/ai/mcp/ToolUtilsTests.java
dimitarproynov: Fixed

models/spring-ai-google-genai/src/test/java/org/springframework/ai/google/genai/CreateGeminiRequestTests.java
dimitarproynov: Refined test. Added comment

spring-ai-model/src/test/java/org/springframework/ai/util/json/JsonSchemaGeneratorTests.java
dimitarproynov: Added a comment. It is the "i" in the type name resulting in STRİNG and İNTEGER

spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
dimitarproynov: Yep, you are right. Fixed

spring-ai-tool-search-tool/src/main/java/org/springframework/ai/tool/toolsearch/index/regex/RegexToolIndex.java
dimitarproynov: Switched to use Locale.ROOT and wrote a test for it.

ericbottard:
@dimitarproynov I've added https://github.com/spring-projects/spring-ai/commit/35be70af02829dfea85f7577ab560b933c1d1ffa#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L483 to make sure this kind of error is caught from now on

PR #6567, opened 2026-07-06
Fix `OllamaChatModel` prompt options not inheriting configured model

ilayaperumalg:
When prompt-level `OllamaChatOptions` were provided without an explicit
  model, the options defaulted to `OllamaModel.MISTRAL` regardless of the
  model configured on `OllamaChatModel`. This caused requests to fail with
  `404 - model 'mistral' not found` in environments where only a different
  model is available.

  Align `OllamaChatModel.Builder` with `OpenAiChatModel.Builder` by making
  the `options` field `@Nullable` and resolving it in `build()` rather than
  eagerly initialising it with a hardcoded default. Update `OllamaChatModelIT`
  tests that pass prompt-level options to explicitly set the test model,
  which is the correct practice when overriding specific options at the
  prompt level.

### DIFF
--- a/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
+++ b/models/spring-ai-ollama/src/main/java/org/springframework/ai/ollama/OllamaChatModel.java
@@ -504,7 +504,7 @@ public class OllamaChatModel implements ChatModel {
 
 		private @Nullable OllamaApi ollamaApi;
 
-		private OllamaChatOptions options = OllamaChatOptions.builder().build();
+		private @Nullable OllamaChatOptions options;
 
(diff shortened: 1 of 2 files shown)


nicolaskrier:
Hi @ilayaperumalg and @sdeleuze, I had opened #6319 one month ago regarding this issue but I closed it because it seemed `404 - model 'mistral' not found` error had disappeared.

I understand now that @ilayaperumalg's solution is the good one but I wonder if it would make sense to promote more recent models for Mistral AI in `OllamaModel`.

sdeleuze:
For 2.0.x, I think it makes sense to keep current defaults to not break existing use case. Let's consider upgrading the default in 2.1 if that makes sense.

PR #6587, opened 2026-07-09
Restore Jedis 7.0 compatibility for Redis integrations

### LINKED ISSUE #6584: Cannot auto-configure redis chat memory due to unresolved jedis module
**Bug description**
`jedis` does not provide `RedisClient`

Image

**Environment**

- Spring AI: 2.0.0
- Java: 25
- jedis: 7.0.0

**Steps to reproduce**
1. Add to pom.xml:

```xml
<dependency>
    <artifactId>spring-ai-starter-model-openai</artifactId>
</dependency>
<dependency>
    <artifactId>spring-ai-starter-model-chat-memory-repository-redis</artifactId>
</dependency>
```

2. Add `RedisChatMemoryRepository` and observe the issue that the bean is not found

Image

**Expected behavior**

A bean for `RedisChatMemoryRepository` should be created by default when `spring-ai-starter-model-chat-memory-repository-redis` is added to `pom.xml`. However, `RedisClient` is not found in `jedis`, the bean could not be initialized.

No.

jewoodev:
Redis chat memory currently references `RedisClient`, which is only available in newer Jedis versions. Spring Boot 4.0.x manages Jedis 7.0.0, so the Redis chat memory starter can fail during auto-configuration before `RedisChatMemoryRepository` is created.

This changes the Redis integrations to use `UnifiedJedis` as the client-facing type and creates default clients with `JedisPooled`, which is available in Jedis 7.0.0. Newer Jedis clients that extend `UnifiedJedis` can still be provided by users.

The same compatibility adjustment is applied across Redis chat memory, Redis vector store, Redis semantic cache, tests, and docs.

Validated with Jedis 7.0.0 by running dependency resolution and package verification for the affected Redis modules and auto-configurations.

Fixes #6584

### DIFF
--- a/auto-configurations/models/chat/memory/repository/spring-ai-autoconfigure-model-chat-memory-repository-redis/src/main/java/org/springframework/ai/model/chat/memory/repository/redis/autoconfigure/RedisChatMemoryRepositoryAutoConfiguration.java
+++ b/auto-configurations/models/chat/memory/repository/spring-ai-autoconfigure-model-chat-memory-repository-redis/src/main/java/org/springframework/ai/model/chat/memory/repository/redis/autoconfigure/RedisChatMemoryRepositoryAutoConfiguration.java
@@ -16,7 +16,8 @@
 
 package org.springframework.ai.model.chat.memory.repository.redis.autoconfigure;
 
-import redis.clients.jedis.RedisClient;
+import redis.clients.jedis.JedisPooled;
+import redis.clients.jedis.UnifiedJedis;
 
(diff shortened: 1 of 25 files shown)


quaff:
Spring AI is base on Spring Boot 4.1 which uses jedis 7.4 not 7.0.

jewoodev:
The reported application resolves Jedis 7.0.0 through its dependency management. This PR uses `UnifiedJedis`, which is available in both Jedis 7.0 and 7.4.

PR #6589, opened 2026-07-10
Cache Pinecone `Index` connection in `PineconeVectorStore`

TimurRakhmatullin86:
`PineconeVectorStore` called `Pinecone.getIndexConnection(indexName)` on every `add`, `delete`, and `similaritySearch` operation. In the Pinecone Java client, `getIndexConnection` performs a blocking control-plane REST call (`describeIndex`) to resolve the index host on every invocation. Under bulk ingestion or query load this adds an extra HTTP round-trip of latency to each data-plane operation and runs into control-plane rate limits (HTTP 429).

The repeated call is pure overhead: the underlying gRPC connection is already cached in the client's static `connectionsMap` (`computeIfAbsent`), so the host resolved by every `describeIndex` call after the first one is discarded.

Changes:

- Resolve the `Index` connection lazily once and reuse it across operations (`volatile` field with double-checked locking). If resolution fails, the field stays unset and the next operation retries.
- The cached connection is intentionally never closed by the store: `Index.close()` closes the shared connection and removes it from the client's static `connectionsMap`, which would affect other consumers of the same index. The store did not close per-operation connections before this change either, so lifecycle behavior is unchanged.
- Remove dead code in `similaritySearch`: a `QueryRequest` builder was constructed but never used (the query is issued via `queryByVector`), along with the now-unused `io.pinecone.proto.QueryRequest` import.
- Add `PineconeVectorStoreTests` (no network) verifying the connection is resolved exactly once across add/delete/search and lazily on first use.

### DIFF
--- a/vector-stores/spring-ai-pinecone-store/src/main/java/org/springframework/ai/vectorstore/pinecone/PineconeVectorStore.java
+++ b/vector-stores/spring-ai-pinecone-store/src/main/java/org/springframework/ai/vectorstore/pinecone/PineconeVectorStore.java
@@ -25,8 +25,8 @@ import java.util.Optional;
 import com.google.protobuf.Struct;
 import com.google.protobuf.Value;
 import com.google.protobuf.util.JsonFormat;
+import io.pinecone.clients.Index;
 import io.pinecone.clients.Pinecone;
-import io.pinecone.proto.QueryRequest;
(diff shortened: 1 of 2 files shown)


TimurRakhmatullin86:
Friendly ping — happy to rebase or make any changes if needed.

PR #6596, opened 2026-07-10
Scope Vertex AI credentials to cloud-platform

### LINKED ISSUE #6595: GoogleGenAiChatAutoConfiguration builds Vertex AI credentials unscoped → invalid_scope on token refresh
### Bug description

In Vertex AI mode, the google-genai starter applies the service-account credentials loaded from `spring.ai.google.genai.credentials-uri` **without an OAuth scope**, so the access-token refresh is rejected by Google with `invalid_scope`.

`GoogleGenAiChatAutoConfiguration#configureVertexAi(...)` builds credentials via `GoogleCredentials.fromStream(is)` and passes them straight to the client builder:

```java
private void configureVertexAi(Client.Builder builder, GoogleGenAiConnectionProperties props) throws IOException {
    ...
    builder.project(props.getProjectId()).location(props.getLocation()).vertexAI(true);
    if (props.getCredentialsUri() != null) {
        try (var is = props.getCredentialsUri().getInputStream()) {
            builder.credentials(GoogleCredentials.fromStream(is));   // <-- not scoped
        }
    }
}
```

`GoogleCredentials.fromStream(...)` on a service-account key returns a **scopeless** `ServiceAccountCredentials`. Vertex AI requires the `https://www.googleapis.com/auth/cloud-platform` scope, so token refresh fails:

```
com.google.auth.oauth2.GoogleAuthException: Error getting access token for service account: 400 Bad Request
POST https://oauth2.googleapis.com/token
{"error":"invalid_scope","error_description":"Invalid OAuth scope or ID token audience provided."}
iss: <service-account>@<project>.iam.gserviceaccount.com
```

This only affects the explicit `credentials-uri` path (the ADC path is scoped elsewhere), and there is currently **no way to fix it via configuration**: `GoogleGenAiConnectionProperties` exposes no scope property, and the `googleGenAiClient` bean offers no `Client.Builder` customizer hook (it is only `@ConditionalOnMissingBean`), so the sole workaround is to replace the entire `Client` bean.

Related: #5242 (`credentials-uri` not applied to the client in 1.1.2). In 2.0.0 the credentials **are** applied — but unscoped — so this is the follow-on defect.

### Environment

- Spring AI: **2.0.0** (`spring-ai-starter-model-google-genai` / `spring-ai-autoconfigure-model-google-genai:2.0.0`)
- google-genai SDK: `com.google.genai:google-genai:1.58.0`
- Spring Boot: 4.1.x
- Java: 21
- Mode: Vertex AI (`spring.ai.google.genai.vertex-ai=true`), credentials via `spring.ai.google.genai.credentials-uri`
- No vector store involved

### Steps to reproduce

1. Configure Vertex AI mode with an explicit service-account key (not ADC):
   ```properties
   spring.ai.google.genai.vertex-ai=true
   spring.ai.google.genai.project-id=<project>
   spring.ai.google.genai.location=us-central1
   spring.ai.google.genai.credentials-uri=file:/path/to/service-account.json
   spring.ai.model.chat=google-genai
   ```
2. Autowire `ChatModel` (or `ChatClient.Builder`) and make any call, e.g. `chatClient.prompt("hi").call().content()`.
3. The call fails: `GoogleAuthException: Error getting access token for service account: 400 Bad Request` / `{"error":"invalid_scope"}`.

### Expected behavior

Service-account credentials supplied via `spring.ai.google.genai.credentials-uri` should authenticate Vertex AI calls successfully. The autoconfiguration should scope them to `cloud-platform` (as the legacy `vertex-ai-gemini` starter did), e.g.:

```java
builder.credentials(GoogleCredentials.fromStream(is)
        .createScoped(java.util.List.of("https://www.googleapis.com/auth/cloud-platform")));
```

Ideally also expose a way to configure this without replacing the client — a scope property and/or an `ObjectProvider<Consumer<Client.Builder>>` customizer hook on the `googleGenAiClient` bean.

### Minimal Complete Reproducible example

Workaround that confirms the root cause — supplying a scoped `Client` (the starter's bean is `@ConditionalOnMissingBean`, so it backs off) makes the same call succeed:

```java
@TestConfiguration
class ScopedGoogleGenAiClientConfig {

    @Bean
    Client googleGenAiClient(Environment env) throws IOException {
        var credentials = GoogleCredentials
                .fromStream(new FileInputStream(env.getProperty("spring.ai.google.genai.credentials-uri-path")))
                .createScoped(List.of("https://www.googleapis.com/auth/cloud-platform"));  // <-- the missing step
        return Client.builder()
                .project(env.getProperty("spring.ai.google.genai.project-id"))
                .location(env.getProperty("spring.ai.google.genai.location"))
                .vertexAI(true)
                .credentials(credentials)
                .build();
    }
}
```

With the stock autoconfiguration (no custom `Client`), the same setup throws `invalid_scope`; adding `.createScoped(cloud-platform)` is the only difference that makes it pass.

subhashpolisetti:
### Problem

In Vertex AI mode, the google-genai starter loads the service-account
credentials from `spring.ai.google.genai.credentials-uri` and currently
applies them to the client without an OAuth scope. `GoogleCredentials.fromStream`
returns scopeless credentials, so Vertex AI rejects the access-token
refresh with `invalid_scope` and no chat request can complete:

    GoogleAuthException: Error getting access token for service account:
    400 Bad Request POST https://oauth2.googleapis.com/token
    {"error":"invalid_scope"}

Only the explicit `credentials-uri` path is affected; the Application
Default Credentials path is scoped by the underlying SDK. There is no
configuration workaround today: there is no scope property, and the
`Client` bean is `@ConditionalOnMissingBean` with no customizer, so the
only escape is replacing the entire bean.

### Fix

Scope the credentials loaded from `credentials-uri` to `cloud-platform`,
matching the scope the SDK applies to Application Default Credentials.

### Testing

Adds `GoogleGenAiChatAutoConfigurationTests`, which is fully
self-contained (no network, no real key, no environment variables — it
generates a throwaway RSA keypair and a synthetic service-account JSON):

- one test pins the root cause: credentials from
  `GoogleCredentials.fromStream` carry no scope;
- one test verifies the fix: the loaded credentials are scoped to
  `cloud-platform`.

I also verified the behavior end to end against a real Google Cloud
service-account key. The unit test can only assert the scope offline,
because `invalid_scope` is returned by a live token exchange — the exact
call the SDK makes (`refreshIfExpired()` on the provided credentials):

```java
// unscoped credentials (current behavior)
GoogleCredentials.fromStream(key).refreshIfExpired();
// 400 Bad Request POST https://oauth2.googleapis.com/token
// {"error":"invalid_scope","error_description":"Invalid OAuth scope or ID token audience provided."}

// scoped to cloud-platform (the fix)
GoogleCredentials.fromStream(key)
    .createScoped(List.of("https://www.googleapis.com/auth/cloud-platform"))
    .refreshIfExpired();
// access token obtained
```

Fixes #6595

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-google-genai/src/main/java/org/springframework/ai/model/google/genai/autoconfigure/chat/GoogleGenAiChatAutoConfiguration.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-google-genai/src/main/java/org/springframework/ai/model/google/genai/autoconfigure/chat/GoogleGenAiChatAutoConfiguration.java
@@ -17,6 +17,8 @@
 package org.springframework.ai.model.google.genai.autoconfigure.chat;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
 
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.genai.Client;
@@ -41,6 +43,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
+import org.springframework.core.io.Resource;
 import org.springframework.core.retry.RetryTemplate;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
@@ -64,6 +67,8 @@ import org.springframework.util.StringUtils;
 public class GoogleGenAiChatAutoConfiguration {
 
 	private static final Log logger = LogFactory.getLog(GoogleGenAiChatAutoConfiguration.class);
+
+	private static final String CLOUD_PLATFORM_SCOPE = "https://www.googleapis.com/auth/cloud-platform";
 
 	@Bean
 	@ConditionalOnMissingBean
@@ -116,10 +121,17 @@ public class GoogleGenAiChatAutoConfiguration {
 
 		builder.project(props.getProjectId()).location(props.getLocation()).vertexAI(true);
 
-		if (props.getCredentialsUri() != null) {
-			try (var is = props.getCredentialsUri().getInputStream()) {
-				builder.credentials(GoogleCredentials.fromStream(is));
-			}
+		Resource credentialsUri = props.getCredentialsUri();
+		if (credentialsUri != null) {
+			builder.credentials(loadScopedCredentials(credentialsUri));
+		}
+	}
+
+	// Credentials from GoogleCredentials.fromStream are scopeless; Vertex AI rejects the
+	// token refresh with invalid_scope unless they are scoped to cloud-platform.
+	static GoogleCredentials loadScopedCredentials(Resource credentialsUri) throws IOException {
+		try (InputStream inputStream = credentialsUri.getInputStream()) {
+			return GoogleCredentials.fromStream(inputStream).createScoped(List.of(CLOUD_PLATFORM_SCOPE));
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


mikecollard (approved):
lgtm

PR #6598, opened 2026-07-11
Make JSON schema strict configurable in OpenAiChatModel

### LINKED ISSUE #6564: Why is the `strict` setting of `json_schema` set to `true` by default instead of being configurable?
It was configurable in 1.0, so why did 2.0 make it non-configurable?

Image

OneZero-Y:
Fixes #6564

## Summary

In Spring AI 2.0, `OpenAiChatModel` hardcoded `strict(true)` when building JSON schema response format requests. In 1.x this was configurable; users who need non-strict JSON schema output had no way to disable it.

This PR adds an optional `strict` property to `OpenAiChatModel.ResponseFormat` and applies it when constructing the OpenAI SDK request. When unset, behavior remains backward compatible (`strict` defaults to `true`).

## Changes

- Add `strict` to `OpenAiChatModel.ResponseFormat` (getter, setter, builder)
- Use `ResponseFormat.strict` in `OpenAiChatModel#createRequest` instead of hardcoding `true`
- Add unit tests in `OpenAiChatModelTests` and `OpenAiChatOptionsTests`

## Configuration

Programmatic:
```java
OpenAiChatModel.ResponseFormat.builder()
    .jsonSchema(schema)
    .strict(false)
    .build();
```

Properties:

```yaml
spring:
  ai:
    openai:
      chat:
        options:
          response-format:
            type: json_schema
            json-schema: |
              {"type":"object","properties":{"name":{"type":"string"}}}
            strict: false
```

## Test plan

- [x] ./mvnw -pl models/spring-ai-openai test '-Dtest=OpenAiChatModelTests#jsonSchemaResponseFormatStrict*,OpenAiChatOptionsTests#testSetOutputSchema*'
- [x] Default strict=true when unset
- [x] strict=false propagated to the outgoing request

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -768,7 +768,8 @@ public final class OpenAiChatModel implements ChatModel {
 					ResponseFormatJsonSchema.JsonSchema.Builder jsonSchemaBuilder = ResponseFormatJsonSchema.JsonSchema
 						.builder();
 					jsonSchemaBuilder.name("json_schema");
-					jsonSchemaBuilder.strict(true);
+					Boolean strict = responseFormat.getStrict();
+					jsonSchemaBuilder.strict(strict != null ? strict : true);
 
 					ResponseFormatJsonSchema.JsonSchema.Schema schema = objectMapper.readValue(jsonSchemaString,
 							ResponseFormatJsonSchema.JsonSchema.Schema.class);
@@ -1237,12 +1238,15 @@ public final class OpenAiChatModel implements ChatModel {
 	 * @author Hyunjoon Choi
 	 * @author Jonghoon Park
 	 * @author Sebastien Deleuze
+	 * @author Bishen Yu
 	 */
 	public static class ResponseFormat {
 
 		private Type type = Type.TEXT;
 
 		private @Nullable String jsonSchema;
+
+		private @Nullable Boolean strict;
 
 		public Type getType() {
 			return this.type;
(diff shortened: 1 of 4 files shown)


OneZero-Y:
Hey @dimitarproynov  — would you mind taking a look when you get a chance? This restores a configurable strict on JSON schema response format (was hardcoded to true in 2.0). Happy to adjust anything.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -1282,6 +1299,16 @@ public Builder jsonSchema(String jsonSchema) {
 				return this;
 			}
 
+			/**
+			 * Whether to enable strict schema adherence for JSON schema response format.
```
dimitarproynov, line 1325: Please, add a line in the javadoc for the user not to be confused with OpenAiChatOptions.builder().strict(), which applies to the function calling.

Can you also, while you are it, add one line at the OpenAiChatOptions.builder().strict() javadoc to not be confused with OpenAiChatOptions.builder().responseFormat(ResponseFormat.builder().strict())

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
```diff
@@ -1237,13 +1238,16 @@ private static long getCreated(ChatCompletionChunk chunk) {
...
 	 * @author Sebastien Deleuze
+	 * @author Bishen Yu
 	 */
 	public static class ResponseFormat {
 
 		private Type type = Type.TEXT;
 
 		private @Nullable String jsonSchema;
 
+		private @Nullable Boolean strict;
```
dimitarproynov, line 1249: Could you, please, generate equals and hashCode methods for ResponseFormat? I know the absence is not related to this change, but it affects it with another field and the ResponseFormat.equals() is used in the OpenAiChatOptions.equals()

dimitarproynov (changes requested):
Thank you for your contribution! Overall a minimal and well-tested change, quite solid. Just a few nits below.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
OneZero-Y: Done — ResponseFormat.strict is for JSON schema response format, and OpenAiChatOptions.strict is for tool/function calling.

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
OneZero-Y: Added equals/hashCode on ResponseFormat based on type, jsonSchema, and strict.

dimitarproynov (approved):

PR #6600, opened 2026-07-11
Fix ParagraphManager unconditionally printing PDF outline to stdout

### LINKED ISSUE #6599: ParagraphPdfDocumentReader    unconditionally prints PDF Table of Contents to stdout
While testing ParagraphPdfDocumentReader, I noticed that creating the reader prints the entire PDF outline (Table of Contents) to the console.

This happens automatically during reader initialization without any configuration or debug option.

The behavior comes from the ParagraphManager constructor.

At  line 67, the constructor calls

Image

it executes every time a ParagraphManager instance is created.

ParagraphPdfDocumentReader always prints the Table of Contents to the console. If you're reading many PDFs, the logs become messy and difficult to follow.

**Expected Behavior**

When creating a `ParagraphPdfDocumentReader`, no output should be written to the console during initialization.

If the PDF outline (Table of Contents) is useful for debugging, it should only be shown when debug logging is explicitly enabled.

**Current Behavior**

Every time a `ParagraphPdfDocumentReader` is created, the complete PDF outline (Table of Contents) is printed to `System.out`.

After investigating the source code, I found that this happens in the `ParagraphManager` constructor (around line 67), where the following line executes unconditionally on every new instance:

    printParagraph(this.rootParagraph, System.out);

**Why This Matters**

Since the output is written directly to `System.out`, it cannot be controlled through the application's logging configuration.

When processing multiple PDF files, the console becomes filled with the Table of Contents for every document, making logs noisy and difficult to read.

I also noticed that other classes in the same module — `PagePdfDocumentReader` and `ForkPDFLayoutTextStripper` — use the project's logging framework instead of writing directly to the console, so this behavior seems inconsistent.

**Possible Improvement**

Use the existing Commons Logging framework with a debug-level guard, consistent with the pattern already used in sibling classes:

    private static final Log logger = LogFactory.getLog(ParagraphManager.class);

    if (logger.isDebugEnabled()) {
        printParagraph(this.rootParagraph, debugStream);
    }

This would make the output opt-in via standard Spring logging configuration instead of always printing to stdout.

**Environment**
- Spring AI version: 1.0.0
- Java version: 17
- OS: Windows 11

pavan542935:
Fixes #6599 

## Problem
`ParagraphManager` constructor calls `printParagraph(this.rootParagraph, System.out)` 
unconditionally at line 67, printing the entire PDF Table of Contents to stdout 
every time a `ParagraphPdfDocumentReader` is created. There is no way to suppress 
this output through logging configuration.

## Fix
Wrapped the `printParagraph` call with a `logger.isDebugEnabled()` guard using 
Commons Logging — consistent with the pattern already used in 
`PagePdfDocumentReader` and `ForkPDFLayoutTextStripper`.

## Changes
- Added `Log` and `LogFactory` imports to `ParagraphManager.java`
- Added `private static final Log logger = LogFactory.getLog(ParagraphManager.class)`
- Wrapped `printParagraph` call with `if (logger.isDebugEnabled())`

### DIFF
--- a/document-readers/spring-ai-pdf-document-reader/src/main/java/org/springframework/ai/reader/pdf/config/ParagraphManager.java
+++ b/document-readers/spring-ai-pdf-document-reader/src/main/java/org/springframework/ai/reader/pdf/config/ParagraphManager.java
@@ -21,6 +21,7 @@ import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.List;
 
+
 import org.apache.pdfbox.pdmodel.PDDocument;
 import org.apache.pdfbox.pdmodel.PDPage;
 import org.apache.pdfbox.pdmodel.PDPageTree;
@@ -31,6 +32,8 @@ import org.jspecify.annotations.Nullable;
 
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 /**
  * The ParagraphManager class is responsible for managing the paragraphs and hierarchy of
@@ -44,6 +47,9 @@ public class ParagraphManager {
 	/**
 	 * Root of the paragraphs tree.
 	 */
+	private static final Log logger = LogFactory.getLog(ParagraphManager.class);
+
+
 	private final Paragraph rootParagraph;
 
 	private final PDDocument document;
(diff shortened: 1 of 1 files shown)


sobychacko:
We're going to close this in favor of #6632, which addresses the same issue but goes a bit further: it routes the output through Commons Log at DEBUG level (rather than keeping the System.out call behind a guard), so the output can actually be controlled via logging configuration, which was the core ask in #6599. It also aligns with the pattern
already used in `PagePdfDocumentReader` and `ForkPDFLayoutTextStripper`, and adds a regression test.

Appreciate the contribution and hope to see more from you. Thanks!

PR #6605, opened 2026-07-13
Fix AnthropicChatOptions.timeout(Duration) being ignored outside model construction

### LINKED ISSUE #6604: AnthropicChatOptions.timeout(Duration) set via ChatClient (default or per-request) is silently ignored — only construction-time timeout applies

**Bug description**

The official docs show this as valid, working usage:

```java
var chatOptions = AnthropicChatOptions.builder()
    .model("claude-sonnet-4-20250514")
    .maxTokens(1024)
    .timeout(Duration.ofSeconds(120))
    .build();
```

**We followed this exact documented pattern** — set `.timeout(Duration.ofSeconds(N))` via `ChatClient.Builder.defaultOptions()` and via per-request `ChatClient.prompt().options(...)` — and it did not work. The configured `Duration` is silently accepted (no error, no warning) but has **zero effect** on the actual HTTP timeout used for the call.

The *only* place a `timeout` Duration actually takes effect is `AnthropicChatModel` **construction time** — i.e. the `spring.ai.anthropic.timeout` property, consumed once by `AnthropicChatAutoConfiguration` when it builds the shared `AnthropicChatModel` bean. Nothing set afterwards — not `ChatClient.Builder.defaultOptions()`, not a per-request `ChatClient.prompt().options(...)` override — has any effect, contrary to what the docs show.

**Practical impact**: because the Anthropic SDK retries transient failures up to `maxRetries` (default 2, so 3 total attempts), an unconfigured/short construction-time timeout compounds silently: with the SDK's `DEFAULT_TIMEOUT` (60s) and no `spring.ai.anthropic.timeout` set, any call that legitimately needs more than 60s (e.g. `web_search` tool use, large `maxTokens` generations) reliably fails after ~180s (3 × 60s) — no matter what `AnthropicChatOptions.timeout(...)` is set via `ChatClient`, following the documented pattern exactly.

**Environment**

- Spring AI version: 2.0.0 (confirmed still present on `main` as of this report)
- Module: `spring-ai-anthropic`, `spring-ai-autoconfigure-model-anthropic`
- Java 21

**Steps to reproduce**

1. Start a local HTTP server that responds to `POST /v1/messages` after a fixed delay (e.g. 2s), returning a minimal valid `Message` JSON body.
2. Build an `AnthropicChatModel` with a **short** construction-time timeout (e.g. 500ms) — deliberately too short for the delayed response above.
3. Call it through `ChatClient` with a **longer** per-request timeout (e.g. 5s), exactly as documented — this should legitimately allow the 2s-delayed mock response to succeed.

See **Minimal Complete Reproducible example** below for the exact code.

**Expected behavior**

Per the documented example, the per-request `.timeout(Duration.ofSeconds(5))` should apply, and the call should succeed after ~2s (mock server delay < 5s per-request timeout).

**Actual behavior**

The call fails at ~500ms × 3 attempts — the construction-time 500ms timeout is used for every attempt; the per-request 5s value is completely ignored. Confirmed via an `AnthropicHttpClientBuilderCustomizer` interceptor logging the live `chain.readTimeoutMillis()`/`chain.call().timeout()` at call time: it consistently reports the construction-time value (500ms), never the per-request value (5000ms), across all 3 retry attempts.

```java
try (MockWebServer server = new MockWebServer()) {
    server.enqueue(new MockResponse().setHeadersDelay(2, TimeUnit.SECONDS)
        .setResponseCode(200)
        .setBody("""
            {"id":"msg-test","type":"message","role":"assistant","content":[{"type":"text","text":"ok"}],"model":"claude-haiku-4-5-20251001","stop_reason":"end_turn","usage":{"input_tokens":1,"output_tokens":1}}
            """));
    server.start();

    AnthropicChatModel chatModel = AnthropicChatModel.builder()
            .apiKey("test-key")
            .model("claude-haiku-4-5-20251001")
            .maxTokens(10)
            .timeout(Duration.ofMillis(500)) // short construction-time timeout
            .build())
        .build();
    ChatClient chatClient = ChatClient.builder(chatModel).build();

    // Per the docs, this 5s per-request timeout should let the 2s-delayed call
    // succeed. Instead, it's ignored and the call fails at ~500ms.
    String content = chatClient.prompt()
        .user("hello")
        .options(AnthropicChatOptions.builder().timeout(Duration.ofSeconds(5)))
        .call()
        .content();
}
```

**Root cause (traced via bytecode + source)**

`AnthropicChatAutoConfiguration.anthropicChatModel(...)` only threads `AnthropicConnectionProperties.timeout` (the `spring.ai.anthropic.timeout` property) into the `AnthropicChatOptions` used to build the shared `AnthropicChatModel` bean, once, at startup:

```java
if (connectionProperties.getTimeout() != null) {
    builder.timeout(connectionProperties.getTimeout());
}
...
var chatModel = AnthropicChatModel.builder().options(options)...build();
```

But `AnthropicChatModel.internalCall()`/`internalStream()` (the per-call path used by every `ChatClient` request) never reads the merged runtime `AnthropicChatOptions.getTimeout()` to build a per-call `com.anthropic.core.RequestOptions` — it always calls `messageService.withRawResponse().create(request)` / `createStreaming(request)` (no explicit `RequestOptions` argument), so `RequestOptions.none()` is used, and the actual HTTP client behavior is whatever was baked into the shared `AnthropicClient` at construction time.

**Fix**

We have a working fix + regression tests: #6605.

Related: #512 (general HTTP client configuration umbrella issue) and #4970 (open PR adding `connect-timeout`/`read-timeout` config *properties*, still construction-time-only) don't cover this specific case.

seeun0210:

* [x] Add a Signed-off-by line to each commit (`git commit -s`) per the [DCO](https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring#how-to-use-developer-certificate-of-origin)
* [x] Rebase your changes on the latest `main` branch and squash your commits
* [x] Add/Update unit tests as needed
* [x] Run a build and make sure all tests pass prior to submission

Thank you upfront!

---

## Summary

Fixes #6604.

`AnthropicChatModel.internalCall()`/`internalStream()` never forwarded the effective `AnthropicChatOptions.getTimeout()` (whether set via `ChatClient.Builder.defaultOptions()` or a per-request `ChatClient.prompt().options(...)` override) into a `com.anthropic.core.RequestOptions` for the underlying Anthropic Java SDK call. Both call sites always used the SDK's default `RequestOptions`, so the only timeout that ever had any effect was the one baked into the shared `AnthropicChatModel`/`AnthropicClient` once, at construction time (via `spring.ai.anthropic.timeout`). This contradicts the documented usage, which shows `.timeout(Duration)` as a per-call `AnthropicChatOptions` setting.

**Why this gap existed**: The shared `AnthropicClient`/`AnthropicChatModel` bean is (correctly) built once at startup, so the timeout baked into its HTTP client is fixed at that point. When per-call `AnthropicChatOptions.timeout(Duration)` was documented as an override, the wiring to actually forward it into the SDK's per-call `RequestOptions` (which the SDK already supports without rebuilding the client) was never added — the two call sites just kept using the SDK's single-arg `create()`/`createStreaming()` overloads.

**Why per-call overrides above 60s matter**: Some calls legitimately need longer than the SDK's 60s default:
- `web_search` tool use, where the model can issue multiple search round-trips (query → fetch → read → continue generating) within a single API call.
- Generations with a large `maxTokens`, since output tokens are produced sequentially, so wall-clock time scales with output length.
- Claude Skills (`AnthropicSkillContainer`) invoking code execution to generate/edit complex `xlsx`/`pptx`/`docx`/`pdf` documents.

Without this fix, an app has no way to grant just those calls more time — raising `spring.ai.anthropic.timeout` globally is the only lever, and even then, an unconfigured/short construction-time timeout compounds silently across the SDK's automatic retries (default `maxRetries: 2` → 3 attempts), so a call can fail after ~3× the configured timeout with no indication why.

- Adds a `requestOptionsFor(Prompt)` helper that resolves the merged (default + per-request) `AnthropicChatOptions.getTimeout()` into an explicit `RequestOptions`, passed to both `messageService.create(params, requestOptions)` and `createStreaming(params, requestOptions)`.
- Updates `AnthropicChatModelTests`' Mockito stubs to match the new two-arg SDK call signature (mechanical change, no behavior change to the tests themselves).
- Adds `AnthropicChatModelTimeoutTests` with 3 cases: construction-time timeout bounds a call; a per-request `ChatClient` timeout overrides a too-short construction-time default; a `ChatClient` default-options timeout overrides a too-short construction-time default. All 3 fail without the fix (verified) and pass with it.

## Test plan

- [x] `./mvnw -pl models/spring-ai-anthropic test` — full module suite (123 tests) passes
- [x] New `AnthropicChatModelTimeoutTests` fail without the fix (confirmed via `git stash` on the production change) and pass with it
- [x] Verified no real Anthropic API calls are needed — regression tests use `MockWebServer` with `setHeadersDelay(...)` for deterministic timing

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -16,6 +16,7 @@
 
 package org.springframework.ai.anthropic;
 
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
@@ -29,6 +30,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import com.anthropic.client.AnthropicClient;
 import com.anthropic.client.AnthropicClientAsync;
 import com.anthropic.core.JsonValue;
+import com.anthropic.core.RequestOptions;
 import com.anthropic.core.http.HttpResponseFor;
 import com.anthropic.core.http.StreamResponse;
 import com.anthropic.models.messages.Base64ImageSource;
@@ -330,7 +332,9 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 			// stream start) can be captured. The SDK exposes this as a blocking
 			// StreamResponse, so events are pulled on a boundedElastic worker.
 			Flux<ChatResponse> chatResponseFlux = Mono
-				.fromFuture(() -> this.anthropicClientAsync.messages().withRawResponse().createStreaming(request))
+				.fromFuture(() -> this.anthropicClientAsync.messages()
+					.withRawResponse()
+					.createStreaming(request, requestOptionsFor(prompt)))
 				.flatMapMany(rawResponse -> {
 					streamingState.setRateLimit(AnthropicRateLimit.from(rawResponse.headers()));
 					StreamResponse<RawMessageStreamEvent> streamResponse = rawResponse.parse();
@@ -552,7 +556,7 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 
 				HttpResponseFor<Message> rawResponse = this.anthropicClient.messages()
 					.withRawResponse()
-					.create(request);
+					.create(request, requestOptionsFor(prompt));
 				Message message = rawResponse.parse();
 				RateLimit rateLimit = AnthropicRateLimit.from(rawResponse.headers());
 
@@ -587,6 +591,40 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 	}
 
 	/**
+	 * Resolves the effective {@link AnthropicChatOptions} for a prompt — the merged
+	 * default/per-request options set on {@link Prompt#getOptions()}, or an empty
+	 * {@link AnthropicChatOptions} if the prompt carries a different {@link ChatOptions}
+	 * implementation.
+	 * @param prompt the prompt with message history and options
+	 * @return the effective Anthropic chat options
+	 */
+	private static AnthropicChatOptions resolveAnthropicOptions(Prompt prompt) {
+		ChatOptions options = prompt.getOptions();
+		return options instanceof AnthropicChatOptions anthropicOptions ? anthropicOptions
+				: AnthropicChatOptions.builder().build();
+	}
+
+	/**
+	 * Resolves the per-call {@link RequestOptions} (currently just {@code timeout} — the
+	 * SDK's {@code RequestOptions} doesn't expose {@code maxRetries}/{@code proxy}/custom
+	 * headers, so those remain construction-time-only settings) from the effective
+	 * {@link AnthropicChatOptions} on the given prompt. Without this, a {@code timeout}
+	 * set via {@link AnthropicChatOptions#getTimeout()} — whether on {@code ChatClient}
+	 * default options or a per-request override — would silently have no effect: the
+	 * Anthropic Java SDK only honors the {@code timeout} supplied on this
+	 * {@link RequestOptions} argument; a call made without one falls back to whatever
+	 * {@link com.anthropic.core.Timeout} the client was built with once, at construction
+	 * time (see {@link AnthropicSetup}).
+	 * @param prompt the prompt with message history and options
+	 * @return request options carrying the resolved timeout, or
+	 * {@link RequestOptions#none()} if none is set
+	 */
+	private static RequestOptions requestOptionsFor(Prompt prompt) {
+		Duration timeout = resolveAnthropicOptions(prompt).getTimeout();
+		return timeout != null ? RequestOptions.builder().timeout(timeout).build() : RequestOptions.none();
+	}
+
+	/**
 	 * Creates a {@link MessageCreateParams} request from a Spring AI {@link Prompt}. Maps
 	 * message types to Anthropic format: TOOL messages become user messages with
 	 * {@link ToolResultBlockParam}, and ASSISTANT messages with tool calls become
@@ -599,9 +637,7 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 
 		MessageCreateParams.Builder builder = MessageCreateParams.builder();
 
-		ChatOptions options = prompt.getOptions();
-		AnthropicChatOptions requestOptions = options instanceof AnthropicChatOptions anthropicOptions
-				? anthropicOptions : AnthropicChatOptions.builder().build();
+		AnthropicChatOptions requestOptions = resolveAnthropicOptions(prompt);
 
 		// Set required fields
 		builder.model(requestOptions.getModel()).maxTokens(requestOptions.getMaxTokens());
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -146,6 +146,7 @@ import org.springframework.util.MimeType;
  * @author Sebastien Deleuze
  * @author Ilayaperumal Gopinathan
  * @author Jewoo Shin
+ * @author Seeun Kim
  * @since 1.0.0
  * @see AnthropicChatOptions
  * @see <a href="https://docs.anthropic.com/en/api/messages">Anthropic Messages API</a>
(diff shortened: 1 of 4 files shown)


seeun0210:
While working on this, I noticed the same pattern likely exists in other provider modules — e.g. `OpenAiChatModel.internalCall()`/`internalStream()` also call `this.openAiClient.chat().completions().create(request)` / `.createStreaming(request)` with no `com.openai.core.RequestOptions` argument, even though the OpenAI Java SDK exposes the same `create(params, RequestOptions)` two-arg overload that this fix uses for Anthropic. So `OpenAiChatOptions.getTimeout()` set via `ChatClient` (default or per-request) is probably silently ignored there too, for the same reason.

If this fix direction looks right to maintainers, I'm happy to apply the same pattern to OpenAI (and check other provider modules for the same issue) in follow-up PRs.

seeun0210:
Hi @sobychacko — thanks for triaging #6604 and assigning it to the 2.0.1 milestone!

Just wanted to flag that this PR already has a working fix + regression tests ready for review whenever you get a chance. Happy to adjust the approach if you'd prefer a different direction (e.g. bundling it with other providers as mentioned in my follow-up comment on this PR). No rush — just wanted to make sure it's on your radar alongside the issue.

Thanks!

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -586,6 +590,40 @@ public ChatResponse internalCall(Prompt prompt, @Nullable ChatResponse previousC
...
+	 * default options or a per-request override — would silently have no effect: the
+	 * Anthropic Java SDK only honors the {@code timeout} supplied on this
+	 * {@link RequestOptions} argument; a call made without one falls back to whatever
+	 * {@link com.anthropic.core.Timeout} the client was built with once, at construction
+	 * time (see {@link AnthropicSetup}).
+	 * @param prompt the prompt with message history and options
+	 * @return request options carrying the resolved timeout, or
+	 * {@link RequestOptions#none()} if none is set
+	 */
+	private static RequestOptions requestOptionsFor(Prompt prompt) {
```
sobychacko, line 602: Too much javadoc on a private method. Can you cut it down to what is absolutely necessary? We usually prefer this much detail only on the public API's.

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -586,6 +590,40 @@ public ChatResponse internalCall(Prompt prompt, @Nullable ChatResponse previousC
...
 
+	/**
+	 * Resolves the effective {@link AnthropicChatOptions} for a prompt — the merged
+	 * default/per-request options set on {@link Prompt#getOptions()}, or an empty
+	 * {@link AnthropicChatOptions} if the prompt carries a different {@link ChatOptions}
+	 * implementation.
+	 * @param prompt the prompt with message history and options
+	 * @return the effective Anthropic chat options
+	 */
+	private static AnthropicChatOptions resolveAnthropicOptions(Prompt prompt) {
```
sobychacko, line 594: In-line these comments instead of javadoc. This level of comment details are not required for private methods.

models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/AnthropicChatModelTimeoutTests.java
```diff
@@ -0,0 +1,209 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Verifies that {@link AnthropicChatOptions#getTimeout()} actually bounds the HTTP call
+ * whether it's set at {@link AnthropicChatModel} construction time, as a
+ * {@link ChatClient} default option, or as a per-request {@code ChatClient} override —
+ * not just at construction time.
+ *
+ * @author seeun0210
```
sobychacko: Use your full real name.

sobychacko (changes requested):
Looks good. I added some minor nit-picks. You can also add your name (full real name) as an author to all the classes you modified in the PR.

seeun0210:
Thanks for the review @sobychacko — addressed all three points: trimmed the private-method Javadoc on `resolveAnthropicOptions`/`requestOptionsFor` to brief inline comments, and added my full name as an author to the classes touched in this PR. PTAL 🙏

On the follow-up I raised earlier: I went and confirmed the **same pattern in `OpenAiChatModel`** — `internalCall()`/`internalStream()` call `.create(request)`/`.createStreaming(request)` with no `com.openai.core.RequestOptions`, while `OpenAiChatOptions` exposes `getTimeout()`. So a timeout set via `ChatClient` (default or per-request) is silently ignored there too, identical to #6604. I also swept the other model modules and OpenAI is the only other one on this SDK two-arg `create` pattern.

Happy to fix it — would you prefer it **bundled into this PR** or as a **separate follow-up PR** (with its own issue)? Whatever is easier to review on your end.

seeun0210:
Quick follow-up on my question above: I see #6648 has already been opened to propagate the timeout for the OpenAI models, so no need for a separate OpenAI PR from me — this one can stay scoped to Anthropic. 👍 I left a note over there linking the two since they share the same root cause.

PR #6615, opened 2026-07-14
Reduce `RedisVectorStore` search payloads by omitting unused embeddings

ultramancode:
`RedisVectorStore` currently asks Redis to include each matched document's stored embedding in the response for similarity, text, and range searches. 

Vector searches still need the stored embeddings inside Redis to find matching documents, but Spring AI does not use the returned copy when converting any result to a `Document`.

As a result, Redis sends the full vector for every match even though Spring AI never uses it. 

In a local Redis Stack payload check, a single 3,072-dimensional embedding added about 37 KB to the returned data. 
This overhead grows with the number of results requested through `topK`.

This change removes only the unused embedding from search responses. 
Stored embeddings are still used for indexing and vector search, and the returned `Document` results are unchanged. 

Tests verify the generated `RETURN` fields and cover similarity, text, and range search paths, including a custom embedding field name.


ilayaperumalg (approved):

ilayaperumalg:
@ultramancode Thanks for the PR!

ultramancode:
Thanks for taking the time to review this, @ilayaperumalg — really appreciate it!

PR #6623, opened 2026-07-15
Omit exception messages from error response bodies in the MCP server transports

dimitarproynov:
Tool failure messages are preserved by the handling in SyncMcpToolMethodCallback.apply/AbstractSyncMcpToolMethodCallback.createSyncErrorResult

This change stops exposing exception messages for failures outside the tool handler pipeline (session initialization/deletion, response serialization, transport plumbing)

### DIFF
--- a/mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/server/webflux/transport/WebFluxSseServerTransportProvider.java
+++ b/mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/server/webflux/transport/WebFluxSseServerTransportProvider.java
@@ -453,7 +453,7 @@ public final class WebFluxSseServerTransportProvider implements McpServerTranspo
 					// return ServerResponse.ok().build();
 					return ServerResponse.status(HttpStatus.INTERNAL_SERVER_ERROR)
 						.bodyValue(McpError.builder(McpSchema.ErrorCodes.INTERNAL_ERROR)
-							.message(error.getMessage())
+							.message("Internal server error. Check server logs for details.")
 							.build());
 				});
 			}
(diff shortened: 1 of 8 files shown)


mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/server/webflux/transport/WebFluxSseServerTransportProvider.java
ilayaperumalg, line 343: Though not in the scope of this PR, I wonder if this part of the error message handling should be taken care as well?

ilayaperumalg: Also, I see `WebFluxStreamableServerTransportProvider` requires similar changes when catching `ServerTransportSecurityException`

dimitarproynov: I'm not so certain here the ServerTransportSecurityException is thrown by a user supplied implementation of ServerTransportSecurityValidator, which throws the above exception – if validation fails. This is user controlled and I assume it is something that the user would want to transfer to the other side of the wire.

I assume the user won't log any sensitive details, if they do that is an application error.

ilayaperumalg: yeah, I was thinking from the security point of view but I agree that this is user controlled path and sometimes, it could be intentional to pass the message as is.

ilayaperumalg (approved):

PR #6638, opened 2026-07-18
Support `thinking` and `reasoningEffort` control for `DeepSeekChatModel`

guanxuc:


guanxuc:
Hi @sdeleuze , @dimitarproynov 

Could you please take a look this, thanks!

models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/api/DeepSeekApi.java
```diff
@@ -577,6 +590,274 @@ public static Object FUNCTION(String functionName) {
...
+				this.thinking = thinking;
+				return this;
+			}
+
+			/**
+			 * Build the {@link ChatCompletionRequest}.
+			 * @return a new {@link ChatCompletionRequest}.
+			 */
+			public ChatCompletionRequest build() {
+				Assert.notNull(this.messages, "Messages must not be null");
```
dimitarproynov, line 851: Since this.messages defaults to a empty ArrayList, this assertion will fire only if the user purposefully set messages to null. Also since the default is an empty ArrayList, a user may forget to set messages and send a request. 

I'd make the default null, so users **have** to set messages. 

Not a blocker as the only user DeepSeekChatModel.createRequest, always sets messages, but this builder is public API surface.

models/spring-ai-deepseek/src/main/java/org/springframework/ai/deepseek/api/DeepSeekApi.java
```diff
@@ -577,6 +590,274 @@ public static Object FUNCTION(String functionName) {
...
+
+			/**
+			 * Set the thinking mode configuration.
+			 * @param thinking the thinking configuration.
+			 * @return this builder.
+			 */
+			public Builder thinking(@Nullable Thinking thinking) {
+				this.thinking = thinking;
+				return this;
+			}
```
dimitarproynov, line 844: DeepSeek's docs state "Thinking mode does not support the temperature, top_p, presence_penalty, or frequency_penalty parameters." 

There is no validation or warning when a caller sets thinking(ENABLED) alongside those parameters, they'll be silently sent together and is not clear what is going to happen.

Please, implement some kind of validation for this.

models/spring-ai-deepseek/src/test/java/org/springframework/ai/deepseek/chat/DeepSeekChatModelIT.java
dimitarproynov, line 62: Has this test been ran? Neither the PR description, nor the commit message mention any testing.

guanxuc: Yes, I've added integration tests (`thinkingEnabledTest`, `thinkingDisabledTest`, `reasoningEffortMaxTest`, and `reasoningEffortHighTest`) to verify the `thinking` and `reasoningEffort` behavior. All tests pass except `DeepSeekChatModelIT#prefixCompletionTest`, which is a pre-existing issue unrelated to this change. I've opened a separate PR (#6703) to address it.
image

guanxuc:
Hi @dimitarproynov ,
Thanks for your review and good suggestions! I have addressed your comments, please take another look.

PR #6639, opened 2026-07-18
Support non-String URI variable parameters for @McpResource methods

### LINKED ISSUE #6617: [Feature request] Support non-String params for resources
Consider this McpResource.

```java
@McpResource(uri = "example.com/users/{id}")
public ResourceContents users(int id) {
    // return something
}
```

It will throw an exception at startup

```
Caused by: java.lang.IllegalArgumentException: URI variable parameters must be of type String: users in com.example.MyMcpTest, parameter of type int is not valid
```

This is bad for 2 reasons:

- It goes against the existing Spring patterns that people are used to, such as in RestControllers
- It creates boilerplate to have to parse parameters manually

Example boilerplate

```java
@McpResource(uri = "example.com/users/{id}")
public ResourceContents users(String id) {
    try {
        var idAsInt = Integer.parseInt(id);
        // now do the lookup
    }
    catch (NumberFormatException e) {
        throw McpError.builder(McpSchema.ErrorCodes.INVALID_PARAMS)
            .message("ID must be an integer")
            .build();
    }
}
```

## Suggested change

I suggest that params use existing Converter beans to automatically convert from a String into the method's declared type. If this is not possible (e.g. due to NumberFormatException), then map into an McpError automatically.

OneZero-Y:
Fixes #6617

## Summary

`@McpResource` methods with URI template variables currently require those parameters to be `String`. Declaring a primitive or wrapper type (e.g. `int id`) fails at startup with:

> URI variable parameters must be of type String

This forces users to manually parse URI variables, which is inconsistent with Spring MVC-style path variable handling.

This PR removes the `String`-only restriction and converts URI variable values to the declared parameter type using Spring's `DefaultConversionService`. Invalid values are surfaced as `McpError(INVALID_PARAMS)` via the existing resource callback error handling.

## Changes

* Add `McpUriVariableValueConverter` to convert URI variable strings to the declared parameter type
* Remove the `String`-only validation in `AbstractMcpResourceMethodCallback`
* Convert URI variable values when building method arguments in `buildArgsWithUriVariables`
* Add unit tests for successful `int` conversion and invalid value handling

## Usage

Before (manual parsing required):

```java
@McpResource(uri = "example.com/users/{id}")
public ResourceContents users(String id) {
    int userId = Integer.parseInt(id);
    // ...
}
```
After:

```java
@McpResource(uri = "example.com/users/{id}")
public ResourceContents users(int id) {
    // id is already converted
}
```
##  Test plan
- [x]  ./mvnw -pl mcp/mcp-annotations test -Dtest=SyncMcpResourceMethodCallbackTests#testCallbackWithIntegerUriVariable,SyncMcpResourceMethodCallbackTests#testCallbackWithInvalidIntegerUriVariable,SyncMcpResourceMethodCallbackTests#testCallbackWithUriVariables -Dsurefire.failIfNoSpecifiedTests=false
- [x]   @McpResource with int URI variable parameter registers successfully at startup
- [x]   Valid URI value (42) is converted and passed to the handler method
- [x]  Invalid URI value (not-a-number) returns McpError
- [x]  Existing String URI variable behavior unchanged

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/resource/AbstractMcpResourceMethodCallback.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/resource/AbstractMcpResourceMethodCallback.java
@@ -368,23 +368,6 @@ public abstract class AbstractMcpResourceMethodCallback {
 							+ " in " + method.getDeclaringClass().getName() + ". URI variables: " + this.uriVariables);
 		}
 
-		// Check that all non-special parameters are String type (for URI variables)
-		for (Parameter param : parameters) {
-			// Skip @McpProgressToken annotated parameters
-			if (param.isAnnotationPresent(McpProgressToken.class)) {
-				continue;
-			}
-
-			Class<?> paramType = param.getType();
-			if (!McpSyncRequestContext.class.isAssignableFrom(paramType)
-					&& !McpAsyncRequestContext.class.isAssignableFrom(paramType) && !isExchangeOrContextType(paramType)
-					&& !ReadResourceRequest.class.isAssignableFrom(paramType)
-					&& !McpMeta.class.isAssignableFrom(paramType) && !String.class.isAssignableFrom(paramType)) {
-				throw new IllegalArgumentException("URI variable parameters must be of type String: " + method.getName()
-						+ " in " + method.getDeclaringClass().getName() + ", parameter of type " + paramType.getName()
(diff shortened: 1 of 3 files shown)


OneZero-Y:
Hey @dimitarproynov — would you mind taking a look when you get a chance? This lets @McpResource URI variables bind to non-String parameter types (e.g. int) via DefaultConversionService, instead of failing at startup. Invalid values surface as McpError(INVALID_PARAMS).

PR #6642, opened 2026-07-18
Add structured transcription results

### LINKED ISSUE #6640: TranscriptionModel returns plain text for all response formats, discarding structured data (diarized, verbose, JSON)
**Expected Behavior**

When calling `TranscriptionModel.call()` with `AudioTranscriptionOptions` configured to use a structured response format, such as `DIARIZED_JSON` or `VERBOSE_JSON`, the resulting `AudioTranscriptionResponse` should preserve the structured transcription data returned by the provider instead of exposing only the transcript text.

Spring AI already follows a similar pattern in the chat domain: `Generation.getOutput()` returns an `AssistantMessage` rather than a raw `String`, allowing callers to access text, tool calls, media, and additional properties.

The transcription abstraction should likewise provide access to semantic transcription data such as:

* transcript text;
* language;
* duration;
* timestamped segments;
* speaker annotations;
* timestamped words.

Possible API designs consistent with existing Spring AI patterns include the following.

### Approach A: Enrich `AudioTranscriptionMetadata`

Keep `AudioTranscription` as `ModelResult<String>` and expose structured data through `AudioTranscriptionMetadata`.

This would preserve the current output contract while extending the metadata abstraction:

```java
public interface AudioTranscriptionMetadata extends ResultMetadata {

    @Nullable
    Duration getDuration();

    @Nullable
    String getLanguage();

    List<AudioTranscriptionSegment> getSegments();

    List<AudioTranscriptionWord> getWords();

    <T> @Nullable T get(String key);
}
```

Provider-neutral transcription records could be introduced:

```java
public record AudioTranscriptionSegment(
    @Nullable String id,
    @Nullable String speaker,
    Duration start,
    Duration end,
    String text
) {
}
```

```java
public record AudioTranscriptionWord(
    String text,
    Duration start,
    Duration end
) {
}
```

The OpenAI implementation could then map the structured response into the metadata associated with the result:

```java
var metadata = AudioTranscriptionMetadata.builder()
    .duration(response.asDiarized().duration())
    .segments(mapSegments(response.asDiarized().segments()))
    .build();

return new AudioTranscriptionResponse(transcript, metadata);
```

### Approach B: Introduce a structured output type

Keep `AudioTranscription` as the `ModelResult` implementation, but replace its raw `String` output with a provider-neutral transcription value:

```java
public final class AudioTranscription
        implements ModelResult<AudioTranscriptionResult> {

    private final AudioTranscriptionResult output;

    private final AudioTranscriptionMetadata metadata;

    // ...
}
```

```java
public record AudioTranscriptionResult(
    String text,
    @Nullable String language,
    @Nullable Duration duration,
    List<AudioTranscriptionSegment> segments,
    List<AudioTranscriptionWord> words
) {
}
```

The responsibilities would then be separated as follows:

* `AudioTranscriptionResponse`
  * `AudioTranscription`
    * `output: AudioTranscriptionResult`
      * `text`
      * `language`
      * `duration`
      * `segments`
      * `words`
    * `metadata`
      * technical or provider-specific result metadata

This would follow the same general model used by `Generation`, whose output is an `AssistantMessage` containing the semantic result rather than only its textual representation.

### Approach C: Enrich `AudioTranscription` directly

Keep `AudioTranscription` as `ModelResult<String>` so that `getOutput()` continues to return the transcript text, while adding structured transcription fields directly to the result:

```java
public final class AudioTranscription implements ModelResult<String> {

    private final String text;

    private final @Nullable String language;

    private final @Nullable Duration duration;

    private final List<AudioTranscriptionSegment> segments;

    private final List<AudioTranscriptionWord> words;

    private final AudioTranscriptionMetadata metadata;

    @Override
    public String getOutput() {
        return this.text;
    }

    public @Nullable String getLanguage() {
        return this.language;
    }

    public @Nullable Duration getDuration() {
        return this.duration;
    }

    public List<AudioTranscriptionSegment> getSegments() {
        return this.segments;
    }

    public List<AudioTranscriptionWord> getWords() {
        return this.words;
    }

    // ...
}
```

**Current Behavior**

`OpenAiAudioTranscriptionModel.call()` extracts only the textual content from every `TranscriptionCreateResponse` and wraps that text in an `AudioTranscription`:

```java
TranscriptionCreateResponse response =
        this.openAiClient.audio().transcriptions().create(params);

String text = extractText(response);
AudioTranscription transcript = new AudioTranscription(text);

return new AudioTranscriptionResponse(
        transcript,
        new AudioTranscriptionResponseMetadata()
);
```

As a result, structured fields returned by the provider are not represented by the Spring AI abstraction.

For `DIARIZED_JSON`, this includes data such as:

* duration;
* segments;
* speaker annotations;
* segment timestamps.

For `VERBOSE_JSON`, this includes data such as:

* language;
* duration;
* segments;
* words;
* timestamps.

`AudioTranscriptionMetadata` currently cannot carry this information because it is only a marker interface:

```java
public interface AudioTranscriptionMetadata extends ResultMetadata {

    AudioTranscriptionMetadata NULL = AudioTranscriptionMetadata.create();

    static AudioTranscriptionMetadata create() {
        return new AudioTranscriptionMetadata() {};
    }
}
```

This differs from abstractions such as `ChatGenerationMetadata`, which expose defined metadata fields and support additional key-value data.

Consequently, there is currently no provider-neutral way to access structured transcription results through Spring AI.

**Context**

We are building an application that performs diarized audio transcription. We initially planned to use Spring AI 2.0.0, but the current text-only result prevents us from accessing speaker annotations and segment timestamps through the Spring AI model abstraction.

The alternatives considered were:

* **Calling the provider SDK directly:** this exposes the complete response, but bypasses the Spring AI abstraction.
* **Parsing the raw response manually:** this is fragile, provider-specific, and depends on serialization or deserialization details that are not part of the Spring AI contract.
* **Extending `OpenAiAudioTranscriptionModel`:** this is not currently possible because the class is `final`.

The chat abstraction already preserves structured semantic output through `AssistantMessage`, including text, tool calls, media, and properties. Structured audio transcription formats would benefit from an equivalent provider-neutral representation for text, language, timestamps, segments, words, and speaker information.

wskr00:
## Summary

- Add `AudioTranscriptionResult`, `AudioTranscriptionSegment`, and `AudioTranscriptionWord` to expose structured transcription data.
- Preserve language, duration, segments, words, and speaker annotations from OpenAI verbose and diarized responses.
- Preserve diarized segment data for streaming responses.
- Update the transcription reference documentation.
- Close audio resource input streams and simplify redundant option checks.

## Breaking Change

`AudioTranscription.getOutput()` now returns `AudioTranscriptionResult` instead of `String`.

Use `.text()` when only the transcription text is needed:

```java
String text = response.getResult().getOutput().text();
```

## Testing

- `./mvnw process-sources`
- `./mvnw javadoc:javadoc`
- `./mvnw clean package`
- `./mvnw -pl spring-ai-docs antora`

Closes #6640

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-openai/src/test/java/org/springframework/ai/model/openai/autoconfigure/OpenAiAudioTranscriptionAutoConfigurationIT.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-openai/src/test/java/org/springframework/ai/model/openai/autoconfigure/OpenAiAudioTranscriptionAutoConfigurationIT.java
@@ -51,7 +51,7 @@ public class OpenAiAudioTranscriptionAutoConfigurationIT {
 				AudioTranscriptionResponse response = transcriptionModel
 					.call(new AudioTranscriptionPrompt(new ClassPathResource("/speech.flac")));
 				assertThat(response.getResults()).hasSize(1);
-				assertThat(response.getResult().getOutput()).isNotBlank();
+				assertThat(response.getResult().getOutput().text()).isNotBlank();
 			});
 	}
 
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
@@ -174,7 +174,7 @@ public final class OpenAiAudioTranscriptionModel implements TranscriptionModel {
 				}
 			}));
 
-		return chunks.map(event -> {
+		return chunks.filter(event -> !event.isTranscriptTextDone()).map(event -> {
 			AudioTranscription transcript = new AudioTranscription(toTranscriptionResult(event));
 			return new AudioTranscriptionResponse(transcript, new AudioTranscriptionResponseMetadata());
 		});
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioTranscriptionModel.java
@@ -19,6 +19,7 @@ package org.springframework.ai.openai;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
@@ -28,7 +29,13 @@ import com.openai.client.OpenAIClientAsync;
 import com.openai.core.MultipartField;
 import com.openai.models.audio.transcriptions.TranscriptionCreateParams;
 import com.openai.models.audio.transcriptions.TranscriptionCreateResponse;
+import com.openai.models.audio.transcriptions.TranscriptionDiarized;
+import com.openai.models.audio.transcriptions.TranscriptionDiarizedSegment;
+import com.openai.models.audio.transcriptions.TranscriptionSegment;
 import com.openai.models.audio.transcriptions.TranscriptionStreamEvent;
+import com.openai.models.audio.transcriptions.TranscriptionTextSegmentEvent;
+import com.openai.models.audio.transcriptions.TranscriptionVerbose;
+import com.openai.models.audio.transcriptions.TranscriptionWord;
 import io.micrometer.observation.ObservationRegistry;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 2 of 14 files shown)


wskr00:
Converted this PR to draft because the OpenAI Java SDK currently treats `diarized_json` as a string response and wraps the raw JSON in `Transcription.text`.

I've found a upstream [issue](https://github.com/openai/openai-java/issues/652) and a [PR](https://github.com/openai/openai-java/pull/744). Once the SDK issue is resolved, I will update this PR accordingly.

wskr00:
Implementation in #6697.

PR #6656, opened 2026-07-21
Close OpenAI stream response on cancellation

### LINKED ISSUE #6654: OpenAiChatModel does not close AsyncStreamResponse on Reactor cancellation
**Bug description**

  `OpenAiChatModel.internalStream()` creates an OpenAI Java SDK
  `AsyncStreamResponse<ChatCompletionChunk>` inside `Flux.create()`, subscribes to
  it, and then discards the response handle.

  No `FluxSink.onCancel(...)` or `FluxSink.onDispose(...)` callback closes the
  `AsyncStreamResponse`.

  As a result, cancelling the downstream Reactor subscription stops downstream
  signals, but does not close the OpenAI SDK stream or its underlying HTTP
  response. The SDK continues consuming the SSE response until the provider sends
  EOF, the request times out, or the whole client is closed.

  This can waste network and provider resources after an HTTP client disconnects
  or an application explicitly cancels model generation.

  The relevant code currently has this shape:

  ```java
  Flux<ChatCompletionChunk> chunks = Flux.create(sink ->
          this.openAiClientAsync.chat()
                  .completions()
                  .createStreaming(request)
                  .subscribe(sink::next)
                  .onCompleteFuture()
                  .whenComplete((unused, throwable) -> {
                      if (throwable != null) {
                          sink.error(throwable);
                      }
                      else {
                          sink.complete();
                      }
                  }));
  ```

  The OpenAI Java SDK already exposes `AsyncStreamResponse.close()` for releasing
  the underlying stream, but `OpenAiChatModel` no longer retains the response
  reference after subscription.

  **Environment**

  - Spring AI: `2.0.1-SNAPSHOT`
  - Resolved snapshot build: `2.0.1-20260717.095612-74`
  - Spring Boot: `4.1.0`
  - OpenAI Java SDK: `4.41.0`
  - Reactor Core: `3.8.6`
  - OkHttp: `4.12.0`
  - Java: OpenJDK `25.0.2`
  - OS: macOS `26.4`, arm64
  - Vector store: none

  The same `internalStream()` implementation is also present on the current
  Spring AI `main` branch.

  **Steps to reproduce**

  1. Start an OpenAI-compatible SSE endpoint that continuously returns chat
     completion chunks without closing the response.
  2. Create an `OpenAiChatModel` using that endpoint.
  3. Subscribe to `model.stream(prompt)`.
  4. Cancel after the first emitted `ChatResponse`, for example with `take(1)`.
  5. Observe that the server can continue writing SSE chunks because the
     underlying HTTP response was not closed.
  6. The connection only closes after server EOF, timeout, or client shutdown.

  **Expected behavior**

  Cancelling or disposing the downstream Reactor subscription should call
  `AsyncStreamResponse.close()`.

  The underlying HTTP response should be closed promptly, and the provider should
  observe the client disconnect without waiting for EOF or request timeout.

  A possible fix is to retain the SDK response and register it with the Reactor
  sink:

  ```java
  Flux<ChatCompletionChunk> chunks = Flux.create(sink -> {
      var response = this.openAiClientAsync.chat()
              .completions()
              .createStreaming(request);

      sink.onDispose(response::close);

      try {
          response.subscribe(sink::next)
                  .onCompleteFuture()
                  .whenComplete((unused, throwable) -> {
                      if (throwable != null) {
                          sink.error(throwable);
                      }
                      else {
                          sink.complete();
                      }
                  });
      }
      catch (RuntimeException failure) {
          if (!sink.isCancelled()) {
              sink.error(failure);
          }
      }
  });
  ```

  `onDispose` is preferable to only `onCancel`, because it releases the SDK
  resource after cancellation, error, or normal termination. The SDK close
  operation is idempotent.

  ```java
  package example;

  import static org.junit.jupiter.api.Assertions.assertNotNull;
  import static org.junit.jupiter.api.Assertions.assertTrue;

  import com.sun.net.httpserver.HttpServer;
  import java.io.IOException;
  import java.net.InetSocketAddress;
  import java.time.Duration;
  import java.util.concurrent.CountDownLatch;
  import java.util.concurrent.TimeUnit;
  import java.util.concurrent.atomic.AtomicBoolean;
  import org.springframework.ai.openai.OpenAiChatModel;
  import org.springframework.ai.openai.OpenAiChatOptions;

  class OpenAiChatModelCancellationTest {

      @Test
      void reactorCancellationShouldCloseTheSdkStream() throws Exception {
          var serverObservedDisconnect = new CountDownLatch(1);
          var keepWriting = new AtomicBoolean(true);

          var server = HttpServer.create(
                  new InetSocketAddress("127.0.0.1", 0), 0);

          server.createContext("/v1/chat/completions", exchange -> {
              exchange.getResponseHeaders()
                      .add("Content-Type", "text/event-stream");
              exchange.sendResponseHeaders(200, 0);

              try (var body = exchange.getResponseBody()) {
                  while (keepWriting.get()) {
                      var frame = """
                              data: {"id":"chunk-1","object":"chat.completion.chunk","created":1,"model":"test-model","choices":[{"index":0,"delta":{"role":"assistant","content":"chunk"},"finish_reason":null}]}

                              """;

                      body.write(frame.getBytes(StandardCharsets.UTF_8));
                      body.flush();
                      Thread.sleep(25);
                  }
              }
              catch (IOException disconnected) {
                  serverObservedDisconnect.countDown();
              }
              catch (InterruptedException interrupted) {
                  Thread.currentThread().interrupt();
              }
          });

          server.start();

          try {
              var baseUrl = "http://127.0.0.1:"
                      + server.getAddress().getPort()
                      + "/v1";

              var options = OpenAiChatOptions.builder()
                      .baseUrl(baseUrl)
                      .apiKey("test-key")
                      .model("test-model")
                      .maxRetries(0)
                      .timeout(Duration.ofSeconds(30))
                      .build();

              var model = OpenAiChatModel.builder()
                      .options(options)
                      .build();

              // take(1) cancels the upstream Reactor subscription after the
              // first ChatResponse.
              var response = model.stream(new Prompt("hello"))
                      .take(1)
                      .blockLast(Duration.ofSeconds(5));

              assertNotNull(response);

              // This assertion fails with the current implementation. The server
              // continues writing because AsyncStreamResponse.close() was never
              // called.
              assertTrue(
                      serverObservedDisconnect.await(2, TimeUnit.SECONDS),
                      "The HTTP stream remained open after Reactor cancellation"
              );
          }
          finally {
              keepWriting.set(false);
              server.stop(0);
          }
      }
  }
  ```

gaoxiaolei-s59:
Closes #6654

## Summary

- retain the OpenAI SDK asynchronous stream response while bridging it to Reactor
- close the SDK response whenever the Reactor sink is disposed
- add a regression test for downstream cancellation of a stream that has not completed

## Testing

- `OpenAiChatModelTests`: 23 tests passed
- focused cancellation regression test passed
- Spring Java Format and Checkstyle passed for the affected module

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -35,6 +35,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.openai.client.OpenAIClient;
 import com.openai.client.OpenAIClientAsync;
 import com.openai.core.JsonValue;
+import com.openai.core.http.AsyncStreamResponse;
 import com.openai.errors.OpenAIInvalidDataException;
 import com.openai.models.FunctionDefinition;
 import com.openai.models.FunctionParameters;
@@ -290,19 +291,20 @@ public final class OpenAiChatModel implements ChatModel {
 			}
 
 			// Convert from AsyncStreamResponse<ChatCompletionChunk> to Flux<CCC>
-			Flux<ChatCompletionChunk> chunks = Flux.<ChatCompletionChunk>create(sink -> this.openAiClientAsync.chat()
-				.completions()
-				.createStreaming(request)
-				.subscribe(sink::next)
-				.onCompleteFuture()
-				.whenComplete((unused, throwable) -> {
+			Flux<ChatCompletionChunk> chunks = Flux.<ChatCompletionChunk>create(sink -> {
+				AsyncStreamResponse<ChatCompletionChunk> response = this.openAiClientAsync.chat()
+					.completions()
+					.createStreaming(request);
+				sink.onDispose(response::close);
+				response.subscribe(sink::next).onCompleteFuture().whenComplete((unused, throwable) -> {
 					if (throwable != null) {
 						sink.error(throwable);
 					}
 					else {
 						sink.complete();
 					}
-				}));
+				});
+			});
 
 			// Next, aggregate CCCs that deal with tool calls together
 			AtomicBoolean isInsideTool = new AtomicBoolean(false);
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -36,6 +36,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.openai.client.OpenAIClient;
 import com.openai.client.OpenAIClientAsync;
 import com.openai.core.JsonValue;
+import com.openai.core.http.AsyncStreamResponse;
 import com.openai.errors.OpenAIInvalidDataException;
 import com.openai.models.FunctionDefinition;
 import com.openai.models.FunctionParameters;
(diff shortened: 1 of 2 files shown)


tzolov (approved):
Thanks @gaoxiaolei-s59, I made some small improvements on top of your PR

PR #6678, opened 2026-07-23
Preserve cache usage metadata in MessageAggregator

### LINKED ISSUE #6667: MessageAggregator loses cacheReadInputTokens/cacheWriteInputTokens in streaming
## Bug description

PR #5657 (merged in 7cfad4559be4dbb4e4c60afe081cc3d169e631bb) added cache metrics to the `Usage` interface (`getCacheReadInputTokens()` / `getCacheWriteInputTokens()`), and also updated `DefaultUsage` and `UsageCalculator`.

But looks like `MessageAggregator` was forgotten. It aggregates usage with its own internal record that has only 3 fields:

```java
// spring-ai-model/src/main/java/org/springframework/ai/chat/model/MessageAggregator.java (main, ~line 210)
public record DefaultUsage(Integer promptTokens, Integer completionTokens, Integer totalTokens) implements Usage {
```

So after `MessageAggregator.aggregate(...)` the aggregated response always has `getCacheReadInputTokens() == null` and `getCacheWriteInputTokens() == null`, even if every chunk in the stream had cache metrics. Original `nativeUsage` is also lost, it gets replaced by a map with the same 3 fields.

The main problem is observability. Chat models do streaming observation like this:

```java
return new MessageAggregator().aggregate(chatResponseFlux, observationContext::setResponse);
```

and `DefaultChatModelObservationConvention` adds the `gen_ai.usage.cache_read.input_tokens` tag only when the getter is not null. So with sync `call()` the span has this tag, but with `stream()` it is never there for the same request. Per-chunk `Usage` objects inside the `Flux<ChatResponse>` are correct, only the final aggregated one is broken. In result any dashboard for cache hit-rate shows 0% cache usage for streaming calls.

I found this when I was mapping GigaChat `precached_prompt_tokens` to `DefaultUsage.cacheReadInputTokens` in the spring-ai-gigachat adapter (ai-forever/spring-ai-gigachat#120), but it is not related to specific provider, repro below needs only spring-ai-model.

## Environment

Spring AI 2.0.0 (GA), Java 21. The record on current `main` is the same.

## Steps to reproduce

```java
@Test
void aggregatedUsageDropsCacheMetrics() {
    Usage chunkUsage = new DefaultUsage(10, 5, 15, null, 42L, 7L);
    ChatResponse chunk = ChatResponse.builder()
            .generations(List.of(new Generation(new AssistantMessage("hi"))))
            .metadata(ChatResponseMetadata.builder().usage(chunkUsage).build())
            .build();

    AtomicReference<ChatResponse> aggregated = new AtomicReference<>();
    new MessageAggregator().aggregate(Flux.just(chunk), aggregated::set).blockLast();

    Usage usage = aggregated.get().getMetadata().getUsage();
    System.out.println(usage.getClass());
    System.out.println(usage.getPromptTokens() + " / " + usage.getCompletionTokens() + " / " + usage.getTotalTokens());
    System.out.println(usage.getCacheReadInputTokens() + " / " + usage.getCacheWriteInputTokens());
    System.out.println(usage.getNativeUsage());
}
```

Output:

```
class org.springframework.ai.chat.model.MessageAggregator$DefaultUsage
10 / 5 / 15
null / null
{promptTokens=10, totalTokens=15, completionTokens=5}
```

Token counts are ok, but cacheReadInputTokens=42 and cacheWriteInputTokens=7 became null.

## Expected behavior

Aggregated `Usage` keeps cache metrics the same way as prompt/completion/total tokens, so streaming observation reports `gen_ai.usage.cache_read.input_tokens` same as sync.

## Possible fix

Add two `Long` cache fields to the internal record (same logic as for token counts, last chunk with non-empty value wins). Or maybe reuse `org.springframework.ai.chat.metadata.DefaultUsage` instead of the private record, this would also keep the real `nativeUsage`. I can try to prepare a PR if some of these approaches is ok for you.

Maybe related: #3875 (also about lost response data in streaming observation, but that one is about `observation.stop()` ordering).

Will happy to discuss and open PR.

vinkiYu:
## Summary

- preserve cache read and cache write input token counts when aggregating streaming chat responses
- preserve the provider-native usage object in the aggregated response
- retain the last non-null cache/native usage value reported by streaming chunks
- add regression tests for preservation, multiple chunks, and absent cache metrics

## Testing

- `./mvnw.cmd -pl spring-ai-model -Dtest=MessageAggregatorTests test`

Closes [#6667](https://github.com/spring-projects/spring-ai/issues/6667)

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/model/MessageAggregator.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/model/MessageAggregator.java
@@ -25,6 +25,7 @@ import java.util.function.Consumer;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.jspecify.annotations.Nullable;
 import reactor.core.publisher.Flux;
 
 import org.springframework.ai.chat.messages.AssistantMessage;
@@ -70,6 +71,9 @@ public class MessageAggregator {
 		AtomicReference<Integer> metadataUsagePromptTokensRef = new AtomicReference<>(0);
 		AtomicReference<Integer> metadataUsageGenerationTokensRef = new AtomicReference<>(0);
 		AtomicReference<Integer> metadataUsageTotalTokensRef = new AtomicReference<>(0);
+		AtomicReference<@Nullable Long> metadataUsageCacheReadInputTokensRef = new AtomicReference<>();
+		AtomicReference<@Nullable Long> metadataUsageCacheWriteInputTokensRef = new AtomicReference<>();
+		AtomicReference<@Nullable Object> metadataNativeUsageRef = new AtomicReference<>();
 
 		AtomicReference<PromptMetadata> metadataPromptMetadataRef = new AtomicReference<>(PromptMetadata.empty());
 		AtomicReference<RateLimit> metadataRateLimitRef = new AtomicReference<>(new EmptyRateLimit());
@@ -88,6 +92,9 @@ public class MessageAggregator {
 			metadataUsagePromptTokensRef.set(0);
 			metadataUsageGenerationTokensRef.set(0);
 			metadataUsageTotalTokensRef.set(0);
+			metadataUsageCacheReadInputTokensRef.set(null);
+			metadataUsageCacheWriteInputTokensRef.set(null);
+			metadataNativeUsageRef.set(null);
 			metadataPromptMetadataRef.set(PromptMetadata.empty());
 			metadataRateLimitRef.set(new EmptyRateLimit());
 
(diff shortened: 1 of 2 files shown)


CodeTrainerMan:
Good catch on the root cause — this is the right place to fix it.

I had opened #7021 for the same aggregation path; I closed it as a duplicate of this one (and of #7009) to avoid a third competing patch. While working on it I hit two edge cases that this patch and #7009 do not currently cover, so leaving them here in case they are useful:

1. `EmptyUsage.getNativeUsage()` returns `Map.of()`, not `null`:

```java
@Override
public Object getNativeUsage() {
    return Map.of();
}
```

So the `if (usage.getNativeUsage() != null)` guard lets a chunk carrying an *empty* usage overwrite a real native usage recorded by an earlier chunk. In streams where the provider reports usage on an early chunk and the remaining chunks default to `EmptyUsage`, the aggregated response ends up with an empty map. Gating on "did this chunk actually report tokens" avoids it, e.g. `usage.getPromptTokens() > 0 || usage.getCompletionTokens() > 0 || usage.getTotalTokens() > 0`.

2. The 3-field `MessageAggregator.DefaultUsage` record being replaced returns a *synthetic* map:

```java
@Override
public Map<String, Integer> getNativeUsage() {
    Map<String, Integer> usage = new HashMap<>();
    usage.put("promptTokens", promptTokens());
    usage.put("completionTokens", completionTokens());
    usage.put("totalTokens", totalTokens());
    return usage;
}
```

Switching unconditionally to `org.springframework.ai.chat.metadata.DefaultUsage` therefore changes `getNativeUsage()` from that synthetic map to `null` whenever no provider-native usage was recorded — a behavioural regression for code that reads `getNativeUsage()` on streaming responses today. Keeping the synthetic map as the fallback preserves the current behaviour.

Two regression tests that would pin both down: a stream with a real usage chunk followed by `EmptyUsage` chunks, and a stream that only carries token counts with no native usage.

PR #6693, opened 2026-07-27
Use the negotiated protocol version on streamable HTTP reconnect

### LINKED ISSUE #6573: WebFlux streamable HTTP reconnect GET /mcp uses the latest client protocol version instead of the negotiated one
**Bug description**

In the WebFlux streamable HTTP MCP client transport, the background reconnect `GET /mcp` request does not consistently use the negotiated protocol version.

After `initialize`, the server may negotiate the protocol down to an older supported version. Subsequent `POST` requests correctly use that negotiated version, but the reconnect `GET /mcp` request still sends the client's latest supported version instead.

This creates an internal inconsistency in the transport: the same session continues with different `mcp-protocol-version` values depending on the request type.

---

**What I verified**

I reproduced this against the existing WebFlux version-negotiation integration test setup.

Using a client configured with:

- `2025-11-25`
- `2263-03-18`

and a server that negotiates down to `2025-11-25`:

- the `initialize` request advertises the latest client-supported version, which is expected
- subsequent `POST` requests use `2025-11-25`, which is correct
- the reconnect `GET /mcp` request uses `2263-03-18`, which appears incorrect

I also noticed there is already a `FIXME` in the existing integration test mentioning this exact gap.

---

**Relevant code**

The reconnect GET path reads the protocol version from Reactor context and falls back to `latestSupportedProtocolVersion`:

- `WebClientStreamableHttpTransport.reconnect(...)`

The POST path also reads from context, but in practice it keeps the negotiated version when triggered from the initialized request flow.

This makes the reconnect GET path behave differently from the rest of the negotiated session.

---

**Steps to reproduce**

1. Use `WebClientStreamableHttpTransport` with supported protocol versions:
   - `2025-11-25`
   - `2263-03-18`
2. Initialize a client against a test server that negotiates down to `2025-11-25`
3. Trigger a normal tool call after initialization
4. Record outgoing transport requests
5. Compare the `mcp-protocol-version` header on:
   - non-initialize `POST` requests
   - reconnect `GET /mcp` request

---

**Actual behavior**

The reconnect `GET /mcp` request sends:

```text
mcp-protocol-version: 2263-03-18
```

while the rest of the established session uses:

```text
mcp-protocol-version: 2025-11-25
```

---

**Expected behavior**

Once the protocol has been negotiated, all subsequent requests for that transport session, including reconnect `GET /mcp`, should use the negotiated protocol version rather than the client's latest supported version.

---

**Why this looks like a real bug**

There is already an inline comment in the existing integration test acknowledging the problem:

```java
// FIXME: Set the correct protocol version on GET /mcp
```

So this seems to be a known gap in the current implementation rather than an ambiguous interpretation.

---

**Environment**

- Spring AI: current `main` branch
- Module: `mcp/transport/mcp-spring-webflux`

---

**Notes**

I reproduced this locally with a temporary regression test added to the existing integration test suite and confirmed that it fails consistently until the reconnect GET path is corrected.

dlwldn30:
Resolves #6573.

## Problem

`WebFluxStreamableHttpVersionNegotiationIT` carries this on `main`, with the assertion filtered down to `POST` so it can pass:

```java
// FIXME: Set the correct protocol version on GET /mcp
assertThat(calls)
    .filteredOn(c -> !c.body().contains("\"method\":\"initialize\"") && c.method().equals(HttpMethod.POST))
    .hasSize(2)
```

When the server negotiates an older version, the background `GET /mcp` reconnect still sends the client's latest supported version. `reconnect(...)` reads `NEGOTIATED_PROTOCOL_VERSION` from the Reactor context, but it is subscribed while the `initialize` exchange is still in flight, so the value is not there yet and it falls back to `latestSupportedProtocolVersion`.

## Changes

- persist the negotiated version in `WebClientStreamableHttpTransport` and fall back to it when the context has none
- defer the post-initialization reconnect until the negotiated version is known
- drop the FIXME and the `HttpMethod.POST` filter, so the assertion covers `GET /mcp` as well

## Testing

```
./mvnw -Dmaven.build.cache.enabled=false -Pintegration-tests \
  -pl mcp/transport/mcp-spring-webflux \
  -Dit.test=WebFluxStreamableHttpVersionNegotiationIT -DfailIfNoTests=false verify
```

Reverting only the production file makes `usesServerSupportedVersion` fail with the defect itself: the `GET` carries `mcp-protocol-version: 2263-03-18` while both `POST`s carry the negotiated `2025-11-25`. With the fix, both tests pass.

## Notes

Most of the diff is re-indentation, because the `Flux.deferContextual` lambda became a block. `git diff -w`, or GitHub's "Hide whitespace", shows `+41/-18` rather than `+108/-85`.

Two decisions I would rather leave to you:

- whether `negotiatedProtocolVersion` should be cleared when the session closes
- whether deferring the reconnect to the next `sendMessage` is acceptable, since it relies on the client sending `notifications/initialized` after `initialize`

Not addressed here: the session-delete `DELETE` in `createTransportSession()` still sends `latestSupportedProtocolVersion` unconditionally. Happy to cover it in a follow-up.

### DIFF
--- a/mcp/transport/mcp-spring-webflux/src/test/java/org/springframework/ai/mcp/WebFluxStreamableHttpVersionNegotiationIT.java
+++ b/mcp/transport/mcp-spring-webflux/src/test/java/org/springframework/ai/mcp/WebFluxStreamableHttpVersionNegotiationIT.java
@@ -39,7 +39,6 @@ import reactor.netty.http.server.HttpServer;
 import org.springframework.ai.mcp.client.webflux.transport.WebClientStreamableHttpTransport;
 import org.springframework.ai.mcp.server.webflux.transport.WebFluxStreamableServerTransportProvider;
 import org.springframework.ai.mcp.utils.McpTestRequestRecordingExchangeFilterFunction;
-import org.springframework.http.HttpMethod;
 import org.springframework.http.server.reactive.HttpHandler;
 import org.springframework.http.server.reactive.ReactorHttpHandlerAdapter;
 import org.springframework.web.reactive.function.client.WebClient;
@@ -157,13 +156,17 @@ class WebFluxStreamableHttpVersionNegotiationIT {
 
 			McpSchema.CallToolResult response = client.callTool(McpSchema.CallToolRequest.builder("test-tool").build());
 
+			Awaitility.await()
+				.atMost(Duration.ofSeconds(5))
+				.until(() -> this.recordingFilterFunction.getCalls()
+					.stream()
+					.filter(c -> !c.body().contains("\"method\":\"initialize\""))
+					.count() >= 3);
+
 			var calls = this.recordingFilterFunction.getCalls();
-			// Initialize tells the server the Client's latest supported version
-			// FIXME: Set the correct protocol version on GET /mcp
-			assertThat(calls)
-				.filteredOn(c -> !c.body().contains("\"method\":\"initialize\"") && c.method().equals(HttpMethod.POST))
-				// POST notification/initialized ; POST tools/call
-				.hasSize(2)
+			assertThat(calls).filteredOn(c -> !c.body().contains("\"method\":\"initialize\""))
+				// GET /mcp ; POST notification/initialized ; POST tools/call
+				.hasSize(3)
 				.map(McpTestRequestRecordingExchangeFilterFunction.Call::headers)
 				.allSatisfy(headers -> assertThat(headers).containsEntry("mcp-protocol-version",
 						ProtocolVersions.MCP_2025_11_25));
--- a/mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransport.java
+++ b/mcp/transport/mcp-spring-webflux/src/main/java/org/springframework/ai/mcp/client/webflux/transport/WebClientStreamableHttpTransport.java
@@ -54,6 +54,7 @@ import org.reactivestreams.Publisher;
 import reactor.core.Disposable;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+import reactor.util.context.ContextView;
 import reactor.util.function.Tuple2;
 import reactor.util.function.Tuples;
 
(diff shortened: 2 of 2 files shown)


dlwldn30:
Hi @sdeleuze, sorry to ping you directly — I noticed you have been working in this area recently.

No rush at all, but I wanted to ask whether there is anything I can do to help move this one along. It has been open since July 27, and the linked issue #6573 is still open.

I have just rewritten the description to make it quicker to review: it quotes the `FIXME` in `WebFluxStreamableHttpVersionNegotiationIT` that this closes, gives the command to reproduce the failure locally, and points out that most of the diff is re-indentation. The branch is still clean against `main`, so no rebase is needed.

If the scope or the size is a concern, I would be glad to adjust it however you prefer. Thanks for all your work on the project!

dlwldn30:
Hi @dimitarproynov, sorry for the direct ping — you merged #6539 and #6716 recently, so you are probably the right person to ask about this one.

Open since July 27 with no triage, and the linked issue #6573 has been open since July 8. The fix is one production file: `WebClientStreamableHttpTransport` persists the negotiated protocol version and defers the post-initialization reconnect until it is known, so the background `GET /mcp` stops sending `latestSupportedProtocolVersion`. That also let me drop the `FIXME` and the `HttpMethod.POST` filter in `WebFluxStreamableHttpVersionNegotiationIT`, which was masking the defect.

Two open decisions from the description are the only things I am unsure about, and either is a one-line answer for you:

- whether `negotiatedProtocolVersion` should be cleared when the session closes
- whether deferring the reconnect to the next `sendMessage` is acceptable, given it relies on the client sending `notifications/initialized`

Happy to take either direction, and to fold in the `DELETE` in `createTransportSession()` if you want that covered here rather than as a follow-up. Is there anything else I can do to help move this along?

dlwldn30:
Hi @dimitarproynov, checking in again since it's been about two weeks. The two open questions from my last comment — whether `negotiatedProtocolVersion` should be cleared on session close, and whether deferring the reconnect to the next `sendMessage` is acceptable — are still the only blockers I'm aware of on my end. Happy to go either direction, or split one off as a follow-up if that's easier. Thanks for your patience.

dimitarproynov:
@dlwldn30 sorry, I was very busy with the most requested feature in the issues tracker.

@Kehrlann can you, please, take a look at this change.

mcp/transport/mcp-spring-webflux/src/test/java/org/springframework/ai/mcp/WebFluxStreamableHttpVersionNegotiationIT.java
```diff
@@ -157,13 +156,17 @@ void usesServerSupportedVersion() {
 
 			McpSchema.CallToolResult response = client.callTool(McpSchema.CallToolRequest.builder("test-tool").build());
 
+			Awaitility.await()
+				.atMost(Duration.ofSeconds(5))
+				.until(() -> this.recordingFilterFunction.getCalls()
+					.stream()
+					.filter(c -> !c.body().contains("\"method\":\"initialize\""))
+					.count() >= 3);
+
```
Kehrlann:
```suggestion
			// GET /mcp is triggered asynchronously
			Awaitility.await()
				.atMost(Duration.ofSeconds(5))
				.until(() -> this.recordingFilterFunction.getCalls()
					.stream()
					.anyMatch(r -> r.method().equals(HttpMethod.GET)));
```

PR #6701, opened 2026-07-29
Parse `long` literals in `FilterExpressionTextParser`

### LINKED ISSUE #4705: FilterExpressionTextParser treats all number literals as Integer, failing on Long values

wxggzz:
**What**

`FilterExpressionTextParser` parsed every bare integer literal as `Integer`, so a filter expression whose value exceeds the int range — e.g. a `long` metadata id used when deleting by id — failed with:

```
```

**Fix**

`visitIntegerConstant` now falls back to `Long` when the literal doesn't fit in an `int`, while still returning `Integer` for smaller values (backward compatible). Literals with an explicit `L` suffix keep going through the existing `visitLongConstant` path, and values inside `IN`/`NIN` arrays are covered too since they delegate to the same visitor.

**Tests**

Added `testLongValueLiteral` covering `Long.MAX_VALUE` / `Long.MIN_VALUE` and asserting that in-range values still parse as `Integer`. All 16 parser tests pass locally:

```
./mvnw -pl spring-ai-vector-store -am -Dtest=FilterExpressionTextParserTests -Dsurefire.failIfNoSpecifiedTests=false test
```

Fixes #4705

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
@@ -231,7 +231,14 @@ public class FilterExpressionTextParser {
 
 		@Override
 		public Filter.Operand visitIntegerConstant(FiltersParser.IntegerConstantContext ctx) {
-			return new Filter.Value(Integer.valueOf(ctx.getText()));
+			String text = ctx.getText();
+			try {
+				return new Filter.Value(Integer.valueOf(text));
+			}
+			catch (NumberFormatException ex) {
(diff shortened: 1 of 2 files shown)


sobychacko:
Looks like this is a duplicate of https://github.com/spring-projects/spring-ai/pull/5996. Closing it; feel free to re-open if there is a need. Thanks!

PR #6338, opened 2026-06-08
Support `none` and `required` `tool_choice` options in `OpenAiChatModel`

guanxuc:
Closes: #6332

PR #6703, opened 2026-07-29
Preserve AssistantMessage subclass fields on copy

guanxuc:
# Root Cause

`Prompt.instructionsCopy()` rebuilt every `AssistantMessage` using the base `AssistantMessage.builder()`, which downcast provider-specific subclasses (DeepSeek, Bedrock, Anthropic) to a plain `AssistantMessage`, silently dropping their extra state:
- **DeepSeek**: `reasoningContent` and `prefix`
- **Bedrock**: `reasoningContents`
- **Anthropic**: `thinkingContents`

It also dropped `media` for all assistant messages. As a result, multi-turn and tool-calling flows that route through `Prompt.copy()` / `mutate()` (e.g., `DeepSeekChatModel.buildRequestPrompt` when no per-call options are set) silently lost reasoning context on subsequent requests.

# Fix

Following the same pattern as `UserMessage` / `SystemMessage`, `AssistantMessage` now exposes `copy()` backed by `mutate().build()`, and each stateful subclass overrides `mutate()` to carry its own fields through the copy:
- **`AssistantMessage`**: `copy()` + `mutate()` preserving `content`, `metadata`, `toolCalls`, and `media`.
- **`DeepSeekAssistantMessage`**: preserves `reasoningContent` and `prefix`.
- **`BedrockAssistantMessage`**: preserves `reasoningContents`.
- **`AnthropicChatModel.AnthropicAssistantMessage`**: adds a `Builder` and overrides `mutate()` to preserve `thinkingContents`.

`Prompt.instructionsCopy()` now delegates to `assistantMessage.copy()`, so subclass identity and fields are fully preserved.

# Tests

All unit tests pass, covering `AssistantMessage.copy()` across the base class and each subclass. Run them with:
```sh
./mvnw -am -pl spring-ai-model -Dtest=org.springframework.ai.chat.messages.AssistantMessageTests -Dsurefire.failIfNoSpecifiedTests=false -Dmaven.build.cache.enabled=false test
./mvnw -am -pl models/spring-ai-deepseek -Dtest=org.springframework.ai.deepseek.DeepSeekAssistantMessageTests -Dsurefire.failIfNoSpecifiedTests=false -Dmaven.build.cache.enabled=false test
./mvnw -am -pl models/spring-ai-bedrock-converse -Dtest=org.springframework.ai.bedrock.converse.BedrockAssistantMessageTests -Dsurefire.failIfNoSpecifiedTests=false -Dmaven.build.cache.enabled=false test
./mvnw -am -pl models/spring-ai-anthropic -Dtest=org.springframework.ai.anthropic.AnthropicAssistantMessageTests -Dsurefire.failIfNoSpecifiedTests=false -Dmaven.build.cache.enabled=false test
```

# Related Issue
Follow up #6638, rerun `DeepSeekChatModelIT#prefixCompletionTest` and now the unit test pass.
```sh
./mvnw -am -pl models/spring-ai-deepseek -Dtest=org.springframework.ai.deepseek.chat.DeepSeekChatModelIT#prefixCompletionTest -Dsurefire.failIfNoSpecifiedTests=false -Dmaven.build.cache.enabled=false test
```

image

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -1584,7 +1584,7 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 
 			@Override
 			public AnthropicAssistantMessage build() {
-				Assert.hasText(this.content, "content cannot be null or empty");
+				Assert.notNull(this.content, "content cannot be null");
 				return new AnthropicAssistantMessage(this.content, this.toolCalls, this.thinkingContents);
 			}
 
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -147,6 +147,7 @@ import org.springframework.util.MimeType;
  * @author Ilayaperumal Gopinathan
  * @author Jewoo Shin
  * @author Seeun Kim
+ * @author guan xu
  * @since 1.0.0
  * @see AnthropicChatOptions
  * @see <a href="https://docs.anthropic.com/en/api/messages">Anthropic Messages API</a>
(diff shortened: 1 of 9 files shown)


guanxuc:
Hi @dimitarproynov ,

Could you please take another look at this PR? It is a follow-up to #6338.

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -1554,12 +1555,41 @@ boolean hasThinkingContents() {
...
+			}
+
+			Builder thinkingContents(List<AnthropicThinkingContent> thinkingContents) {
+				this.thinkingContents = thinkingContents;
+				return self();
+			}
+
+			@Override
+			public AnthropicAssistantMessage build() {
+				Assert.hasText(this.content, "content cannot be null or empty");
```
dimitarproynov: This is a regression. See line 494 where we call:

 buildAssistantMessage("", toolCalls, streamingState.getThinkingContents());

Meaning, that if somebody calls the .copy() on that Anthropic AssistantMessage that uses thinking, the above assertion will fire.

Please, drop the assertion.

models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/AnthropicAssistantMessageTests.java
```diff
@@ -0,0 +1,67 @@
...
+ * Unit tests for {@link AnthropicChatModel.AnthropicAssistantMessage}.
+ *
+ * @author guan xu
+ */
+class AnthropicAssistantMessageTests {
+
+	@Test
+	void copyPreservesThinkingContents() {
+		AnthropicChatModel.AnthropicAssistantMessage original = AnthropicChatModel.AnthropicAssistantMessage.builder()
+			.content("hello")
```
dimitarproynov: Maybe set the content to an empty string here, to verify the case works.

dimitarproynov (changes requested):

Some remarks below:

guanxuc:
Hi @dimitarproynov ,

Thanks for the review and the great point! I've fixed the incorrect assertion and now only assert non-null for null-safe ability. And also updated the unit test to verify empty content handling.

dimitarproynov (approved):

PR #6705, opened 2026-07-29
Preserve tool schema property order in Converse API

### LINKED ISSUE #6696: ToolConfiguration input schema loses JSON property declaration order (ConverseApiUtils#convertMapToDocument uses HashMap)
### Environment
- Spring AI version: 2.0.0
- Module: spring-ai-bedrock-converse

### Description
When a `ToolCallback`'s JSON input schema is converted into a Bedrock `Document` for the Converse API's `ToolConfiguration`, the property insertion order declared in the schema JSON (and preserved by Jackson's default `LinkedHashMap` parsing) is discarded, because
`ConverseApiUtils#convertMapToDocument` collects entries with a plain `Collectors.toMap(...)` call and no map-supplier argument, which defaults to `HashMap`:

https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/api/ConverseApiUtils.java

```java
private static Document convertMapToDocument(Map<String, Object> value) {
    Map<String, Document> attr = value.entrySet()
        .stream()
        .collect(Collectors.toMap(e -> e.getKey(), e -> convertObjectToDocument(e.getValue())));
    return Document.fromMap(attr);
}
```

This means the "properties" object of a tool's `inputSchema` is sent to Bedrock with an arbitrary (`String.hashCode()`-bucket-driven) key order instead of the order declared in the schema JSON. For example, a 3-key `LinkedHashMap` for `{evidenceText, justification, valid}`
consistently comes out as `[valid, evidenceText, justification]` regardless of the declared order.

### Impact
Property order isn't semantically meaningful for JSON Schema validation itself, but several real prompting techniques for LLM tool-calling rely on the field order actually presented to the model (e.g. asking the model to reason through evidence/justification fields before a
final decision field, a common "chain-of-thought via schema field order" pattern). Because this order is silently rearranged, those techniques stop working reliably as soon as the schema is routed through Bedrock Converse, with no error or warning — it just quietly produces worse model output.

### Expected behavior
The map passed to `Document.fromMap(...)` should preserve the original iteration order of the input `Map` (e.g. by using
`Collectors.toMap(Map.Entry::getKey, e -> convertObjectToDocument(e.getValue()), (a, b) -> a, LinkedHashMap::new)`), so that any `LinkedHashMap`/ordered-map input keeps its property order end-to-end.

### Suggested fix
```java
private static Document convertMapToDocument(Map<String, Object> value) {
    Map<String, Document> attr = value.entrySet()
        .stream()
        .collect(Collectors.toMap(
            Map.Entry::getKey,
            e -> convertObjectToDocument(e.getValue()),
            (a, b) -> a,
            LinkedHashMap::new));

    return Document.fromMap(attr);
}
```

yyyCode:
`ConverseApiUtils#convertMapToDocument` collected map entries with a plain `Collectors.toMap(...)`, which returns a `HashMap` and therefore discards the insertion order of the source map (preserved by Jackson as a `LinkedHashMap`).

As a result, the `properties` of a tool's `inputSchema` were sent to Bedrock in an arbitrary, `hashCode`-driven order instead of the declared order.

This change collects into a `LinkedHashMap` so the declaration order is preserved, and adds a test covering both top-level and nested maps.

Fixes #6696

### DIFF
--- a/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/api/ConverseApiUtils.java
+++ b/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/api/ConverseApiUtils.java
@@ -19,6 +19,7 @@ package org.springframework.ai.bedrock.converse.api;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -100,7 +101,8 @@ public final class ConverseApiUtils {
 	private static Document convertMapToDocument(Map<String, Object> value) {
 		Map<String, Document> attr = value.entrySet()
 			.stream()
-			.collect(Collectors.toMap(e -> e.getKey(), e -> convertObjectToDocument(e.getValue())));
+			.collect(Collectors.toMap(Map.Entry::getKey, e -> convertObjectToDocument(e.getValue()), (a, b) -> b,
+					LinkedHashMap::new));
 
 		return Document.fromMap(attr);
 	}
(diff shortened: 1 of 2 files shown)


yyyCode:
@rozza Could I get a quick approve to kick off the CI tests? Thanks!

PR #6709, opened 2026-07-30
Skip empty text content blocks for Bedrock user messages

### LINKED ISSUE #6695: Bedrock Converse: Empty text content is sent to API and being rejected
**Bug description**
Spring AI always creates a text content block from user messages that are sent to AWS Bedrock Converse API, even if the text of the user message is null or empty. This results in an HTTP 400 error being returned from the Converse API, if a user message with only images but no text was sent, with the error message stating that there must be no empty text content block. Sending the same request without the empty text content block to the API via Postman results in a valid result.

**Environment**
Spring AI version: 2.0.0
Used Model Provider: Amazon Bedrock Converse API
Java version: 21 (Zulu JDK 21.48.16)

**Steps to reproduce**
1. Setup Spring AI to access a model like Claude Sonnet 4.6 via Amazon Bedrock.
2. Create a Prompt with a simple system message and one user message containing an empty text content (null is not allowed) and at least one image as media.
3. Call the Bedrock Chat Model with the prompt.

**Expected behavior**
There is no text content block created, the request is accepted by the API and a generation response is returned successfully.

**Minimal code example**
```
chatModel.call(
                    new Prompt(
                            List.of(
                                    SystemMessage.builder().text("Describe the image").build(),
                                    UserMessage.builder()
                                            .text("")
                                            .media(Media.builder()
                                                .mimeType(MediaType.IMAGE_PNG)
                                                .data(Base64.getDecoder().decode("<PNG as Base64>"))
                                                .build())
                                            .build()
                            ),
                            BedrockChatOptions.builder()
                                    .model("global.anthropic.claude-sonnet-4-6")
                                    .maxTokens(1000)
                                    .build()
                    )
            );
```

chabinhwang:
## Summary

`BedrockProxyChatModel` always added a text `ContentBlock` for a `UserMessage`, even when the message had no text.
The Converse API rejects an empty text content block, so a user message that carries only media failed with an HTTP 400.

The assistant message branch of the same method already guards with `StringUtils.hasText(...)`; this applies the same guard to the user message branch.

Fixes #6695

## Reproduction

From the issue: a prompt whose user message has `text("")` and one image, sent to a vision-capable model through Bedrock, is rejected by the API. The same request without the empty text block succeeds.

```java
UserMessage.builder()
    .text("")
    .media(Media.builder().mimeType(MediaType.IMAGE_PNG).data(pngBytes).build())
    .build()
```

## Testing

Three tests added to `BedrockProxyChatModelTest`, all going through `createRequest`, so they assert on the actual `ConverseRequest` payload:

* `mediaOnlyUserMessageOmitsEmptyTextContentBlock` — empty text plus image yields a single image block and no text block
* `blankUserMessageTextOmitsEmptyTextContentBlock` — whitespace-only text is treated the same way
* `userMessageWithTextAndMediaKeepsBothContentBlocks` — regression guard: text plus image still yields both blocks in order

Verified locally:

* `./mvnw -Dmaven.build.cache.enabled=false -pl models/spring-ai-bedrock-converse clean package` — `Tests run: 87, Failures: 0, Errors: 0`, checkstyle and `spring-javaformat` pass
* Reverted the main-code change and re-ran only the three new tests: the two empty-text tests fail (`Tests run: 3, Failures: 2`) and the regression guard passes, confirming the tests cover the reported defect

`BedrockProxyChatModelIT` requires AWS credentials and was not executed.

## Note on the all-empty case

A user message with neither text nor media now produces a message with an empty content list. That request was already rejected by the API before this change (as an empty text block), so this is not a regression, and it is out of scope here.

### DIFF
--- a/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
+++ b/models/spring-ai-bedrock-converse/src/main/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModel.java
@@ -288,7 +288,12 @@ public class BedrockProxyChatModel implements ChatModel {
 				List<ContentBlock> contents = new ArrayList<>();
 				if (message instanceof UserMessage) {
 					var userMessage = (UserMessage) message;
-					contents.add(ContentBlock.fromText(userMessage.getText()));
+					// The Converse API rejects empty text content blocks, so only send
+					// the text when there is any. A user message may legitimately carry
+					// media only (gh-6695).
+					if (StringUtils.hasText(userMessage.getText())) {
+						contents.add(ContentBlock.fromText(userMessage.getText()));
+					}
 
 					if (!CollectionUtils.isEmpty(userMessage.getMedia())) {
 						List<ContentBlock> mediaContent = userMessage.getMedia()
--- a/models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModelTest.java
+++ b/models/spring-ai-bedrock-converse/src/test/java/org/springframework/ai/bedrock/converse/BedrockProxyChatModelTest.java
@@ -415,6 +415,58 @@ class BedrockProxyChatModelTest {
 		assertThat(tools.get(1).cachePoint().ttlAsString()).isEqualTo("1h");
 	}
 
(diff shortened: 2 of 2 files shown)


chabinhwang:
Superseded by #6946. Same change, rebased onto current `main`.

PR #6710, opened 2026-07-30
Map `audio/mpeg` to the OpenAI mp3 input audio format

chabinhwang:
## Summary

The OpenAI audio input format was chosen with `mimeType.contains("mp3")`, falling back to WAV for anything else.
`audio/mpeg` is the media type registered for MP3 (RFC 3003) and does not contain the substring `mp3`, so MP3 audio declared that way was silently sent to OpenAI labeled as `wav`.

This moves the decision into `toInputAudioFormat(String)`, which accepts `audio/mp3` and `audio/mpeg` (case-insensitively) for MP3 and warns when an audio type that is neither MP3 nor WAV is forwarded as WAV.

## Why `audio/mpeg` matters

`Media.Format` exposes no audio constants, so callers build the MIME type themselves, and `audio/mpeg` is what standard detection produces for an `.mp3` file.
Spring AI's own speech documentation uses `audio/mpeg` for MP3 output, so feeding generated speech back in as chat input hits exactly this path.
The failure is silent at the Spring AI layer — the request is built and sent with the wrong format declaration.

The reference documentation previously listed only `audio/mp3` and `audio/wav`; it now lists `audio/mpeg` as well.

## Testing

Two parameterized tests added to `OpenAiChatModelTests`, both going through `createRequest` so they assert on the `ChatCompletionCreateParams` actually sent:

* `userMessageMp3MediaMapsToMp3InputAudioFormat` — `audio/mp3`, `audio/mpeg`, `audio/MPEG`
* `userMessageWavMediaMapsToWavInputAudioFormat` — `audio/wav`, `audio/x-wav`, `audio/vnd.wave`

Both also assert the base64 payload is unchanged.

Verified locally:

* `./mvnw -Dmaven.build.cache.enabled=false -pl models/spring-ai-openai clean package` — `Tests run: 176, Failures: 0, Errors: 0`, checkstyle and `spring-javaformat` pass
* Reverted only the `audio/mpeg` handling and re-ran the new tests: the `audio/mpeg` and `audio/MPEG` cases fail with `expected: mp3 but was: wav`, while the `audio/mp3` and WAV cases pass either way

Integration tests requiring an OpenAI API key were not executed.

## Scope

Only the input audio format resolution and the corresponding documentation line changed. The output audio path and the transcription/speech models are untouched.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -575,9 +575,7 @@ public final class OpenAiChatModel implements ChatModel {
 										.inputAudio(ChatCompletionContentPartInputAudio.builder()
 											.inputAudio(ChatCompletionContentPartInputAudio.InputAudio.builder()
 												.data(fromAudioData(media.getData()))
-												.format(mimeType.contains("mp3")
-														? ChatCompletionContentPartInputAudio.InputAudio.Format.MP3
-														: ChatCompletionContentPartInputAudio.InputAudio.Format.WAV)
+												.format(toInputAudioFormat(mimeType))
 												.build())
 											.build()
 											.inputAudio())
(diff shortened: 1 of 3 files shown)


chabinhwang:
Superseded by #6945. Rebased onto current `main` to resolve merge conflicts in `OpenAiChatModelTests` after later OpenAI timeout / RequestOptions work landed.

PR #6716, opened 2026-08-01
Apply MCP HTTP client request customizer beans

### LINKED ISSUE #6715: `McpSyncHttpClientRequestCustomizer` and `McpAsyncHttpClientRequestCustomizer` beans are silently ignored by the MCP client transport auto-configuration
## Bug description

`StreamableHttpHttpClientTransportAutoConfiguration` no longer picks up `McpSyncHttpClientRequestCustomizer` or `McpAsyncHttpClientRequestCustomizer` beans from the application context.
A bean of either type is created, never wired to any transport, and never invoked.
No warning is logged, so the only symptom is a missing header on every outbound MCP request.

This changed in commit `272c73612` ("Introduce `McpClientCustomizer` abstraction", PR #5606).
Before that commit the auto-configuration took both types as `ObjectProvider` parameters and installed them on the transport builder:

```java
asyncHttpRequestCustomizer.ifUnique(transportBuilder::asyncHttpRequestCustomizer);
syncHttpRequestCustomizer.ifUnique(transportBuilder::httpRequestCustomizer);
```

Those two parameters were removed and replaced by the `McpClientCustomizer` loop.
The commit message describes replacing `McpSyncClientCustomizer` / `McpAsyncClientCustomizer` and using the new abstraction "in all transports to customize the transport builders".
It doesn't mention dropping request-customizer bean support, so this looks incidental rather than intended.

`McpClientCustomizer` is not an equivalent replacement for the removed behavior:

- It customizes the **builder**, not the request. To add a header you have to know that `HttpClientStreamableHttpTransport.Builder#asyncHttpRequestCustomizer` exists and call it yourself.
- The builder setter **assigns** rather than composes, so two `McpClientCustomizer` beans that both set a request customizer silently overwrite each other. Last one wins, with no warning.
- It is transport-type specific. A `McpClientCustomizer<HttpClientStreamableHttpTransport.Builder>` does nothing if the app is on the WebFlux transport, and the generic type gives no compile-time or startup-time signal.

The request-customizer interfaces are still public API in `mcp-core` 2.0.0, still documented, and the SDK ships `DelegatingMcpSyncHttpClientRequestCustomizer` and `DelegatingMcpAsyncHttpClientRequestCustomizer` for composing several of them. Spring AI just doesn't use any of it.

## Environment

- Spring AI: reproduced on `main` (`2.0.1-SNAPSHOT`, commit `db45fc548`) and on 2.0.0. The last working version is 2.0.0-M2.
- MCP Java SDK: `io.modelcontextprotocol.sdk:mcp-core` 2.0.0
- Module: `spring-ai-autoconfigure-mcp-client-httpclient`
- Class: `org.springframework.ai.mcp.client.httpclient.autoconfigure.StreamableHttpHttpClientTransportAutoConfiguration`
- Java: 17
- Build: Maven, no vector store involved

## Steps to reproduce

1. Add `spring-ai-starter-mcp-client` and configure one streamable-HTTP connection:

   ```yaml
   spring:
     ai:
       mcp:
         client:
           streamable-http:
             connections:
               server1:
                 url: http://localhost:8080
   ```

2. Declare a request customizer bean:

   ```java
   @Bean
   McpAsyncHttpClientRequestCustomizer asyncRequestCustomizer() {
       return (builder, method, endpoint, body, context) -> {
           builder.setHeader("X-Async", "applied");
           return Mono.just(builder);
       };
   }
   ```

3. Start the application and call any MCP tool.

The `X-Async` header is absent from every request. The bean's lambda never runs. Nothing is logged.

Same result with a `McpSyncHttpClientRequestCustomizer` bean.

## Expected behavior

Every `McpSyncHttpClientRequestCustomizer` and `McpAsyncHttpClientRequestCustomizer` bean is applied to every transport the auto-configuration creates, composed rather than overwritten, honoring `@Order`.

If the removal is intentional and the beans are meant to stay unused, then a startup warning instead:

```java
if (syncHttpRequestCustomizer.getIfAvailable() != null || asyncHttpRequestCustomizer.getIfAvailable() != null) {
    logger.warn("Found McpSyncHttpClientRequestCustomizer or McpAsyncHttpClientRequestCustomizer beans. "
            + "These are no longer applied automatically. Register them on the transport builder from a "
            + "McpClientCustomizer instead.");
}
```

The second option is cheap and would have turned this into a one-minute fix.

## Minimal Complete Reproducible example

Failing tests in `StreamableHttpHttpClientTransportAutoConfigurationTests`, in module `spring-ai-autoconfigure-mcp-client-httpclient`.
The transport exposes no getter for its request customizer, so the test reads the field reflectively and invokes it, just as the existing `getStreamableHttpEndpoint` helper in that class reads `endpoint`.

```java
@Test
void asyncHttpRequestCustomizerBeanIsApplied() {
    this.applicationContext.withUserConfiguration(AsyncRequestCustomizerConfiguration.class)
        .withPropertyValues("spring.ai.mcp.client.streamable-http.connections.server1.url=http://localhost:8080")
        .run(context -> {
            List<NamedClientMcpTransport> transports = context.getBean("streamableHttpHttpClientTransports",
                    List.class);
            assertThat(headersAppliedBy((HttpClientStreamableHttpTransport) transports.get(0).transport()))
                .containsEntry("X-Async", "applied");
        });
}

@Configuration
static class AsyncRequestCustomizerConfiguration {

    @Bean
    McpAsyncHttpClientRequestCustomizer asyncRequestCustomizer() {
        return (builder, method, endpoint, body, context) -> {
            builder.setHeader("X-Async", "applied");
            return Mono.just(builder);
        };
    }

}

/**
 * Invokes the transport's request customizer against a throwaway request builder and
 * returns the headers it set.
 */
private Map<String, String> headersAppliedBy(HttpClientStreamableHttpTransport transport) {
    Field privateField = ReflectionUtils.findField(HttpClientStreamableHttpTransport.class,
            "httpRequestCustomizer");
    ReflectionUtils.makeAccessible(privateField);
    McpAsyncHttpClientRequestCustomizer customizer = (McpAsyncHttpClientRequestCustomizer) ReflectionUtils
        .getField(privateField, transport);

    URI endpoint = URI.create("http://localhost:8080/mcp");
    HttpRequest.Builder builder = HttpRequest.newBuilder(endpoint);
    HttpRequest.Builder customized = Mono
        .from(customizer.customize(builder, "POST", endpoint, "{}", McpTransportContext.EMPTY))
        .block();

    Map<String, String> headers = new LinkedHashMap<>();
    customized.build().headers().map().forEach((name, values) -> headers.put(name, String.join(",", values)));
    return headers;
}
```

On `main` this fails:

```
[ERROR] StreamableHttpHttpClientTransportAutoConfigurationTests.asyncHttpRequestCustomizerBeanIsApplied

Expecting map:
  {}
to contain entries:
  ["X-Async"="applied"]
but could not find the following map entries:
  ["X-Async"="applied"]
```

I have four of these locally, covering the async bean, the sync bean, both beans together, and `@Order` between two async beans. All four fail on `main` with the same empty map. A fifth pins the existing precedence: a `McpClientCustomizer` that calls `asyncHttpRequestCustomizer(...)` still wins over the beans, which pass both before and after the fix.

## Suggested fix

Take both `ObjectProvider`s again and compose the beans with the SDK's delegating customizer:

```java
@Bean
public List<NamedClientMcpTransport> streamableHttpHttpClientTransports(
        McpStreamableHttpClientProperties streamableProperties, ObjectProvider<JsonMapper> jsonMapperProvider,
        ObjectProvider<McpSyncHttpClientRequestCustomizer> syncRequestCustomizers,
        ObjectProvider<McpAsyncHttpClientRequestCustomizer> asyncRequestCustomizers,
        ObjectProvider<McpClientCustomizer<HttpClientStreamableHttpTransport.Builder>> transportCustomizers) {

    List<McpAsyncHttpClientRequestCustomizer> requestCustomizers = requestCustomizers(syncRequestCustomizers,
            asyncRequestCustomizers);

    // ... per connection, before the transportCustomizers loop:
    if (!requestCustomizers.isEmpty()) {
        transportBuilder
            .asyncHttpRequestCustomizer(new DelegatingMcpAsyncHttpClientRequestCustomizer(requestCustomizers));
    }
}

private static List<McpAsyncHttpClientRequestCustomizer> requestCustomizers(
        ObjectProvider<McpSyncHttpClientRequestCustomizer> syncRequestCustomizers,
        ObjectProvider<McpAsyncHttpClientRequestCustomizer> asyncRequestCustomizers) {

    List<McpAsyncHttpClientRequestCustomizer> requestCustomizers = new ArrayList<>();
    syncRequestCustomizers.orderedStream()
        .map(McpAsyncHttpClientRequestCustomizer::fromSync)
        .forEach(requestCustomizers::add);
    asyncRequestCustomizers.orderedStream().forEach(requestCustomizers::add);
    return requestCustomizers;
}
```

Notes on the shape:

- `orderedStream()` rather than plain iteration, so `@Order` and `Ordered` work. The old code used `ifUnique`, which silently dropped everything when more than one bean was present.
- Sync customizers run before async ones. Interleaving the two by a single global order isn't reliably available from two separate `ObjectProvider`s, and the old code had a fixed precedence too.
- Request customizers are installed before the `McpClientCustomizer` loop, so any existing customizer that calls `asyncHttpRequestCustomizer(...)` keeps winning. That keeps the change backward compatible.

With this applied, the whole test class passes: 13 tests, 0 failures.

I left `SseHttpClientTransportAutoConfiguration` alone since it is `@Deprecated(since = "2.0.0", forRemoval = true)`.

Tests from https://github.com/spring-projects/spring-ai/pull/6716 reproduce the issue.

iuliiasobolevska:
`McpSyncHttpClientRequestCustomizer` and `McpAsyncHttpClientRequestCustomizer` beans defined in the application context are not applied to the streamable HTTP client transport.
Wire them into the transport builder through `DelegatingMcpAsyncHttpClientRequestCustomizer` so that every such bean is applied to each outbound request.

Fixes #6715.

### DIFF
--- a/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-httpclient/src/main/java/org/springframework/ai/mcp/client/httpclient/autoconfigure/StreamableHttpHttpClientTransportAutoConfiguration.java
+++ b/auto-configurations/mcp/spring-ai-autoconfigure-mcp-client-httpclient/src/main/java/org/springframework/ai/mcp/client/httpclient/autoconfigure/StreamableHttpHttpClientTransportAutoConfiguration.java
@@ -22,6 +22,9 @@ import java.util.List;
 import java.util.Map;
 
 import io.modelcontextprotocol.client.transport.HttpClientStreamableHttpTransport;
+import io.modelcontextprotocol.client.transport.customizer.DelegatingMcpAsyncHttpClientRequestCustomizer;
+import io.modelcontextprotocol.client.transport.customizer.McpAsyncHttpClientRequestCustomizer;
+import io.modelcontextprotocol.client.transport.customizer.McpSyncHttpClientRequestCustomizer;
 import io.modelcontextprotocol.json.jackson3.JacksonMcpJsonMapper;
 import tools.jackson.databind.json.JsonMapper;
 
@@ -52,6 +55,8 @@ import org.springframework.context.annotation.Bean;
  * connections
  * <li>Configures JsonMapper for JSON serialization/deserialization
  * <li>Supports multiple named server connections with different URLs
+ * <li>Applies all {@link McpSyncHttpClientRequestCustomizer} and
+ * {@link McpAsyncHttpClientRequestCustomizer} beans to every outbound request
  * <li>Applies {@link McpClientCustomizer<HttpClientStreamableHttpTransport.Builder>}
  * beans to each transport builder.
  * </ul>
@@ -75,14 +80,26 @@ public class StreamableHttpHttpClientTransportAutoConfiguration {
 	 * <li>A new HttpClient instance
 	 * <li>Server URL from properties
 	 * <li>JsonMapper for JSON processing
+	 * <li>All available {@link McpSyncHttpClientRequestCustomizer} and
+	 * {@link McpAsyncHttpClientRequestCustomizer} beans, composed into a single request
+	 * customizer
 	 * <li>All available
 	 * {@link McpClientCustomizer<HttpClientStreamableHttpTransport.Builder>} beans
 	 * applied with the connection name and transport builder
 	 * </ul>
+	 *
+	 * <p>
+	 * Request customizer beans are installed on the builder before the
+	 * {@link McpClientCustomizer} beans run, so a customizer that calls
+	 * {@code asyncHttpRequestCustomizer(...)} itself still takes precedence.
 	 * @param streamableProperties the Streamable HTTP client properties containing server
 	 * configurations
 	 * @param jsonMapperProvider the provider for JsonMapper or a new instance if not
 	 * available
+	 * @param syncRequestCustomizers provider for
+	 * {@link McpSyncHttpClientRequestCustomizer} beans
+	 * @param asyncRequestCustomizers provider for
+	 * {@link McpAsyncHttpClientRequestCustomizer} beans
 	 * @param transportCustomizers provider for
 	 * {@link McpClientCustomizer<HttpClientStreamableHttpTransport.Builder>} beans
 	 * @return list of named MCP transports
@@ -90,9 +107,14 @@ public class StreamableHttpHttpClientTransportAutoConfiguration {
 	@Bean
 	public List<NamedClientMcpTransport> streamableHttpHttpClientTransports(
 			McpStreamableHttpClientProperties streamableProperties, ObjectProvider<JsonMapper> jsonMapperProvider,
+			ObjectProvider<McpSyncHttpClientRequestCustomizer> syncRequestCustomizers,
+			ObjectProvider<McpAsyncHttpClientRequestCustomizer> asyncRequestCustomizers,
 			ObjectProvider<McpClientCustomizer<HttpClientStreamableHttpTransport.Builder>> transportCustomizers) {
 
 		JsonMapper jsonMapper = jsonMapperProvider.getIfAvailable(JsonMapper::shared);
+
+		List<McpAsyncHttpClientRequestCustomizer> requestCustomizers = requestCustomizers(syncRequestCustomizers,
+				asyncRequestCustomizers);
 
 		List<NamedClientMcpTransport> streamableHttpTransports = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


dimitarproynov:

PR #6718, opened 2026-08-01
Make AnthropicChatModel internal methods private

### LINKED ISSUE #6700: AnthropicChatModel exposes `internalStream` method publicly
While doing some API mapping of the Spring AI project for my job (I work in security research @ Veracode) I noticed that the Anthropic implementation of `ChatModel` appears to expose a method that has a naming prefix that would indicate it probably is not intended to have `public` access:

https://github.com/spring-projects/spring-ai/blob/9b76ba9efca3fd07633d3d90b6163a79f0c05acc/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java#L307

Checking other implementations, and just perusing the code a bit It seems that this was mentioned in an update guide:

https://github.com/spring-projects/spring-ai/blob/9b76ba9efca3fd07633d3d90b6163a79f0c05acc/spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc#the-access-level-of-model-internal-methods-changed-to-private

I was going to open a PR, but the guide says to avoid opening a PR for straightforward changes, and this looks like it meets that criteria. 

Cheers!

arimu1:

Thank you upfront!

## Description

`AnthropicChatModel.internalStream` (and `internalCall`) were left `public` after the model internal methods were supposed to be `private` (see upgrade notes). Other chat model implementations already use `private`.

This change restricts both methods to `private`, matching:

* upgrade notes: "All `internalCall` and `internalStream` methods in model classes have been changed to `private`"
* OpenAI, Ollama, Bedrock Converse, Mistral, DeepSeek, Google GenAI

Maintainer @guanxuc confirmed on #6700 that these should be private.

No external callers needed adjustment; both methods are only used within `AnthropicChatModel` (`call`/`stream` and recursive tool-call loops).

Fixes #6700

## Testing

```
./mvnw -pl models/spring-ai-anthropic -am test -DskipITs
```

Results: **124 tests, 0 failures, 0 errors** (JDK 17.0.19)

## AI assistance

Created with AI assistance and reviewed by a human.

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -304,7 +304,7 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 	 * chunks and the final response with tool call information or the model's final
 	 * answer.
 	 */
-	public Flux<ChatResponse> internalStream(Prompt prompt, @Nullable ChatResponse previousChatResponse) {
+	private Flux<ChatResponse> internalStream(Prompt prompt, @Nullable ChatResponse previousChatResponse) {
 
 		return Flux.deferContextual(contextView -> {
 			MessageCreateParams request = createRequest(prompt, true);
@@ -541,7 +541,7 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 	 * turn.
 	 * @return The final {@link ChatResponse} after all tool calls (if any) are resolved.
 	 */
-	public ChatResponse internalCall(Prompt prompt, @Nullable ChatResponse previousChatResponse) {
+	private ChatResponse internalCall(Prompt prompt, @Nullable ChatResponse previousChatResponse) {
 
 		MessageCreateParams request = createRequest(prompt, false);
 


models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -304,7 +304,7 @@ public Flux<ChatResponse> stream(Prompt prompt) {
 	 * chunks and the final response with tool call information or the model's final
 	 * answer.
 	 */
-	public Flux<ChatResponse> internalStream(Prompt prompt, @Nullable ChatResponse previousChatResponse) {
+	private Flux<ChatResponse> internalStream(Prompt prompt, @Nullable ChatResponse previousChatResponse) {
```
kuntal1461, line 308: Looks right. The only caller is `stream()` on line 283, and since the class is `final` there's no subclass path either — matches what Ollama, Mistral, DeepSeek, and the rest all do.

Nit (pre-existing, not a blocker): the Javadoc still says "called recursively to support multi-turn tool calling" — that's been stale since `a48dc60af` moved the recursive loop into `ToolCallingManager`. Worth a follow-up cleanup but doesn't need to hold up this PR.

models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
```diff
@@ -541,7 +541,7 @@ else if (streamingState.isTrackingThinking()) {
 	 * turn.
 	 * @return The final {@link ChatResponse} after all tool calls (if any) are resolved.
 	 */
-	public ChatResponse internalCall(Prompt prompt, @Nullable ChatResponse previousChatResponse) {
+	private ChatResponse internalCall(Prompt prompt, @Nullable ChatResponse previousChatResponse) {
```
kuntal1461, line 545: Same story — only `call()` on line 262 uses this. Good to have both locked down.

Nit: same stale "called recursively" Javadoc as above.

kuntal1461 (approved):
## Code Review — Issue #6700: AnthropicChatModel exposes `internalStream` and `internalCall` publicly

**Verdict: Approve — no blocking findings.**

### Context

`AnthropicChatModel` was rewritten in March 2026 to use the official Anthropic Java SDK (`b0ce25aed`). That rewrite reintroduced `public` access on `internalCall` and `internalStream`, undoing the earlier privacy fix (`da3bec4a8`, Feb 2026). Every other chat model in the project (`OpenAiChatModel`, `OllamaChatModel`, `MistralAiChatModel`, `DeepSeekChatModel`, `GoogleGenAiChatModel`, `BedrockProxyChatModel`) already declares these methods `private`, and the upgrade notes explicitly state:

> All `internalCall` and `internalStream` methods in model classes have been changed to `private`.

This PR is the minimal, correct fix to restore that contract for `AnthropicChatModel`.

### Call-site analysis

A full-codebase grep confirms:
- `internalCall` is called **only** from `call()` (line 262, same class).
- `internalStream` is called **only** from `stream()` (line 283, same class).
- No external callers, no test callers, no reflection usage.
- The class is `final`, so no subclass extension concerns.
- Neither method appears in `ChatModel` or `StreamingChatModel`, so there is no interface override.

### Testing

Ran locally:

```
./mvnw -pl models/spring-ai-anthropic test
Tests run: 124, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS
```

### Minor observation (pre-existing, does not block)

The Javadoc on both methods still says *"This method is called recursively to support multi-turn tool calling."* That description became stale when `a48dc60af` ("Remove internal tool execution from AnthropicChatModel") delegated recursive tool-call handling to `ToolCallingManager`. The PR does not need to fix this, but a follow-up Javadoc polish would improve accuracy.

### Commit message

Follows the project convention correctly: uppercase verb, no `fix:` prefix, no PR number in the title, wrapped description, `Fixes #6700`, and `Signed-off-by` for DCO.

**No blocking code-review findings. Ready to merge after maintainer triage.**

kuntal1461:
Looks good to me. When the Anthropic module was rewritten to use the official Java SDK (`b0ce25aed`), these two methods accidentally came back as `public`, undoing the earlier fix from `da3bec4a8`. Every other chat model already has them private and the upgrade notes say they should be, so this is clearly the right thing to do.

Checked both call sites — nothing outside this class calls either method, no reflection usage, no interface override. Class is `final` so subclassing isn't a concern either.

Ran `./mvnw -pl models/spring-ai-anthropic test` locally — 124 tests, all green.

One pre-existing nit on the Javadoc: both methods still mention "called recursively" which stopped being true after `a48dc60af`. Not a blocker, just worth tidying up at some point.

PR #6733, opened 2026-08-04
Support audio streaming in OpenAiAudioSpeechModel

tzolov:
- The stream() used to buffer the full response and wrap it in a single Flux element.
- Now the request stream_format=audio and emit chunks as they arrive from the OpenAI SDK's InputStream instead.
- Tests and reference docs updated.
- Instructions how to bridge the Flux into binary input stream are provided along with demo IT.

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioSpeechModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioSpeechModel.java
@@ -18,12 +18,15 @@ package org.springframework.ai.openai;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InterruptedIOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 
 import com.openai.client.OpenAIClient;
 import com.openai.core.http.Headers;
+import com.openai.core.http.HttpResponse;
 import com.openai.models.audio.speech.SpeechCreateParams;
 import com.openai.models.audio.speech.SpeechModel;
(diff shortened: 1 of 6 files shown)


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioSpeechModel.java
```diff
@@ -100,20 +107,105 @@ public byte[] call(String text) {
...
+		catch (InterruptedIOException e) {
+			// The blocking read was interrupted because the downstream subscriber
+			// cancelled (e.g. a consumer stopped reading early). This is a normal
+			// shutdown path, not a failure, so complete quietly instead of raising
+			// sink.error(), which would otherwise surface as a spurious "dropped
+			// error" once nothing is left to receive it.
+			Thread.currentThread().interrupt();
+			sink.complete();
+			return state;
+		}
```
dimitarproynov: There is an issue here: java.net.SocketTimeoutException extends InterruptedIOException

A legitimate network outage that causes java.net.SocketTimeoutException will be reported as sink.complete(), rather than sink.error() to the caller.

Maybe, even, add a test case for SocketTimeoutException specifically

dimitarproynov (changes requested):

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiAudioSpeechModel.java
tzolov: Thanks @dimitarproynov . Actually there is a bigger issue in detecting explicit cancelation from an error state. Will submit a commit to harden those conditions

PR #6734, opened 2026-08-04
Harden ToolCallLimits ITs against parallel tool calls

tzolov:
ChatClient's .content() only returns the first Generation, and when the model batches parallel calls to the same tool, the first (successful) call sorts before the limit-breach notice, hiding it from any single-result read.
Assert against the full chatResponse().getResults() instead, so the tests validate the actual enforcement regardless of whether the model batches or sequences its tool calls.


ilayaperumalg (approved):

tzolov:
Replaced by https://github.com/spring-projects/spring-ai/pull/6734

PR #6737, opened 2026-08-05
Fix ToolCallingAdvisor streaming loop dropping doBeforeStream mutations

kezhenxu94:
internalStream() captured `toolCallingChatOptions` from the incoming request before calling the doBeforeStream() hook, then threaded that stale, pre-mutation reference all the way through
streamWithToolCallResponses() and handleToolCallRecursion() into the actual toolCallingManager.executeToolCalls(...) call - on every round of the streaming tool-calling loop, not just the first. Any ToolCallingChatOptions mutation a subclass hook applied in doBeforeStream() (extra tool callbacks, toolContext entries) was therefore invisible at tool-execution time, even though that same mutated options object was what actually got sent to the chat model.

This silently breaks ToolSearchToolCallingAdvisor: its prepareIteration() (invoked from its doBeforeStream() override) injects a synthetic toolSearchTool ToolCallback into the options for the current iteration so the model can search the indexed tool set. Because that mutation never reached executeToolCalls(), the model's call to toolSearchTool couldn't be resolved from the (stale) options and fell through to the configured ToolCallbackResolver - which has no way to resolve a callback that only ever existed on the per-iteration mutated options, since it isn't a registered bean. The resolver's "unavailable" fallback response is plain text, not the JSON tool-name array extractToolNameReferences() expects, so it throws a StreamReadException and the whole streamed chat response fails with an Aggregation Error whenever the model decides to call toolSearchTool.

Fix: derive the options passed into streamWithToolCallResponses() from finalRequest (the request as mutated by doBeforeStream()) instead of the pre-mutation snapshot, so tool execution and the model call agree on what tools/context are available for each iteration.

Added streamToolExecutionSeesOptionsMutatedByDoBeforeStreamHook to ToolCallingAdvisorTests, which reproduces the bug with a minimal doBeforeStream-mutating subclass (independent of the tool-search-advisor module) and asserts the Prompt passed to executeToolCalls carries the callback injected by the hook.

Signed-off-by: kezhenxu94 <kezhenxu94@apache.org>

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ToolCallingAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/ToolCallingAdvisor.java
@@ -289,8 +289,18 @@ public class ToolCallingAdvisor implements CallAdvisor, StreamAdvisor, ToolAdvis
 
 			Flux<ChatClientResponse> responseFlux = chainCopy.nextStream(processedRequest);
 
+			// Use the options doBeforeStream() produced on finalRequest - not the
+			// pre-doBeforeStream toolCallingChatOptions captured above - so that tool
+			// callbacks/context a subclass hook injects for this iteration (e.g. a
+			// dynamically resolved tool that only exists on the mutated options, not via
+			// any globally registered ToolCallbackResolver) are visible when the tool
+			// calls in the model's response are actually executed below.
+			ToolCallingChatOptions processedOptions = (ToolCallingChatOptions) finalRequest.prompt().getOptions();
+			Assert.notNull(processedOptions,
+					"redundant check that should never fail (doBeforeStream must not change the options type away from ToolCallingChatOptions), but here to help NullAway");
+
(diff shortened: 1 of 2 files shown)


kezhenxu94:
Hi @tzolov , please let me know if this fix makes sense, I get an error for the ToolSearchTool after upgrading to latest snapshot and suspect that was brought by https://github.com/spring-projects/spring-ai/commit/029af3796e0289c8e1a1bd73f5a4a1463f7469fb#diff-23bfc7a39d7cf73c6336086b090aae892e9f0c7cb3bab8bc918736d303b890a0

kezhenxu94:
@tzolov thank you for rerunning the tests, just verified locally it passed and thought it was a flaky test 😄

tzolov (approved):
Thanks @kezhenxu94 , I added some additional updates to handle the non-streaming case as well

kezhenxu94:
> Thanks @kezhenxu94 , I added some additional updates to handle the non-streaming case as well

Thank you for review and revise the change!

PR #6759, opened 2026-08-07
Fix `OpenAiAudioTranscriptionModel` leaking the stream on cancellation

subhashpolisetti:
`OpenAiAudioTranscriptionModel.stream()` creates the SDK `AsyncStreamResponse` inside `Flux.create`
but discards the handle. Cancelling the returned `Flux` then stops the downstream signals without
closing the SDK stream, so the underlying HTTP response stays open until the server sends EOF or the
request times out.

Retain the response and close it from `FluxSink#onDispose`, releasing it on cancellation, error, or
completion. Same defect as #6654 on the chat path, here on the transcription path.

## Testing

Added `streamClosesResponseWhenSubscriptionIsCancelled`, which fails without the fix (the discarded
response is never closed) and passes with it; existing transcription streaming tests are unaffected.
It needs no credentials  a mock async client returns an open `AsyncStreamResponse`, the subscription
is disposed, and the response is asserted closed.

`./mvnw -pl models/spring-ai-openai clean test` passes (203 tests).


tzolov (approved):
Thanks for addressing this issue @subhashpolisetti 
Approach looks good to me

PR #6779, opened 2026-08-10
Make jtokkit optional in spring-ai-commons

### LINKED ISSUE #6778: Make jtokkit optional for spring-ai-commons consumers
## Problem

spring-ai-commons declares jtokkit as a normal compile dependency. Because commons sits beneath broad parts of Spring AI, applications that never tokenize text still receive the roughly 3.1 MiB tokenizer JAR. A concrete example is an MCP-only Spring AI server: it uses the MCP infrastructure but no TokenTextSplitter or token-count batching and nevertheless packages jtokkit.

The classes that actually use jtokkit are limited and identifiable. Keeping the dependency mandatory at the commons boundary makes unrelated consumers pay the binary size and supply-chain cost.

## Proposed change

- Mark jtokkit optional in spring-ai-commons.
- Keep modules that expose tokenizer-backed behavior self-contained by declaring jtokkit directly. spring-ai-model needs it for TokenCountBatchingStrategy; spring-ai-client-chat already declares it directly.
- Preserve behavior for users of the tokenizer features while allowing unrelated consumers to omit the JAR naturally.

## Compatibility

This is not a request to remove tokenizer support. Modules/classes that directly require jtokkit continue to carry a direct dependency. The change only prevents the broad commons module from exporting it transitively to consumers that do not need those APIs.

I have a focused two-POM change and have built and tested the affected reactor modules on OpenJDK 26.

gregjotau:
Closes #6778.

Marks jtokkit optional at the broad spring-ai-commons boundary and adds an explicit dependency to spring-ai-model, which directly uses it in TokenCountBatchingStrategy. spring-ai-client-chat already declares jtokkit directly.

This preserves tokenizer behavior for the modules that require it while letting unrelated consumers such as MCP-only servers omit the roughly 3.1 MiB tokenizer JAR.

Validation:
- affected reactor package build on OpenJDK 26
- spring-ai-commons and spring-ai-model test reactor: 834 tests, 0 failures/errors

### DIFF
--- a/spring-ai-commons/pom.xml
+++ b/spring-ai-commons/pom.xml
@@ -59,6 +59,7 @@
 			<groupId>com.knuddels</groupId>
 			<artifactId>jtokkit</artifactId>
 			<version>${jtokkit.version}</version>
+			<optional>true</optional>
 		</dependency>
 
 		<!-- test dependencies -->
--- a/spring-ai-model/pom.xml
+++ b/spring-ai-model/pom.xml
@@ -20,6 +20,13 @@
 			<version>${project.parent.version}</version>
 		</dependency>
 
+		<!-- Required by TokenCountBatchingStrategy; optional in spring-ai-commons. -->
+		<dependency>
+			<groupId>com.knuddels</groupId>
+			<artifactId>jtokkit</artifactId>
+			<version>${jtokkit.version}</version>
+		</dependency>
+
 		<dependency>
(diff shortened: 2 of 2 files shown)


gregjotau:
Closing because we are removing the Spring AI/MCP integration that exposed this transitive dependency in our application. The optionality improvement is still valid in isolation, but we no longer have a concrete downstream need and do not want to consume maintainer attention without one.

PR #6784, opened 2026-08-11
Reserve Redis chat memory timestamps atomically

dimitarproynov:
RedisChatMemoryRepository.add() reserved a single timestamp slot per call in Redis and then derived the remaining keys for a message batch locally by incrementing an in-memory counter. Only the first key in a batch was ever coordinated through Redis, so two add() calls for the same conversation could compute overlapping timestamps, and one call would silently overwrite messages already written by the other.

Change the counter reservation to atomically claim a whole block of timestamps sized to the batch in a single Lua script (via INCRBY), so every message in a batch gets a slot that no other caller, even a concurrent one, can also claim.


ilayaperumalg (approved):

ilayaperumalg:
Rebased and merged via c988e72ab7282bcc352b2155e8cb27ff4b5e0fda

PR #6785, opened 2026-08-11
Fail clearly on out-of-range OpenAI usage counts

dimitarproynov:
getDefaultUsage() converted server-supplied prompt/completion/total token counts from long to int via Math.toIntExact, which throws a bare ArithmeticException("integer overflow") with no indication of which field or value caused it.

Wrap the conversion so an out-of-range count fails with a clear IllegalStateException naming the field and the offending value. Token usage feeds cost tracking and cumulative usage calculations downstream, so an implausible count should fail loudly rather than being silently truncated into a smaller-but-still-wrong number.


deshpande-pavan:
Thanks for the fix! 

I noticed OpenAiEmbeddingModel.java has the same pattern — getDefaultUsage() at line 282–284 also calls Math.toIntExact() on promptTokens and totalTokens without any overflow handling. 

Would it make sense to apply the same toIntTokenCount() helper there too for consistency?

PR #6788, opened 2026-08-12
Normalize advisor observation contextual names

### LINKED ISSUE #6750: `DefaultAdvisorObservationConvention` produces a span name with stray spaces
### Bug description

The span produced for `ToolCallingAdvisor` is named `tool _calling `: there is a
space before `_calling` and a trailing space at the end. Tool calling is on by
default, so this appears in every trace where the model calls a tool.

### Environment

Spring AI 2.0.0, Spring Boot 4.1.0, Java 21. Anthropic chat model with tool
callbacks supplied by an MCP client. Traces exported over OTLP to the
OpenTelemetry Collector. The same code is present on `main`.

### Steps to reproduce

1. Build a `ChatClient` with tool callbacks and tracing to an OTLP endpoint.
2. Send a prompt that makes the model call a tool.
3. Read the span names.

From the Collector's debug exporter, piped through `cat -A` so that `$` marks
end of line:

    Name           : tool _calling $
    Name           : spring_ai chat_client$
    Name           : call$

### Expected behavior

A span name with no embedded or trailing whitespace, for example `tool_calling`.

### Minimal Complete Reproducible example

https://github.com/kycasdzxc/spring-ai-otel-example

`docker compose up -d`, run the app, then ask a question that triggers a tool
call. The malformed name appears in the collector output and in Tempo.

The minimal condition is narrower than that project: any advisor whose
`getName()` returns a string containing spaces reproduces it. `ToolCallingAdvisor`
is built in, so no custom advisor is needed.

### Root cause

`DefaultAdvisorObservationConvention#getContextualName`:

```java
return ParsingUtils.reConcatenateCamelCase(context.getAdvisorName(), "_")
    .replace("_around_advisor", "")
    .replace("_advisor", "");
```

`reConcatenateCamelCase` splits on camelCase boundaries, so it expects an
identifier such as `ToolCallingAdvisor`. `ToolCallingAdvisor#getName()` returns
the display form `"Tool Calling Advisor"`, and splitting that leaves the original
spaces attached to each token:

    "Tool Calling Advisor"
      -> ["Tool ", "Calling ", "Advisor"]
      -> "tool _calling _advisor"
      -> replace("_advisor", "") -> "tool _calling "

Advisors whose name contains no spaces are unaffected: `call` stays `call`.

Either normalising whitespace in `getContextualName` or having advisors supply an
identifier-shaped name would fix it.

AzazelSensei:
## Summary
- Strip whitespace from advisor names in `DefaultAdvisorObservationConvention.getContextualName()` before camel-case concatenation
- Prevents stray spaces in span names such as `tool _calling ` for `Tool Calling Advisor`
- Adds a focused unit test covering the `Tool Calling Advisor` case

Fixes #6750

## Test plan
- [x] `./mvnw -pl spring-ai-client-chat -am -Dtest=DefaultAdvisorObservationConventionTests -Dsurefire.failIfNoSpecifiedTests=false test`
- [ ] Verify OTLP / Tempo span name is `tool_calling` with tool callbacks enabled

### DIFF
--- a/spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/advisor/observation/DefaultAdvisorObservationConventionTests.java
+++ b/spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/advisor/observation/DefaultAdvisorObservationConventionTests.java
@@ -64,6 +64,25 @@ class DefaultAdvisorObservationConventionTests {
 	}
 
 	@Test
+	void contextualNameNormalizesWhitespaceForStructuredOutputValidationAdvisor() {
+		AdvisorObservationContext observationContext = AdvisorObservationContext.builder()
+			.chatClientRequest(ChatClientRequest.builder().prompt(new Prompt("Hello")).build())
+			.advisorName("Structured Output Validation Advisor")
+			.build();
+		assertThat(this.observationConvention.getContextualName(observationContext))
+			.isEqualTo("structured_output_validation");
+	}
+
+	@Test
+	void contextualNameStripsAroundAdvisorSuffix() {
+		AdvisorObservationContext observationContext = AdvisorObservationContext.builder()
+			.chatClientRequest(ChatClientRequest.builder().prompt(new Prompt("Hello")).build())
+			.advisorName("MyAroundAdvisor")
+			.build();
+		assertThat(this.observationConvention.getContextualName(observationContext)).isEqualTo("my");
+	}
+
+	@Test
 	void supportsAdvisorObservationContext() {
 		AdvisorObservationContext observationContext = AdvisorObservationContext.builder()
 			.chatClientRequest(ChatClientRequest.builder().prompt(new Prompt("Hello")).build())
--- a/spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/advisor/observation/DefaultAdvisorObservationConventionTests.java
+++ b/spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/advisor/observation/DefaultAdvisorObservationConventionTests.java
@@ -55,6 +55,15 @@ class DefaultAdvisorObservationConventionTests {
 	}
 
 	@Test
+	void contextualNameNormalizesWhitespaceInAdvisorName() {
+		AdvisorObservationContext observationContext = AdvisorObservationContext.builder()
+			.chatClientRequest(ChatClientRequest.builder().prompt(new Prompt("Hello")).build())
+			.advisorName("Tool Calling Advisor")
+			.build();
+		assertThat(this.observationConvention.getContextualName(observationContext)).isEqualTo("tool_calling");
+	}
+
+	@Test
 	void supportsAdvisorObservationContext() {
 		AdvisorObservationContext observationContext = AdvisorObservationContext.builder()
 			.chatClientRequest(ChatClientRequest.builder().prompt(new Prompt("Hello")).build())
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/observation/DefaultAdvisorObservationConvention.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/observation/DefaultAdvisorObservationConvention.java
@@ -26,6 +26,7 @@ import org.springframework.ai.observation.conventions.AiProvider;
 import org.springframework.ai.observation.conventions.SpringAiKind;
 import org.springframework.ai.util.ParsingUtils;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * Default implementation of the {@link AdvisorObservationConvention}.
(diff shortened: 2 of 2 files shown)


kycasdzxc:
Ran the reported input through `ParsingUtils.CAMEL_CASE` to check the surrounding behaviour, and two things came up that might be worth folding in.

**`StructuredOutputValidationAdvisor` hits the same bug and isn't covered.**

It is the other advisor in the codebase that overrides `getName()` with a display string:

```java
// StructuredOutputValidationAdvisor.java:109
return "Structured Output Validation Advisor";
```

On `main` that produces:

```
splitCamelCaseToLower -> ["structured ", "output ", "validation ", "advisor"]
join("_")             -> "structured _output _validation _advisor"
.replace("_advisor")  -> "structured _output _validation "
```

This PR fixes it correctly (`structured_output_validation`), so this is only a suggestion to add it to the new test. It is the second of the two real instances, and pinning both makes the regression boundary explicit.

**The `_around_advisor` branch is still untested.**

The only contextual-name test on `main` is `contextualName()` with `"MyName"` -> `"my_name"`, which reaches neither `replace`. The test added here is the first to exercise `_advisor`. Since `trimAllWhitespace` changes what reaches both, pinning `_around_advisor` as well would be cheap.

One limitation worth recording, though I do not think it needs to block this: `trimAllWhitespace` only recovers an identifier when every word is capitalized. `"Simple Logger advisor"` becomes `simple_loggeradvisor`: no stray spaces, so the reported bug is fixed, but the suffix no longer strips. This is not a regression, since `main` gives `simple _logger advisor` for the same input, equally unstripped. It is the same assumption in a narrower form, in that `getContextualName()` expects `Advisor#getName()` to be identifier-shaped. That is the open question from #6750, and this change does not settle it either way, so it is likely a maintainer call.

spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/advisor/observation/DefaultAdvisorObservationConventionTests.java
```diff
@@ -54,6 +54,15 @@ void contextualName() {
...
 
+	@Test
+	void contextualNameNormalizesWhitespaceInAdvisorName() {
+		AdvisorObservationContext observationContext = AdvisorObservationContext.builder()
+			.chatClientRequest(ChatClientRequest.builder().prompt(new Prompt("Hello")).build())
+			.advisorName("Tool Calling Advisor")
+			.build();
+		assertThat(this.observationConvention.getContextualName(observationContext)).isEqualTo("tool_calling");
+	}
+
```
kycasdzxc, line 65: Both cases are test-only, so this is just pinning behaviour the PR already fixes.

`StructuredOutputValidationAdvisor` is the other advisor that returns a display-form name, so it is the second real instance of the reported bug. The `_around_advisor` branch has no coverage today — the existing `contextualName()` test reaches neither `replace`, and `trimAllWhitespace` changes what arrives at both.

```suggestion

	@Test
	void contextualNameNormalizesWhitespaceForStructuredOutputValidationAdvisor() {
		AdvisorObservationContext observationContext = AdvisorObservationContext.builder()
			.chatClientRequest(ChatClientRequest.builder().prompt(new Prompt("Hello")).build())
			.advisorName("Structured Output Validation Advisor")
			.build();
		assertThat(this.observationConvention.getContextualName(observationContext))
			.isEqualTo("structured_output_validation");
	}

	@Test
	void contextualNameStripsAroundAdvisorSuffix() {
		AdvisorObservationContext observationContext = AdvisorObservationContext.builder()
			.chatClientRequest(ChatClientRequest.builder().prompt(new Prompt("Hello")).build())
			.advisorName("MyAroundAdvisor")
			.build();
		assertThat(this.observationConvention.getContextualName(observationContext)).isEqualTo("my");
	}
```

Reworded my earlier notes as suggestions so they're easier to apply. Test-only, and entirely up to you whether to take them.

kycasdzxc (commented):
Reworded my earlier notes as suggestions so they're easier to apply. Test-only, and entirely up to you whether to take them.

spring-ai-client-chat/src/test/java/org/springframework/ai/chat/client/advisor/observation/DefaultAdvisorObservationConventionTests.java
AzazelSensei: Thanks — added the two suggested tests.

AzazelSensei:
Thanks for the suggestions. Pushed the two extra tests (Structured Output Validation Advisor + AroundAdvisor suffix).

kycasdzxc:
@AzazelSensei Glad they were useful. Since the suggestions were applied manually,
the co-author trailer didn't get added automatically. Could you amend
the commit to include it?

```
Co-authored-by: kycasdzxc <105793579+kycasdzxc@users.noreply.github.com>
```

AzazelSensei:
Amended — added your Co-authored-by.

PR #6791, opened 2026-08-13
Support generic output schemas in async MCP tool providers

wxggzz:
## Summary

Generate MCP output schemas for parameterized reactive return types in both stateful and stateless async tool providers.

When an `@McpTool(generateOutputSchema = true)` method returns a type such as `Mono<List<Result>>`, the reactive payload is a `ParameterizedType`. Both async providers converted non-`Class<?>` payloads to `null` before calling `ClassUtils`, causing tool registration to fail with `IllegalArgumentException: Class must not be null`.

This change:

- preserves the existing primitive and simple-value filtering for class payloads;
- generates schemas from `ParameterizedType` payloads;
- covers both stateful and stateless async providers;
- verifies the generated array schema and the structured tool result.

This supersedes and extends #5920 by covering both async provider variants and adding end-to-end structured-content assertions.

## Testing

- `./mvnw -Dmaven.build.cache.enabled=false -pl mcp/mcp-annotations -am test` (1,378 tests, 0 failures, 2 skipped)
- `./mvnw clean package` (169 reactor modules, BUILD SUCCESS)

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/provider/tool/AsyncMcpToolProvider.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/provider/tool/AsyncMcpToolProvider.java
@@ -17,6 +17,7 @@
 package org.springframework.ai.mcp.annotation.provider.tool;
 
 import java.lang.reflect.Method;
+import java.lang.reflect.ParameterizedType;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


wxggzz:
Hi maintainers, gentle follow-up on this PR. It fixes output schema generation for parameterized reactive return types in both async MCP tool provider variants, with tests covering schema generation and structured content. The PR is currently mergeable, and I would appreciate a review when you have a chance. Happy to rebase or make any requested adjustments. Thanks!

Issue #6171, opened 2026-05-26
`spring-ai-starter-model-google-genai` incorrectly declares dependency on `spring-ai-google-genai-embedding`


jxblum:
As a user, if I am only using Google (Vertex AI) Generative AI for **chat**, and **not** using "embedding model(s)", then I **do not** require a dependency on `spring-ai-google-genai-embedding` (see [here](https://github.com/spring-projects/spring-ai/blob/v2.0.0-M7/starters/spring-ai-starter-model-google-genai/pom.xml#L74-L78)).

The only Spring AI Google dependency in `spring-ai-starter-model-google-genai` should be `spring-ai-google-genai` ([here](https://github.com/spring-projects/spring-ai/blob/v2.0.0-M7/starters/spring-ai-starter-model-google-genai/pom.xml#L51-L55)).

Conversely, there is (now) a separate starter for using Google (Vertex AI) GenAI for Embeddings, namely `spring-ai-starter-model-google-genai-embedding`, and this starter appropriately defines the Google GenAI Embedding dependency, [here](https://github.com/spring-projects/spring-ai/blob/v2.0.0-M7/starters/spring-ai-starter-model-google-genai-embedding/pom.xml#L51-L55).

jxblum:
Note, with only a dependency on:


```xml
<dependency>
    <groupId>org.springframework.ai</groupId>
    <artifactId>spring-ai-starter-model-google-genai</artifactId>
</dependency>
```

And declaring a test class similar to:

```java
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.NONE, properties = {
    "spring.ai.google.genai.api-key=mockApiKey",
    "spring.ai.google.genai.location=us-west1",
    "spring.ai.google.genai.project-id=mockGeminiProjectId",
})
@SuppressWarnings("unused")
public class ExampleIntegrationTests {
```

This can lead to the following exception on initialization:


```text
...
ava.lang.IllegalStateException: Failed to load ApplicationContext for [MergedContextConfiguration@652b3733 testClass = com.example.ExampleIntegrationTests
...
..
.
[...]

PR #6794, opened 2026-08-13
Guard `GoogleGenAiEmbeddingConnectionAutoConfiguration` with the model enablement property

hoyo1744:
`GoogleGenAiEmbeddingConnectionAutoConfiguration` is not guarded by the
model enablement property, so it is applied whenever the artifact is on
the classpath. When nothing is configured it falls back to Vertex AI mode
and fails the context:

```
java.lang.IllegalArgumentException: Google GenAI project-id must be set!
```

This means adding `spring-ai-starter-model-google-genai-embedding` breaks
startup for applications that do not use Google GenAI embeddings, and
`spring.ai.model.embedding.text=none` does not help because the condition
is never evaluated.

## Reproduce

1. Add `spring-ai-starter-model-google-genai-embedding`.
2. Set `spring.ai.model.embedding.text=none` (or leave everything unset).
3. Start the application.

## Relation to previous reports

The same symptom was reported in #6130 and closed as a duplicate of
#6171, which removed the embedding artifact from the chat starter. That
resolves the reported case, where the artifact was pulled in
unintentionally.

It does not cover this one. Here the embedding starter is added on
purpose, because the application does use Google GenAI embeddings in some
environments and selects another provider in others. Since the artifact
is wanted on the classpath, dependency removal is not an option and the
missing condition is the only thing left.

This is exactly what was pointed out in #6130:

> There is `@ConditionalOnProperty` annotation on
> `GoogleGenAiTextEmbeddingAutoConfiguration`, so it should disable it,
> but not on `GoogleGenAiEmbeddingConnectionAutoConfiguration`, which is
> not ideal because it can't be disabled like this for now.

Unlike #6156, which gated the auto-configuration on the presence of
credentials, this reuses the existing model enablement property so that
the connection and the model it feeds share one switch.

See #6130
See #6171

## Inconsistency

The sibling auto-configurations in the same module are guarded, the
connection one is not:

| Auto-configuration | Property condition |
| --- | --- |
| `GoogleGenAiChatAutoConfiguration` | `spring.ai.model.chat` |
| `GoogleGenAiTextEmbeddingAutoConfiguration` | `spring.ai.model.embedding.text` |
| `GoogleGenAiImageAutoConfiguration` | `spring.ai.model.image` |
| `GoogleGenAiEmbeddingConnectionAutoConfiguration` | none |

## Change

Applies the same condition already used by
`GoogleGenAiTextEmbeddingAutoConfiguration`, so the connection and the
model it feeds share a single switch. `matchIfMissing = true` keeps the
current default-on behaviour, so applications that configure nothing are
unaffected.

Adds `GoogleGenAiEmbeddingConnectionAutoConfigurationTests` covering
explicit enable, default enable, `none`, another provider, and that the
context starts when nothing is configured.

## Note

`GoogleGenAiImageConnectionAutoConfiguration` has the same missing
condition. I left it out because its bean method does not fail fast the
same way and I could not confirm a broken startup for it. Happy to
extend this PR if you would like both aligned.

This may also be a small piece of the broader auto-configuration
activation discussion raised in #6130 and #5989.

### DIFF
--- a/auto-configurations/models/spring-ai-autoconfigure-model-google-genai/src/main/java/org/springframework/ai/model/google/genai/autoconfigure/embedding/GoogleGenAiEmbeddingConnectionAutoConfiguration.java
+++ b/auto-configurations/models/spring-ai-autoconfigure-model-google-genai/src/main/java/org/springframework/ai/model/google/genai/autoconfigure/embedding/GoogleGenAiEmbeddingConnectionAutoConfiguration.java
@@ -22,9 +22,12 @@ import com.google.auth.oauth2.GoogleCredentials;
 import com.google.genai.Client;
 
 import org.springframework.ai.google.genai.embedding.GoogleGenAiEmbeddingConnectionDetails;
+import org.springframework.ai.model.SpringAIModelProperties;
+import org.springframework.ai.model.SpringAIModels;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.util.Assert;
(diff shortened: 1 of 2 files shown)


hoyo1744:
I opened #6973 to cover the general case, since the same missing condition affects
`GoogleGenAiImageConnectionAutoConfiguration` and
`VertexAiEmbeddingConnectionAutoConfiguration` as well. This PR stays the narrow
fix for the embedding connection.

Two things I was able to confirm on current `main` since opening this:

The note at the bottom of the description is now resolved.
`GoogleGenAiImageConnectionAutoConfiguration` does fail fast after all. With
nothing configured it reaches the final `else` in
`googleGenAiImageConnectionDetails` and throws
`IllegalStateException: Incomplete Google GenAI configuration`, so it breaks
startup the same way. It takes the identical single-property fix, and I am happy
to fold it into this PR.

`VertexAiEmbeddingConnectionAutoConfiguration` is the one that does not, because
`VertexAiEmbeddingConnectionDetails` is consumed by both
`VertexAiTextEmbeddingAutoConfiguration` and
`VertexAiMultiModalEmbeddingAutoConfiguration`. Guarding it on
`spring.ai.model.embedding.text` alone would break multimodal-only applications,
so it needs a condition over both properties. That is the part I left for #6973.

No rush on my side, but let me know if you would rather see the three aligned in
one PR, or a shared mechanism as discussed in #6130. I can rework this either way.

Issue #6770, opened 2026-08-07
`MongoChatMemoryRepository.saveAll()` can lose conversation history on insert failure


ricardohsmello:
### **Bug Description**
`MongoChatMemoryRepository.saveAll()` updates a conversation by deleting the existing messages and then inserting the new message list:

```java
deleteByConversationId(conversationId);
this.mongoTemplate.insert(conversations, Conversation.class);
``` 
These operations are executed independently, without a transaction.

If the delete succeeds and the subsequent insert fails, the existing messages have already been removed, leaving the conversation without its previously persisted history.

### **Environment**
Spring AI: main (2.0.1-SNAPSHOT)
Module: spring-ai-model-chat-memory-repository-mongodb
MongoDB: 8.x
Spring Data MongoDB / MongoTemplate

### **Steps to reproduce**

1. Persist an initial conversation using MongoChatMemoryRepository.
2. Call saveAll() again for the same conversation.
3. Make the `MongoTemplate.insert(...) `operation fail after `deleteByConversationId()` succeeds.
4. Read the conversation again. 

The original conversation is no longer available because the delete operation was completed before the insert failed.

### **Expected behavior**
Updating a persisted conversation should not remove the existing conversation if the replacement messages cannot be saved.

The delete and insert operations should be atomic, so that if the insert fails, the previous conversation remains available.

`JdbcChatMemoryRepository` already provides this behavior by executing the equivalent delete-and-insert flow inside a `TransactionTemplate`, allowing the delete to be rolled back when the insert fails.

### **Minimal Complete Reproducible example**
The following test, added to `MongoChatMemoryRepositoryIT`, reproduces the issue by forcing the insert operation to fail after the existing conversation has already been deleted:

```java
@Test
void shouldPreserveExistingMessagesWhenInsertFails() {
	var conversationId = UUID.randomUUID().toString();
	var existingMessages = List.<Message>of(
			new UserMessage("First message"),
			new AssistantMessage("Second message"));

	this.chatMemoryRepository.saveAll(conversationId, existingMessages);

	var failingMongoTemplate = spy(this.mongoTemplate);
	doThrow(new DataIntegrityViolationException("Insert failed"))
		.when(failingMongoTemplate)
		.insert(anyCollection(), eq(Conversation.class));

	var failingRepository = MongoChatMemoryRepository.builder()
		.mongoTemplate(failingMongoTemplate)
		.build();

	assertThatThrownBy(() ->
[...]

PR #6798, opened 2026-08-16
Make `Neo4jChatMemoryRepository.saveAll` atomic

dlwldn30:
## Problem

`Neo4jChatMemoryRepository.saveAll()` replaces a conversation by calling
`deleteByConversationId()` first. That method opens its own session, runs the
two delete statements in a transaction and commits. Only then does `saveAll()`
open a second session to write the replacement messages.

If that write fails, the delete is already committed, so the conversation is
left with neither its previous messages nor the new ones.

`JdbcChatMemoryRepository` already runs the equivalent delete-and-insert flow
inside a `TransactionTemplate`. The same defect on the MongoDB repository is
tracked in #6770.

## Solution

Extract the delete statements into
`deleteConversation(SimpleQueryRunner, String)` and run them inside the same
write transaction as the inserts. `deleteByConversationId()` keeps its own
transaction and reuses that method, so its behaviour is unchanged.

Both statements are `MATCH ... DETACH DELETE`, and the inserts derive
`msg.idx` from a count taken inside the transaction, so the combined unit of
work stays safe under the driver's retry behaviour.

## Testing

`Neo4jChatMemoryRepositoryIT.saveAllKeepsTheExistingConversationWhenAWriteFails`
saves two messages, then attempts a replacement whose second message cannot be
serialized — the same `Optional`-typed metadata that the existing
`saveAssistantMessageWithOptionalMetadataFails` relies on.

Reverting only the production change and keeping the test shows the
conversation is wiped:

    Expecting actual:
      []
    to contain exactly (and in same order):
      ["First message", "Second message"]

With the fix the two original messages survive the failed write.

    ./mvnw -Dmaven.build.cache.enabled=false -Pintegration-tests \
      -pl memory-repositories/spring-ai-model-chat-memory-repository-neo4j verify

32 tests run, 0 failures, 0 errors, 0 skips. The downstream
`spring-ai-autoconfigure-model-chat-memory-repository-neo4j` module passes too
(3 tests). `spring-javaformat:apply` and `checkstyle:check` reported no
violations.

## Follow-up (not in this PR)

`deleteByConversationId()` is the only method here still using an unmanaged
transaction, so unlike the others it does not get the driver's retry behaviour
on transient failures. That is unrelated to atomicity, so I left it out — happy
to follow up separately if you would like it aligned with the rest of the class.

---

Contributed on behalf of [Goatshave](https://github.com/Goatshave).

### DIFF
--- a/memory-repositories/spring-ai-model-chat-memory-repository-neo4j/src/main/java/org/springframework/ai/chat/memory/repository/neo4j/Neo4jChatMemoryRepository.java
+++ b/memory-repositories/spring-ai-model-chat-memory-repository-neo4j/src/main/java/org/springframework/ai/chat/memory/repository/neo4j/Neo4jChatMemoryRepository.java
@@ -26,6 +26,7 @@ import java.util.UUID;
 import java.util.stream.Collectors;
 
 import org.neo4j.driver.Session;
+import org.neo4j.driver.SimpleQueryRunner;
 import org.neo4j.driver.Transaction;
 import org.neo4j.driver.TransactionContext;
 
@@ -124,12 +125,12 @@ public final class Neo4jChatMemoryRepository implements ChatMemoryRepository {
 
 	@Override
 	public void saveAll(String conversationId, List<Message> messages) {
-		// First delete existing messages for this conversation
-		deleteByConversationId(conversationId);
-
-		// Then add the new messages
+		// Replace the conversation in a single transaction. Deleting through a
+		// separate, already committed transaction loses the previous messages when
+		// the write that follows it fails.
 		try (Session s = this.config.getDriver().session()) {
 			s.executeWriteWithoutResult(tx -> {
+				deleteConversation(tx, conversationId);
 				for (Message m : messages) {
 					addMessageToTransaction(tx, conversationId, m);
 				}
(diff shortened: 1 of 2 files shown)


dlwldn30:
Hi @sobychacko, sorry for the direct ping — you made the last change to this file, so you are probably the right person to ask.

Open since August 16, no triage yet. `Neo4jChatMemoryRepository.saveAll()` calls `deleteByConversationId()` first, which commits its own transaction before the replacement messages are written. If that write fails, the conversation is left with neither the old messages nor the new ones. `JdbcChatMemoryRepository` already runs the same delete-and-insert inside a `TransactionTemplate`, and #6770 tracks the same defect on MongoDB.

`Neo4jChatMemoryRepositoryIT.saveAllKeepsTheExistingConversationWhenAWriteFails` leaves an empty conversation if the production change alone is reverted.

It still merges cleanly, though the branch is 16 commits behind `main` now — happy to rebase if that helps. Thanks for your work on the project.

dlwldn30:
Hi @sobychacko, sorry for another ping — it's been about three weeks since my last comment with no triage yet. `Neo4jChatMemoryRepository.saveAll()` still isn't wrapped in a transaction the way `JdbcChatMemoryRepository` is, so a failed write after `deleteByConversationId()` still leaves an empty conversation. The branch is still clean against `main`. No rush, just wanted to keep this on your radar. Thanks again for your time.

PR #6813, opened 2026-08-18
Fix `FilterExpressionTextParser` shared error state

### LINKED ISSUE #6807: `FilterExpressionTextParser` shares mutable error state across instances
**Bug description**

  `FilterExpressionTextParser` assigns the same mutable `DescriptiveErrorListener.INSTANCE` to every parser instance.

  Each call to `parse()` clears, modifies, and reads the listener's shared `errorMessages` list. Concurrent parsing through separate parser instances can therefore interfere with diagnostic collection performed by another invocation.

  `CopyOnWriteArrayList` makes individual list operations thread-safe, but it does not isolate the compound `clear → parse → join` sequence.

  If the maintainers agree that this is a bug, I would be happy to work on the fix and submit a pull request with regression tests.

  **Environment**

  - Spring AI: `2.0.1-SNAPSHOT`
  - Tested artifact: `spring-ai-vector-store-2.0.1-20260808.133500-105.jar`
  - Java: 17.0.17
  - ANTLR runtime: 4.13.2
  - Vector store: none required

  **Steps to reproduce**

  1. Create two separate `FilterExpressionTextParser` instances.
  2. Retrieve the `errorListener` assigned to each instance.
  3. Observe that both parser instances use the same listener.
  4. Run invalid parse operations concurrently and observe that both calls operate
     on the same `errorMessages` collection.

  Relevant implementation:

  ```java
  public FilterExpressionTextParser(ANTLRErrorStrategy handler) {
        this.errorListener = DescriptiveErrorListener.INSTANCE;
        this.errorHandler = handler;
  }
  ```

  ```java
  parser.removeErrorListeners();
  this.errorListener.errorMessages.clear();
  parser.addErrorListener(this.errorListener);
  ```

  ```java
  var msg = String.join("", this.errorListener.errorMessages);
  ```

  **Expected behavior**

  Each parser invocation, or at least each parser instance, should collect its syntax error messages independently. Concurrent parsing should not clear or modify diagnostics belonging to another invocation.

  The following JUnit test fails because both parser instances receive the same
  mutable error listener:

  ```java
  import java.lang.reflect.Field;

  class FilterExpressionTextParserConcurrencyTests {

        @Test
        void parserInstancesShouldNotShareMutableErrorListener() throws Exception {
                var firstParser = new FilterExpressionTextParser();
                var secondParser = new FilterExpressionTextParser();

                Field listenerField =
                                FilterExpressionTextParser.class.getDeclaredField("errorListener");
                listenerField.setAccessible(true);

                Object firstListener = listenerField.get(firstParser);
                Object secondListener = listenerField.get(secondParser);

                assertThat(firstListener).isNotSameAs(secondListener);
        }

  }
  ```

  Actual result:

  ```text
  Expecting actual:
    org.springframework.ai.vectorstore.filter.FilterExpressionTextParser$DescriptiveErrorListener@...
  not to be the same as:
    org.springframework.ai.vectorstore.filter.FilterExpressionTextParser$DescriptiveErrorListener@...
  ```

kdkrkwhr:
## Overview

`FilterExpressionTextParser` used a shared singleton `DescriptiveErrorListener`,
so every parser instance wrote into the same mutable `errorMessages` list. A
failed parse on one instance could leak or clear error messages observed by
another instance, including under concurrent use.

This scopes the error listener to the parser instance so each one owns its own
error state.

Fixes #6807

## Changes

- Construct a per-instance `DescriptiveErrorListener` instead of the shared `INSTANCE`.
- Remove the unused `DescriptiveErrorListener.INSTANCE` constant.
- Add a `getErrorMessages()` accessor (test-only) to assert instance isolation.
- Add regression tests:
  - `testErrorStateIsNotSharedAcrossInstances`
  - `testParallelParsingDoesNotMixErrorState` (8-thread concurrent parse)

## Verification

`FilterExpressionTextParserTests` passes: 17/17, 0 Checkstyle violations.

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
@@ -109,8 +109,6 @@ public class FilterExpressionTextParser {
 
 	private static final String WHERE_PREFIX = "WHERE";
 
-	private final DescriptiveErrorListener errorListener;
-
 	private final ANTLRErrorStrategy errorHandler;
 
 	private final Map<String, Filter.Expression> cache = new ConcurrentHashMap<>();
@@ -120,7 +118,6 @@ public class FilterExpressionTextParser {
 	}
 
 	public FilterExpressionTextParser(ANTLRErrorStrategy handler) {
-		this.errorListener = DescriptiveErrorListener.INSTANCE;
 		this.errorHandler = handler;
 	}
 
@@ -141,9 +138,11 @@ public class FilterExpressionTextParser {
 		var tokens = new CommonTokenStream(lexer);
 		var parser = new FiltersParser(tokens);
 
+		// Per-invocation error listener: each parse() call owns its own error state,
+		// so concurrent calls on the same instance do not share mutable error state.
+		var errorListener = new DescriptiveErrorListener();
 		parser.removeErrorListeners();
-		this.errorListener.errorMessages.clear();
-		parser.addErrorListener(this.errorListener);
+		parser.addErrorListener(errorListener);
 
 		if (this.errorHandler != null) {
 			parser.setErrorHandler(this.errorHandler);
@@ -157,7 +156,7 @@ public class FilterExpressionTextParser {
 			return filterExpression;
 		}
 		catch (ParseCancellationException e) {
-			var msg = String.join("", this.errorListener.errorMessages);
+			var msg = String.join("", errorListener.errorMessages);
 			var rootCause = NestedExceptionUtils.getRootCause(e);
 			throw new FilterExpressionParseException(msg, rootCause);
 		}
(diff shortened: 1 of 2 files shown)


qixiangme:
Regarding @ViniciusKoiti's comment on #6807, since the listener is now per-instance, would concurrent `parse()` calls on the same `FilterExpressionTextParser` instance still share the same error state?

Would a per-invocation listener make sense for covering that case as well?

If this makes sense, I'd be happy to work on this case.

kdkrkwhr:
Thanks @qixiangme  good catch. You're right.

The per-instance change fixes cross-instance leakage, but a shared instance would still
mix state across concurrent parse() calls. I'll move the listener to a per-invocation local
inside parse() — that covers both cases and drops the clear() call. Will update the PR and
adjust the test to use a single shared instance.

kdkrkwhr:
Done, moved the listener into parse() (per-invocation, as suggested in the issue discussion) and added a same-instance regression test. Ready for another look.

Issue #6762, opened 2026-08-07
Streaming tool calls fail with No value present in OpenAiChatModel$ChunkMerger


zhangmr-zmr:
When using ChatClient.stream().content() with tools enabled, the stream fails in MessageAggregator while merging tool call chunks.

Exception:
java.util.NoSuchElementException: No value present

Failing component:
org.springframework.ai.openai.OpenAiChatModel$ChunkMerger.chunkToChatCompletion

Observed stack:
OpenAiChatModel$ChunkMerger.lambda$chunkToChatCompletion$6(OpenAiChatModel.java:1130)
OpenAiChatModel$ChunkMerger.lambda$chunkToChatCompletion$7(OpenAiChatModel.java:1134)
OpenAiChatModel$ChunkMerger.lambda$chunkToChatCompletion$8(OpenAiChatModel.java:1123)
MessageAggregator - Aggregation Error

ERROR openai-stream-handler-thread-0 org.springframework.ai.chat.model.MessageAggregator - Aggregation Error

java.util.NoSuchElementException: No value present
    at java.base/java.util.Optional.get(Optional.java:143)
    at org.springframework.ai.openai.OpenAiChatModel$ChunkMerger.lambda$chunkToChatCompletion$6(OpenAiChatModel.java:1130)
    at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
    at java.base/java.util.stream.ReferencePipeline.toList(ReferencePipeline.java:627)
    at org.springframework.ai.openai.OpenAiChatModel$ChunkMerger.lambda$chunkToChatCompletion$7(OpenAiChatModel.java:1134)
    at java.base/java.util.Optional.ifPresent(Optional.java:178)
    at org.springframework.ai.openai.OpenAiChatModel$ChunkMerger.lambda$chunkToChatCompletion$8(OpenAiChatModel.java:1123)

AlexYrkn:
Confirming on 2.0.0 GA — same line numbers as the report, so same build.

Environment
- Spring AI 2.0.0, Java 25
- openai-java-core 4.39.1, reactor-core 3.8.6
- Provider: OpenAI-compatible gateway (Timeweb AI Gateway)
- Reproduces with two different model families: DeepSeek and Qwen.
  Not provider-specific to a single vendor.

**Full stack trace (untruncated)**

[...]

PR #6817, opened 2026-08-19
Ignore empty OpenAI stream tool calls in messages

liushenling2001:
Follow-up to #6504.

We hit this with a local vLLM OpenAI-compatible endpoint while streaming from a DeepSeek-style model. The same flow worked against NVIDIA's OpenAI-compatible endpoint, but the local vLLM stream can emit normal text deltas with `tool_calls: []` and can also leave an incomplete streamed tool-call object in the reconstructed chunk.

After #6504, `ChunkMerger.hasToolCall` no longer treats an empty `tool_calls` list as an active tool call. However, the stream path still converts chunk deltas back into `ChatCompletion` messages in `chunkToChatCompletion`. In that conversion, every present tool call was mapped with required `id`, `function`, and `function.name` fields. With the vLLM response shape, that path can still fail during stream aggregation with:

```text
java.lang.IllegalStateException: Tool call function name is missing
    at org.springframework.ai.openai.OpenAiChatModel$ChunkMerger.lambda$chunkToChatCompletion...
```

This change keeps empty or incomplete streamed tool calls out of the reconstructed assistant message. Only tool calls with a non-empty id, function, and function name are mapped to `ChatCompletionMessageToolCall`. Empty `tool_calls: []` text deltas are treated the same as missing tool calls.

Validation:
- `./mvnw.cmd -pl models/spring-ai-openai -Dtest=OpenAiChatModelTests -Dspring-javaformat.skip=true -Dcheckstyle.skip=true test`
- `./mvnw.cmd -pl models/spring-ai-openai -am -DskipTests -DskipITs -Dspring-javaformat.skip=true -Dcheckstyle.skip=true install`

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -1299,25 +1299,28 @@ public final class OpenAiChatModel implements ChatModel {
 					// so they are readable on the message, as in the non-streaming path.
 					.putAllAdditionalProperties(cccc.delta()._additionalProperties());
 				cccc.delta().toolCalls().ifPresent(ccctcs -> {
-					msgBuilder.toolCalls(ccctcs.stream().map(tc -> {
+					List<ChatCompletionMessageToolCall> toolCalls = ccctcs.stream().flatMap(tc -> {
+						if (tc.function().isEmpty() || tc.id().filter(StringUtils::hasText).isEmpty()) {
+							return Stream.empty();
+						}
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks for the PR!

Closing in favor of #6857, which changes the same tool-call mapping block in `ChunkMerger.chunkToChatCompletion` for #6762 and also updates the three tests that assert the old throwing behavior: `chunkToChatCompletionRequiresToolCallId`, `chunkToChatCompletionRequiresToolCallFunctionName` and `chunkToChatCompletionRequiresToolCallFunction`. The commit is also missing a `Signed-off-by` line for DCO.

One idea from this PR is worth carrying over: skipping an incomplete tool call is safer than forwarding one with a blank name, since a blank name fails later in `DefaultToolCallingManager`. A review comment to that effect on #6857 would be welcome.

Thanks again!

PR #6820, opened 2026-08-19
Document MCP client scope and session boundaries

dimitarproynov:
Add a "Client Scope and Session Boundaries" section to the MCP client boot starter docs covering the scope of the autoconfigured beans, the consequences of the shared session, when per-user protocol isolation is actually needed, and how to obtain it. Cross-reference it from the MCP client security docs and record the limitation alongside the existing autoconfiguration caveats.


ilayaperumalg (approved):

ilayaperumalg:
Rebased and merged via 5fec219c57ac50b0595cc46d4f92789203a3ad66

PR #6822, opened 2026-08-19
Find @McpTool methods on AOP-proxied beans

### LINKED ISSUE #4405: MCP and AOP cannot be used simultaneously.
when i user mcp tools,can't use spring aop

the tool's instance can't proxy by cglib

minsuking:
## Summary

`AbstractMcpToolProvider` scans annotated methods using `bean.getClass().getDeclaredMethods()`.

When the bean is wrapped in a Spring AOP proxy, `bean.getClass()` returns the proxy class rather than the original target class. In the case of a CGLIB proxy, the overridden target method does not retain the `@McpTool` annotation, causing the tool to be silently skipped during registration.

This change resolves the bean to its AOP target class before scanning for annotated methods.

## Why

This mirrors the existing behavior in `AnnotationProviderUtil.beanMethods()`, which is already used by the Spring-facing annotation providers such as:

* `SyncMcpAnnotationProviders`
* `AsyncMcpAnnotationProviders`

The lower-level tool providers under `provider.tool` did not apply the same AOP-aware lookup.

Since all four tool provider variants — sync/async and stateless/stateful — inherit from `AbstractMcpToolProvider`, the fix can be applied in one place.

## Testing

Added a test that wraps a bean containing an `@McpTool` method with a real Spring `ProxyFactory` proxy and verifies that the tool is still discovered.

* Current `main`: **0 tools found**
* With this change: **tool is discovered successfully**

## Scope

This PR is intentionally limited to tool providers, matching the reported issue.

The resource, prompt, complete, and progress providers currently use the same proxy-unaware pattern and may require similar treatment in a follow-up.

  
Fixes #4405

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/provider/tool/AbstractMcpToolProvider.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/provider/tool/AbstractMcpToolProvider.java
@@ -24,6 +24,7 @@ import io.modelcontextprotocol.json.McpJsonMapper;
 import io.modelcontextprotocol.util.Assert;
 
 import org.springframework.ai.mcp.annotation.McpTool;
+import org.springframework.aop.support.AopUtils;
 
 /**
  * @author Christian Tzolov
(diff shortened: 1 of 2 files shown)


minsuking:
Hi, following up on this PR. The MCP integration workflow expired while waiting for approval, so the reported failure is not a test failure.

This change still addresses the AOP-proxied @McpTool discovery issue and includes a regression test. Happy to rebase or adjust the implementation if needed. Thanks!

PR #6824, opened 2026-08-20
Bind NOT tighter than AND and OR in filter DSL

tm2-web:
The NOT alternative was declared last in the left-recursive booleanExpression rule, which in ANTLR4 gives it the lowest precedence. As a result "NOT a == 1 AND b == 2" parsed as NOT(a == 1 AND b == 2) rather than (NOT a == 1) AND b == 2.

FilterHelper.negate then applies De Morgan to that tree, so the surplus negation turns the AND into an OR and documents that should be excluded are returned instead. Parsing succeeds and nothing is logged, which makes the wrong result set silent.

Move the grouping and NOT alternatives above AND and OR so that NOT binds tightest, in line with the SQL WHERE semantics this language documents itself as a subset of. Expressions that parenthesise the NOT operand are unaffected.

Thank you upfront!


sobychacko:
Hi @tm2-web, thanks for the contribution and for the clear explanation of the `NOT` precedence issue.

This overlaps completely with #6818, which came in before and makes the exact same change (the grammar reordering, the regenerated parser, and the same three tests). We'll track the fix there, so I'm going to close this one as a duplicate.

Really appreciate you taking the time, and hope to see more PRs from you.

PR #6838, opened 2026-08-23
Fix Media JSON round trips

### LINKED ISSUE #6834: `Document` JSON serialization fails for URI-backed media
**Bug description**

I ran into a serialization failure while storing a `Document` that contains URI-backed media.

`Document` is documented as supporting a Jackson JSON round trip, but serialization currently calls `Media#getDataAsByteArray()`. That method throws when the media data is a URI.

**Environment**

- Spring AI: `2.0.2-SNAPSHOT`
- Commit: `fd3fd6ec700340b90c3d516be57c6e6c87dd7df1`
- Java: Microsoft OpenJDK 21.0.12
- macOS 26.5.1, Apple Silicon

**Steps to reproduce**

Running the following program reproduces the problem:

```java
import java.net.URI;

import org.springframework.ai.content.Media;
import org.springframework.ai.document.Document;
import org.springframework.ai.util.JacksonUtils;
import org.springframework.util.MimeTypeUtils;

public class MediaDocumentJsonRepro {

	public static void main(String[] args) throws Exception {
		var media = Media.builder()
			.mimeType(MimeTypeUtils.IMAGE_PNG)
			.data(URI.create("https://example.com/image.png"))
			.name("image")
			.build();

		var document = Document.builder().id("media-doc").media(media).build();
		var mapper = JacksonUtils.getDefaultJsonMapper();

		String json = mapper.writeValueAsString(document);
		System.out.println(mapper.readValue(json, Document.class));
	}

}
```

The call to `writeValueAsString` fails with:

```text
Document["media"] -> Media["dataAsByteArray"]
IllegalStateException: Media data is not a byte[]
```

**Expected behavior**

The document should serialize without Jackson treating `getDataAsByteArray()` as a JSON property. The resulting JSON should also be readable back into a `Document` while preserving the media data.

The program above is self-contained and reproduces the failure against the current `main` branch.

yaner-here:
Closes #6834.

This PR provides a simpler fix, only edits 13 lines:

- Add `@JsonIgnore` to `getDataAsByteArray()`, stop `Jackson` from calling this helper method as a JSON property.
- Use **native** Spring's `MimeType` serialization/deserialization for `MineType` property.

with a simple unittest attached.

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/content/Media.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/content/Media.java
@@ -20,7 +20,12 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import org.jspecify.annotations.Nullable;
+import tools.jackson.databind.annotation.JsonSerialize;
+import tools.jackson.databind.ser.std.ToStringSerializer;
 
 import org.springframework.core.io.Resource;
 import org.springframework.util.Assert;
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks for this fix, @yaner-here. This overlaps with #6836, which addresses the same issue (#6834) with the same core change: adding `@JsonIgnore` to `getDataAsByteArray()` and a private `@JsonCreator` so `Media` can be read back from JSON. Your approach using `@JsonSerialize(ToStringSerializer.class)` on `getMimeType()` is nicely concise; one thing to note is that the `@JsonCreator` here accepts `mimeType` only as a `String`, so it depends on that serializer always producing a string shape, whereas #6836 normalizes the several shapes Jackson may materialize and adds coverage for `byte[]` and invalid-data cases. We are going to consolidate on #6836 for that broader coverage, so I will close this one. Really appreciate the contribution and hope you will send more our way.

Issue #6665, opened 2026-07-22
`GoogleGenAiChatModel` throws `NoSuchElementException` for models that omit `modelVersion` in `GenerateContentResponse` (e.g. Vertex Gemma MaaS)


joshuamoffat:
**Bug description**

`GoogleGenAiChatModel` builds its chat response metadata with
`generateContentResponse.modelVersion().get()` (in `internalCall`, and again in
`internalStream`). Gemini responses always carry `modelVersion`, but Vertex AI also
serves non-Gemini publisher models through the same `generateContent` surface — for
example the managed/serverless Gemma models
(`publishers/google/models/gemma-4-26b-a4b-it-maas`) — and their responses **omit
`modelVersion`** (and `responseId`). The empty `Optional` makes every `call()` throw
`NoSuchElementException: No value present`, even though the model responded
successfully (including with a fully honored `responseSchema`).

Two adjacent accessors have the same unguarded pattern and may bite on other
response shapes:

- `candidate.content().get().parts().get()` in `responseCandidateToGeneration`
  (fails on candidates with no content/parts, e.g. safety-blocked responses)
- Gemma also returns `"role": "assistant"` where Gemini returns `"role": "model"` —
  it did not cause a failure here, but the response shape is not identical to Gemini's

**Environment**

- Spring AI 2.0.0 (`spring-ai-starter-model-google-genai`)
- Java 25
- Vertex AI, global endpoint, model `publishers/google/models/gemma-4-26b-a4b-it-maas`
- The unguarded `.get()` calls are still present on `main` at the time of filing

**Steps to reproduce**

1. Configure `GoogleGenAiChatModel` against a Vertex project with model
   `publishers/google/models/gemma-4-26b-a4b-it-maas` (Gemma 4 26B, managed/serverless
   in Model Garden).
2. `chatModel.call(new Prompt("hi"))`
3. `NoSuchElementException: No value present` is thrown from the
   response-to-`ChatResponse` mapping.

**Expected behavior**

The call succeeds. When `modelVersion` is absent from the response, the metadata
should fall back to the requested model name (or empty), e.g.
`generateContentResponse.modelVersion().orElse(requestedModel)`, instead of failing
the whole call after a successful generation.

**Minimal Complete Reproducible example**

The response shape is easiest to show with a direct REST call to the same endpoint
the SDK uses. Request (global endpoint, `:generateContent`, Gemma MaaS model):

```json
{
[...]

PR #6861, opened 2026-08-26
Fix NoSuchElementException in `GoogleGenAiChatModel` on empty response candidates

symphony-enrico:
### Summary
This PR fixes an unhandled `java.util.NoSuchElementException: No value present` crash that occurs in `GoogleGenAiChatModel` when a response candidate contains empty content.

### Issue Description
When using `GoogleGenAiChatModel` with native Google Search Grounding (Web Search) enabled, or when Gemini triggers safety-blocked filters, the API response containing candidate structures can return empty content (i.e., `candidate.content().isPresent()` is `false`).

When this happens:
1. `isFunctionCall` evaluates to `false`.
2. The execution path falls into the `else` block of `responseCandidateToGeneration`.
3. Spring AI calls `candidate.content().get()` blindly without checking `.isPresent()`, throwing:
   ```
   java.util.NoSuchElementException: No value present
       at java.base/java.util.Optional.get(Unknown Source)
       at org.springframework.ai.google.genai.GoogleGenAiChatModel.responseCandidateToGeneration(GoogleGenAiChatModel.java:594)
   ```

### Fix Details
- Introduced a robust, early safeguard in `GoogleGenAiChatModel#responseCandidateToGeneration` to verify if `candidate.content().isEmpty()`.
- If empty, the method cleanly builds and returns an empty `Generation` alongside its metadata (including finish reason and candidate index), preventing any blind `.get()` calls or crashes.
- Added a full unit test suite `GoogleGenAiChatModelTests` checking this behavior with mocked candidates.

### DIFF
--- a/models/spring-ai-google-genai/src/main/java/org/springframework/ai/google/genai/GoogleGenAiChatModel.java
+++ b/models/spring-ai-google-genai/src/main/java/org/springframework/ai/google/genai/GoogleGenAiChatModel.java
@@ -519,6 +519,9 @@ public class GoogleGenAiChatModel implements ChatModel, DisposableBean {
 	}
 
 	protected List<Generation> responseCandidateToGeneration(Candidate candidate) {
+		if (candidate == null) {
+			return List.of();
+		}
 
 		// TODO - The candidateIndex (e.g. choice must be assigned to the generation).
 		int candidateIndex = candidate.index().orElse(0);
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks for this fix, @symphony-enrico. This addresses the same `NoSuchElementException` in `GoogleGenAiChatModel.responseCandidateToGeneration` as #6757, which was opened a few weeks earlier and links the originating issue #6665. Both changes make an absent-content candidate return a single empty `Generation` with finish-reason metadata instead of throwing, so the end effect is equivalent. #6757 also includes a test that exercises the full `call()` path, so we plan to move forward with that one and will close this in its favor. We really appreciate you taking the time here and hope you will send more contributions.

Issue #4422, opened 2025-09-19
Generated JSON schema fails strict mode checks of Azure


bwaldvogel:
After upgrading Spring AI from `1.0.1` to `1.0.2`, my chat client began throwing this exception:

```text
400 Bad Request: {
  "error": {
    "message": "Invalid schema for response_format 'custom_schema': In context=('properties', [MASKED]), 'additionalProperties' must be supplied and set to false.",
    "type": "invalid_request_error",
    "param": "response_format",
    "code": null
  }
}
```

The problem seems to arise because the `strict` property is now mapped correctly (probably changed via https://github.com/spring-projects/spring-ai/issues/3928), which exposes previously “invalid” JSON schemas.

I’m looking for the correct way to generate a JSON schema that satisfies strict mode—that is, where `additionalProperties` is set to `false` **in every nested object** and every field is listed in `required`.

So far I’ve configured the response format like this:

```java
AzureOpenAiResponseFormat.builder()
    .type(AzureOpenAiResponseFormat.Type.JSON_SCHEMA)
    .jsonSchema(JsonSchemaGenerator.generateForType(JSON_SCHEMA_CLASS))
    .build();
```

However, `JsonSchemaGenerator.generateForType(JSON_SCHEMA_CLASS)` sets `additionalProperties=false` only on the **top-level** object, not on nested ones.

I also tried `BeanOutputConverter.getJsonSchema`. That method correctly sets `additionalProperties=false` on nested objects, but it doesn’t mark every field as required, leading to a different exception:

```text
400 Bad Request: {
  "error": {
    "message": "Invalid schema for response_format 'custom_schema': In context=('properties', [MASKED]), 'required' must be supplied and must include every key in properties. Missing '[MASKED]'.",
    "type": "invalid_request_error",
    "param": "response_format",
    "code": null
  }
}
```

PR #6865, opened 2026-08-26
Backfill `additionalProperties: false` in OpenAI strict tool schemas

JamesBLewis:
TLDR; When a caller turns on OpenAI's strict tool-calling mode, Spring AI rewrites each tool's JSON schema to meet OpenAI's strict-mode contract but it only handles half of it. It fills in the "required" list and makes optional fields nullable, yet never adds `additionalProperties: false`, which OpenAI also demands on every object in the schema. Tools whose schemas were generated by Spring AI happen to work because the generator adds it earlier, but tools from MCP servers or with hand-written schemas get rejected by the OpenAI API. This change makes the strict-mode rewrite fill in `additionalProperties: false` wherever it is missing.

## Problem

`OpenAiChatModel#applyStrictModeRequirements` rewrites a tool's input schema when `OpenAiChatOptions#strict(true)` is set: it backfills `required` with every property key and widens previously optional properties to nullable types, recursing into `$defs`, nested object properties, and array items.

OpenAI's strict function-calling mode additionally requires `additionalProperties: false` on every object level, and rejects requests otherwise with:

> Invalid schema for function '...': 'additionalProperties' is required to be supplied and to be false.

The rewrite never emits it. Schemas produced by `JsonSchemaGenerator` are unaffected because `forbidAdditionalProperties` already adds it recursively by default — which is why the gap is invisible on the `@Tool`/`FunctionToolCallback` path. Any schema from another source hits the 400 under strict mode:

- MCP tool schemas (servers rarely emit `additionalProperties: false`)
- hand-written `ToolDefinition#inputSchema` JSON
- schemas generated with `SchemaOption.ALLOW_ADDITIONAL_PROPERTIES_BY_DEFAULT`

See #4422

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -1054,12 +1054,16 @@ public final class OpenAiChatModel implements ChatModel {
 	/**
 	 * Rewrites an object schema in place to satisfy OpenAI's strict function-calling
 	 * mode: every property must be listed in "required", with optionality expressed via a
-	 * nullable type rather than omission from "required". {@link JsonSchemaGenerator}
+	 * nullable type rather than omission from "required", and every object level must pin
+	 * {@code "additionalProperties"} to {@code false}. {@link JsonSchemaGenerator}
 	 * produces conventional JSON Schema instead - it omits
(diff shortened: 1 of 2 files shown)


ilayaperumalg (approved):

ilayaperumalg:
@JamesBLewis Thanks for the fix! Rebased and merged via e5e277fd0

PR #6871, opened 2026-08-27
Close `WebMvcStreamableServerTransportProvider` transports on disconnect

### LINKED ISSUE #6860: [BUG] WebMvcStreamableServerTransportProvider: session and socket leak when clients disconnect without DELETE
# [BUG] WebMvcStreamableServerTransportProvider: session and socket leak when clients disconnect without DELETE

## Description

`WebMvcStreamableServerTransportProvider` leaks MCP sessions and their associated TCP sockets permanently when a client disconnects without sending an explicit `DELETE /mcp` request. Over time this exhausts server file descriptors and makes the application unresponsive.

The root cause is a combination of:

1. `sessions.remove(sessionId)` is only called inside `handleDelete` — no other code path evicts a session.
2. `SseBuilder.onTimeout()` only logs; it does not close the stream or remove the session.
3. SSE write failures (`SseBuilder.send()` throwing) do not trigger transport cleanup.
4. `KeepAliveScheduler` swallows ping failures via `.onErrorComplete()` and never removes dead sessions.
5. There is no session TTL or background reaper.

Any client that goes away without a clean DELETE (network drop, process kill, laptop sleep, load-balancer timeout) leaks a session and socket permanently.

## Environment

- Spring AI: 1.1.4 – 1.1.8 (all use `WebMvcStreamableServerTransportProvider`)
- Spring Boot: 3.4.x / 3.5.x
- MCP Java SDK: 0.17.0 – 0.18.3 (bundled via Spring AI)
- Transport: Streamable HTTP with `spring-ai-starter-mcp-server-webmvc`
- Server: Tomcat (both embedded and standalone WAR)
- OS: Linux (Kubernetes)

## Production Evidence

Reproduced independently by two users in production:

### User 1 (reported in [spring-projects/spring-ai#6384](https://github.com/spring-projects/spring-ai/issues/6384))

- 138 CLOSE-WAIT sockets accumulated within minutes under moderate load
- Tomcat listen backlog full (151/150), health checks timing out
- `connectionTimeout` ineffective because connections are held by `startAsync()`
- Only recovered after ~2 hours via kernel `tcp_keepalive_time`

### User 2 (reported in [modelcontextprotocol/java-sdk#1021](https://github.com/modelcontextprotocol/java-sdk/issues/1021#issuecomment-5351378121))

- Measured over 48 hours on a single host:

| elapsed | `tomcat.connections.current` | open FDs |
|---|---|---|
| baseline | 2 | 72 |
| +~2h | 20 | 91 |
| +~48h | 6373 | 6443 |

- FD count only ratchets up, never decreases during idle periods
- Earlier instance reached 8290 FDs after ~4.5 days
- Log signature: `Failed to send keep-alive ping to session ...: Stream unavailable for session <id>`
- On shutdown: `Failed to complete SSE builder for session {}: The response object has been recycled`

## Steps to Reproduce

1. Deploy an MCP server using `spring-ai-starter-mcp-server-webmvc` with Streamable HTTP transport.
2. Connect an MCP client, complete initialization.
3. Terminate the client without sending `DELETE` (e.g. `kill -9`, network drop).
4. Repeat steps 2–3.
5. Observe: session count and socket count only increase; keep-alive warnings appear once per interval per dead session.

## Expected Behavior

When a client disconnects (SSE write failure, timeout, or async error):

1. The current SSE stream/transport should be closed.
2. The servlet async context should be completed, releasing the Tomcat NIO channel.
3. After repeated keep-alive failures or a configurable TTL, the logical MCP session should be evicted from `sessions`.
4. The server-side socket should be closed, preventing CLOSE-WAIT accumulation.

## Root Cause Analysis

In `WebMvcStreamableServerTransportProvider`:

```java
// onTimeout — only logs, does not close
sseBuilder.onTimeout(() -> log.debug("SSE connection timed out for session: {}", sessionId));

// onComplete — closes listeningStream, but only this one path
sseBuilder.onComplete(() -> listeningStream.close());

// sendMessage write failure — calls sseBuilder.error(), does not close transport
catch (Exception e) {
    sseBuilder.error(e);  // does not trigger session/transport cleanup
}
```

In `KeepAliveScheduler`:

```java
// ping failure is swallowed, session never removed
.onErrorComplete()
```

There is no `sessions.remove()` outside of `handleDelete`, no TTL, and no reaper.

## Relationship to MCP Java SDK #1027

I previously filed [spring-projects/spring-ai#6384](https://github.com/spring-projects/spring-ai/issues/6384) for this issue. A Spring AI maintainer (`@sdeleuze`) closed it, directing the fix to the MCP Java SDK side. I then submitted [modelcontextprotocol/java-sdk#1027](https://github.com/modelcontextprotocol/java-sdk/pull/1027) to address the servlet Streamable HTTP transport lifecycle.

However, as clarified in the [MCP Java SDK 1.0.0 release notes](https://github.com/modelcontextprotocol/java-sdk/releases/tag/v1.0.0) (February 2026):

> "Moved `mcp-spring-webflux` and `mcp-spring-webmvc` to Spring AI 2.0" ([#805](https://github.com/modelcontextprotocol/java-sdk/pull/805))

The Spring MVC transport has been maintained by Spring AI since February 2026. PR #1027 only fixes `HttpServletStreamableServerTransportProvider` in `mcp-core` — a different code path that is **not used by `spring-ai-starter-mcp-server-webmvc`**. Therefore, #1027 does not address this issue for any Spring AI WebMVC user.

## Proposed Fix

I would like to contribute a PR to address this. The fix mirrors the approach in [MCP SDK #1027](https://github.com/modelcontextprotocol/java-sdk/pull/1027):

1. **`onTimeout` → close the listening stream / transport**, not just log.
2. **`onError` → close the transport**, completing the async context.
3. **SSE write failure → call `transport.close()`**, not just `sseBuilder.error()`.
4. **Keep-alive ping failure → evict the dead session** from `sessions` after N consecutive failures (or integrate with session TTL logic similar to [#1028](https://github.com/modelcontextprotocol/java-sdk/pull/1028)).
5. **Add regression tests** covering client disconnect, timeout, and write-failure scenarios.

The fix should be minimal and non-breaking — closing only the affected physical stream/transport, while preserving the logical session for legitimate reconnects within a bounded window.

If maintainers agree with this direction, I can submit a PR against `main` targeting the `mcp/transport/mcp-spring-webmvc` module.

## Related

- [spring-projects/spring-ai#6384](https://github.com/spring-projects/spring-ai/issues/6384) — my original report (closed, was redirected to MCP SDK)
- [modelcontextprotocol/java-sdk#1021](https://github.com/modelcontextprotocol/java-sdk/issues/1021) — upstream issue with production data from two independent users
- [modelcontextprotocol/java-sdk#1027](https://github.com/modelcontextprotocol/java-sdk/pull/1027) — my PR fixing the servlet transport (does not cover WebMVC)
- [modelcontextprotocol/java-sdk#1028](https://github.com/modelcontextprotocol/java-sdk/pull/1028) — session TTL / keep-alive eviction (servlet only)
- [modelcontextprotocol/java-sdk#805](https://github.com/modelcontextprotocol/java-sdk/pull/805) — migration of Spring transports to Spring AI (February 2026)

lxq19991111:
Fixes #6860

Makes the WebMVC Streamable HTTP SSE transport close when the SSE response completes, times out, or errors, and routes SSE write failures through the transport `close()` path instead of only reporting the error to the `SseBuilder`.

It mirrors modelcontextprotocol/java-sdk#1021: listening streams close on timeout/error as well as completion, replay and streaming responses close on lifecycle events, and every lifecycle event uses the same idempotent `close()` (no separate `markClosed()` flag).

Key design point: a write failure closes only the current HTTP/SSE transport and does NOT remove the logical MCP session. Removing the session on a request-specific write failure would break the next request carrying the same `mcp-session-id` and Last-Event-ID replay, so session eviction is intentionally left to DELETE / keep-alive eviction.

  ## Testing

Rewrote both tests without reflection, using a package-private transport factory seam (the provider exposes a `RouterFunction`, so the SDK-style servlet entry-point mocking is not applicable here):

- `sseLifecycleCallbacksCloseStreamOnCompleteTimeoutAndError` asserts all three lifecycle callbacks trigger the close callback.
- `sseWriteFailureClosesOnlyCurrentTransportWithoutRemovingSession` initializes a session over the router, triggers an SSE write failure, verifies the transport is completed, and then deletes the session with the same id, asserting 200 to prove the session was retained.

  ```bash
  mvn -pl mcp/transport/mcp-spring-webmvc -am \
      -Dtest=WebMvcStreamableServerTransportProviderTests \
      -Dsurefire.failIfNoSpecifiedTests=false test

Module unit tests pass, and WebMcpStreamableAsyncServerTransportIT, WebMcpStreamableSyncServerTransportIT, and WebMvcStreamableIT pass.

### DIFF
--- a/mcp/transport/mcp-spring-webmvc/src/main/java/org/springframework/ai/mcp/server/webmvc/transport/WebMvcStreamableServerTransportProvider.java
+++ b/mcp/transport/mcp-spring-webmvc/src/main/java/org/springframework/ai/mcp/server/webmvc/transport/WebMvcStreamableServerTransportProvider.java
@@ -352,18 +352,13 @@ public final class WebMvcStreamableServerTransportProvider implements McpStreama
 
 		try {
 			return ServerResponse.sse(sseBuilder -> {
-				sseBuilder.onTimeout(() -> {
-					if (logger.isDebugEnabled()) {
-						logger.debug("SSE connection timed out for session: " + sessionId);
-					}
-				});
-
-				WebMvcStreamableMcpSessionTransport sessionTransport = new WebMvcStreamableMcpSessionTransport(
-						sessionId, sseBuilder);
+				WebMvcStreamableMcpSessionTransport sessionTransport = createSessionTransport(sessionId, sseBuilder);
 
 				// Check if this is a replay request
 				if (!request.headers().header(HttpHeaders.LAST_EVENT_ID).isEmpty()) {
 					String lastId = request.headers().asHttpHeaders().getFirst(HttpHeaders.LAST_EVENT_ID);
+
+					registerSseLifecycle(sseBuilder, sessionId, sessionTransport::close);
 
 					try {
 						session.replay(lastId)
@@ -379,7 +374,7 @@ public final class WebMvcStreamableServerTransportProvider implements McpStreama
 									if (logger.isErrorEnabled()) {
 										logger.error("Failed to replay message: " + e.getMessage());
 									}
-									sseBuilder.error(e);
+									sessionTransport.close();
 								}
 							});
 					}
@@ -387,7 +382,7 @@ public final class WebMvcStreamableServerTransportProvider implements McpStreama
 						if (logger.isErrorEnabled()) {
 							logger.error("Failed to replay messages: " + e.getMessage());
 						}
-						sseBuilder.error(e);
+						sessionTransport.close();
 					}
 				}
 				else {
@@ -395,12 +390,7 @@ public final class WebMvcStreamableServerTransportProvider implements McpStreama
 					McpStreamableServerSession.McpStreamableServerSessionStream listeningStream = session
 						.listeningStream(sessionTransport);
 
-					sseBuilder.onComplete(() -> {
-						if (logger.isDebugEnabled()) {
-							logger.debug("SSE connection completed for session: " + sessionId);
-						}
-						listeningStream.close();
-					});
+					registerSseLifecycle(sseBuilder, sessionId, listeningStream::close);
 				}
 			}, Duration.ZERO);
 		}
@@ -487,19 +477,10 @@ public final class WebMvcStreamableServerTransportProvider implements McpStreama
 			else if (message instanceof McpSchema.JSONRPCRequest jsonrpcRequest) {
 				// For streaming responses, we need to return SSE
 				return ServerResponse.sse(sseBuilder -> {
-					sseBuilder.onComplete(() -> {
-						if (logger.isDebugEnabled()) {
-							logger.debug("Request response stream completed for session: " + sessionId);
-						}
-					});
-					sseBuilder.onTimeout(() -> {
-						if (logger.isDebugEnabled()) {
-							logger.debug("Request response stream timed out for session: " + sessionId);
-						}
-					});
+					WebMvcStreamableMcpSessionTransport sessionTransport = createSessionTransport(sessionId,
+							sseBuilder);
 
-					WebMvcStreamableMcpSessionTransport sessionTransport = new WebMvcStreamableMcpSessionTransport(
-							sessionId, sseBuilder);
+					registerSseLifecycle(sseBuilder, sessionId, sessionTransport::close);
 
 					try {
 						session.responseStream(jsonrpcRequest, sessionTransport)
@@ -510,7 +491,7 @@ public final class WebMvcStreamableServerTransportProvider implements McpStreama
 						if (logger.isErrorEnabled()) {
 							logger.error("Failed to handle request stream: " + e.getMessage());
 						}
-						sseBuilder.error(e);
+						sessionTransport.close();
 					}
 				}, Duration.ZERO);
 			}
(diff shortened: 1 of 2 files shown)


lxq19991111:
Pushed a follow-up commit propagating SSE write failures: the transport still closes only the current connection and keeps the logical session, but `sendMessage` now surfaces an `McpTransportException` instead of completing normally, so callers stop writing to a stream that no longer leads anywhere. This matches what modelcontextprotocol/java-sdk#1130 settled on upstream. Broadcasting is unaffected: `notifyClients` logs and swallows per-session failures, and the keep-alive scheduler completes ping failures, so a single write failure cannot break other sessions.

Two notes on how this relates to modelcontextprotocol/java-sdk#1130 ("Streamable HTTP: Add session sweeping and close hanging streams"). It landed after the v2.0.1 tag, and it only changes `HttpServletStreamableServerTransportProvider` in `mcp-core`, not the WebMVC provider maintained here, so upgrading the MCP SDK alone does not fix `DELETE`-less disconnects for `spring-ai-starter-mcp-server-webmvc` users, and this PR is still needed. When Spring AI moves to a release containing that work, the WebMVC transport can be aligned with the new `releaseTransport()` and response-stream APIs separately.

The write-failure test now asserts the propagated exception and no longer verifies that the `SseBuilder` was completed on the mock. That assertion was a false positive: `DefaultSseBuilder` sets `sendFailed` on the first failed write, after which `complete()` returns immediately, so the mock verified a call that cannot happen on a real connection.

PR #6878, opened 2026-08-30
Add streaming time-to-first-chunk observability

### LINKED ISSUE #6670: Observability: report time to first chunk for streaming calls
**Expected Behavior**

For streaming chat calls Spring AI observability should report time to first chunk, as defined in the current OpenTelemetry GenAI semantic conventions. The spec has two signals for it:

* span attribute `gen_ai.response.time_to_first_chunk` (double, seconds, recommended for streaming requests): https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-spans.md
* client histogram metric `gen_ai.client.operation.time_to_first_chunk`: https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-metrics.md

Both are measured from the moment the client issues the request to the moment the first chunk arrives from the response stream.

**Current Behavior**

Spring AI measures only the duration of the whole operation (`gen_ai.client.operation.duration`). For streaming this number is dominated by the generation time of the full response, so it says nothing about how long the user waited before the first token appeared. Time to first chunk is the main latency indicator for streaming UX and there is no way to get it from Spring AI observability today:

* `ChatModelObservationContext` has no field for this value
* `DefaultChatModelObservationConvention` and `AiObservationAttributes` do not know the key
* model implementations have no hook to record the timing into the observation, so a custom `ObservationConvention` does not help either

**Context**

We maintain a third party model adapter for Spring AI and went through the current GenAI semconv to check what telemetry is missing. Teams that run streaming chat in production usually alert on time to first token, right now they have to instrument it manually around `ChatClient`/`ChatModel` calls.

Possible implementation: the streaming path already creates the `Observation` before subscribing to the provider `Flux`, so the framework could capture the timestamp of the first emitted `ChatResponse` (for example with `doOnNext` on the first element) and put the delta into the observation context. Then the default convention emits the span attribute, and a meter handler can produce the histogram.

Related issues about aligning with the current GenAI semconv: #6668 (`gen_ai.provider.name` rename), #6669 (`gen_ai.conversation.id`), #5002 and #6051 (input/output messages on spans). The initial semconv adoption (#1174) was done against semconv 1.27.0, which did not have this attribute yet.

Will happy to discuss and open PR.

samirsahoo:
For streaming ChatModel calls, record the elapsed time between the start of the model request and the first ChatResponse chunk, following the OpenTelemetry GenAI semantic conventions.

The value is captured once, in ChatModelObservationContext, using System.nanoTime() from context creation (inside Flux.deferContextual, right before the provider request) until the first emitted chunk. The first-chunk detection is done in a new
MessageAggregator.aggregate(Flux, Consumer, Runnable) overload, the single point every streaming ChatModel already funnels its chunk Flux through, so no timing logic is duplicated across providers. Each of the seven streaming ChatModels only passes
observationContext::recordTimeToFirstChunk to that overload.

The DefaultChatModelObservationConvention emits the span attribute gen_ai.response.time_to_first_chunk (double, seconds) only for streaming calls that received at least one chunk. The ChatModelMeterObservationHandler records the
gen_ai.client.operation.time_to_first_chunk timer (unit: seconds) with the same low-cardinality tags as gen_ai.client.operation.duration. Both signals are absent for non-streaming calls and for empty streams, and a value measured before a mid-stream failure is preserved.

Fixes #6670

### DIFF
--- a/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
+++ b/models/spring-ai-anthropic/src/main/java/org/springframework/ai/anthropic/AnthropicChatModel.java
@@ -361,7 +361,8 @@ public final class AnthropicChatModel implements ChatModel, StreamingChatModel {
 			// @formatter:on
 
 			// Aggregate streaming responses and handle tool execution on final response
-			return new MessageAggregator().aggregate(flux, observationContext::setResponse);
+			return new MessageAggregator().aggregate(flux, observationContext::setResponse,
+					observationContext::recordTimeToFirstChunk);
 		});
 	}
 
--- a/models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/chat/AnthropicChatModelObservationIT.java
+++ b/models/spring-ai-anthropic/src/test/java/org/springframework/ai/anthropic/chat/AnthropicChatModelObservationIT.java
@@ -150,11 +150,14 @@ public class AnthropicChatModelObservationIT {
 			.hasBeenStarted()
 			.hasBeenStopped();
 		if (streaming) {
-			observationAssert.hasHighCardinalityKeyValue(HighCardinalityKeyNames.REQUEST_STREAM.asString(), "true");
+			observationAssert.hasHighCardinalityKeyValue(HighCardinalityKeyNames.REQUEST_STREAM.asString(), "true")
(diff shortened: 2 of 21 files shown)


sobychacko:
Closing in favor of #6672, opened earlier with the same metric, the same attribute and the same hooks across the same seven models.

The deciding difference is thread safety: the timing value here can be written and read on different threads without synchronisation, so the metric can come back empty. #6672 handles that.

Thanks!

PR #6895, opened 2026-09-01
Support tool call messages in `MongoChatMemoryRepository`

kezhenxu94:
`MongoChatMemoryRepository` filtered out `ToolResponseMessage` and `AssistantMessage` instances carrying tool calls, because the persisted document had no field to hold tool call identifiers, names, arguments or response data.
Applications using tool calling silently lost part of their conversation history.

This was a stopgap rather than a MongoDB limitation, introduced alongside the equivalent JDBC and Cassandra changes.
Unlike JDBC, MongoDB needs no schema migration to store the missing data, and two built-in repositories already support tool calls, so this brings MongoDB to parity with them.

## Changes

`Conversation.Message` now stores `toolCalls` and `toolResponses`, reusing `AssistantMessage.ToolCall` and `ToolResponseMessage.ToolResponse` the same way `RedisChatMemoryRepository` does, so a complete tool-calling exchange round-trips unchanged.
Null assistant content is preserved rather than coerced to an empty string, which is the usual shape of a message that only carries tool calls.

Messages also gain a sequence identifier.
Every message in a `saveAll` batch shared a single BSON timestamp, which only has millisecond precision, and MongoDB does not guarantee the order of documents with duplicate sort keys.
Reads now sort by timestamp and sequence identifier.
This matters specifically for tool calling, because providers reject a tool response that is not immediately preceded by the assistant message requesting it.
The compound index created by `MongoChatMemoryIndexCreatorAutoConfiguration` was extended so the new sort is still served by the index.

## Backwards compatibility

Documents written by earlier versions have neither new field.

* `sequenceId` stays a primitive `int`. Spring Data passes `null` for an absent field and a record cannot declare a default, so with a plain canonical constructor, reading any pre-existing document fails with `MappingInstantiationException`. A package-private `@PersistenceCreator` constructor taking a nullable `Integer` normalizes the absent value to zero, which keeps the record component and its accessor primitive.
* Assistant documents without `toolCalls` are read back with an empty tool call list.
* Tool documents cannot be reconstructed, so they are still skipped, and empty stubs never reach the LLM context. This is distinguished by the field being absent rather than empty, so a tool message genuinely saved without responses stores an empty array and is preserved.
* The previous constructor signatures are retained as delegating constructors, since both records are public API.

## Tests

`MongoChatMemoryRepositoryIT` grows from 8 to 30 tests, covering round-trips of tool calls and tool responses, ordering within a message and across a conversation, parallel tool calls with correlated responses, multi-turn tool-calling conversations, JSON/unicode/large payloads in arguments and response data, metadata preservation, storage-level document structure, replace semantics, and each of the backwards compatibility paths above using raw legacy BSON documents.

Verified locally with:

```
./mvnw -pl memory-repositories/spring-ai-model-chat-memory-repository-mongodb -am \
  -Pintegration-tests -Dfailsafe.failIfNoSpecifiedTests=false \
  -Dit.test=MongoChatMemoryRepositoryIT verify
```

30/30 pass, along with `MongoChatMemoryAutoConfigurationIT`, `javadoc:javadoc` and `spring-javaformat:validate`.

## Note for reviewers

This reverses the deliberate decision in #6399, which filtered these messages across JDBC, Cassandra and MongoDB, and the reference documentation currently points tool-calling users to the Spring AI Session project.
I am happy to close this if the intent is for `ChatMemoryRepository` to stay tool-call-free, or to extend the same approach to the other repositories if parity is preferred.

The equivalent JDBC and Cassandra limitations are untouched here.
I also noticed two pre-existing issues while comparing implementations, both out of scope for this PR and neither addressed: `MongoChatMemoryRepository` drops media entirely, and `RedisChatMemoryRepository` appears to throw an NPE when reading an assistant message saved with null content, since Gson omits the null `content` key and the read path calls `json.get("content").getAsString()` without a guard.

See #6399

### DIFF
--- a/auto-configurations/models/chat/memory/repository/spring-ai-autoconfigure-model-chat-memory-repository-mongodb/src/main/java/org/springframework/ai/model/chat/memory/repository/mongo/autoconfigure/MongoChatMemoryIndexCreatorAutoConfiguration.java
+++ b/auto-configurations/models/chat/memory/repository/spring-ai-autoconfigure-model-chat-memory-repository-mongodb/src/main/java/org/springframework/ai/model/chat/memory/repository/mongo/autoconfigure/MongoChatMemoryIndexCreatorAutoConfiguration.java
@@ -34,10 +34,11 @@ import org.springframework.data.mongodb.core.index.IndexOperations;
 
 /**
  * Class responsible for creating proper MongoDB indices for the ChatMemory. Creates a
- * main index on the conversationId and timestamp fields, and a TTL index on the timestamp
- * field if the TTL is set in properties.
+ * main index on the conversationId, timestamp and sequenceId fields, and a TTL index on
+ * the timestamp field if the TTL is set in properties.
  *
  * @author Łukasz Jernaś
+ * @author kezhenxu94
  * @see MongoChatMemoryProperties
  * @since 1.1.0
  */
@@ -70,7 +71,9 @@ public class MongoChatMemoryIndexCreatorAutoConfiguration {
 
 	private void createMainIndex() {
 		var indexOps = this.mongoTemplate.indexOps(Conversation.class);
-		var index = new Index().on("conversationId", Sort.Direction.ASC).on("timestamp", Sort.Direction.DESC);
+		var index = new Index().on("conversationId", Sort.Direction.ASC)
+			.on("timestamp", Sort.Direction.DESC)
+			.on("sequenceId", Sort.Direction.DESC);
 
 		// Use reflection to handle API differences across Spring Data MongoDB versions
 		createIndexSafely(indexOps, index);
--- a/memory-repositories/spring-ai-model-chat-memory-repository-mongodb/src/main/java/org/springframework/ai/chat/memory/repository/mongo/Conversation.java
+++ b/memory-repositories/spring-ai-model-chat-memory-repository-mongodb/src/main/java/org/springframework/ai/chat/memory/repository/mongo/Conversation.java
@@ -17,20 +17,77 @@
 package org.springframework.ai.chat.memory.repository.mongo;
 
 import java.time.Instant;
+import java.util.List;
 import java.util.Map;
 
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.ai.chat.messages.AssistantMessage;
+import org.springframework.ai.chat.messages.ToolResponseMessage;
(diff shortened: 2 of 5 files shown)


kezhenxu94:
@tzolov can you please take a look? Thanks

kezhenxu94:
Some more context on why the sequence identifier is part of this PR, since it reads like an unrelated change riding along with tool call support.

It is a fix for a real ordering bug, independent of tool calls.

`saveAll` stamps each message with `Instant.now()` as it maps the list to documents, and reads sort by that timestamp. A turn is saved in one call, so the whole list is mapped within a few microseconds, and BSON stores a date to millisecond precision: every message of a turn lands on the same millisecond. The sort then has nothing to order them by, MongoDB is free to break the tie however it likes, and a two-message conversation can come back answer-first.

That is not just a display problem. The same collection is what is handed to the model as the conversation history, so a scrambled read tells it that it answered before it was asked, and the next turn is reasoning about a transcript that never happened.

I hit this in an application before working on this PR, and ended up writing a replacement repository over the same `ai_chat_memory` collection purely to get deterministic ordering. This change is that fix, upstreamed, so the workaround is no longer needed.

The approach is the one `JdbcChatMemoryRepository` already takes with its `sequence_id` column: store the position and order by it. `timestamp` stays the primary sort key so that documents written by earlier versions, which have no `sequenceId`, read exactly as they did before, and the next `saveAll` on a conversation rewrites the whole of it with positions.

It is also a prerequisite for the tool call support in this PR rather than merely adjacent to it. A tool response has to be read back immediately after the assistant message carrying the call that produced it, or the provider rejects the request. Persisting tool calls on top of a non-deterministic order would have shipped a latent failure.

CryoThrust:
The sequence field fixes the ordering semantics, but the new compound index direction does not match the read query. `findByConversationId()` sorts `timestamp ASC, sequenceId ASC`, while `createMainIndex()` creates `{ conversationId: 1, timestamp: -1, sequenceId: -1 }`. After the equality prefix, Mongo can reverse-scan that index as `{ timestamp: 1, sequenceId: 1 }` only when every remaining direction is reversed together; here it actually is every remaining (wait, is current DESC/DESC reverse is ASC/ASC) - please ignore this comment if I have missed an index-planner rule. My initial concern was mixed direction, but both fields are currently DESC, so reverse scan should satisfy the ASC query. The implementation therefore looks aligned.

PR #6906, opened 2026-09-03
Map embedded provider errors to OpenAI exceptions

### LINKED ISSUE #6900: Improper handling of error responses from providers like OpenRouter
**Bug description**
OpenRouter as API provider facilitates "additionalProperties" in ChatCompletion response to indicate about errors. The following is an example:

```
ChatCompletion{id=gen-xxxxxx, choices=, created=, model=, object_=, serviceTier=, systemFingerprint=, usage=, additionalProperties={error={message=Upstream error from Nvidia: Service temporarily overloaded, code=502}}}
```

The current implementation fails with the exception below:

```
com.openai.errors.OpenAIInvalidDataException: `choices` is not set
	at com.openai.core.JsonField.getRequired$openai_java_core(Values.kt:174)
	at com.openai.models.chat.completions.ChatCompletion.choices(ChatCompletion.kt:81)
	at org.springframework.ai.openai.OpenAiChatModel.lambda$internalCall$2(OpenAiChatModel.java:207)
	at io.micrometer.observation.Observation.observe(Observation.java:634)
	at org.springframework.ai.openai.OpenAiChatModel.internalCall(OpenAiChatModel.java:203)
	at org.springframework.ai.openai.OpenAiChatModel.call(OpenAiChatModel.java:182)
	at ....
```

Which prevents from proper handling the "rate limiting" cases in the code and has overall very few data to create a solid error handling mechanism in agents.

The framework should try to cater for catching these cases and wrapping them into proper errors.

**Environment**
Java 25
Spring AI version: 2.0.0-RC2
OpenRouter as provider

**Steps to reproduce**
It is hard to provide steps to reproduce, the specification is here - https://openrouter.ai/docs/api_reference/responses/error-handling

Would require an account on OpenRouter and a state of provider where it's not able to respond.

**Expected behavior**
These cases should map to corresponding exceptions from openai-java-core package.

Unfortunately very hard to provide something like this.

jianran:
OpenAI-compatible providers such as OpenRouter can return a successful HTTP response whose body contains an `error` extension and omits `choices`. Detect that payload before accessing required completion fields and translate its status code into the corresponding openai-java exception type, preserving the provider message for callers.

A regression test covers an embedded 502 response without choices.

Fixes #6900

Tested with: `./mvnw -pl models/spring-ai-openai -Dtest=OpenAiChatModelTests -DargLine=-Djdk.attach.allowAttachSelf=true test`

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -38,7 +38,17 @@ import com.openai.client.OpenAIClientAsync;
 import com.openai.core.JsonValue;
 import com.openai.core.RequestOptions;
 import com.openai.core.http.AsyncStreamResponse;
+import com.openai.core.http.Headers;
+import com.openai.errors.BadRequestException;
+import com.openai.errors.InternalServerException;
+import com.openai.errors.NotFoundException;
 import com.openai.errors.OpenAIInvalidDataException;
+import com.openai.errors.PermissionDeniedException;
+import com.openai.errors.RateLimitException;
+import com.openai.errors.UnauthorizedException;
+import com.openai.errors.UnexpectedStatusCodeException;
+import com.openai.errors.UnprocessableEntityException;
+import com.openai.models.ErrorObject;
 import com.openai.models.FunctionDefinition;
 import com.openai.models.FunctionParameters;
 import com.openai.models.ReasoningEffort;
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks for this @jianran. This overlaps heavily with #6908, which was opened around the same time and takes the same approach (a `throwIfErrorResponse` check after `completions().create(...)` that maps an embedded `error` payload to the matching openai-java exception). #6908 covers a bit more ground: it also guards on `_choices().isMissing()` before inspecting the error, and it adds tests for the 429, missing-code, and valid-choices cases in addition to the 502 path you covered. To avoid duplicate work we're going to consolidate on #6908 and close this one. Really appreciate you digging into #6900, and please keep the contributions coming.

PR #6912, opened 2026-09-03
Fix `FilterExpressionTextParser` for keys starting with the `where` letters

### LINKED ISSUE #6909: `FilterExpressionTextParser` fails to parse expressions whose first key starts with the letters `where`
**Bug description**

`FilterExpressionTextParser` prepends the compulsory `WHERE` keyword only when the expression does not already start with it, but the check is a plain string prefix test:

```java
if (!textFilterExpression.toUpperCase(Locale.ROOT).startsWith(WHERE_PREFIX)) {
```

A metadata key that merely starts with the letters `where` — such as `whereabouts` or `where_clause` — satisfies the prefix check, so `WHERE` is never prepended. The lexer then tokenizes the identifier as `IDENTIFIER` rather than the `WHERE` token required by the grammar rule `where : WHERE booleanExpression EOF`, and parsing fails.

**Environment**

- Spring AI `2.0.2-SNAPSHOT`, `main` at `0933bd0`
- Module: `spring-ai-vector-store`
- Class: `org.springframework.ai.vectorstore.filter.FilterExpressionTextParser`

**Steps to reproduce**

```java
new FilterExpressionTextParser().parse("whereabouts == 'BG'");
```

fails with:

```text
FilterExpressionTextParser$FilterExpressionParseException
Caused by: org.antlr.v4.runtime.InputMismatchException
	at org.springframework.ai.vectorstore.filter.antlr4.FiltersParser.where(FiltersParser.java:186)
```

while e.g. `parse("country == 'BG'")` works.

**Expected behavior**

`whereabouts == 'BG'` parses to `EQ(Key("whereabouts"), Value("BG"))` like any other key. The prefix check should only match the standalone `WHERE` keyword (e.g. `^\s*where\b`, case-insensitive), not any identifier that happens to start with those letters.

harrisleesh:
## Motivation

`FilterExpressionTextParser` prepends the compulsory `WHERE` keyword using a plain string prefix check, so an expression whose first key merely starts with the letters `where` — e.g. `whereabouts == 'BG'` or `where_clause == 'BG'` — never gets the keyword prepended and fails with an `InputMismatchException`.

## Changes

- Only skip prepending when the expression actually starts with the standalone `WHERE` keyword (`^\s*where\b`, case-insensitive), not any identifier starting with those letters.
- Added a regression test covering `whereabouts` and `where_clause` keys. It fails on `main` and passes with the fix; the full `spring-ai-vector-store` module test suite passes (163 tests).

Fixes #6909

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
@@ -18,10 +18,10 @@ package org.springframework.ai.vectorstore.filter;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.regex.Pattern;
 
 import org.antlr.v4.runtime.ANTLRErrorStrategy;
 import org.antlr.v4.runtime.BailErrorStrategy;
@@ -109,6 +109,8 @@ public class FilterExpressionTextParser {
 
 	private static final String WHERE_PREFIX = "WHERE";
 
+	private static final Pattern WHERE_PREFIX_PATTERN = Pattern.compile("^\\s*where\\b", Pattern.CASE_INSENSITIVE);
+
 	private final DescriptiveErrorListener errorListener;
 
 	private final ANTLRErrorStrategy errorHandler;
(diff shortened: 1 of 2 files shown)


CryoThrust:
One compatibility edge to consider: the new pattern uses `CASE_INSENSITIVE`, so inputs such as `WhErE country == 'BG'` are treated as already prefixed. However, `Filters.g4` currently defines `WHERE` as only `'WHERE' | 'where'`; the parser then rejects mixed-case variants because the prefix is not added. Could the implementation either limit matching to the grammar-supported spellings or normalize a recognized standalone keyword to `WHERE` before parsing? A regression test for mixed-case input would make the intended contract explicit.

harrisleesh:
Good catch, thanks! You're right that a mixed-case spelling like `WhErE` matched the case-insensitive pattern but was then rejected by the lexer, which only knows the `WHERE`/`where` spellings (the previous `toUpperCase().startsWith(...)` check had the same blind spot). I've pushed a commit that normalizes a matched keyword to the uppercase spelling instead of leaving it untouched, so `WhErE country == 'BG'` now parses as well. Added a test covering `WhErE` and `Where`.

PR #6913, opened 2026-09-03
Copy the metadata map in `Document.mutate()`

### LINKED ISSUE #6910: `Document.mutate()` shares the metadata map with the original document
**Bug description**

`Document.mutate()` is documented as "allowing selective modification without altering the original", but it passes the document's internal metadata map into the new builder by reference:

```java
public Builder mutate() {
    return new Builder().id(this.id).text(this.text).media(this.media).metadata(this.metadata).score(this.score);
}
```

`Builder.metadata(String key, Object value)` then calls `put` directly on that map, so mutating the builder mutates the original `Document`'s metadata.

**Environment**

- Spring AI `2.0.2-SNAPSHOT`, `main` at `0933bd0`
- Module: `spring-ai-commons`
- Class: `org.springframework.ai.document.Document`

**Steps to reproduce**

```java
Document original = Document.builder().text("Original text").metadata("original", "value").build();

Document mutated = original.mutate().metadata("new", "metadata").build();

// fails: original now contains the "new" key as well
assertThat(original.getMetadata()).doesNotContainKey("new");
```

**Expected behavior**

`mutate()` seeds the builder with a copy of the metadata map (`new HashMap<>(this.metadata)`), so builder modifications never leak into the original document. Note that `mutate()` for the other fields already behaves this way because they are immutable or reassigned on build.

harrisleesh:
## Motivation

`Document.mutate()` is documented as "allowing selective modification without altering the original", but it passes the document's internal metadata map to the new builder by reference. `Builder.metadata(key, value)` puts directly into that map, so mutating the builder silently mutates the original document's metadata.

## Changes

- `mutate()` now seeds the builder with a copy of the metadata map.
- Added a regression test asserting the original document is unchanged after `mutate().metadata(...)`. It fails on `main` and passes with the fix; the full `spring-ai-commons` module test suite passes (237 tests).

Fixes #6910

### DIFF
--- a/spring-ai-commons/src/main/java/org/springframework/ai/document/Document.java
+++ b/spring-ai-commons/src/main/java/org/springframework/ai/document/Document.java
@@ -357,7 +357,11 @@ public class Document {
 	 * @return a builder initialised from this document's state
 	 */
 	public Builder mutate() {
-		return new Builder().id(this.id).text(this.text).media(this.media).metadata(this.metadata).score(this.score);
+		return new Builder().id(this.id)
+			.text(this.text)
+			.media(this.media)
+			.metadata(new HashMap<>(this.metadata))
+			.score(this.score);
 	}
 
 	@Override
--- a/spring-ai-commons/src/test/java/org/springframework/ai/document/DocumentTests.java
+++ b/spring-ai-commons/src/test/java/org/springframework/ai/document/DocumentTests.java
@@ -280,6 +280,16 @@ public class DocumentTests {
 	}
 
 	@Test
+	void testMutateDoesNotAlterOriginalMetadata() {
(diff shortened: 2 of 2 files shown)


sobychacko:
Thanks for the fix @harrisleesh. This overlaps with #6901, which was opened a day earlier and targets the same `Document.mutate()` metadata aliasing bug (issue #6910). Both use a defensive `new HashMap<>(...)` copy. #6901 places the copy in `Builder.metadata(Map)` rather than at the `mutate()` call site, so it also covers external callers passing a map they still own, and it adds a second test for the `HashSet` eviction case (since metadata is part of `equals`/`hashCode`). We are going to keep #6901 for that broader coverage.

I am going to close this one as a duplicate. Thanks for taking the time, and please keep the contributions coming.

PR #6917, opened 2026-09-04
Fix MariaDB schema validation for default schema

jhpark1227:
## Problem

`MariaDBSchemaValidator` supports leaving the schema name unset, in which case the
table is expected to live in the connection's default schema. That path never worked.

`isTableExists` bound the literal string `"SCHEMA()"` as a query *parameter*:

```java
this.jdbcTemplate.queryForObject(sql, Integer.class,
        (schemaName == null) ? "SCHEMA()" : schemaName, tableName);
```

A bind parameter is a value, not SQL, so the predicate evaluated as
`TABLE_SCHEMA = 'SCHEMA()'` and matched nothing. The column lookup in
`validateTableSchema` had the same defect from the other direction: it bound
`schemaName` while it was still `null`, so `TABLE_SCHEMA = NULL` would have matched
nothing either, had it been reached — the table check above already threw first.

The result is that with `schemaValidation` enabled and no `schemaName` configured,
`MariaDBVectorStore.afterPropertiesSet()` fails for a table that is present:

```
java.lang.IllegalStateException: Table 'vector_store' does not exist in schema 'null'
```

The workaround is to set `schemaName` explicitly, which should not be required.
Note the `null` in the message — the log and exception messages on this path reported
the unresolved name rather than the schema actually being validated. The `CREATE TABLE`
statement suggested when a column is missing had the opposite problem: it dropped the
schema qualifier entirely, so it never named the schema being validated either.

## Solution

Resolve the schema name once, up front, before any lookup:

```java
private String resolveSchemaName(@Nullable String schemaName) {
    if (schemaName != null) {
        return schemaName;
    }
    String currentSchema = this.jdbcTemplate.queryForObject("SELECT SCHEMA()", String.class);
    Assert.state(currentSchema != null, "No schema name configured and no schema selected on the connection");
    return currentSchema;
}
```

`SELECT SCHEMA()` is evaluated by the server as SQL, which is what the original code
intended. A connection with no database selected returns `NULL` there; that is a
configuration error rather than a missing table, so it fails with a distinct message
instead of being reported as `schema 'null'`.

The resolved name is then used for both `INFORMATION_SCHEMA` queries, for the log and
exception messages, and for the suggested DDL, which is now always schema-qualified.
`isTableExists` now takes a non-null schema name, so the `@Nullable` handling exists in
exactly one place.

This also drops two stray arguments from the vector support probe:

```java
this.jdbcTemplate.queryForObject("SELECT vec_distance_euclidean(x'0000803f', x'0000803f')",
        Integer.class, schemaName, tableName);
```

That statement has no placeholders, so `schemaName` and `tableName` were bound to
nothing. Harmless, but misleading to read.

## Testing

`MariaDBSchemaValidatorIT` starts a MariaDB container, creates a `vector_store` table
in the container's default schema and validates it three ways: with no schema name,
with the schema name given explicitly, and against a table that does not exist — the
last asserting that the message now names the resolved schema.

The IT needs `spring-boot-testcontainers` for `@ServiceConnection`, added as a test
dependency. It requires only Docker, no API key.

### DIFF
--- a/vector-stores/spring-ai-mariadb-store/pom.xml
+++ b/vector-stores/spring-ai-mariadb-store/pom.xml
@@ -53,6 +53,12 @@
 		</dependency>
 
 		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-testcontainers</artifactId>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
 			<groupId>org.testcontainers</groupId>
 			<artifactId>testcontainers</artifactId>
 			<scope>test</scope>
--- a/vector-stores/spring-ai-mariadb-store/src/main/java/org/springframework/ai/vectorstore/mariadb/MariaDBSchemaValidator.java
+++ b/vector-stores/spring-ai-mariadb-store/src/main/java/org/springframework/ai/vectorstore/mariadb/MariaDBSchemaValidator.java
@@ -45,13 +45,21 @@ public class MariaDBSchemaValidator {
 		this.jdbcTemplate = jdbcTemplate;
 	}
 
-	private boolean isTableExists(@Nullable String schemaName, String tableName) {
+	private String resolveSchemaName(@Nullable String schemaName) {
+		if (schemaName != null) {
+			return schemaName;
+		}
+		String currentSchema = this.jdbcTemplate.queryForObject("SELECT SCHEMA()", String.class);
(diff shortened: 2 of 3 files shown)


ilayaperumalg (approved):

ilayaperumalg:
@jhpark1227 Thanks for the PR fixing this issue!

PR #6919, opened 2026-09-05
Make DefaultUsage nativeUsage serialization tolerant

### LINKED ISSUE #2535: DefaultUsage#nativeUsage can crash SimpleLoggingAdvisor

### LINKED ISSUE #5682: No serializer found for class software.amazon.awssdk.services.bedrockruntime.model.TokenUsage

sobychacko:
nativeUsage holds an opaque provider object that Jackson may not be able to serialize, which breaks any JSON view of a ChatResponse such as SimpleLoggerAdvisor output or a controller response.

Serialize the value into a TokenBuffer first and fall back to its toString() representation if that fails, so the enclosing response always serializes. Output for already serializable values is unchanged.

Fixes #2535
Fixes #5682

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/metadata/DefaultUsage.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/metadata/DefaultUsage.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import org.jspecify.annotations.Nullable;
+import tools.jackson.databind.annotation.JsonSerialize;
 
 /**
  * Default implementation of the {@link Usage} interface.
@@ -157,6 +158,7 @@ public class DefaultUsage implements Usage {
 	@Override
 	@JsonProperty("nativeUsage")
 	@JsonInclude(JsonInclude.Include.NON_NULL)
+	@JsonSerialize(using = NativeUsageSerializer.class)
 	public @Nullable Object getNativeUsage() {
 		return this.nativeUsage;
 	}
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/metadata/NativeUsageSerializer.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/metadata/NativeUsageSerializer.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2023-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.ai.chat.metadata;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import tools.jackson.core.JsonGenerator;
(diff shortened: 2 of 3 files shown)


CryoThrust:
The tolerant serializer addresses the failure mode well. One follow-up concern is the fallback contract: `String.valueOf(value)` is completely provider-controlled and unbounded, yet this value can be emitted in a controller response or `SimpleLoggerAdvisor` log. Some SDK/protobuf `toString()` implementations include verbose diagnostic fields (and provider objects may contain credential-like metadata), so the fallback could turn a serialization failure into a large or sensitive response.

Could we make the fallback policy explicit, for example by bounding the rendered value to a small maximum and/or using a safe type marker when the value is not a scalar usage representation? A regression test with a `toString()` that returns a long credential-like string would make the intended API/logging boundary clear. The normal serializable-object path can remain unchanged.

CryoThrust:
The only failing check is `MCP integration tests`; the failure is isolated to `mcp/transport/mcp-spring-webflux` `WebFluxStreamableIT.testRootsSuccess(String)[2]` with a 500 from `POST /otherPath/mcp/message`. The same run reports 540 tests with one error, while the branch build and DCO checks pass. This appears unrelated to `spring-ai-model`/`DefaultUsage`; a rerun or maintainer-side MCP test diagnosis should be enough to separate the gate from this change.

sobychacko:
@CryoThrust  Thanks for looking at this and for the feedback.

After thinking through this, I don't think a cap is needed since `nativeUsage` is always the provider's own usage record (`TokenUsage` for Bedrock, `UsageMetadata` for Google, the `Usage` classes in the Anthropic and OpenAI SDKs), and `toString()` on those is a few labeled ints, the same numbers already in promptTokens/completionTokens next to it. The field is `@JsonProperty` on purpose and already serializes in full for most providers. Bedrock and Google users were just getting an exception instead.

Picking a limit would be a guess, and if we ever hit it we'd be cutting off the only readable output those users get. If some SDK's usage type turns out to have a huge toString(), we can add truncation then. What do you think? Thanks.

ilayaperumalg (approved):

PR #6931, opened 2026-09-07
Handle missing fields in `ChunkMerger`

### LINKED ISSUE #6928: ChunkMerger.chunkToChatCompletion throws OpenAIInvalidDataException when a streamed chunk omits `id` or `model` (only `created` is guarded)
### Bug description                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                  
  `OpenAiChatModel.ChunkMerger.chunkToChatCompletion` reads three required fields off                                                                                                                                                                             
  `ChatCompletionChunk`, but only guards one of them:                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                  
  ```java                                                                                                                                                                                                                                                         
  // OpenAiChatModel.java:1326-1333 (2.0.1)                                                                                                                                                                                                                       
  return ChatCompletion.builder()                                                                                                                                                                                                                                 
      .id(chunk.id())               // unguarded                                                                                                                                                                                                                  
      .choices(choices)                                                                                                                                                                                                                                           
      .created(getCreated(chunk))   // guarded                                                                                                                                                                                                                    
      .model(chunk.model())         // unguarded                                                                                                                                                                                                                  
      .usage(...)                                                                                                                                                                                                                                                 
      .putAllAdditionalProperties(chunk._additionalProperties())                                                                                                                                                                                                  
      .build();                                                                                                                                                                                                                                                   
  // OpenAiChatModel.java:1340-1347                                                                                                                                                                                                                               
  private static long getCreated(ChatCompletionChunk chunk) {                                                                                                                                                                                                     
      try {                                                                                                                                                                                                                                                       
          return chunk.created();                                                                                                                                                                                                                                 
      }                                                                                                                                                                                                                                                           
      catch (OpenAIInvalidDataException ex) {                                                                                                                                                                                                                     
          return 0L;                                                                                                                                                                                                                                              
      }                                                                                                                                                                                                                                                           
  }                                                                                                                                                                                                                                                               
  ```                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                  
  All three accessors go through `JsonField.getRequired(...)` in `openai-java-core` and therefore                                                                                                                                                                 
  throw `OpenAIInvalidDataException` when the field is absent (verified against                                                                                                                                                                                   
  `openai-java-core` 4.49.0 bytecode: `ChatCompletionChunk.id()`, `.model()` and `.created()` all                                                                                                                                                                 
  call `JsonField.getRequired$openai_java_core`).                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                  
  The existence of `getCreated()` shows the code already anticipates providers that omit required                                                                                                                                                                 
  fields — the same defensive treatment is simply missing for `id` and `model`.                                                                                                                                                                                   
                                                                                                                                                                                                                                                                  
  **Impact:** with an OpenAI-compatible gateway whose final (usage-only) chunk carries no `id`, the                                                                                                                                                               
  entire stream fails **after the assistant message has been fully delivered**. In our case 95                                                                                                                                                                    
  content deltas had already been emitted and the answer was complete; the exception surfaces at the                                                                                                                                                              
  very end, `MessageAggregator` logs `Aggregation Error`, and the `Flux` terminates with `onError`.                                                                                                                                                               
  Downstream that turns a fully successful request into a failed one (error event pushed to clients,                                                                                                                                                              
  error-rate metrics polluted).                                                                                                                                                                                                                                   

 ### Environment                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                  
  - Spring AI **2.0.1**                                                                                                                                                                                                                                           
  - `openai-java-core` **4.49.0**                                                                                                                                                                                                                                 
  - Spring Boot 4.1.0, Java 25                                                                                                                                                                                                                                    
  - Model: a DeepSeek model served through a third-party OpenAI-compatible gateway                                                                                                                                                                                
    (`ChatModel` = `OpenAiChatModel`, streaming, tool calling enabled)                                                                                                                                                                                            
                                                                                                                                                                                                                                                                  
  ### Steps to reproduce                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                  
  Any OpenAI-compatible endpoint whose final SSE chunk omits `id` triggers it. A unit test against                                                                                                                                                                
  `ChunkMerger` is probably the cleanest repro: feed it a `ChatCompletionChunk` whose `_id()` is                                                                                                                                                                  
  absent (e.g. deserialized from `{"choices":[],"usage":{...}}`) and observe the exception.                                                                                                                                                                       
                                                                                                                                                                                                                                                                  
  ### Stack trace                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                  
  ```                                                                                                                                                                                                                                                             
  com.openai.errors.OpenAIInvalidDataException: `id` is not set                                                                                                                                                                                                   
      at com.openai.core.JsonField.getRequired$openai_java_core(Values.kt:174)                                                                                                                                                                                    
      at com.openai.models.chat.completions.ChatCompletionChunk.id(ChatCompletionChunk.kt:91)                                                                                                                                                                     
      at org.springframework.ai.openai.OpenAiChatModel$ChunkMerger.chunkToChatCompletion(OpenAiChatModel.java:1327)                                                                                                                                               
      at reactor.core.publisher.FluxMap$MapConditionalSubscriber.onNext(FluxMap.java:208)                                                                                                                                                                         
      at reactor.core.publisher.FluxBufferPredicate$BufferPredicateSubscriber.onNextNewBuffer(FluxBufferPredicate.java:316)                                                                                                                                       
      ...                                                                                                                                                                                                                                                         
      at com.openai.core.http.TrackedHandler.onNext(PhantomReachableClosingAsyncStreamResponse.kt:53)                                                                                                                                                             
  ```                                                                                                                                                                                                                                                             

 ### Suggested fix                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                  
  Apply the `getCreated` pattern to the other two required fields, e.g.                                                                                                                                                                                           
                                                                                                                                                                                                                                                                  
  ```java                                                                                                                                                                                                                                                         
  private static String getId(ChatCompletionChunk chunk) {                                                                                                                                                                                                        
      try {                                                                                                                                                                                                                                                       
          return chunk.id();                                                                                                                                                                                                                                      
      }                                                                                                                                                                                                                                                           
      catch (OpenAIInvalidDataException ex) {                                                                                                                                                                                                                     
          return "";                                                                                                                                                                                                                                              
      }                                                                                                                                                                                                                                                           
  }                                                                                                                                                                                                                                                               
  ```                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                  
  `ChatCompletionChunk` also exposes `_id()` / `_model()` (raw `JsonField`) and `isValid()`, so the                                                                                                                                                               
  check can be done without relying on exceptions if that is preferred. Carrying over the id from an                                                                                                                                                              
  earlier chunk in the same buffered group would preserve more information than a placeholder, since                                                                                                                                                              
  `ChunkMerger` already has the whole group in hand.                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                  
  ### Related                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                  
  Same class, third variant of this failure mode: #6591 (continuation frames with `"id": ""` treated                                                                                                                                                              
  as new tool calls, fixed in 2.0.1) was about tool-call merging; this one is about required                                                                                                                                                                      
  top-level fields on the chunk itself. The underlying assumption — that every chunk conforms to the                                                                                                                                                              
  official OpenAI schema — does not hold for third-party OpenAI-compatible gateways, which is worth                                                                                                                                                               
  hardening across `ChunkMerger` rather than field by field.                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                  
  ### Workaround                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                  
  Not interceptable from application code (`ChunkMerger` sits inside `internalStream`). We tolerate                                                                                                                                                               
  it downstream instead: once a chunk has reported a `finish_reason` other than `TOOL_CALLS`, the                                                                                                                                                                 
  model has stopped generating and only a usage frame remains, so any later parse failure is treated                                                                                                                                                              
  as a benign end-of-stream condition (logged at WARN) rather than a request failure.

yesod-a:
ChatCompletionChunk fields can be absent in streamed responses from
OpenAI-compatible providers. Handle missing id and model consistently
with the existing fallback for created.

Adds a regression test using a deserialized chunk without those fields.

Fixes #6928

Tests: ./mvnw -pl models/spring-ai-openai -am `n  -Dtest=OpenAiChatModelChunkMergerTests `n  -Dsurefire.failIfNoSpecifiedTests=false test

### DIFF
--- a/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
+++ b/models/spring-ai-openai/src/main/java/org/springframework/ai/openai/OpenAiChatModel.java
@@ -1334,14 +1334,32 @@ public final class OpenAiChatModel implements ChatModel {
 			}).toList();
 
 			return ChatCompletion.builder()
-				.id(chunk.id())
+				.id(getId(chunk))
 				.choices(choices)
 				.created(getCreated(chunk))
-				.model(chunk.model())
+				.model(getModel(chunk))
 				.usage(chunk.usage()
 					.orElse(CompletionUsage.builder().promptTokens(0).completionTokens(0).totalTokens(0).build()))
 				.putAllAdditionalProperties(chunk._additionalProperties())
 				.build();
+		}
+
+		private static String getId(ChatCompletionChunk chunk) {
+			try {
+				return chunk.id();
+			}
+			catch (OpenAIInvalidDataException ex) {
+				return "";
+			}
+		}
+
+		private static String getModel(ChatCompletionChunk chunk) {
+			try {
+				return chunk.model();
+			}
+			catch (OpenAIInvalidDataException ex) {
+				return "";
+			}
 		}
 
 		/**
--- a/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelChunkMergerTests.java
+++ b/models/spring-ai-openai/src/test/java/org/springframework/ai/openai/OpenAiChatModelChunkMergerTests.java
@@ -17,7 +17,9 @@
 package org.springframework.ai.openai;
 
 import java.util.List;
+import java.util.Map;
 
+import com.openai.core.JsonValue;
 import com.openai.models.chat.completions.ChatCompletion;
 import com.openai.models.chat.completions.ChatCompletionChunk;
 import com.openai.models.chat.completions.ChatCompletionChunk.Choice;
@@ -169,6 +171,17 @@ class OpenAiChatModelChunkMergerTests {
 		assertThat(functionToolCall.id()).isEqualTo("call_1");
 		assertThat(functionToolCall.function().name()).isEqualTo("get_weather");
 		assertThat(functionToolCall.function().arguments()).isEmpty();
+	}
+
+	@Test // gh-6928
+	void chunkToChatCompletionUsesEmptyDefaultsForMissingIdAndModel() {
+		ChatCompletionChunk chunk = JsonValue.from(Map.of("choices", List.of(), "created", 1L))
+			.convert(ChatCompletionChunk.class);
+
+		ChatCompletion completion = OpenAiChatModel.ChunkMerger.chunkToChatCompletion(chunk);
+
+		assertThat(completion.id()).isEmpty();
+		assertThat(completion.model()).isEmpty();
 	}
 
 	private static ChatCompletionChunk chunk(@Nullable FinishReason finishReason, ToolCall... toolCalls) {


sobychacko:
Thanks for this fix, @yesod-a. This overlaps with #6936, which addresses the same issue (#6928) by making `id` and `model` fall back to empty strings in `ChunkMerger.chunkToChatCompletion` when a provider omits them. The two differ mainly in mechanism: your PR catches `OpenAIInvalidDataException` around the accessors, while #6936 reads the raw optional fields via `chunk._id().asKnown().orElse("")` and `chunk._model().asKnown().orElse("")`, avoiding the exception path, and its test also asserts the `created` fallback. Given that, we plan to move forward with #6936 and close this one. We really appreciate you digging into this, and more PR's are welcomed. Thanks!

PR #6941, opened 2026-09-08
Clean `MapOutputConverter` responses with the shared `ResponseTextCleaner`

anjeongkyun:
`MapOutputConverter` strips code fences with its own inline check:

```java
if (text.startsWith("```json") && text.endsWith("```")) {
    text = text.substring(7, text.length() - 3);
}
```

That only matches one exact shape. `BeanOutputConverter` in the same package already goes through `ResponseTextCleaner`, which handles the rest. Running both over the shapes models actually emit:

| response | `MapOutputConverter` | `BeanOutputConverter` |
|---|---|---|
| ` ```json\n{...}\n``` ` | ok | ok |
| `\n` before/after the fence | **MessageConversionException** | ok |
| trailing spaces after the fence | **MessageConversionException** | ok |
| ` ```JSON ` | **MessageConversionException** | ok |
| ` ``` ` with no language tag | **MessageConversionException** | ok |
| bare JSON, no fence | ok | ok |

A trailing newline is enough to break it, and that is a very ordinary thing for a model to emit. `<think>` blocks aren't handled either, since `ThinkingTagCleaner` is part of the same chain.

So this drops the inline check and runs the response through the cleaner instead. The default chain lived in a private method on `BeanOutputConverter`; I moved it to `ResponseTextCleaner.defaultCleaner()` so both use one definition rather than a second copy drifting from the first, and added a constructor taking a custom cleaner, mirroring `BeanOutputConverter`.

`MapOutputConverter` had no test class, which is presumably how this survived. Added one covering the fenced shapes above, a thinking tag, and a custom cleaner.

`mvn -pl spring-ai-model test` → 844 tests, green.

### Related, not included

`ListOutputConverter` doesn't clean at all, and it fails more quietly — the fence markers become list elements:

```
"```\nfoo, bar, baz\n```"     -> [```, foo, bar, baz, ```]
"```csv\nfoo, bar, baz\n```"  -> [```csv, foo, bar, baz, ```]
```

I left it out because every cleaner in the chain trims, and `ListOutputConverterTest.csvWithOnlyWhitespace` asserts that `"   \t\n   "` converts to a single blank element rather than an empty list. Fixing the fences changes that, so it seemed like your call rather than mine. Happy to do it in a follow-up if you want the behaviour changed, or to scope a fence-only cleaner that preserves it.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/converter/BeanOutputConverter.java
@@ -161,12 +161,7 @@ public class BeanOutputConverter<T> implements StructuredOutputConverter<T> {
 	 * @return a composite text cleaner with default cleaning strategies
 	 */
 	private static ResponseTextCleaner createDefaultTextCleaner() {
-		return CompositeResponseTextCleaner.builder()
-			.addCleaner(new WhitespaceCleaner())
-			.addCleaner(new ThinkingTagCleaner())
-			.addCleaner(new MarkdownCodeBlockCleaner())
-			.addCleaner(new WhitespaceCleaner()) // Final trim after all cleanups
-			.build();
+		return ResponseTextCleaner.defaultCleaner();
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


sobychacko:
Closing as fully covered by #6975, which you stacked on top of this one. Reviewing one PR is simpler than two. Thanks for the efforts!

anjeongkyun:
Makes sense, #6975 carries both commits so nothing is lost by closing this one. Thanks for taking a look.

PR #6943, opened 2026-09-09
docs: document OpenAI reasoning effort property

CryoThrust:
## Summary\n\nDocument the existing spring.ai.openai.chat.reasoning-effort property in the OpenAI chat options table.\n\nThe documentation explains that the value is forwarded to the OpenAI Chat Completions API and must be supported by the selected model. It also calls out the tool-calling compatibility case where the value none may be required for models that do not support tools with reasoning enabled.\n\nThis is documentation-only; no runtime behavior changes.


CryoThrust:
This is a documentation-only change in the OpenAI chat options table. The property is already implemented and forwarded by OpenAiChatModel; this PR documents its purpose and model/API compatibility without changing runtime behavior. DCO is passing.

ilayaperumalg (approved):

CryoThrust:
Thanks for reviewing. DCO and the required checks are green; this remains a documentation-only change and is ready for merge.

ilayaperumalg:
@CryoThrust Thanks for the PR!

PR #6949, opened 2026-09-09
Avoid duplicate tool error messages

### LINKED ISSUE #6948: [Bug - MCP server] @McpTool error messages are emitted twice when the thrown exception has no cause
**Bug description**

The `@McpTool` method callbacks build their error text by concatenating the exception's message with its root cause's message:

```java
// AbstractSyncMcpToolMethodCallback#createSyncErrorResult
protected CallToolResult createSyncErrorResult(Exception e) {
    Throwable rootCause = findCauseUsingPlainJava(e);
    return CallToolResult.builder()
        .isError(true)
        .addTextContent(e.getMessage() + System.lineSeparator() + rootCause.getMessage())
        .build();
}
```

`findCauseUsingPlainJava` starts from the throwable itself and only walks `getCause()`:

```java
Throwable rootCause = throwable;
while (rootCause.getCause() != null && rootCause.getCause() != rootCause) {
    rootCause = rootCause.getCause();
}
return rootCause;
```

So for an exception with **no cause** it returns the exception itself, `rootCause.getMessage()` is the same string as `e.getMessage()`, and the client receives every such error twice, separated by a line separator.

This affects the most ordinary case there is: a tool that validates its own input and throws a bare `IllegalArgumentException` carrying an actionable message.

Affected paths:

- `SyncMcpToolMethodCallback` / `SyncStatelessMcpToolMethodCallback` → `AbstractSyncMcpToolMethodCallback#createSyncErrorResult`
- `AsyncMcpToolMethodCallback` / `AsyncStatelessMcpToolMethodCallback` → `AbstractAsyncMcpToolMethodCallback#createAsyncErrorResult` (identical concatenation)

Tool methods that *return* a reactive type take a different path — `AbstractAsyncMcpToolMethodCallback#toErrorResultOrPropagate` — which emits the message once, as `"Error invoking method: %s"`. So the same server currently formats tool errors two different ways depending on a tool's return type, and only one of them duplicates.

**Environment**

Spring AI 2.0.1. Also verified still present on `main` (`AbstractSyncMcpToolMethodCallback.java` is unchanged), so this is not fixed by upgrading.

For context: #6456 / #6534 reworked *which* exceptions reach `createSyncErrorResult` (`McpError` and `UndeclaredThrowableException` now propagate instead of being converted), but left the message construction untouched.

**Steps to reproduce**

```java
@Service
public class ToolService {

    @McpTool(name = "find_thing", description = "Looks something up")
    public String findThing(String name) {
        throw new IllegalArgumentException("No thing named '" + name + "' found. Use list_things to see valid names.");
    }
}
```

Call the tool. `content[0].text` is:

```
No thing named 'x' found. Use list_things to see valid names.
No thing named 'x' found. Use list_things to see valid names.
```

**Expected behavior**

The message once:

```
No thing named 'x' found. Use list_things to see valid names.
```

Appending the root cause is genuinely useful when the thrown exception wraps another, so the fix is to append it only when it adds information — e.g.:

```java
protected CallToolResult createSyncErrorResult(Exception e) {
    Throwable rootCause = findCauseUsingPlainJava(e);
    String text = (rootCause == e || Objects.equals(rootCause.getMessage(), e.getMessage()))
            ? e.getMessage()
            : e.getMessage() + System.lineSeparator() + rootCause.getMessage();
    return CallToolResult.builder().isError(true).addTextContent(text).build();
}
```

and the same in `createAsyncErrorResult`.

ReuDa:
The `@McpTool` method callbacks built their error text by appending the root cause message to the exception message.
findCauseUsingPlainJava returns the throwable itself when it has no cause, so an exception thrown directly by a tool - the common case for input validation - was reported twice, separated by a line separator.

Append the root cause message only when it adds information, via a shared createErrorResultMessage helper used by both createSyncErrorResult and createAsyncErrorResult.

Fixes #6948

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractAsyncMcpToolMethodCallback.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractAsyncMcpToolMethodCallback.java
@@ -181,11 +181,7 @@ public abstract class AbstractAsyncMcpToolMethodCallback<T, RC extends McpReques
 	 * @return A Mono<CallToolResult> representing the error
 	 */
 	protected Mono<CallToolResult> createAsyncErrorResult(Exception e) {
-		Throwable rootCause = findCauseUsingPlainJava(e);
-		return Mono.just(CallToolResult.builder()
-			.isError(true)
-			.addTextContent(e.getMessage() + System.lineSeparator() + rootCause.getMessage())
-			.build());
+		return Mono.just(CallToolResult.builder().isError(true).addTextContent(createErrorResultMessage(e)).build());
 	}
 
 	/**
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
@@ -232,15 +232,24 @@ public abstract class AbstractMcpToolMethodCallback<T, RC extends McpRequestCont
 	 * Creates the message for an error {@link CallToolResult}, appending the root cause
 	 * message only when it adds information. An exception with no cause is its own root
 	 * cause, so appending unconditionally would report the same message twice.
+	 * <p>
+	 * Messages may be {@code null}: an exception thrown without one. Only the non-null
+	 * side is used, and when neither side has a message the exception's class name is
+	 * returned, since the text of an error result must not be null.
 	 * @param e The exception that occurred
-	 * @return The error message string
+	 * @return The error message string, never {@code null}
 	 */
 	protected String createErrorResultMessage(Throwable e) {
 		Throwable rootCause = findCauseUsingPlainJava(e);
-		if (rootCause == e || Objects.equals(rootCause.getMessage(), e.getMessage())) {
-			return e.getMessage();
+		String message = e.getMessage();
+		String rootCauseMessage = (rootCause == e) ? null : rootCause.getMessage();
+		if (rootCauseMessage == null || rootCauseMessage.equals(message)) {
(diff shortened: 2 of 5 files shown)


CryoThrust (commented):
The shared helper is a good direction and covers the duplicate/no-cause case. One edge case from the issue matrix remains: when the outer exception or root cause has a null message, string concatenation currently renders the literal `null` (and can produce an unnecessary separator). Could you add a focused test and decide the desired fallback, for example using the non-null side when only one message exists and an empty/safe message when both are null? This keeps the error result deterministic for ordinary exceptions without messages.

ReuDa:
Good catch, thanks — it turned out to be worse than a cosmetic `null`. For an exception with no message *and* no cause, the helper returned `null`, and `CallToolResult.Builder#addTextContent` rejects that with `IllegalArgumentException: text must not be null`. That happens inside the `catch`, so it escaped `apply()` entirely: a message-less exception became a hard failure instead of an error result, on both the sync and async paths. My first commit introduced that, so this was well worth flagging.

Pushed a second commit:

- only the non-null side is used, so a cause without a message no longer appends `\nnull`
- when neither the exception nor its root cause has a message, it falls back to `e.getClass().getName()` rather than an empty string. An empty error body tells the model nothing, whereas `java.lang.IllegalStateException` is at least actionable — happy to switch to `""` if you'd prefer to keep it minimal.
- three tests each in the sync and async exception-handling classes; I checked all six fail against the previous commit, the no-message one as the escaping `IllegalArgumentException` described above.

One case I deliberately left out: `new IllegalStateException(cause)` derives its message from `cause.toString()`, so the root message still appears twice in different forms and the equality check can't catch it. Handling that needs a `contains`/`endsWith` heuristic, which felt like scope creep here — happy to look at it separately if you think it's worth doing.

CryoThrust (approved):
Reviewed the follow-up implementation. The shared formatter now avoids duplicate messages, handles null outer/root messages without emitting literal null, provides a non-null class-name fallback, and has focused sync/async coverage for the relevant matrix. No further issues found.

PR #6953, opened 2026-09-09
Copy message lists when saving in-memory chat history

### LINKED ISSUE #6952: `InMemoryChatMemoryRepository` retains the caller's mutable message list
**Bug description**

`InMemoryChatMemoryRepository.saveAll` stores the supplied `List<Message>` directly. Mutating that list after `saveAll` silently changes the stored conversation without another repository write. In contrast, `findByConversationId` already returns a defensive copy.

**Environment**

Spring AI `2.0.2-SNAPSHOT`, commit `7374b63fcb6fd0e1950e70c9e04192a6826f2f2a` (current `main` when checked before submission). Windows 11, JDK 17.0.20.1. No model service or vector store is required.

**Steps to reproduce**

Save a mutable list with `saveAll`, clear the original list, and retrieve the conversation. The saved message has disappeared even though no repository write or deletion occurred after the save.

Place this JUnit test in `spring-ai-model/src/test/java/org/springframework/ai/chat/memory/`:

```java
package org.springframework.ai.chat.memory;

import java.util.ArrayList;
import java.util.List;
import org.springframework.ai.chat.messages.Message;
import org.springframework.ai.chat.messages.UserMessage;

class ChatMemoryListIsolationTests {
    @Test
    void savedMessagesSurviveCallerListMutation() {
        var repository = new InMemoryChatMemoryRepository();
        Message message = new UserMessage("Hello");
        List<Message> messages = new ArrayList<>(List.of(message));
        repository.saveAll("conversation", messages);
        messages.clear();
        assertThat(repository.findByConversationId("conversation")).containsExactly(message);
    }
}
```

```text
./mvnw -pl spring-ai-model -am -Dtest=ChatMemoryListIsolationTests -Dsurefire.failIfNoSpecifiedTests=false test
```

On Windows, use `mvnw.cmd`. The equivalent regression added to the existing repository test class was executed locally against the unmodified implementation: 1 test, 1 failure, 0 errors; actual history was `[]`.

**Expected behavior**

Expected: the repository retains the saved message until a subsequent repository write or deletion. Actual: the stored history becomes empty. Applications that clear or reuse their message buffer can therefore lose or replace previously saved conversation history.

A defensive snapshot at the write boundary (`List.copyOf(messages)`) would isolate the stored list from subsequent structural mutations by the caller. The existing validation already rejects null lists and null elements. This does not attempt to deep-copy individual `Message` objects.

The proposed regression test is `InMemoryChatMemoryRepositoryTests.saveMessagesCopiesCallerOwnedList`. With the defensive copy, all 31 repository/message-window tests and the scoped reactor build pass. Source inspection, implementation, and test preparation used AI assistance; the patch was reviewed and approved by the contributor before submission.

pengmoubuaixuexi:
Fixes #6952

`InMemoryChatMemoryRepository.saveAll` retains a caller-owned list, so clearing or reusing that list after saving silently changes the stored conversation. Store a defensive list snapshot and add a regression test that clears the original list after saving and checks that the saved message remains.

The existing null validation and overwrite behavior are preserved. This snapshots the list structure; it does not deep-copy messages.

Validation: the new regression fails against the unmodified implementation. After the fix, all 31 repository/message-window tests pass. The scoped reactor build passes with formatting and Checkstyle checks:

```text
./mvnw.cmd -B -pl spring-ai-model -am -Dtest=InMemoryChatMemoryRepositoryTests,MessageWindowChatMemoryTests -Dsurefire.failIfNoSpecifiedTests=false clean package
```

This does not include the full repository test suite or external integration tests. Environment: Windows 11, JDK 17.0.20.1, upstream commit `7374b63fcb6fd0e1950e70c9e04192a6826f2f2a`.

AI assistance was used for source inspection, implementation, and test preparation. The contributor reviewed and approved the patch before submission.

### DIFF
--- a/spring-ai-model/src/main/java/org/springframework/ai/chat/memory/InMemoryChatMemoryRepository.java
+++ b/spring-ai-model/src/main/java/org/springframework/ai/chat/memory/InMemoryChatMemoryRepository.java
@@ -51,7 +51,7 @@ public final class InMemoryChatMemoryRepository implements ChatMemoryRepository 
 		Assert.hasText(conversationId, "conversationId cannot be null or empty");
 		Assert.notNull(messages, "messages cannot be null");
 		Assert.noNullElements(messages, "messages cannot contain null elements");
-		this.chatMemoryStore.put(conversationId, messages);
+		this.chatMemoryStore.put(conversationId, List.copyOf(messages));
 	}
 
 	@Override
--- a/spring-ai-model/src/test/java/org/springframework/ai/chat/memory/InMemoryChatMemoryRepositoryTests.java
+++ b/spring-ai-model/src/test/java/org/springframework/ai/chat/memory/InMemoryChatMemoryRepositoryTests.java
@@ -85,6 +85,18 @@ public class InMemoryChatMemoryRepositoryTests {
 	}
 
 	@Test
+	void saveMessagesCopiesCallerOwnedList() {
+		String conversationId = UUID.randomUUID().toString();
+		Message message = new UserMessage("Hello");
+		List<Message> messages = new ArrayList<>(List.of(message));
+
+		this.chatMemoryRepository.saveAll(conversationId, messages);
+		messages.clear();
+
+		assertThat(this.chatMemoryRepository.findByConversationId(conversationId)).containsExactly(message);
+	}
+
+	@Test
 	void findNonExistingConversation() {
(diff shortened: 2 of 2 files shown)


ilayaperumalg (approved):

ilayaperumalg:
@pengmoubuaixuexi Thanks for the PR fixing the issue.

PR #6955, opened 2026-09-10
Parse out-of-range integer filter literals as Long

### LINKED ISSUE #4705: FilterExpressionTextParser treats all number literals as Integer, failing on Long values

Lubaoshuai:
## Summary

`FilterExpressionTextParser` parses every unsuffixed integer literal with `Integer.valueOf`, so a filter containing a value outside the int range fails with `NumberFormatException` at parse time - independent of the vector store backend.

```
```

This blocks filtering on `long` metadata values (for example deleting documents by a `long` id), as reported in #4705.

## Changes

- `visitIntegerConstant` keeps parsing with `Integer.valueOf` first, so in-range literals keep producing `Integer` values (fully backward compatible with existing converters and tests).
- When the literal does not fit an `int`, it falls back to `Long.valueOf`, matching the behavior of the already-supported explicit `l`/`L` suffix (`visitLongConstant`).
- Literals beyond the `long` range still raise `NumberFormatException`, unchanged.

## Testing

- Added `testIntegerLiterals` in `FilterExpressionTextParserTests` covering: in-range literals stay `Integer`, `Long.MAX_VALUE`/`Long.MIN_VALUE` parse as `Long`, the explicit `L` suffix keeps working, and large literals work inside `IN` lists.
- `FilterExpressionTextParserTests`: 16/16 pass.
- Full `spring-ai-vector-store` module suite: 163 tests, 0 failures.

Fixes #4705

### DIFF
--- a/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
+++ b/spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/filter/FilterExpressionTextParser.java
@@ -231,7 +231,13 @@ public class FilterExpressionTextParser {
 
 		@Override
 		public Filter.Operand visitIntegerConstant(FiltersParser.IntegerConstantContext ctx) {
-			return new Filter.Value(Integer.valueOf(ctx.getText()));
+			String literal = ctx.getText();
+			try {
+				return new Filter.Value(Integer.valueOf(literal));
+			}
+			catch (NumberFormatException ex) {
+				return new Filter.Value(Long.valueOf(literal));
+			}
 		}
 
 		@Override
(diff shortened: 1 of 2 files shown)


ilayaperumalg (approved):

ilayaperumalg:
@Lubaoshuai Thanks for the PR! Rebased and merged via f320b16ab3efc4f03033e498936e65e123f6a9c3

PR #6956, opened 2026-09-10
Emit tool exception messages once when root cause adds nothing

### LINKED ISSUE #6948: [Bug - MCP server] @McpTool error messages are emitted twice when the thrown exception has no cause

Lubaoshuai:
## Summary

The `@McpTool` method callbacks build their error text by concatenating the thrown exception's message with its root cause's message. `findCauseUsingPlainJava` returns the exception itself when it has no cause, so a bare exception (for example an `IllegalArgumentException` carrying an actionable validation message) reaches the client **twice**, separated by a line separator - as reported in #6948.

## Changes

- Added `buildErrorText` to `AbstractMcpToolMethodCallback`: it appends the root cause message only when it adds information (root cause differs from the thrown exception and its message differs), and shared it from `AbstractSyncMcpToolMethodCallback#createSyncErrorResult` and `AbstractAsyncMcpToolMethodCallback#createAsyncErrorResult`, which duplicated the same concatenation.
- Wrapped exceptions with a distinct root cause keep the existing two-line output, so no diagnostic information is lost.

## Testing

- Added `exceptionWithoutCauseIsEmittedOnce` (asserts the exact single message) and `wrappedExceptionRetainsDistinctRootCauseMessage` (asserts both messages survive) to `SyncMcpToolMethodCallbackExceptionHandlingTests` and `AsyncMcpToolMethodCallbackExceptionHandlingTests`.
- Full `mcp/mcp-annotations` module suite: 1380 tests, 0 failures.

Fixes #6948

### DIFF
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractAsyncMcpToolMethodCallback.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractAsyncMcpToolMethodCallback.java
@@ -181,11 +181,7 @@ public abstract class AbstractAsyncMcpToolMethodCallback<T, RC extends McpReques
 	 * @return A Mono<CallToolResult> representing the error
 	 */
 	protected Mono<CallToolResult> createAsyncErrorResult(Exception e) {
-		Throwable rootCause = findCauseUsingPlainJava(e);
-		return Mono.just(CallToolResult.builder()
-			.isError(true)
-			.addTextContent(e.getMessage() + System.lineSeparator() + rootCause.getMessage())
-			.build());
+		return Mono.just(CallToolResult.builder().isError(true).addTextContent(buildErrorText(e)).build());
 	}
 
 	/**
--- a/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
+++ b/mcp/mcp-annotations/src/main/java/org/springframework/ai/mcp/annotation/method/tool/AbstractMcpToolMethodCallback.java
@@ -211,6 +211,21 @@ public abstract class AbstractMcpToolMethodCallback<T, RC extends McpRequestCont
 	}
 
 	/**
+	 * Builds the text conveyed to the model for an exception thrown by a tool method. The
+	 * root cause message is appended only when it adds information, so an exception
+	 * without a distinct root cause is not reported twice.
+	 * @param e The exception that occurred
+	 * @return The error text
+	 */
+	protected String buildErrorText(Exception e) {
+		Throwable rootCause = findCauseUsingPlainJava(e);
+		if (rootCause == e || Objects.equals(rootCause.getMessage(), e.getMessage())) {
+			return e.getMessage();
+		}
+		return e.getMessage() + System.lineSeparator() + rootCause.getMessage();
+	}
+
+	/**
 	 * Determines if the given parameter type is an exchange or context type that should
 	 * be injected. Subclasses must implement this method to specify which types are
 	 * considered exchange or context types.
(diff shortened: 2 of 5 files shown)


Lubaoshuai:
Friendly ping: the "PR Check" / "MCP Integration Tests" workflows are in `action_required` and need a maintainer workflow approval to run. DCO is green and the branch merges cleanly. Thanks!

sobychacko:
Hey @Lubaoshuai, thanks for taking the time on this one, and for the really clear write-up.

Turns out this covers the same ground as #6949, which fixes the same issue (#6948) in basically the same way: a shared helper on `AbstractMcpToolMethodCallback` that only tacks on the root-cause message when it actually adds something, used by both the sync and async paths. That one also handles the null-message cases (falling back to the class name when there's no message, since the result text can't be null), so we're going to go with it.

Closing this in favor of #6949, but I really appreciate the contribution and the analysis. Would love to see more PRs from you down the road!

PR #6957, opened 2026-09-10
make internalCall and internalStream private

controllerface:
Changes two methods which are clearly internal, from having `public `access to having `private` access.


controllerface:
closes #6700

PR #6974, opened 2026-09-12
Avoid calling multiple times the model to retrieve tool calls to perform while using structured output validation

nicolaskrier:
## Description

I noticed that the model is called repeatedly on each structured output validation attempt even when the response already contains tool calls to perform. In this scenario, a single call suffices, so this change stops the retry loop as soon as the response contains tool calls.

## Reproduction

I performed some successful tests of these changes with my [tools example](https://github.com/nicolaskrier/spring-ai-examples/tree/main/tools-example) by using Mistral AI and Ollama model providers. You can find below screenshots of what I observed on Phoenix.

### Behavior of my tools example before the fix

Screenshot 2026-09-12 at 17 23 05

### Behavior of my tools example after the fix

Screenshot 2026-09-12 at 18 24 49

## Testing

I added a unit test in `StructuredOutputValidationAdvisorTests` class.

## Notes

It could be worth testing with other model providers.

### DIFF
--- a/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/StructuredOutputValidationAdvisor.java
+++ b/spring-ai-client-chat/src/main/java/org/springframework/ai/chat/client/advisor/StructuredOutputValidationAdvisor.java
@@ -60,6 +60,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Christian Tzolov
  * @author Jewoo Shin
+ * @author Nicolas Krier
  */
 public final class StructuredOutputValidationAdvisor implements CallAdvisor, StreamAdvisor {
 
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks for the contribution @nicolaskrier, and for the detailed before/after tracing with your tools example. This overlaps with #6588, which was opened earlier (back in July) and fixes the same issue in `StructuredOutputValidationAdvisor`, where a tool-call response inside the validation retry loop caused the model to be re-invoked instead of handing the response to the outer tool-calling advisor.

The two approaches reach the same result. #6588 also happens to cover the harder case where a tool call arrives after a validation failure, in addition to the first-attempt case, so we are going to consolidate on that one and close this in its favor.

Really appreciate you digging into this, and please keep the contributions coming.

nicolaskrier:
Thank you for the feedback @sobychacko, and sorry for having missed the other PR related to this issue.

I hope the fix for this issue will be available soon, as it improves the speed of this kind of AI workflow.

PR #6977, opened 2026-09-13
Handle JSON null string while using structured output validation

nicolaskrier:
## Description

I noticed that structured output validation fails when a model emits "null" as its JSON output. These changes handle a null JSON node as a passed validation instead of triggering unnecessary retries.

## Reproduction

I performed some successful tests of these changes with my [tools example](https://github.com/nicolaskrier/spring-ai-examples/tree/main/tools-example) by using Ollama and Mistral AI model providers. As a first step, I applied #6974 fix. You can find below screenshots of what I observed on Phoenix.

### Behavior of my tools example before the fix with Ollama

Screenshot 2026-09-13 at 09 34 53

### Behavior of my tools example after the fix with Ollama

Screenshot 2026-09-13 at 11 13 57

### Behavior of my tools example before the fix with Mistral AI

Screenshot 2026-09-13 at 13 37 42

### Behavior of my tools example after the fix with Mistral AI

Screenshot 2026-09-13 at 13 44 54

## Testing

I added a unit test in `StructuredOutputValidationAdvisorTests` class.

## Notes

- It could be worth testing with other model providers.
- With Mistral AI, the model is invoked multiple times to retrieve tool calls to perform. This appears to be a separate issue worth investigating. I haven't yet identified why this occurs.
- The model may not follow the system prompt instructing it to return "null" if the tool call response is null. In that case, it returns a JSON response.


nicolaskrier:
I figured out that I had to modify [the system prompt](https://github.com/nicolaskrier/spring-ai-examples/blob/main/tools-example/src/main/resources/prompts/system-prompt.txt#L4) to have `mistral-medium-latest` behaving as intended. Now it works fine for Ollama and Mistral AI model providers. Nevertheless, I think having to define this kind of system prompt is cumbersome. I wonder if something could be done on Spring AI side to ease this behavior.

PR #6997, opened 2026-09-15
Introduce MessagePart in AbstractMessage

dimitarproynov:
Give AssistantMessage, UserMessage and ToolResponseMessage an ordered List<MessagePart> as their shared representation, so text, reasoning, tool calls, tool results and media can interleave and round-trip in the order a LLM provider produced them.

Introduce the MessagePart hierarchy (TextPart, ReasoningPart, ToolCallPart, ToolResultPart, MediaPart, UnknownPart). ReasoningPart and ToolCallPart can carry an opaque, provider-tagged payload (OpaquePayload) for data that must be replayed verbatim, such as an Anthropic thinking signature or a Gemini thought signature. UnknownPart keeps the raw JSON of blocks an adapter does not model yet, so nothing is dropped silently. Parts serialize with Jackson 3 using a "type" discriminator, with Media round-tripping through a DTO that preserves its data kind (bytes, string or URL).

The change is additive: existing constructors, builders and accessors (getText(), getToolCalls(), getMedia(), hasToolCalls(), getResponses()) keep working as views over the parts list, and messages built through legacy constructors or setters get their parts in the legacy order regardless of setter call order. Deprecated fields are kept for subclasses that read them directly. A message built without a text part has an empty, not null, text; ToolResponseMessage keeps its historical guarantee of a never-null text.

Equality and hashCode now compare parts, which brings media into UserMessage equality. Media therefore gains value equality over id, MIME type and data, with byte arrays compared by content and the name excluded, since it defaults to a random value per instance.

Streaming chat models can now deliver a part in increments. StreamingParts stamps a part with its provider content-block index and whether it is a delta; MessageAggregator merges indexed parts per block, appending deltas, replacing complete parts and payloads, and stripping the transport attributes from the aggregated message. Indexed parts are grouped by response id so that aggregating across tool-calling rounds does not merge unrelated groups together, with a heuristic fallback when a model stamps no id. Unindexed parts keep the previous concatenation behavior, including not carrying media on unindexed chunks.

AOT runtime hints register the new part types for reflection, and AiRuntimeHints' nested-class discovery no longer recurses forever on nested classes that extend their enclosing Jackson type. Docs gain a Message Parts section and a 2.1.0 upgrade-notes section covering the parts model, behavioral changes, deprecations and the streamToolCallDeltas flag.

Build and tests passed.
Integration tests for OpenAI - passed, others IT runs pending


ericbottard:
A couple of remarks/questions before I even read the code:

> messages built through legacy constructors or setters get their parts in the legacy order regardless of setter call order.

Is that really what we want (in the long run)? My understanding was that is broken (because we lost the interleaving). But maybe I'm not understanding what you're talking about here correctly.

> Equality and hashCode now compare parts

I'm not sure we want to invest too much in that space. I see no reason for those classes to be keys in hashmaps. nor for them to be found in lists, etc by "semantic equality". NOT overriding Object.equals()/hc() should be fine IMO, but maybe I'm missing something here cc @tzolov

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/part/ReasoningPart.java
```diff
@@ -0,0 +1,75 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * Model reasoning (thinking) that preceded or accompanied the answer.
+ * <p>
+ * The provider-neutral {@code text} and {@code summary} are for display. The
+ * {@link #payload()} carries what the producing provider requires to be replayed verbatim
+ * (a signature, encrypted content, a thought signature) and is only meaningful to that
+ * provider; see {@link #replayableTo(String)}. A {@code redacted} part has no text and
```
ericbottard: Not sure I understand this correctly, but if I do, there is no need (and there is a risk of desynch) to keep a separate boolean. Isn't this equivalent to text==null?

dimitarproynov:
> Is that really what we want (in the long run)? My understanding was that is broken (because we lost the interleaving). But maybe I'm not understanding what you're talking about here correctly.

In subsequent changes we will migrate all ChatModel implementations to use the MessagePart abstraction and not to use the legacy constructors. They are there to provide backward compatibility.

> > Equality and hashCode now compare parts

Equality is used in the isMemoryAlreadyInPrompt in the MessageChatMemoryAdvisor

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AbstractMessage.java
```diff
@@ -61,17 +70,86 @@ public abstract class AbstractMessage implements Message {
...
+
+	/**
+	 * Joins the text parts without a separator, matching the historical concatenation of
+	 * provider text blocks. A message with parts but no text part (reasoning only, tool
+	 * calls only) has an empty text, so streamed chunks and tool-call turns never report
+	 * {@code null} and the common {@code map(getText).collect(joining())} idiom keeps
+	 * working; only a message without any part has a {@code null} text, as a legacy
+	 * message built from a {@code null} content always had.
+	 */
+	private static @Nullable String joinText(List<MessagePart> parts) {
```
ericbottard, line 118: This does the correct thing but is way too hard to read.
It is used only once, and can (or it should if our API is to be elegant) be replaced by
```java
		this.select(TextPart.class).map(TextPart::text).collect(Collectors.joining());

```

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AssistantMessage.java
```diff
@@ -43,6 +59,14 @@ public class AssistantMessage extends AbstractMessage implements MediaContent {
 
 	private final List<ToolCall> toolCalls;
 
+	private final List<ReasoningPart> reasoning;
```
ericbottard: I'm not sure I understand the reasoning (ah ah) here. I thought the point of this PR was to always une a SINGLE list of parts, so that things can be interleaved. Why have a second, separate field?

If this value is derived, it should have a getter, but not a field IMHO (see comment elsewhere)

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AbstractMessage.java
```diff
@@ -61,17 +70,86 @@ public abstract class AbstractMessage implements Message {
...
+		Assert.notNull(parts, "Parts must not be null");
 		Assert.notNull(metadata, "Metadata must not be null");
+		this.parts = List.copyOf(parts);
+		String textContent = joinText(this.parts);
+		if ((messageType == MessageType.SYSTEM || messageType == MessageType.USER) && textContent == null) {
+			Assert.notEmpty(this.parts, "Content must not be null for SYSTEM or USER messages");
+			textContent = "";
+		}
 		this.messageType = messageType;
 		this.textContent = textContent;
```
ericbottard: a value that is derived (or derivable) should have a getter, but not a field moving forward.

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AbstractMessage.java
```diff
@@ -108,12 +186,12 @@ public boolean equals(@Nullable Object o) {
 			return false;
 		}
 		return this.messageType == that.messageType && Objects.equals(this.textContent, that.textContent)
-				&& Objects.equals(this.metadata, that.metadata);
+				&& Objects.equals(this.metadata, that.metadata) && Objects.equals(this.parts, that.parts);
```
ilayaperumalg: Given the textContent is derived from parts, would this be redundant to include textContent as part of equals and hashCode?

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AbstractMessage.java
ericbottard: If this is derived, this should not use a field IMHO

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AbstractMessage.java
ericbottard, line 118: This does the correct thing but is way too hard to read.
It is used just once, and is (or should be if our API is to be elegant) equivalent to
```java
		this.select(TextPart.class).map(TextPart::text).collect(Collectors.joining());

```

ericbottard: the caveat about null can be handled on a different line as well

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AssistantMessage.java
ericbottard: Using a separate list for those goes against the goal of this change IMO.

ericbottard:
> Equality is used in the isMemoryAlreadyInPrompt in the MessageChatMemoryAdvisor

👍

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/part/ReasoningPart.java
dimitarproynov: IMHO what this separate boolean expresses cannot be derived by isEmpty(text), but can be expressed via an attribute (which will survive (de)serialization), thus the "redacted" boolean must be removed.

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AbstractMessage.java
dimitarproynov: textContent however is protected, so removing it will be a breaking change. I will mark it deprecated and marked for removal.

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/AbstractMessage.java
dimitarproynov: Agreed. However Collectors.joining() copies the String contents even if there is just a single TextPart (which is a common case). A large text will be kept twice in memory thus I made a hybrid variant to handle the single TextPart case, but to use a much more readable Streaming API.

ericbottard: Don't underestimate the platform's ability to share the actual char array backing the Strings, etc.
I would not worry about that until proven otherwise. Readability trumps (supposed) performance IMO.

dimitarproynov: True that. Premature optimization is the root of all evil.

ericbottard (approved):
LGTM for now, let's do a first iteration with this

spring-ai-model/src/main/java/org/springframework/ai/chat/messages/part/MessagePart.java
```diff
@@ -0,0 +1,74 @@
...
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * One element of the ordered content of a chat message.
+ * <p>
+ * A message is a list of parts in the order the model produced or the caller supplied
+ * them, so interleavings such as reasoning, tool call, text, tool call survive a round
+ * trip through memory and back to the provider. The hierarchy is sealed: provider
```
tzolov, line 31: The interface is not sealed!
However the Jackson type registry is closed to the six types, and different part type added outside Spring AI deserializes as an UnknownPart.

tzolov (changes requested):
please update the chatmodel.adoc and prompt.adoc and likely add a related section to the upgrade-notes.adoc.

Otherwise, it should be good for now!

Issue #63, opened 2023-10-26
Parent Document Retriever feature.


radhakrishna67:
**Expected Behavior**

Lanchain's Parent Document Retriever feature is expected in Spring AI

**Current Behavior**
Files are chunked and cretead embeddings for them. We would like to have child chunks for parent chunks of the document.

markpollack:
https://python.langchain.com/docs/modules/data_connection/retrievers/parent_document_retriever as a reference.

I'd like to have some RAG evaluation setup in place to better understand how things improve or not based on these types of algorithms.  Putting in the backlog for now. 

Contributions are welcome.

PR #7002, opened 2026-09-16
Add VectorStore.upsert for pre-computed embeddings

### LINKED ISSUE #6989: Add VectorStore.upsert for pre-computed embeddings
Every vector store embeds inside `add()`, so there's no way to write vectors you
  already have (for e.g, from an overnight batch API, another team's pipeline, a migration
  that exports text and vectors together, or a multimodal model over an image/video, etc.)

  So a small, additive method:

  ```
  public record EmbeddedDocument(Document document, float[] embedding) { }

  // VectorStore
  default void upsert(List<EmbeddedDocument> entries) {
      throw new UnsupportedOperationException(getName() + " does not support upsert");
  }
```
  The default throws an exception, so stores that don't implement it are unaffected, and
  add()/Document stay untouched. Same id replaces the row, so re-running with stable ids won't duplicate. Instrumented stores get a `doUpsert` on `AbstractObservationVectorStore` and an UPSERT operation, so it shows up in metrics. 

  For content that lives elsewhere (an image in S3, a doc too big for the row), a new
  `DocumentMetadata.CONTENT_REF` key holds the pointer. The store never reads it; the
  app resolves it after search.

 Scoped for 2.1.0-M1:

  - [ ] EmbeddedDocument + VectorStore.upsert
  - [ ] observation wiring
  - [ ] CONTENT_REF
  - [ ] pgvector + a shared test suite others reuse (Testcontainers only)
  - [ ] Redis, Elasticsearch, Qdrant
  - [ ] docs

  Later: the rest of the stores (Milvus does a plain insert and Bedrock KB has no
  write path, so those are the awkward ones), and deterministic ids / clean re-runs 
  (upsert is what those build on)

  Closes #3540. Replaces #3541 (which changed the doAdd signature) with an additive
  path — thanks @aniketg-21, the pair object came from there. Also helps #1009: embed
  multimodally yourself, then upsert the vector.

### PR DESCRIPTION
**Pull request by sobychacko**
Fixes #6989

User supplied embeddings write path: an EmbeddedDocument record pairing a Document with a caller-supplied vector, and a default VectorStore.upsert that throws so stores opt in. Wired into AbstractObservationVectorStore (UPSERT observation + doUpsert), with replace-by-id implementations for pgvector, Redis, Elasticsearch, and Qdrant. Adds a shared AbstractVectorStoreUpsertTests TCK, the DocumentMetadata.CONTENT_REF key for non-text content, and reference docs.

See also #3540 and #3541

### CODE REVIEW

File: spring-ai-vector-store/src/main/java/org/springframework/ai/vectorstore/observation/VectorStoreObservationContext.java [hunk 1]
```diff
@@ -142,6 +142,10 @@ public enum Operation {
 		 * VectorStore add operation.
 		 */
 		ADD("add"),
+		/**
+		 * VectorStore upsert operation.
+		 */
+		UPSERT("upsert"),
```
tzolov, line 148: Please update the `observability/index.adoc`'s (`low cardinality keys` and `Labels` tables) for the  new `Operation.UPSERT` and the emitted `db.operation.name=upsert`.

tzolov (changes requested):
Thanks @sobychacko, 
I was wondering if we need to put some guards in cases where CONTENT_REF is empty or set on non upsert() execution path? Are those supported invariants? if not maybe a waring message would be useful. 

Also Seems like Redis drops the CONTENT_REF? It returns only metadata declared in `metadataFields` at build time.

vector-stores/spring-ai-redis-store/src/main/java/org/springframework/ai/vectorstore/redis/RedisVectorStore.java
```diff
@@ -385,6 +386,49 @@ public void doAdd(List<Document> documents) {
...
+				fields.put(this.contentFieldName, document.getText());
+				fields.putAll(document.getMetadata());
+				pipeline.jsonSetWithEscape(key(document.getId()), JSON_SET_PATH, fields);
+			}
+			List<Object> responses = pipeline.syncAndReturnAll();
+			Optional<Object> errResponse = responses.stream().filter(Predicate.not(RESPONSE_OK)).findAny();
+			if (errResponse.isPresent()) {
+				String message = MessageFormat.format("Could not upsert document: {0}", errResponse.get());
+				logger.error(message);
+				throw new RuntimeException(message);
```
ilayaperumalg: Other vector stores seem to throw IllegalStateException here. We can be consistent throwing the same even though RuntimeException is still good.

vector-stores/spring-ai-qdrant-store/src/main/java/org/springframework/ai/vectorstore/qdrant/QdrantVectorStore.java
```diff
@@ -201,6 +202,41 @@ public void doAdd(List<Document> documents) {
...
+				return PointStruct.newBuilder()
+					.setId(io.qdrant.client.PointIdFactory.id(UUID.fromString(document.getId())))
+					.setVectors(io.qdrant.client.VectorsFactory.vectors(entry.embedding()))
+					.putAllPayload(toPayload(document))
+					.build();
+			}).toList();
+
+			this.qdrantClient.upsertAsync(this.collectionName, points).get();
+		}
+		catch (InterruptedException | ExecutionException e) {
```
ilayaperumalg: Should we call `Thread.currentThread().interrupt()` when handling InterruptedException?

sobychacko:
@tzolov  Thanks for the review. I will address them. See below.

**Guards / invariants**:  you're right, I never documented the contract.

CONTENT_REF on `add()` looks like valid usage to me: embed a summary or a chunk and keep a pointer to the full article, close to what #63 asks for, so a warning would hit legitimate use.

On a blank value: `CONTENT_REF` is an ordinary metadata entry, and today any metadata value can be blank, on both `add` and `upsert`. Checking this one would make it the only metadata key we validate.

Would documenting the invariants plus a TCK case that round-trips a CONTENT_REF row be enough? There's no behavioral test for it today. Happy to add guards if you'd rather.

**Redis**: The read only asks for the fields declared in `metadataFields`, so undeclared keys get stored but dropped on the way back. `add` does the same, and Redis is the only one of the four stores here that behaves this way.

`content_ref` is a known key, so declaring it covers it:

```java
.metadataFields(MetadataField.tag(DocumentMetadata.CONTENT_REF.value()))
```

I'll put that in the docs and the Redis IT so it runs in CI, plus a warning when a document carries undeclared metadata so it stops being silent. 

@ilayaperumalg Thanks for the review. I will address them.

tzolov (approved):
LGTM, thanks Soby

sobychacko:
Merged via https://github.com/spring-projects/spring-ai/commit/84d9d52e635d16255e1d10e3f4fbf18f9e182b14 and https://github.com/spring-projects/spring-ai/commit/cecb01b141f07d5965bacd7ed44d56d4c0bb7ba5.

PR #7006, opened 2026-09-17
Add OpenAI Responses API support

dimitarproynov:
Add OpenAiResponsesChatModel, a second OpenAI ChatModel talking to /v1/responses, which is the only way to combine reasoning with tool calling on GPT-5.4 and later. The model is stateless: every call sends the whole Prompt, so ChatMemory, advisors and RAG keep working as they do with OpenAiChatModel, and the server-side continuation modes stay out of scope.

A Responses reply is an ordered list of typed items rather than one message, and a reasoning model returns its chain of thought as an opaque encrypted blob that must be handed back verbatim to keep its train of thought across a tool call. Both are carried by the MessagePart list on AssistantMessage, one part per output item and in order, so a reasoning item still precedes the function calls it produced. A reasoning item becomes a ReasoningPart whose OpaquePayload holds the encrypted content, a message item a TextPart, a function call a ToolCallPart, a generated image a MediaPart, and every tool OpenAI ran itself an UnknownPart keeping the item verbatim. The item id, the item status and the message phase travel in the part attributes, since an item cannot be rebuilt without them.

Only function_call items become ToolCalls, so ToolCallingAdvisor never looks for a local callback named web_search. Reasoning is replayed only when its payload came from this provider, because the API rejects unsigned reasoning; a foreign or payload-less reasoning part is skipped with a debug log rather than sent and refused.

Streaming stamps every part with its output index through StreamingParts and lets MessageAggregator rebuild the transcript. A completing item is emitted emptied of whatever already arrived as deltas, so its payload and attributes reach the aggregated part without the text being shown twice, and every chunk carries the response id the aggregator groups indexed parts by.

Also add OpenAiResponsesChatOptions and HostedTool with support for the tools OpenAI executes server-side, structured output, image and PDF input, observability, an auto-configuration selecting the endpoint through spring.ai.openai.chat.api, and a reference documentation page.

Full build and all OpenAI integration tests pass.


models/spring-ai-openai/src/main/java/org/springframework/ai/openai/responses/HostedTool.java
```diff
@@ -0,0 +1,235 @@
...
+ * {@link org.springframework.ai.chat.messages.AssistantMessage.ToolCall tool calls} -
+ * they appear in the turn transcript and, summarized, under
+ * {@link OpenAiResponsesMetadata#HOSTED_TOOL_CALLS} on the generation metadata.
+ * <p>
+ * {@link Raw} is the escape hatch for tools OpenAI ships before Spring AI types them.
+ *
+ * @author Dimitar Proynov
+ * @since 2.1.0
+ */
+public sealed interface HostedTool {
```
ilayaperumalg, line 46: Can we rename `HostedTool` to something like `OpenAIManagedTool` or `BuiltInTool` (what OpenAI Responses refer this as) to add more clarity?

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/responses/OpenAiResponsesChatModel.java
```diff
@@ -0,0 +1,402 @@
...
+			return prompt;
+		}
+		OpenAiResponsesChatOptions.Builder builder = this.options.mutate();
+		builder.combineWith(runtimeOptions.mutate());
+		return prompt.mutate().chatOptions(builder.build()).build();
+	}
+
+	private RequestOptions buildRequestOptions(Prompt prompt) {
+		OpenAiResponsesChatOptions chatOptions = (OpenAiResponsesChatOptions) prompt.getOptions();
+		Assert.state(chatOptions != null, "Prompt options must be OpenAiResponsesChatOptions type");
```
ilayaperumalg, line 273: I think the assertion needs to happen before casting prompt.getOptions in the line before

models/spring-ai-openai/src/main/java/org/springframework/ai/openai/responses/HostedTool.java
```diff
@@ -0,0 +1,235 @@
...
+	 * {@link OpenAiResponsesMetadata#HOSTED_TOOL_CALLS} on the generation metadata
+	 * itself, because an {@code mcp_approval_request} left unanswered ends the turn
+	 * without running the tool.
+	 *
+	 * @param serverLabel the label the model uses to address the server
+	 * @param serverUrl the server URL, or {@code null} when using {@code connectorId}
+	 * @param connectorId an OpenAI-managed connector id, or {@code null}
+	 * @param allowedTools restricts the callable tools, or {@code null} for all of them
+	 * @param headers headers to send to the MCP server, e.g. authorization
+	 * @param requireApproval {@code always} or {@code never}, or {@code null} for the
```
ilayaperumalg, line 148: This contradicts with the `out-of-scope` statement above on requireApproval. Depending on the allowed values, we can introduce assertion at the toTool() method as well.

Issue #7022, opened 2026-09-21
UsageCalculator.getCumulativeUsage() throws NullPointerException when the previous response reports no usage


CodeTrainerMan:
## Bug description

`UsageCalculator.getCumulativeUsage(currentUsage, previousChatResponse)` guards `currentUsage` against being empty, but dereferences the usage taken from the previous response unconditionally:

```java
Usage usageFromPreviousChatResponse = null;
if (previousChatResponse != null) {
    usageFromPreviousChatResponse = previousChatResponse.getMetadata().getUsage();
}
else {
    return currentUsage;
}
// For a valid usage from previous chat response, accumulate it to the current usage.
if (!isEmpty(currentUsage)) {
    Integer promptTokens = currentUsage.getPromptTokens();
    promptTokens += usageFromPreviousChatResponse.getPromptTokens(); // NPE when null
    ...
```

A response can legitimately carry a `null` usage: `ChatResponseMetadata.Builder.usage(...)` performs no null check and `ChatResponse.Builder.from(...)` copies whatever the source metadata holds, so a provider that does not report usage for a given response propagates `null`. Accumulating against such a response fails with:

```
java.lang.NullPointerException: Cannot invoke "org.springframework.ai.chat.metadata.Usage.getPromptTokens()" because "usageFromPreviousChatResponse" is null
	at org.springframework.ai.support.UsageCalculator.getCumulativeUsage(UsageCalculator.java:70)
```

A `null` usage is already an expected state elsewhere in the framework: both `ChatModelMeterObservationHandler` and `DefaultChatModelObservationConvention` null-check `getUsage()`, and `ChatModelMeterObservationHandlerTests#shouldHandleNullUsageGracefully` builds metadata with `ChatResponseMetadata.builder().model("model").usage(null).build()`.

## Steps to reproduce

```java
Usage current = new DefaultUsage(10, 20, 30);
ChatResponse previous = ChatResponse.builder()
    .generations(List.of())
    .metadata(ChatResponseMetadata.builder().usage(null).build())
    .build();

UsageCalculator.getCumulativeUsage(current, previous); // throws NullPointerException
```

## Expected behavior

A previous response that reports no usage contributes nothing, so `currentUsage` should be returned unchanged — the same behaviour already applied when `previousChatResponse` itself is `null`.

## Environment

- Spring AI `main`
- Module: `spring-ai-model`
- Class: `org.springframework.ai.support.UsageCalculator`

[...]

PR #7025, opened 2026-09-21
fix: guard null usage from previous response in UsageCalculator

Rainmemery:
### TLDR

`UsageCalculator.getCumulativeUsage()` dereferences the usage taken from the previous chat response **unconditionally** when `previousChatResponse != null`:

```java
if (previousChatResponse != null) {
    usageFromPreviousChatResponse = previousChatResponse.getMetadata().getUsage();
}
...
promptTokens += usageFromPreviousChatResponse.getPromptTokens(); // NPE
```

A response can legitimately carry a `null` usage — `ChatResponseMetadata.Builder.usage(...)` does no null check and `ChatResponse.Builder.from(...)` copies whatever the source holds, so providers that do not report usage on a given response propagate `null`. Accumulating against such a response throws:

```
NullPointerException: Cannot invoke "org.springframework.ai.chat.metadata.Usage.getPromptTokens()"
because "usageFromPreviousChatResponse" is null
```

This is already an expected state elsewhere in the framework: `ChatModelMeterObservationHandler` and `DefaultChatModelObservationConvention` both null-check `getUsage()`, and `ChatModelMeterObservationHandlerTests#shouldHandleNullUsageGracefully` builds metadata with `usage(null)`.

### Fix

When the previous response reports no usage, there is nothing to accumulate — return the current usage as-is, mirroring the existing `previousChatResponse == null` branch. All other behaviour (isEmpty check, cache-metric accumulation) is unchanged.

### Tests

`getCumulativeUsageReturnsCurrentWhenPreviousResponseHasNullUsage` in `UsageCalculatorTests` — builds a previous `ChatResponse` with `usage(null)` and asserts the current usage is returned unchanged (same instance).

All 18 `UsageCalculatorTests` pass; `spring-javaformat:validate` clean.


Rainmemery:
@ilayaperumalg - could you approve the workflow run for this PR too? It's sitting at action_required (first-time contributor) on head 2906a1d; unit tests and javaformat already pass locally.

Rainmemery:
Closing this in favor of #7023 — @CodeTrainerMan's PR addresses the same issue (#7022, `UsageCalculator.getCumulativeUsage` NPE on null previous usage) with an equivalent guard and was opened earlier (17:38 vs 19:42). Keeping both would duplicate review work for maintainers. Thanks for the fix!

PR #7020, opened 2026-09-21
fix(openai): fall back to list position when embedding response omits index

Rainmemery:
Fixes [#7019](https://github.com/spring-projects/spring-ai/issues/7019).

## Problem
OpenAI-compatible endpoints such as Google Gemini's compatibility mode omit the (redundant) `index` field on embedding responses. `Embedding.index()` then throws `OpenAIInvalidDataException: index is not set` from `OpenAiEmbeddingModel.generateEmbeddingList`, breaking `embeddingModel.embed(...)` entirely for those providers.

## Fix
Resolve the index via `Embedding._index()` and fall back to the position within the response list when the field is missing or null. The OpenAI API guarantees `index` correlates with the position of the input, so spec-compliant responses keep their exact indices.

## Verification
- New unit test `indexFallsBackToPositionWhenProviderOmitsIt` covers a response that mixes a missing index (position fallback) with a present one; full `OpenAiEmbeddingModelTests` passes (2/2).

PR #7016, opened 2026-09-20
Honor MessagePart order in BedrockProxyChatModel user messages

Rainmemery:
Resolves gh-7012

## Summary
`BedrockProxyChatModel.createRequest` read `UserMessage.getText()` and `getMedia()` separately and always emitted the text `ContentBlock` before all media blocks, ignoring the order of `UserMessage.getParts()`. A user message built with

```java
UserMessage.builder()
        .part(MediaPart.of(video))
        .part(TextPart.of("Instructions and question about the video ..."))
        .build();
```

was therefore sent as `[text, video]` instead of the expected `[video, text]`.

This change iterates `userMessage.getParts()` in order and emits one `ContentBlock` per supported part (`TextPart`, `MediaPart`).

## Backward compatibility
Messages built through the existing `text(...)` / `media(...)` builder methods or the legacy constructors keep their previous request output unchanged: their parts are already ordered text-then-media. A user message may still carry media only (gh-6695) — the empty text block is still omitted.

## Testing
- New unit test `userMessagePartsRespectPartOrderWhenMediaPrecedesText` verifies media comes before text when `MediaPart` precedes `TextPart`.
- `BedrockProxyChatModelTest`: 23 tests pass (0 failures, 0 errors).
- `mvn spring-javaformat:validate` passes.

## Checklist
- [x] Signed-off-by present (DCO)
- [x] Rebased on latest main
- [x] Unit tests added and passing

PR #7026, opened 2026-09-21
fix(openai): guard against null usage in embeddings response

### LINKED ISSUE #7019: Compatibility problem with Spring AI OpenAI embeddings and Gemini OpenAI compatibility mode
**Bug description**
When using Spring AI OpenAI with the Google Gemini OpenAI compatibility mode ([as described here](https://geeknet.me/spring-ai-and-the-free-google-gemini-api/)), creating an embedding fails with the following stacktrace:
```
com.openai.errors.OpenAIInvalidDataException: `index` is not set
	at com.openai.core.JsonField.getRequired$openai_java_core(Values.kt:174) ~[openai-java-core-4.49.0.jar:4.49.0]
	at com.openai.models.embeddings.Embedding.index(Embedding.kt:63) ~[openai-java-core-4.49.0.jar:4.49.0]
	at org.springframework.ai.openai.OpenAiEmbeddingModel.generateEmbeddingList(OpenAiEmbeddingModel.java:291) ~[spring-ai-openai-2.0.1.jar:2.0.1]
	at org.springframework.ai.openai.OpenAiEmbeddingModel.generateEmbeddingResponse(OpenAiEmbeddingModel.java:275) ~[spring-ai-openai-2.0.1.jar:2.0.1]
	at org.springframework.ai.openai.OpenAiEmbeddingModel.lambda$call$3(OpenAiEmbeddingModel.java:253) ~[spring-ai-openai-2.0.1.jar:2.0.1]
	at io.micrometer.observation.Observation.observe(Observation.java:634) ~[micrometer-observation-1.17.1.jar:1.17.1]
	at org.springframework.ai.openai.OpenAiEmbeddingModel.call(OpenAiEmbeddingModel.java:249) ~[spring-ai-openai-2.0.1.jar:2.0.1]
	at org.springframework.ai.embedding.EmbeddingModel.embed(EmbeddingModel.java:85) ~[spring-ai-model-2.0.1.jar:2.0.1]
```

While digging into the request, I noticed that Gemini seems to omit `index: 0` from the response:
```
$ curl https://generativelanguage.googleapis.com/v1beta/openai/v1/embeddings -H "Authorization: Bearer $OPENAI_API_KEY" -H "Content-Type: application/json" -d '{ "input": [ "One", "Two" ], "model": "gemini-embedding-001", "encoding_format": "float" }'
{
  "object": "list",
  "data": [
    {
      "object": "embedding",
      "embedding": [
        0.0034968799445778131,
        ...
      ]
    },
    {
      "object": "embedding",
      "index": 1,
      "embedding": [
        0.0098916832357645035,
        ...
      ]
    }
  ],
  "model": "gemini-embedding-001"
}
```

**Note**
I am not sure whose fault this is. One the one hand, it looks like Gemini violates the spec. On the other hand, this looks like a conscious omission, so maybe OpenAI changed the API in a backwards-incompatible way in the past.
Also, it looks like the index is redundant anyway, since it is guaranteed to correspond with the position in the array ([confirmed by OpenAI developers](https://community.openai.com/t/does-the-index-field-on-an-embedding-response-correlate-to-the-index-of-the-input-text-it-was-generated-from/526099/3)).

Possible solution: use 0 if `nativeDatum._index() instanceof JsonMissing`.

**Environment**
Java 21
org.springframework.ai:spring-ai-bom:2.0.1

```java
@SpringBootApplication
public class GeminiEmbeddingExampleApplication implements CommandLineRunner {
		SpringApplication application = new SpringApplication(GeminiEmbeddingExampleApplication.class);
		application.setDefaultProperties(
				Map.of("spring.ai.openai.api-key", "<API key from https://aistudio.google.com/>",
						"spring.ai.openai.base-url", "https://generativelanguage.googleapis.com/v1beta/openai",
						"spring.ai.openai.embedding.model", "gemini-embedding-001"));
		application.run(args);
	}

	@Autowired
	private EmbeddingModel embeddingModel;

	@Override
	public void run(String... args) {
		embeddingModel.embed(List.of("One", "Two"));
	}
}
```

Rainmemery:
Fixes #7019

Gemini's OpenAI-compatible embeddings responses (e.g. `gemini-embedding-*` routed through the completion adapter) may omit `usage`. Without a guard, `getDefaultUsage` throws a `NullPointerException` on every call, making embeddings unusable in those configurations.

This change returns an empty `DefaultUsage` when the native response carries no usage instead of dereferencing `null`.

- `OpenAiEmbeddingModel#getDefaultUsage`: early-return `new DefaultUsage(null, null, null, null)` when `nativeUsage == null`
- New unit test `testEmbeddingWithoutUsageDoesNotFail` covering a response without usage


Rainmemery:
@ilayaperumalg Could you please approve the workflow run for this first-time contributor PR? It fixes a NPE when Gemini OpenAI-compatible embeddings responses omit `usage`. Thanks!

Rainmemery:
@ilayaperumalg gentle nudge — could you approve the workflow runs for this first-time-contributor PR batch when you get a moment? All are small, self-contained fixes with green DCO and locally verified tests:

- #7026 (this PR) — guard against null usage in OpenAI embeddings (Gemini OpenAI-compat)
- #7025 — null-guard cumulative usage in `UsageCalculator`
- #7016 — honor `MessagePart` order in BedrockProxyChatModel user messages
- #7015 — replace message parts on `mutate()` instead of appending

Thanks for the earlier review of #7020!

ewoerner:
@Rainmemery Your PR doesn't fix the issue. `usage()` does not return `null`, it throws an `OpenAIInvalidDataException`. A fix would require calling `_usage()` instead.

Rainmemery:
Thanks @ewoerner — good catch. You're right that `usage()` throws `OpenAIInvalidDataException` when the optional field is absent, so the null guard never fired. Pushed a fix (1f715ab): `OpenAiEmbeddingModel` now reads the raw field via `response._usage()` and unwraps it with `asKnown().orElse(null)` inside `getDefaultUsage(...)`, so an absent `usage` (e.g. Gemini's OpenAI-compatible embeddings responses) yields an empty `DefaultUsage` instead of throwing. Unit tests updated to stub `_usage()` (`JsonField.of(...)` for the present case, `JsonMissing.of()` for the absent case). `mvn test -Dtest=OpenAiEmbeddingModelTests` passes locally (2/2) with NullAway and spring-javaformat clean.

